PR #7, opened 2015-02-24
Add sassdoc to gulp

anthonydillon:
# Details
- None

## Done
- Added sassdoc to gulp script
- Structured the gulp commands

### QA
- Run gulp and see commands available
- Run gulp build is should build the css and docs into a build folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 node_modules/
 npm-delog.log
 css/
+build/
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,4 @@ node_js:
 before_install:
   - gem update --system
   - gem install scss-lint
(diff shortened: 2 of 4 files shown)


nottrobin:
@anthonydillon looks good and QA worked.

Just one thing - we need travis to still do something useful. Could you update `.travis.yml` to say `script: node_modules/gulp/bin/gulp.js sass` so it still checks the sass compiles properly? *or* maybe create a new group `gulp` target called `test` then travis can run `node_modules/gulp/bin/gulp.js test`.

0atman:
works for me! tested locally

nottrobin:
Thanks for updating @anthonydillon that looks great.

Let's just wait for Travis to confirm it works, then we'll merge.

PR #9, opened 2015-02-24
Travis sasslint fail state

nottrobin:
Details
===
Currently Travis isn't failing, even though there's
a syntax error in the sass.

This should fix that by making `gulp sasslint` return
an error state properly when something fails.

Done
---
- Separate `sasslint` from `sass` gulp commands
- Add FailReporter to `sasslint`
- The Gulp Sass target will now return an error status, with a useful error message
- The sass-lite target will return an error in a simpler and more useful way (without breaking watch)

QA
---

Run the following:

``` bash
npm install
node_modules/gulp/bin/gulp.js test || echo "successfully returned an error"
node_modules/gulp/bin/gulp.js sass || echo "successfully returned an error"
```

As long as the Sass still has its syntax error, you should see `successfully returned an error` after the output of each command.

Then run `node_modules/gulp/bin/gulp.js watch`, change one of the sass files and confirm that you see an error message in the watcher without it breaking.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,14 +1,28 @@
-
 var gulp = require('gulp')
     rename = require('gulp-rename'),
     sass = require('gulp-sass'),
     autoprefixer = require('gulp-autoprefixer'),
     notify = require('gulp-notify'),
+    gutil = require('gulp-util'),
     scsslint = require('gulp-scss-lint');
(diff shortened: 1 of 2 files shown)


0atman:
the build fails, but it's clearly because of the very sass lint errors that we're trying to fix

PR #15, opened 2015-02-25
added scss-lint gem installstep to makefile

0atman:
Is this the right thing to do, to bootstrap our gulp setup? Karl's errors were due to an old version of `scss-lint`, which can be installed or upgraded by this command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,2 +1,3 @@
 install:
+	sudo gem install scss-lint
 	npm install


Makefile
```diff
@@ -1,2 +1,3 @@
 install:
```
anthonydillon: Can we have a doc strings please.

anthonydillon:
Shame about that one gem dependence. Thanks :+1: with a comment.

nottrobin:
Not accepting at present. We should look into installing the gem with gulp - having multiple built tools is insane.

PR #17, opened 2015-02-25
Convert tabs to 2 spaces in scss files

nottrobin:
Spacing in scss files looks like a mess in commits, because there are tabs littered everywhere.

This replaces all tab characters with two spaces in SCSS files.

Done
---

- Replaced all tab characters with 2 spaces in scss files

QA
---

- Run `gulp sass`, make sure it still compiles
- Ensure you're okay with the concept of having to merge all the files with everyone else's work

### DIFF
--- a/scss/patterns/_blockquotes.scss
+++ b/scss/patterns/_blockquotes.scss
@@ -8,7 +8,7 @@ blockquote {
       color: $cool-grey;
       padding-left: 10px;
       padding-right: 10px;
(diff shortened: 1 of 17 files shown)


0atman:
feeling in the room: we can find and replace tabs at any point, lets do this when everyone's at a commited state, perhaps tomorrow morning?

cwebteam:
No prob

nottrobin:
Cool

PR #19, opened 2015-02-25
Scss linting recusion

deadlight:
# SCSS linting recursion

## Done

Changed the gulpfile to recurse over all scss files, not just the root.

## QA

run "gulp test" and then ensure that it's reading deeper files in the file structure.


grahambancroft:
Lgtmii

PR #21, opened 2015-02-26
Make it so

0atman:
# Details
After yesterday's debacle with different versions of gulp, npm and scss-lint, I've codified our requirements in a Makefile.

## Done
*
## QA
Type `make clean install` in the root. If you're up to date, not much output. If you're behind, you'll update. The command is cross-platform and idempotent (thanks for the syntax @nottrobin) so it is safe to run `make install` every time you want to be sure.


Makefile
```diff
@@ -1,2 +1,10 @@
 install:
-	npm install
+	@echo "Installing missing gulp deps..."
+	@type ${'gulp'} || sudo npm install -g gulp
```
nottrobin: Why `${'gulp'}` and not `gulp`?

0atman: because I'm a noob

0atman:
@nottrobin updated, thanks

nottrobin:
Yes I agree that it turns out `gulp` can't sensibly install the gem itself. This is a real shame 'cos it'd be nice to include one and only one build tool.

Given that we're probably going down the `make` route, I have two points:

- First, this won't solve the problem you describe, because you don't do anything to make sure `scss-lint` is any particular version. The problem here is that if we want to include `gem`s of specific versions we should really be doing it with a `Gemfile` using [bundler](http://bundler.io/), but now things are just getting out of hand. What do we do if someone has the wrong version on their system? We can't go changing their system version, that'd be really irresponsible. The ideal is to install the correct version locally (e.g. with bundler) and then make sure `gulp-scss-lint` is using that version instead. But I've no idea if that's possible.
- You're installing a system version of `gulp` as well as the `gulp` dependency specified in the gulpfile itself. I agree it should be installed at the system level, and probably removed from the gulpfile. Could you do this?
- Do we want to be opinionated about the version of `gulp` people have installed? Once again we can't have too much control over which version people use, but we do need to know which versions to test with in our `.travis.yml`.

PR #28, opened 2015-02-26
Inline disabled some linting exceptions as they are required

anthonydillon:
# Details
- None

## Done
- Disabled some linting exceptions inline

### QA
- See there are no linting issues when running ```gulp test --file patterns/_vertical-divider.scss```


nottrobin:
Well the QA command you suggested had tons of output - it still seems to be linting everything.

But I tried `gulp test | grep '_vertical-divider.scss'` and that produced no output so I'm satisfied.

:+1:

anthonydillon:
@nottrobin thanks

PR #36, opened 2015-02-26
Lint structure

anthonydillon:
# Details
- None

## Done
- Linted the structure scss file

### QA
- Run ```gulp test --file patterns/structure.scss``` and check there is no output.


deadlight:
LGTM

PR #40, opened 2015-02-26
Changing test file to full path

anthonydillon:
# Details
- None

## Done
- Changed the path required for the test param to include scss/ so its the full path

### QA
- Run ```gulp test``` and see if checks all files
- Run ```gulp test --file scss/patterns/_image-centered.scss``` and check it only checks that file
- Try tab completing the file path as its relative


nottrobin:
Yep. :+1:

PR #42, opened 2015-02-27
Fix the box-solid-grey issue

anthonydillon:
# Details
- Issue: https://github.com/ubuntudesign/web-style-guide/issues/41

## Done
- Replaced the use of box-solid-grey with mid-grey

### QA
- Run ```gulp build``` and check it generates the css without an error


deadlight:
LGTM

PR #45, opened 2015-02-27
Lint inline logos

anthonydillon:
# Details
- None

## Done
- Linted the inline-logos

### QA
- Run ```gulp test --file scss/patterns/_inline-logos.scss``` and check there is no warnings


richmccartney:
All looks fan-dabby-dozy +1

anthonydillon:
@richmccartney thanks

PR #52, opened 2015-02-27
Fixed the row padding

anthonydillon:
# Details
- None

## Done
- Moved row styles from grid to row styles

### QA
- Do ```gulp build``` and check the index.html rows have good padding


deadlight:
LGTM

PR #62, opened 2015-02-27
Update readme install instructions; change "install-dependencies" to "dependencies" to save peoples fingers.

nottrobin:
Done:

- Changed name of npm dependencies script
- Update README to have commands which actually work

QA:

- Run the dependencies command from the readme.


deadlight:
:+1:

PR #82, opened 2015-03-16
Add list module

grahambancroft:
Done:

Includes the lists styles and wraps it in a mixin

To test:

gulp test


richmccartney:
Add the mixin include to _vanilla.scss then good to go.

PR #96, opened 2015-03-16
fixing vanilla mixin include

richmccartney:
## Done

Adding the mixin include

## Test

run ```gulp build```


grahambancroft:
+1

PR #101, opened 2015-03-17
Amend asset path variable. Add fonts.

grahambancroft:
Done:

Amended asset path. Added fonts.

Test:

Run gulp build. Check font is ubuntu.


deadlight:
LGTM +1

PR #103, opened 2015-03-17
reset etc.

deadlight:
Done:
Stripped some styles out of reset and into where they make more sense.

QA:

gulp build


grahambancroft:
LGTM +1

PR #112, opened 2015-03-17
Buttons

richmccartney:
#### What's this PR do?

Rewrites the original button.scss file and adds only the 3 new buttons required; primary, secondary and text.

#### Where should the reviewer start?

Look at _button.scss and view how its been written.

#### How should this be manually tested?

```gulp test``` and ```gulp build```


richmccartney:
@anthonydillon I really want you thoughts on how I have written this. I have had a thought about how to extend things further down into different themes. I think with encapsulating too much into vf-buttons we are stifling the extendability we could have with our mixing and placeholders.

You can see in this you could see use ```@include button(colour, background, border);``` on other classes if its just inside of ```vf-buttons``` we wouldn't be able to use it else where.

As I said would love your thoughts on this.

anthonydillon:
Please re-propose to master.

PR #115, opened 2015-03-18
commented and updated variables

deadlight:
Done:
Added comment block
Changed hard-coded min-width to use a var

QA:
gulp build


grahambancroft:
<3

PR #119, opened 2015-03-18
Framework update

deadlight:
Done:
Added comments to the top of the file

QA:
gulp build (and look at the generated docs, why not)


grahambancroft:
+1 love it

PR #121, opened 2015-03-18
Typography

deadlight:
Done:
Added a header to the file and set the default font size

QA:
gulp build, no errors


grahambancroft:
+1

PR #122, opened 2015-03-18
clean up and add heading comment to rows

deadlight:
Done:
Cleaned and added heading to rows file

QA:
gulp build without errors


0atman:
Looks good to me!

PR #131, opened 2015-03-20
add some examples

deadlight:
Done:
A couple of examples

QA:
gulp build

### DIFF
--- a/scss/modules/_rows.scss
+++ b/scss/modules/_rows.scss
@@ -6,6 +6,10 @@
 
 /// Row element styles
 /// @group Rows
+/// @example
+///   <div class="row">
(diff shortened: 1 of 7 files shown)


scss/modules/_rows.scss
```diff
@@ -6,6 +6,10 @@
 
 /// Row element styles
 /// @group Rows
+/// @example
+///   <div class="row">
```
nottrobin: Really anal point: this shouldn't be indented - the 2-space indent actually comes through in the docs example.

nottrobin:
One tiny comment. Otherwise this is amazing shit. :+1:

PR #135, opened 2015-03-20
Add python script to bump the version of package.json

0atman:
# Details
This new script is called with the params: `major`, `minor`, or `point`, and will increment the correct part of the package.json `version` string.

## QA
Run `./bump_package_version.py point` and observe the new version in `package.json`. Try with major and minor, and see if you can break it!

### DIFF
--- a/bump_package_version.py
+++ b/bump_package_version.py
@@ -1,4 +1,8 @@
 #! /usr/bin/env python
+"""
+A script for increasing the version in an npm project's
+package.json.
+"""
 
 import argparse
 import json
--- a/bump_package_version.py
+++ b/bump_package_version.py
@@ -1,0 +1,47 @@
+#! /usr/bin/env python
+
+import argparse
+import json
+import sys
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "release_level",
+    help=(
+        "Which componant of the version should be incremented."
+        "Options are: major, minor, point."
+    )
+)
+args = parser.parse_args()
+release_level = args.release_level
+
+component = {
+    'major': 0,
+    'minor': 1,
+    'point': 2
+}
+
+if release_level not in component.keys():
+    sys.exit(
+        "release_level must be one of {0}".format(component.keys())
+    )
+
+filename = 'package.json'
+
+with open(filename, 'r') as f:
+    package = json.loads(f.read())
+    print "Old version: {0}".format(package['version'])
+
+with open(filename, 'w') as f:
+    version_list = package['version'].split('.')
+
+    current_level = version_list[component[args.release_level]]
+
+    version_list[component[args.release_level]] = \
+        str(int(current_level) + 1)
+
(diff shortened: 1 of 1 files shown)


nottrobin:
I think we may want to split this out of the project eventually. I don't think we want too much bespoke build stuff polluting the main project. But it's fine in here for now.

0atman:
good point.

anthonydillon:
@0atman Thanks this is good but there are a few issues. When jumping minor and major versions it should reset lesser values to 0. Also does the version have to be applied to the bottom of the package.json. Nice to see the version higher in the file.

anthonydillon:
I'm very sorry...

0atman:
aha! very good point on the resetting lesser bits to 0!

0atman:
yeah, json documents are essentially objects with unordered attributes. The only sane way to machine-output them is alphabetical. In the future we could write some custom json serialisation that puts version and name and others to the top.

nottrobin:
So @0atman I've created a new repository, [vanilla-builder](https://github.com/ubuntudesign/vanilla-builder), with a modified version of your script here, which fixes the resetting to zero issue. Let me know what you think of it. If it looks good to you, close this PR?

0atman:
@nottrobin nice solution, closing.

anthonydillon:
Just so you guys know. I we working with vanilla cloud theme yesterday and
it was a really pain to push upstream a change to vanilla then publish.
Then go to the theme and update the dep and publish. Then go to the site
and update the dep in package.js and npm update.

I wonder if there is a way of scripting that as an internal tool?

nottrobin:
@anthonydillon interesting problem.

So there are a bunch of ways we can manage this.

Firstly, I think this is inevitably going to seem like more of a problem just at the moment when the framework is effectively in "alpha" stage, so you'll be finding and fixing bugs quite often. Hopefully it will become much more stable over time, and rarely will an update to the ubuntu theme actually require any updates to vanilla.

But we do still need a way to make more dynamic development across all the projects easier.

It was my impression that going forward we wouldn't be manually updating the NPM packages for either `vanilla-framework` or the themes ever. This should be done when we actually want to do a real release to the public, with the automated build scripts.

Instead we should have a much simpler method for development. Some ideas:

- we could keep the themes "master" branches' `package.json` pointing at the github repositories instead of locked version numbers, so it's always pulling down the latest version of vanilla, removing the need for an NPM publish
- or, in local development, instead of "npm install" we run a custom make target or something to pull down the *actual git repositories* into the `node_modules` folder, so that you can manage (commit and push to) all the repositories directly.

I still think that going forward it *should* be the exception rather than the rule that you want to edit one of the dependencies when working on a site. If that's not the case, and you actually need to edit the themes for almost every job, then maybe we want to rethink the idea of themes? Maybe we should scrap some of the themes and keep their code just directly in the site project itself?

PR #136, opened 2015-03-20
cleaned up forms (mostly by pressing f9)

deadlight:
Done:
Reordered rules alphabetically

QA:
gulp build check nothing is on fire


grahambancroft:
Smooth operator +1

PR #137, opened 2015-03-20
Fix header in demo

anthonydillon:
- Added header to demo.


deadlight:
Header LGTM with fixed links

PR #140, opened 2015-03-20
Add dependency badges from David-dm.org

nottrobin:
QA
--

Check out [the README](https://github.com/nottrobin/vanilla-framework/blob/david-badges/README.md)! See all those awesome badges!


anthonydillon:
Looks good @nottrobin, QA is good! :+1:

PR #141, opened 2015-03-21
Replace Gitter with IRC badge

nottrobin:
I'm not sure Gitter will ever really be much use.

We could instead just have a link to IRC? What do you think?

QA
--

See [README.md](https://github.com/nottrobin/vanilla-framework/blob/irc-badge/README.md)


anthonydillon:
I agree, were more likely to all remain on IRC.

PR #159, opened 2015-04-29
Converted the brand color to light grey with some affordances for light ...

anthonydillon:
...brand colours

# Done
- Converted the brand color of vanilla from ubuntu orange to light grey
- Added header link color var to set the header link colours
- Changed the site logo from white to orange
- Removed some extra tabbing in global settings
- Button retains the ubuntu orange

## QA 
- Run `gulp build`
- Run `google-chrome demo/index.html` and check the demo header is light grey and looks ok?


richmccartney:
@anthonydillon Yeah looks good to me, the navigation dividers look a bit weird now with the colour change mind you. Would be really good if we could work something out based on what the colour of the navigation bar is to change that. 

Otherwise :+1:

PR #160, opened 2015-04-30
Add extra helper and footer classes

grahambancroft:
## Done

Add new helpers
Add additional footer css

## QA

Run gulp build.


richmccartney:
@grahambancroft There are a few linting issues that needs to be fixed. Everything else is great :+1:

PR #161, opened 2015-04-30
Added inline-logos to vanilla

anthonydillon:
# Done
- Added the inline-logos module
- Added new module to demo
- Linted the inline-logos module

## QA
- Run `gulp build`
- Run `google-chrome demo/index.html`
- Check the inline-logos section looks ok


grahambancroft:
Lgtm

PR #164, opened 2015-05-01
Equal heights

anthonydillon:
# Done
- Added the equal-heights module
- Added it to the demo

## QA
- Check `google-chrome demo/index.html` has a equal height section
- Check the heights are equal


nottrobin:
Code looks good. QA tested and works fine. :+1: :kiss:

PR #194, opened 2015-06-16
Add helper classes

anthonydillon:
# Done
- Added all our helper classes to _helpers.scss
- Documented the helper mixins
- Added demos of the helpers to the bottom demo/index.html

## QA
- Lint and check no errors
- Check the demo works
- Check they are all documented in the docs

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -573,7 +573,7 @@
                     </table>
                 </div>
             </div>
-            <div class="row row-hero no-border">
+            <div class="row row-hero">
                 <h2>Page hero</h2>
                 <div class="six-col">
                     <h1>This is a hero unit title</h1>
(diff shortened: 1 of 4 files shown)


deadlight:
Build error: "no mixin named vf-no-margin"

Lint error: "modules/_helpers.scss:421 [W] FinalNewline: Files should end with a trailing newline"

Also it looks like this can't merge cleanly... sorry.

anthonydillon:
Fixed ... I think

deadlight:
It all lints and the docs cover everything pretty clearly.

I strongly oppose the classes with pixel numbers in their names; as discussed fixing them to make them relative to the gutters or column sizes would be an improvement or just removing them entirely (as seems to be the plan).

But... :+1: from me for now

anthonydillon:
Thanks Karl. I agree with you on the pixel classes. We work out a plan tomorrow.

PR #197, opened 2015-06-17
Remove pixel value pull classes from helper

anthonydillon:
# Details:
- This is a partial of this card: https://canonical.leankit.com/Boards/View/111185042/115572517

## Done
- Removed the pixel value pull mixins from the helpers module

### QA
- Check the .pull-left-20, .pull-right-20, pull-left-40 and .pull-right-40 have been removed
- Check `gulp build` has no errors


deadlight:
The build works :+1:

PR #199, opened 2015-06-17
Fix not for small for medium upwards

grahambancroft:
#Done 

.priority-0/.not-for-small-fix - Add display: block; for medium + to counteract display: none; for small.

##QA

Should just need a visual or a canonical.c/ubuntu.c setup. Without display: block; the image stays hidden.

### DIFF
--- a/scss/modules/_helpers.scss
+++ b/scss/modules/_helpers.scss
@@ -322,8 +322,7 @@
   @media only screen and (min-width : $breakpoint-medium) {
     .priority-0,
     .not-for-small {
-      position: relative;
-      left: auto;
+      display: block;
     }
   }
 }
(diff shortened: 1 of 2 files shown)


scss/modules/_helpers.scss
```diff
@@ -322,8 +322,7 @@
   @media only screen and (min-width : $breakpoint-medium) {
     .priority-0,
     .not-for-small {
-      position: relative;
-      left: auto;
+      display: block;
```
anthonydillon: I think a better method would be to wrap the display: none in a max-width media query. This will set all items to display: block if it has priority-0. This is not always desired.

anthonydillon:
Comment in the code and for-medium is also not working correctly if you could fix up with this mp please.

grahambancroft:
All afore mentioned changes have been made

anthonydillon:
Code LGTM, QA ok :+1: will add a follow up card to fix up for-medium component.

PR #201, opened 2015-06-17
Rows and pulls now respect gutter-width

anthonydillon:
# Done
- Updated the pull modules to use the `$gutter-width` var
- Updated the rows module to use the `$gutter-width` var
- Added all the pull modules to the pattern library (demo)
- Removed pull-top-* as there is no easy way to do that

## QA
- Run `gulp build`
- Check the demo contains all the modules in the helpers module
- Check all view ports that the modules perform as expected
- Check rows are acting the same on all view ports

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -661,6 +661,30 @@
                     <div class="for-small">This should only be visble on small screens</div>
                 </div>
             </div>
+            <div class="row">
+                <h3>.pull-bottom-right</h3>
+                <div class="twelve-col">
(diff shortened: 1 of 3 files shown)


deadlight:
LGTM but the medium and small viewports both scroll horizontally for some reason.

Fix that and it's a :+1: from me.

anthonydillon:
That is releated to https://github.com/ubuntudesign/vanilla-framework/issues/200. Thanks

PR #207, opened 2015-06-24
Remove duplicated image styling

anthonydillon:
# Details
- Card: https://canonical.leankit.com/Boards/View/111185042/115595584

## Done
- Removed display block and clear both from the img element
- I found they are styled correctly in reset already to simply removed the extra styling from the media component

### QA
- Run `gulp check` and see there are no errors
- Check the demo, the image layout should still be correct.


grahambancroft:
Lgtm :+1:

PR #212, opened 2015-07-01
Add list ubuntu and list canonical styles

grahambancroft:
#Done

Add ubuntu and canonical list tick styles to demo

##QA

Run ‘gulp build’ then open the demo.

### DIFF
--- a/scss/modules/_lists.scss
+++ b/scss/modules/_lists.scss
@@ -87,7 +87,7 @@
     li {
       background-repeat: no-repeat;
       background-position: 0 .25em;
-      background-image: svgTick($bullet-color);
+      background-image: svg-tick($bullet-color);
       padding-left: 25px;
     }
   }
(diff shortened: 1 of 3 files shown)


anthonydillon:
Looks good. Shame we cant style the fill directly with css. 

With a little manual trimming I've reduced the SVG a lot. Here is the warm bullet and changing the background color is easy to spot and change.

`<svg xmlns="http://www.w3.org/2000/svg" width="14" height="14" viewBox="0 0 14 14"><circle fill="#AEA79F" cx="7" cy="7" r="7"/><path fill="#fff" d="M6.1 10.813L2.41 8.105l1.184-1.613L5.9 8.187l4.393-4.394 1.414 1.414z"/></svg>`

:+1: with the new svg in base.

anthonydillon:
Sorry after checking the trimmed svg's it looks like the dark and canonical ticks are all the same "#5E2750".

anthonydillon:
Can we see if we can have a tick-list which is styled with brand-color using before maybe. Then we only need one and the themes can extend it to add more styles of it.

grahambancroft:
So now there's just one instance of list-ticks (working title). The tick will always be white and the background is based on $brand-color, if however, $brand-color is below a certain lightness threshold the background color will be darkened.

scss/modules/_lists.scss
```diff
@@ -190,3 +207,26 @@
     margin-left: -$gutter-width * 3;
   }
 }
+
+/// Default list tick pattern styles
+/// @group Lists
+%list-ticks {
```
anthonydillon: It looks like this placeholder is only used once. Would it be better to just have this inline where is it used or are you planning to use it in the future?

scss/modules/_lists.scss
```diff
@@ -190,3 +207,26 @@
...
+    background-repeat: no-repeat;
+    background-position: 0 .25em;
+    padding-left: 25px; 
+  }
+}
+
+// this is to get the variable inserted in an inline svg, we have to strip the # and replace with %23
+// it also checks to see if the color is a light color and adjusts the background color accordingly
+@function url-friendly-color($color) {
+  @if (lightness($color) > 50) {
```
anthonydillon: I think you should check that `$color` is a valid color so it doesnt bust, just errors.

anthonydillon:
@grahambancroft There are some lint issues.

anthonydillon:
LGTM :+1: QA'ed and checked the colour changing it working and tick colour will not reach white.

PR #213, opened 2015-07-03
Made the equal-heights module full width to not wrap with site gutters

anthonydillon:
# Details
- Card: https://canonical.leankit.com/Boards/View/111185042/115595595

## Done
- Added width to the equal height container

### QA
- Run `make build`
- Check the equal-heights section of the demo looks ok
- Check different viewports


grahambancroft:
Lgtm :+1:

PR #216, opened 2015-07-21
Fix med main nav horizontal scroll

grahambancroft:
#Done 
Change ‘width’ to ‘max-width’ so the main nav doesn’t break at $breakpoint-medium

##QA
Pull this and run the site locally.
Resize your browser to 768 or less for that matter, and note that the primary nav doesn’t cause horizontal scroll.

##Details 
Card: https://canonical.leankit.com/Boards/View/111185042/116185595

### DIFF
--- a/scss/modules/_header.scss
+++ b/scss/modules/_header.scss
@@ -73,11 +73,7 @@
       border: 0;
 
       @media only screen and (min-width: $navigation-threshold) {
-        width: $site-max-width;
-      }
-
-      @media only screen and (max-width : $breakpoint-medium) {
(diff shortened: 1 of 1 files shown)


deadlight:
LGTM :+1:

scss/modules/_header.scss
```diff
@@ -73,7 +73,7 @@
       border: 0;
 
       @media only screen and (min-width: $navigation-threshold) {
-        width: $site-max-width;
+        max-width: $site-max-width;
```
anthonydillon: Should we use $navigation-threshold here?

PR #220, opened 2015-07-28
Refactor list-ticks to add border

grahambancroft:
# Done
Refactor list-ticks to add border. Re-use .list styles as placeholder.

## QA 
Pull then gulp build and check that .list and .list-ticks look okay in the demo.


anthonydillon:
Great stuff :+1:

PR #222, opened 2015-07-29
Refactor breadcrumb to remove right margin

grahambancroft:
# Done
Removed the right margin as it’s not the same as on live.

## QA
Pull the gulp build then have a look at the demo and compare with live


deadlight:
:+1:

PR #274, opened 2016-01-13
Add top padding to list-logos to better replicate live

grahambancroft:
# Done
Add top padding to list-logos to better replicate live

## QA
Run 'gulp build' then check the demo, there should be a greater gap between the heading and the list. The issue is more noticable on ubuntu website, this change mirrors live a bit better.


anthonydillon:
LGTM :+1:

PR #275, opened 2016-01-13
Add align-vertically helper

grahambancroft:
# Done
Add align-vertically helper

## QA
Run gulp build, pop on over to demo/index.html and have a lookie at .align-vertically there in the helper section.

## Details
Card: https://canonical.leankit.com/Boards/View/111185042/115594581

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -763,7 +763,7 @@
             <div class="row equal-height">
                 <div class="eight-col">
                     <h2>Vertically and horizontally center an element</h2>
-                    <p class="intro">Sets the element to center an inner element vertically. Has to be used in conjunction with equal-height and if it's an image it'll need align-center.</p>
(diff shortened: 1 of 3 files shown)


anthonydillon:
We need to be able to use align-vertically without relying on equal heights. As we do across the sites.

grahambancroft:
I’m struggling to find an example of where we use .align-vertically without .equal-height, could you point me in the right direction.

anthonydillon:
Code and QA ok but the documentation is missing. Can you update this please.

anthonydillon:
What is the status of this PR?

grahambancroft:
I’ve added align-vertically to the docs so this is ready for re-review

anthonydillon:
Looks good thanks for the doc comments :+1:

Issue #253, opened 2015-11-24
.touch-border doesn't touch the border in small screens


yaili:
See small gutter under placeholder image

![screenshot 2016-01-15 12 45 32](https://cloud.githubusercontent.com/assets/505570/12353645/7a90cdf6-bb88-11e5-8905-855908c70358.png)

PR #281, opened 2016-01-15
Ensure `.touch-border` works on small screens

barrymcgee:
## Done

The rules to float the element and collapse the gutter was previously only kicking in at the medium breakpoint. I've now moved the float out of this media query so it applies by default.

I also refactored the rule to move the media queries inside the selectors to remove the need to duplicate the rule selectors.

## QA

Run `gulp sass` and view the `demo/index.html` page at a width less than 768px. you should see under the heading _.touch-border_ that the placeholder image sits flush on top of the border. (see below)

screenshot 2016-01-15 12 56 23

Issue: #253

### DIFF
--- a/scss/modules/_helpers.scss
+++ b/scss/modules/_helpers.scss
@@ -104,18 +104,15 @@
 ///   </div>
 @mixin vf-touch-border {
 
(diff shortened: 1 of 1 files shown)


anthonydillon:
LGTM :+1:

Issue #265, opened 2015-12-17
[iPhone 6 - iOS] Search input has very rounded corners


anthonydillon:
![bs_realios_mobile_iphone 6-8 0 1](https://cloud.githubusercontent.com/assets/1413534/11877329/8afd8e26-a4e6-11e5-8179-1c7c0461aec5.jpg)

PR #282, opened 2016-01-15
Remove rounded corners from search input example

barrymcgee:
## Done
Removed Safari pseudo-class namespace.. It would appear that this pseudo-class is now deprecated for iOS9 although I can't find any evidence of that online.

## QA
Check out this branch and load `demo/index.html` - the search input example should no longer have rounded corners. Like so;

file 15-01-2016 14 47 37

Fixes: #265

### DIFF
--- a/scss/modules/_forms.scss
+++ b/scss/modules/_forms.scss
@@ -43,8 +43,9 @@
       display: none;
     }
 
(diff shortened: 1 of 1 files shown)


richmccartney:
+1 Looks good

PR #283, opened 2016-01-15
Correct `.for-medium` helper so only display @>768

barrymcgee:
Previously this helper was hiding any content once the large screen breakpoint was met. This fix means the content will be hidden on small screens and only appear once the medium breakpoint is hit.

Fixes: #254


richmccartney:
+1 LGTM

PR #284, opened 2016-01-18
Wrap <textrea> in <form> so they display correctly

### LINKED ISSUE #255: Textareas should be full width in small screens (like other input fields)
In small screens the input fields are full width but the textareas are not, is this on purpose? Eg. Chrome on Android

barrymcgee:
# Details
Textarea elements are styled with a form element so need to be nested inside a <form> element to display consistently like text fields.

## Done
The textarea examples in `demo/index.html` are now wrapped in <form> elements.

## QA
Load up the demo page and check that textarea widths reflect that of text field widths - 100% on small screens, 50% after the tablet breakpoint.

Fixes #255

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -381,13 +381,17 @@
             </div>
             <div class="row no-border">
                 <div class="six-col">
-                    <textarea>Textarea...</textarea>
+                    <form>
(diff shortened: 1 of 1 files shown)


anthonydillon:
LGTM :+1:

Issue #294, opened 2016-02-02
`.row-grey` should not have dotted borders


barrymcgee:
A row with a grey background should not have the dotted black border that vanilla rows have.

PR #295, opened 2016-02-02
Amend .row-grey styling to remove dotted borders

barrymcgee:
## Done

- Add styles to remove dotted borders from row-grey sections
- Add row-grey section to the demo page

## QA

Run code, check demo and see that grey row section should not have a dotted border.

## Notes

Due to CSS' inability to select a preceding element, I'm raising row-grey's margin-top by 1 negative pixel to cover the preceding elements bottom border. Thoughts?

I've added `border: 0` to `.row-grey` when there is already a `.no-border` helper. Should row-grey also contain a rule to zero borders? Personally, I think yes but maybe not? 

Fixes: #294

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -707,6 +707,12 @@
                     <div class="right">right content</div>
                 </div>
             </div>
+            <div class="row row-grey">
+                <h3>.row-grey</h3>
+                    <div class="twelve-col">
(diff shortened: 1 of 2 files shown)


richmccartney:
@barrymcgee +1 apart from my one comment.

PR #311, opened 2016-03-01
Change Ubuntu orange

grahambancroft:
## Done

Change ubuntu orange to the new colour

## QA

Run ‘gulp build’ in the appropriate place and make sure the oranges are #e95420 and not #dd4814 in the demo

## Details

Card: https://trello.com/c/1Z3V717W


grahambancroft:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;1Z3V717W&#x2F;409-update-ubuntu-orange"> Update ubuntu orange</a>

richmccartney:
Looks good 2 points though:

- Can we update the external icon SVG to be the same colour or somehow change the fill colour in our Scss?
- As we're using the same colour in two places should it be a separate variable which can be shared?

grahambancroft:
@richmccartney I’ve updated the external link icon and also the logo. I’ve also created a new variable, let me know what you think of the name.

anthonydillon:
We removed all Ubuntu branding from Vanilla framework. Shouldn't the links be brand colour and the ubuntu orange be set as the brand color in the ubuntu-vanilla-theme?

grahambancroft:
I thought the links and buttons were made to be ubuntu orange as a conscious decision by the design team? If it’s being used on all the sites, which it probably is, it could possibly stay in, maybe?

Also, if we’re removing all ubuntu branding shouldn’t we ditch the logo too?

anthonydillon:
Currently the links are coloured with the `$link-color` and buttons are
coloured with `$primary-button-color`. These two variables should be
updated instead of adding a new variable which is only used in a few
themes. The cloud themes are moving away from the ubuntu orange so there
will be a redundant variable in that case.

richmccartney:
I still feel if we are using colours in multiple places we shouldn't be duplicating it. Can we think about having colour name variables which are a standard and nothing to do with ubuntu / canonical. 

I'd rather know that I have an ```$orange``` variable and I can use it all over than trying to copy and paste a hex value same as we have called things ```$cool-grey```, ```$white``` and ```$light-grey````?

grahambancroft:
I think if it’s Ubuntu orange the variable should be called that, just ```$orange``` could cause confusion in the future, maybe, if we change ubuntu orange again, for instance?

anthonydillon:
Ubuntu Orange is a brand colour. Even if it is used by two themes
(differently) then I think that should be at the theme level. Is there a
problem we are resolving by having Ubuntu orange in Vanilla over having it
in the ubuntu-vanilla-theme as the brand-color and in canonical-theme as
the link-color?

grahambancroft:
I think Richard’s original question was as we have two variables that use the same color, shouldn’t that color be a variable.

anthonydillon:
For me I don't think so as we can hard set the link color to the new hex
and then reuse that var to set the button color. The main use case would be
someone overriding both colours anyway.

Do what you think is best for now. We can revisit if we need to. I am just
against littering the framework with specific brand colours variables which
become redundant outside of Ubuntu and Canonical themes.

grahambancroft:
I’ve gone down the 'set the link color to the new hex and then reuse that var to set the button color’ route

anthonydillon:
Thanks. Nice to discuss it here but if anyone feels the other method would
be better lets discuss and make sure were all on the same page.

grahambancroft:
Could do with a :+1:  on this so I can put it to bed

anthonydillon:
Perfect thank you :+1:

Issue #312, opened 2016-03-04
Include full width strips


barrymcgee:
## Context

Phone and Tablet pages now require full width strip rows and these styles are being duplicated in different codes so it makes sense to move them upstream.

## Expected result

Bespoke styles should not be needed downstream to enable a full width row.

PR #313, opened 2016-03-04
Move Full Width style rules upstream

barrymcgee:
## Done

- Added rules so that rows can stretch the full width of the screen with an inner wrap to constrain content to the the max-width of the site.

## QA
Check code makes sense.

## Details

Fixes: #312


deadlight:
It might be a good idea to add some full-width examples to the demo but these changes look good to me. :+1:

PR #326, opened 2016-04-05
Nest row--grey in row

anthonydillon:
## Done

- Moved row--grey to accompany row so it has higher specificity 

## QA

- Run `gulp build` and check the .grey--row section in the demo looks good

### DIFF
--- a/scss/modules/_rows.scss
+++ b/scss/modules/_rows.scss
@@ -42,13 +42,13 @@
       background: $cool-grey;
       color: $white;
     }
-  }
 
-  .row-grey, // retain to ensure backwards compatibility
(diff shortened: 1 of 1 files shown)


richmccartney:
@anthonydillon Two questions should we change the ```--light``` and ```--dark``` modifiers to follow the same format you've done here? And secondly should we have them in the demo?

Apart form that +1

anthonydillon:
@richmccartney thanks. `--light` and `--dark` are strips where as this is the legacy `row` naming ... We need to talk about rows vs stripes in dev catch up today. As I'm not adding anything, I'll land and add an issue to change when we decide.

PR #333, opened 2016-04-09
Refactor list ticks contrast function

grahambancroft:
## Done

Refactor contrast friendly function to only run if $bullet-color isn’t set

## QA

Run gulp build. Head over to the demo. If $bullet-color doesn’t have a value in _global-settings the contrast friendly function in _lists will revert to using $brand-color. To make sure this is true add a value to $bullet-color which should result in the bullet ticks will use that color.

## Details
Card: https://trello.com/c/jaADfvRV
Issue: https://github.com/ubuntudesign/ubuntu-vanilla-theme/issues/168

### DIFF
--- a/scss/modules/_lists.scss
+++ b/scss/modules/_lists.scss
@@ -68,7 +68,7 @@
     @extend %list-ticks;
     
     li {
-      background-image: svg-tick($bullet-color);
+      background-image: svg-tick($brand-color);
     }
   }
 }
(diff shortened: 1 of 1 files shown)


grahambancroft:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;jaADfvRV&#x2F;755-ticks-are-too-dark"> Ticks are too dark</a>

PR #337, opened 2016-04-11
Add inline-block to grid element

anthonydillon:
## Done
- Added `inline-block` to grid elements

## QA
- `gulp build`
- Check the grid section in the demo looks the same


deadlight:
Looks good :+1:

PR #339, opened 2016-04-11
fix step-lists at smaller viewports

deadlight:
## Done

Modified the step-lists at small and medium viewports.

## QA

```gulp build```
Load the demo in the browser and check that step-lists look good at all viewports


nottrobin:
Parfait! :+1:

PR #340, opened 2016-04-11
Make all settings to !default

anthonydillon:
## Done
- Make all settings !default so they can be overriden

## QA
- `gulp build`
- Check the fonts load of in the demo


deadlight:
Code looks good :+1:

PR #342, opened 2016-04-12
Font size fix

grahambancroft:
## Done

Make list items font size match paragraph size of .875em (14px) on small

## QA

Run gulp build then open the demo and make sure the list items are .875em (14px) on small

## Details

Issue: https://github.com/ubuntudesign/www.ubuntu.com/issues/380

### DIFF
--- a/scss/modules/_typography.scss
+++ b/scss/modules/_typography.scss
@@ -339,7 +339,8 @@
       text-transform: uppercase;
     }
 
-    p {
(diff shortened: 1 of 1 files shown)


anthonydillon:
Makes sense. I assume line-height is a em already so we don't need to change that? :+1:

grahambancroft:
It’s a unitless line height set at 1.5

Issue #358, opened 2016-05-11
Buttons that link to external sites should include external icon


barrymcgee:
## Context
Buttons that link to external sites should include an icon within the button boundaries after the text to indicate to the user they are to be taken to a different site.

## Related ticket
https://github.com/ubuntudesign/www.ubuntu.com/issues/639

anthonydillon:
The way we do that at the moment is to wrap the link text in a span with an
external class. This is because on a button that is display block or mobile
buttons which are 100% width. We want the icon on the next to the text not
aligned to the right on the button itself.

Hope that make sense. If not let me know.

PR #359, opened 2016-05-12
Add .external button pattern

barrymcgee:
## Done
- Added external button pattern to Buttons section of demo
- Tweaked style to ensure white external icon is displayed on primary buttons

## QA
- Run code, check that buttons on demo reflect screenshot below

## Details
Fixes: #358
Related: https://github.com/ubuntudesign/www.ubuntu.com/issues/639

## Screenshots
screenshot 2016-05-12 09 42 43

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -245,6 +245,12 @@
             </div>
             <div class="row no-border">
                 <div class="six-col">
+                    <button class="button--primary" href="#"><span class="external">External button</span></button>
+                </div>
+                <div class="four-col prepend-two last-col">External button</div>
+            </div>
+            <div class="row no-border">
+                <div class="six-col">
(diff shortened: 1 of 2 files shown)


deadlight:
Looks really good, nice and simple.

One minor design issue is that external button and external secondary button look different (as seen in the screenshot you posted).

I think that the 100% 3px should be 100% top. Check this with a designer but the existing inline links on ubuntu.com have the icon aligned to the top of the space rather than the middle (like the external secondary button in your screenshot and not the external button).

barrymcgee:
@deadlight Good spot - I checked this will Lil but forgot to remove the override.

deadlight:
No worries.

Looks perfect now. Tested at all viewports.

:+1:

PR #361, opened 2016-05-16
Add max-width to grid list images

barrymcgee:
## Done
Add a max-width to grid list images to make sure they don't outgrow their place on tablet views

## QA
Pull down, check demo page across breakpoints


deadlight:
QA'd in browser. All looks good.

PR #362, opened 2016-05-17
Change em to rem on p and li defaults

barrymcgee:
## Done

This will ensure that `<p>` tags inside `<li>` tags or vice-versa are not scaled down twice as em sizing is relative to it's parent.

## QA

Code only.

## Details

Fixes: https://github.com/ubuntudesign/www.ubuntu.com/issues/375


grahambancroft:
Lgtm 👍

PR #371, opened 2016-06-06
Add missing .col grid class

anthonydillon:
## Done
Added the missing `.col` class to the grid. Which is used for grid spacing without effecting the grid itself.

## QA
- Run `gulp build`
- Go to the demo and scroll to the Grid section
- Check that the `.col` block at the top has all the same styling as a grid column without a width

## Details
- Fixes https://github.com/ubuntudesign/vanilla-framework/issues/301

### DIFF
--- a/scss/modules/_grid.scss
+++ b/scss/modules/_grid.scss
@@ -49,6 +49,11 @@ $single-column-percent: $single-column-width / $columns-max-width * 100%;
   @include generate-append-classes($columns, 'append-', '');
   @include generate-push-classes($columns, 'push-', '');
   @include generate-pull-classes($columns, 'pull-', '');
+
+  // The base grid class without a width for adding vertical rhythm.
+  .col {
+    @extend %vf-column;
+  }
 }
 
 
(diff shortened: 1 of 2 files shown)


scss/modules/_grid.scss
```diff
@@ -49,6 +49,11 @@ $single-column-percent: $single-column-width / $columns-max-width * 100%;
   @include generate-append-classes($columns, 'append-', '');
   @include generate-push-classes($columns, 'push-', '');
   @include generate-pull-classes($columns, 'pull-', '');
+
+  // The base grid class without a width for adding vertical rhythm.
+  .col {
+    @extend %vf-column;
+  }
```
deadlight: This has exactly the same css as .twelve-col.
1) Do we need both?
2) Shouldn't the rules be merged?

deadlight:
@anthonydillon I need some clarification for some choices - see comment in code.

anthonydillon:
@deadlight You mean you need a docstring which explains a use case?

anthonydillon:
Killing as twelve-col does the same thing.

Issue #218, opened 2015-07-22
Disabled form inputs not visually different from normal inputs.


grahambancroft:
The disabled checkbox and radio inputs are visually the same as enabled.

yaili:
Has this been done? They look just slightly different now.

yaili:
Related to #303 and #239

Issue #303, opened 2016-02-19
Missing disabled button style


yaili:
I can't find a button--disabled style in the demo. Is there one?

richmccartney:
There isn't one currently in the framework but we should also add ```disabled="disabled"``` into the mix. It's more likely a lot of forms will use that format to disable any form elements.

yaili:
Yes, that too. What about the coming soon button here? http://www.ubuntu.com/phone/devices

Issue #239, opened 2015-11-24
Disabled input should show disabled pointer on hover


yaili:
This should happen in text fields, checkboxes, buttons, etc.

PR #375, opened 2016-06-10
Refactor disabled elem styles to include tick elems

barrymcgee:
## Done

Abstracted disabled styles into placeholder variable that can be used for tick elems.

## QA

Run code, check that all disabled elements on the /demo page have 50% opacity and when rolled over with the mouse, the cursor changes to be disabled.

## Details

Fixes: #218 / #239 / #303


deadlight:
works great :shipit:

Issue #188, opened 2015-05-29
input with error message: retain colour when field is focussed


jamiedawsonyoung:
Currently – 

When a form field with a error status is clicked into it loses the colour associated with it. Can we keep the colour until the error is resolved…

barrymcgee:
@jamiedawsonyoung Long wait for this one but I've just proposed a fix. You should see issues like this addressed much more rapidly in future. 🤘

PR #376, opened 2016-06-10
Persist input state styles on focus

barrymcgee:
## Done

Previous to this change, when a state class was added, the corresponding border style would appear but would disappear when the input elem gained focus due to the focus style not being explicitly set and the default fallback of grey showing instead.

## QA

Run code, navigate to /demo and click through each form elem and observe that border style persists on focus.

## Details

Fixes: #188


deadlight:
Tested and working. :+1:

Issue #352, opened 2016-05-05
Add heading classes


yaili:
We agreed to include the classes from **.heading-1** through to **.heading-6** to Vanilla.

barrymcgee:
This is a separate issue to #234

yaili:
Apparently these aren't in Vanilla. Close again if wrong.

anthonydillon:
I know I had seen it https://github.com/ubuntudesign/vanilla-framework/pull/377. 

Let's find out how this was dropped from master. I imagine it was merged into master while working on a future feature branch for v1. If that is the case, then we didn't merge the feature branch but force pushed it. We need to check what else was lost.

yaili:
@anthonydillon Who should I assign to?

anthonydillon:
I'm happy to do it. Merry Christmas!

PR #377, opened 2016-06-10
Add heading classes 1-6

barrymcgee:
## Done

I have purposefully added these classes in addition to the raw element selectors. While this restricted the ability to apply a `heading-1` class to a `h2` element, it will ensure backwards compatibility and that this change is not a large breaking change.

The larger sweep of these styles, decoupling them from raw element selectors will be included in time for v1.0

## QA

Run code, navigate to /demo and check that heading 1-6 are displayed.

## Details

Fixes: #352

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -479,12 +479,12 @@
             <div class="row" id="typography">
                 <h2>Typography</h2>
                 <div class="twelve-col">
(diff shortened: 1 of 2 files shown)


deadlight:
Good stuff. Tested and working. :+1:

Issue #292, opened 2016-01-27
Demo page missing id's for deep linking


anthonydillon:
Each section of the demo page should have an id to help people deep link to each component.

PR #378, opened 2016-06-10
Add deep linking id to each demo section

barrymcgee:
## Done

Add deep linking id to each demo section

## QA

Navigate to /demo and check that you can jump to each section using the relevant id.

## Details

Fixes: #292


deadlight:
All good :shipit:

Issue #418, opened 2016-07-27
Change 'overrides' to 'utilities' in the scss naming


barrymcgee:
As overrides is a subset of "Trumps" in ITCSS world and not everything in this layer is technically an override, it makes more sense to name this layer slightly more generically - for now, @yaili and I have agreed on "Utilities".

Issue #433, opened 2016-08-01
Metalsmith syntax error breaks Travis build


barrymcgee:
After a bit of investigation, I think this is due to a linting error with Metalsmith so have raised an issue in the Metalsmith project - https://github.com/metalsmith/metalsmith/issues/241

PR #434, opened 2016-08-01
Rename layers of ITCSS

barrymcgee:
**Note**: Travis build in `v1-dev` branch due to #433 - hopefully should be sorted this week.

## Done

- Renamed 'Overrides' layer to 'Utilities'
- Renamed ''Tools' layer to 'Layout'

We first decided to renamed Overrides to Utilities as utils are actually as overrides as technically a subset of Utilities. This then meant that it was confusing to have Tools and Utilities as they are very similar to a lot of people. As the Tools are all actually layout based, it makes more sense to rename this layer 'Layout'.

## QA

Pull code and run `gulp develop` to check that there are no errors.

## Details

Fixes: #418

### DIFF
--- a/docs/partials/sidebar.hbt
+++ b/docs/partials/sidebar.hbt
@@ -11,9 +11,9 @@
     {{/each}}
 </ul>
 
-<h3>Tools</h3>
+<h3>Layout</h3>
 <ul class="sidenav-list">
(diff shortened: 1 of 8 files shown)


anthonydillon:
:+1: thanks

PR #436, opened 2016-08-01
Bump Travis version of node to 0.12

barrymcgee:
## Done

Bumped Travis version to 0.12 to hopefully fix build error?

## QA

Check that this PR passes Travis build?

PR #439, opened 2016-08-02
438 - Remove metalsmith

barrymcgee:
## Done

- Removed all Metalsmith artifacts from this repo
- Split Gulp tasks out into separate files

## QA

Sanity check code and test each Gulp command does as expected. Running `gulp` should return list of available commands.

## Details

Fixes: #436


deadlight:
Code changes all look good and gulp build passed

Issue #388, opened 2016-07-23
Audit reset


barrymcgee:
## Context

- Is the reset using the latest version of normalise? 
- Is it overkill given updated browser support?

## Actions

- [x] Browser testing
- [x] Accessibility testing
- [x] Sass linting
- [x] Update documentation
- [x] Convert sizing units to rems
- [x]  Ensure only elements are styled directly

PR #444, opened 2016-08-05
Update reset to use normalize.css v4.2.0

barrymcgee:
## Done

- Updated reset to drop normalize.css wholesale into `.vf-reset` mixin.
- Disabled scss linting around the `.vf-reset` mixin

## Rationale

- Normalise is essentially a 'core' vendor file that we shouldn't overwrite/edit directly or we make it difficult to upgrade easily in the future. 
- Previously, this file had included some opinionated styling including our color variables to set default styling on some patterns. These reset styles should be encapsulated within the pattern styling even if that means some duplication between pattern styling files.
- I've disabled sass linting around this`.vf-mixin` as copying normalize _as is_ will make it much easier to identify changes in future upgrades. 

## QA

Sanity check code.

## Details

Fixes: #388


scss/_base_reset.scss
```diff
@@ -1,408 +1,473 @@
...
+  }
+
+  /**
+   * Remove the border on images inside links in IE 10-.
+   */
+
   img {
-    border: 0;
-    height: auto;
-    max-width: 100%;
```
anthonydillon: We need to retain this styling outside of the reset as its relied on heavy to hold together layouts.

anthonydillon:
Thanks for this @barrymcgee good work! One important comment to address please.

barrymcgee:
Ok, I've now added a `_base_images.scss` file to hold styling directly on images.

Issue #406, opened 2016-07-23
Audit tools - Clearfix


barrymcgee:
## Context

Sanity check source code.

## Audit checklist

- [x] Bemify Sass 
- [x] Convert sizing units to rems
- [x] Namespace styling to `vf-`
- [x] Browser testing
- [x] Accessibility testing
- [x] Sass linting
- [x] Update documentation

PR #445, opened 2016-08-05
Simplify .clearfix layout rule

barrymcgee:
## Done

- This rule was set as %placeholder which was then being called inside a mixin immediately after it which is pointless?
- We can simplify the rule itself as we no longer need to support IE 6/7 and older versions of Opera.

Ref: http://cssmojo.com/the-very-latest-clearfix-reloaded/

Fixes: #406

### DIFF
--- a/scss/_tools_clearfix.scss
+++ b/scss/_tools_clearfix.scss
@@ -1,17 +1,9 @@
-%vf-clearfix {
-  display: block;
+// Force a parent to clear floated children
 
-  &:after {
-    clear: both;
-    content: '\0020';
-    display: block;
-    height: 0;
-    overflow: hidden;
-    visibility: hidden;
-  }
-}
-
-// Hard and soft clearing classes
 @mixin vf-clearfix() {
-  .clearfix { @extend %vf-clearfix; }
+  .clearfix:after {
+    content: '';
+    display: block;
+    clear: both;
+  }
 }


scss/_tools_clearfix.scss
```diff
@@ -1,17 +1,9 @@
...
-    overflow: hidden;
-    visibility: hidden;
-  }
-}
-
-// Hard and soft clearing classes
 @mixin vf-clearfix() {
-  .clearfix { @extend %vf-clearfix; }
+  .clearfix:after {
+    content:"";
```
anthonydillon: Believe the linter requires single quotes.

scss/_tools_clearfix.scss
```diff
@@ -1,17 +1,9 @@
...
+    display:block;
+    clear:both;
```
anthonydillon: These rules need a space after the :

anthonydillon:
Few comments inline. Also this will need a quick QA to confirm the simplified version  still works as before.

barrymcgee:
@anthonydillon Sorry, not fixed linting errors.

I have tested this in isolation myself but I'm still waiting on Robin's on vanillaframework.io to complete before setting up a workflow where a dev can test every change in Vanilla on the Pattern Lib site locally. I don't want it to block me in the meantime though.

anthonydillon:
Thanks :+1:

Issue #429, opened 2016-07-28
Audit .pull-right / left


barrymcgee:
These are deprecated as their use case is too specific for inclusion in Vanilla.

PR #446, opened 2016-08-05
Remove superfluous pull utils

barrymcgee:
## Done

Remove superfluous pull utils as their use cases are too specific for Vanilla.

## QA

Sanity check code.

## Details

Fixes: #429


deadlight:
Code looks good :+1:

Issue #430, opened 2016-07-28
Audit visibility classes


barrymcgee:
To be deprecated;

```css
.priority-0
.not-for-small
.for-tablet
.for-medium
.for-mobile
.for-small
```

These should be replaced with clearer classes which only do what they state - that is toggle a visibility within one breakpoint:

```css
.u-hidden--small
.u-hidden--medium
.u-hidden--large

.u-visible--small
.u-visible--medium
.u-visible--large
```

anthonydillon:
`.priority-0` is replaced with `.for-small`

anthonydillon:
What would visible--small do? Would that only show on small screens and be
hidden on all others? If that is the case what does visible--medium do?
Hidden on small and large screens? Would this require visible--small and
visible--medium to display on both?

barrymcgee:
Yes - `.u-hidden--small` would only be shown on small resolutions as defined by the small breakpoint.

To display on both small and medium breakpoints but not large, the classes needed would be - `<div class="u-visible--small u-visible--medium"></div>`

This may seem verbose but again, to minimise confusion, we have chosen being explicit over brevity. 

In practice, these utils should be used sparingly as most patterns will have their visibility defined inherently.

anthonydillon:
Thanks for explaining. Originally we used a mobile first approach so
for-medium would display on small and medium screens but I like this
verbose method.

barrymcgee:
@anthonydillon Of course, you could also just use the inverse util to display on small/medium but not large - `<div class="u-hidden--large"></div>`

As a rule we should always build mobile first but this may be the only use case where an exception is warranted to use a max-width media query.

anthonydillon:
Good point, like it :+1:.

barrymcgee:
Addressed by #447

PR #447, opened 2016-08-05
Add visibility utility classes

barrymcgee:
## Done

Added visibility utility classes. These allow a develop to positively state whether an element should be visible or hidden within a specific breakpoint.

## QA

- Sanity check code
- Code should be identical in this demo which should show it working; http://codepen.io/barrymcgee/pen/jAZxZp
- Divs with `.u-hidden--*` should be hidden when in `$breakpoint-*`
- Divs with `.u-visible--*` should be visible when in `$breakpoint-*`

## Details

Fixes: #430


anthonydillon:
LGTM :+1: but can you add a space between : and breakpoint in all media queries please.

PR #452, opened 2016-08-09
Remove .no-border

barrymcgee:
## Done

Remove the util class `.no-border` as we'll be removing the presumption that certain elements have borders by default.

## QA

- Check that I've removed all instances
- Check that I haven't removed anything I should not have.


anthonydillon:
LGTM :+1:

PR #454, opened 2016-08-10
Hygiene: Remove ‘settings’ from file titles

barrymcgee:
## Done

Removed the word 'settings' from the title front-matter in `_settings_&.md` files as these are surplus to requirements when they are pulled into the Pattern Lib sidebar as they sit under a sidebar heading which says 'Settings'.

## QA

- Sanity check code
- Fire up vf.io project and check sidebar does not contain items suffixed with ' settings'


anthonydillon:
LGTM :+1:

Issue #409, opened 2016-07-23
Audit tools - row


barrymcgee:
## Context

Sanity check source code.

* Row should not conflate structure and design. 
* Row grey should be deprecated
* Row hero should be deprecated
* Hero should be abstracted out into a separate pattern.
* no-border should be removed - row by default should not have a border

`.no-margin-bottom` and `.no-padding-bottom` utils were created as a sticky-plaster to fix issues with collapsing and double bottom padding/margins on block elements.

This should be investigated more holistically possibly being smarter with the grid.

## Audit checklist

- [ ] Bemify Sass 
- [ ] Convert sizing units to rems
- [ ] Namespace styling to `vf-`
- [ ] Sass linting
- [ ] Browser testing
- [ ] Accessibility testing
- [ ] Update documentation

PR #455, opened 2016-08-10
Cleanup ‘.row’

barrymcgee:
## Done

The class `.row` should be used purely as a layout construct to contain columns of the grid. This changeset removed the verbose styling for `.row` plus related variants and moves the now lean `.row` class into with grid styling.

Note: Some of the styling for light/dark backgrounds which has previously been attached to row variants will be used on the new pattern `.strip` in due course.

## QA

Run code, check that `.row` behaves as described in code and as in this demo: http://codepen.io/barrymcgee/pen/qNLNVV

## Details

Fixes: #409


anthonydillon:
Conflict, sorry.

deadlight:
Tested and QA looks good. :+1:

Issue #423, opened 2016-07-28
Audit .accessibility-aid


barrymcgee:
* `.off-left` should be deprecated

* `.accessibility-aid` should be renamed to `.u-invisible`

PR #459, opened 2016-08-10
Abstract .accessibility-aid as .u-off-screen

barrymcgee:
## Done

- Refactored `.accessibility-aid` as `.u-off-screen` and moved into separate util file.
- Updates docs

## QA

- Sanity check code
- Pull into vf.io pattern lib and check docs load as expected

## Details

Fixes: #423


deadlight:
All tested and looking great

Issue #414, opened 2016-07-27
Audit vertically centre tool


barrymcgee:
## Context

- This should not also horizontally centre elements.

Related: #336, #322 

## Audit checklist

- [ ] Bemify Sass 
- [ ] Convert sizing units to rems
- [ ] Namespace styling to `vf-`
- [ ] Sass linting
- [ ] Browser testing
- [ ] Accessibility testing
- [ ] Update documentation

PR #468, opened 2016-08-17
Add vertical center util

barrymcgee:
## Done

Added a vertical centering util to that the equal height util can be simplified to only make two siblings equal height.

## QA

- Sanity check code. 
- I avoided adding an example in the docs of vf.io to demo this util as it requires a lot of scaffolding CSs to demo. I want to think about a more elegant way to demo utils like this, perhaps using Codepen embeds or similar.

## Details

Fixes: #414

### DIFF
--- a/scss/_utilities_vertically-center.scss
+++ b/scss/_utilities_vertically-center.scss
@@ -1,0 +1,8 @@
+// Vertically align
+@mixin vf-u-vertically-center {
+
+  .u-vertically-center {
+    display: flex !important;
+    align-items: center!important;
+  }
+}
--- a/scss/_utilities_vertically-center.scss
+++ b/scss/_utilities_vertically-center.scss
@@ -1,8 +1,10 @@
 // Vertically align
 @mixin vf-u-vertically-center {
 
+  // scss-lint:disable ImportantRule
+
   .u-vertically-center {
     display: flex !important;
-    align-items: center!important;
+    align-items: center !important;
   }
 }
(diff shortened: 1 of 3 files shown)


scss/_utilities_vertically-center.scss
```diff
@@ -0,0 +1,8 @@
+// Vertically align
+@mixin vf-u-vertically-center {
+
+  .u-vertically-center {
+    display: flex !important;
+    align-items: !important;
```
anthonydillon: This needs a value.

anthonydillon:
Was a decision made to detach vertically-align from equal-height?

scss/_utilities_vertically-center.scss
barrymcgee: D'oh

barrymcgee:
Yes - they perform two distinct functions so should be two distinct utils. They can then be used in conjunction if needs be.

scss/_vanilla.scss
```diff
@@ -71,6 +67,12 @@
   @include vf-tables;
   @include vf-inline-logos;
   @include vf-equal-heights;
+  // Base
+  @include vf-b-blockquotes;
+  @include vf-b-code;
+  @include vf-b-links;
+  @include vf-b-media;
+  @include vf-b-reset;
```
anthonydillon: Why are you adding all these in this branch?

anthonydillon:
Looks good and QA'ed in Codepen (http://codepen.io/anthonydillon/pen/OXqYBW) :+1:

PR #473, opened 2016-08-23
Add strip pattern

### LINKED ISSUE #351: Create 'strip' tool to enable full width background changes
Currently, VF uses variants on 'row' to add light, dark, grey backgrounds to full width pages. 

This should be abstracted into its own tool named "strip" - 

* strips should be full width block constructs
* strips should never be inside rows
* strips should have white backgrounds by default but have two variants:
 - light (light grey background)
 - dark (dark grey background)

barrymcgee:
## Done

- Added `.strip` pattern to allow full-width container.

## QA

- Sanity check code and docs.

## Details

Fixes #351

### DIFF
--- a/docs/patterns/_strip.md
+++ b/docs/patterns/_strip.md
@@ -1,0 +1,16 @@
+---
+collection: patterns
+title: strip
+---
+
+The strip pattern provides a full width strip container in which to wrap a row.
+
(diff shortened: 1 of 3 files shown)


deadlight:
Code looks good

PR #474, opened 2016-08-23
Convert code pattern to use single class modification

### LINKED ISSUE #463: All modifiers should be extended using single class modification convention
We should convert any instances where two classes are required to modify a pattern, such as:

### Multi classes modification
```html
<a href="" class="button">...</a>
<a href="" class="button button--positive">...</a>
```

..to a single class modification convention, such as:

### Single classes modification
```html
<a href="" class="button">...</a>
<a href="" class="button--positive">...</a>
```

barrymcgee:
## Done

Change code pattern to use single class modification convention

## QA

- Sanity check code
- Test with vf.io

## Details

Fixes #463

### DIFF
--- a/docs/patterns/_code-snippets.md
+++ b/docs/patterns/_code-snippets.md
@@ -23,13 +23,13 @@ Code snippets should be used when presenting the user with a small snippet of co
 
 
 ```html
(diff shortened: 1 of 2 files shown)


deadlight:
Good stuff; all good.

PR #480, opened 2016-08-24
Refactor .inline-logos to .p-image-grid 

### LINKED ISSUE #403: Audit inline logos
## Context

- Sanity check source code.
- Remove examples of specific logs and replace with generic asset
- Should possible be renamed to block grid or similar to remove explicit use case of logos, could be used for any list of images

## Audit checklist

- [ ] Browser testing
- [ ] Accessibility testing
- [ ] Sass linting
- [ ] Update documentation
- [ ] Convert sizing units to rems
- [ ] Namespace styling to `vf-`

barrymcgee:
## Done

- Simplified this pattern and removed superfluous styling
- Remove focus on logos to more generic images for multi-use
- Updated naming
- Updated docs

## QA

- Sanity check code
- Sync with vf.io and test pattern

## Issues

Fixes #403


docs/patterns/_image-grid.md
```diff
@@ -0,0 +1,26 @@
+---
+collection: patterns
+title: image grid
+---
+
+This image grid can be used to showcase a group of related image such as logos.
+
+```html
+<div class="p-image-grid">
```
anthonydillon: I think we should demo this as a list.

scss/_patterns_image-grid.scss
```diff
@@ -0,0 +1,15 @@
+// Image grid to showcase group of images
+@mixin vf-p-p-image-grid {
+
+  .p-p-image-grid {
```
anthonydillon: `p-p-`? Is this the pattern to end all patterns?

anthonydillon: `p-p-`? Is this the pattern to end all patterns?

scss/_patterns_image-grid.scss
```diff
@@ -0,0 +1,15 @@
...
+    @extend .clearfix;
+    text-align: center;
+    display: block;
+
+    &__img {
```
anthonydillon: We need to control the max-width and height here. As rarely have the correct sized images. Don't SVGs are used and they can be large.

scss/_patterns_strip.scss
```diff
@@ -1,15 +1,17 @@
 // Full width strip pattern
 @mixin vf-p-strip {
 
-  width: 100%;
-  clear: both;
-  background-color: $color-light;
+  .strip {
```
anthonydillon: Looks like this change is also included in the breadcrumbs PR. With a few comments, please resolve them before landing this branch.

anthonydillon:
Thanks Barry, few comments on the code.

scss/_patterns_image-grid.scss
```diff
@@ -0,0 +1,15 @@
...
+      display: inline-block;
+      text-align: center;
```
barrymcgee: Yep, removing it breaks the pattern...

scss/_patterns_image-grid.scss
barrymcgee: I'd prefer to leave this light touch as responsibility for sourcing and using correctly sized images should be on the author.. by adding height/width stipulations, we leave the pattern more brittle and also open the possibility of people using huge images without realising...

PR #492, opened 2016-08-30
Fix leaky Gulp pipes with Plumber

WillMoggridge:
## Done

Catch exceptions in the Gulp sass job, so that it does not break the
watch command.

The stream will now print sass errors in line and continue to watch for new changes.


barrymcgee:
Code looks good 
LGTM! :shipit:

PR #501, opened 2016-08-30
Change font-weight of links to 400

barrymcgee:
## Details

Fonts should not be bold - they should be font-weight 400.

## QA

Pull down and test that links are not bold


anthonydillon:
LGTM :+1:

PR #503, opened 2016-08-30
Define default definition list style

anthonydillon:
## Done

- Added the default definition list styling

## QA

- Check out lists in vf.io 
- At the bottom there should be definition list 
- Check the styling matches the basic style 

## Details

- Issue: https://github.com/ubuntudesign/vanilla-framework/issues/350
- Design: https://jujucharms.com/wordpress/#charm-config-debug

### DIFF
--- a/docs/base/_lists.md
+++ b/docs/base/_lists.md
@@ -68,3 +68,21 @@ title: lists
     </ul>
     <li>Ordered list item 4</li>
 </ol>
+
+## Definition list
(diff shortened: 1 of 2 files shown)


barrymcgee:
@anthonydillon QA looks good - can you just lift the `<dt>` and `<dd>` styling up to root instead of nesting under `<dl>` to reduce the specificity?

PR #523, opened 2016-08-31
update breakpoint doc

pmahnke:
## Done

* made the doc for breakpoints plural
* added a description
* added examples

## QA

1. go to /settings/breakpoints/ and see that
 * Breakpoints is plural (as per original bug)
 * There is now a description and example of how it is used in media queries

## Details

Fixes [vanillaframework.io bug 23](https://github.com/ubuntudesign/vanillaframework.io/issues/23)


deadlight:
Good stuff

PR #545, opened 2016-09-01
Delete redundant sass-lint:disable no-important

barrymcgee:
## Done

With new sass linting, `no-important` rule has been globally set so individually exempting this rule in each util is no longer required.

## QA

Check that tests still pass.


anthonydillon:
Looks good :+1:

PR #561, opened 2016-09-02
Lists overhaul

deadlight:
## Done

Reworked/added the following lists and items:
 - .list
 - .list__item--icon
 - .list--divided
 - .inline-list
 - .inline-list__item--middot

## QA

Build vanillaframework.io using this sass and check the rendering and code examples on /patterns/lists/

## Details

Partial fix for the issues raised here (with some agreed changes from the original naming suggestions): https://github.com/ubuntudesign/vanilla-framework/issues/349


scss/_patterns_lists.scss
```diff
@@ -1,36 +1,49 @@
...
   @include vf-lists-list;
-  @include vf-lists-list-ticks;
-  @include vf-lists-no-bullets;
+  @include vf-lists-list-divided;
+  @include vf-lists-list-item-state;
   @include vf-lists-combined-list;
   @include vf-lists-inline-list;
+  @include vf-lists-inline-list-middot;
   @include vf-stepped-list;
 }
```
richmccartney: Can we rename the mixins with a ```-p-``` prefix so it would follow what we have done in buttons or card? For example ```@include vf-p-list``` and then it stops the lists list repetition.

scss/_patterns_lists.scss
```diff
@@ -76,22 +89,47 @@
 @mixin vf-lists-inline-list {
   .inline-list {
```
richmccartney: Should this follow the convention of ```.p-list--inline```?

scss/_patterns_lists.scss
```diff
@@ -1,36 +1,49 @@
...
+  @include vf-lists-list-item-state;
   @include vf-lists-combined-list;
   @include vf-lists-inline-list;
+  @include vf-lists-inline-list-middot;
   @include vf-stepped-list;
 }
 
 // List styling using list class
 @mixin vf-lists-list {
   .list {
```
richmccartney: Should this be ```.p-list``` as its a pattern?

scss/_patterns_lists.scss
```diff
@@ -1,36 +1,49 @@
...
   }
 }
 
-// Displays an ubuntu styled list with tick bullets
-@mixin vf-lists-list-ticks {
-  .list-ticks {
-    @extend %list-ticks;
+// A list with separators between items
+@mixin vf-lists-list-divided {
+  .list--divided {
```
richmccartney: See line 14.

scss/_patterns_lists.scss
```diff
@@ -1,36 +1,49 @@
...
 }
 
-// Displays a default styled list with no bullets
-@mixin vf-lists-no-bullets {
-  .no-bullets {
-    list-style: none;
-    margin-left: 0;
+// Displays an ubuntu styled list item with a tick bullet
+@mixin vf-lists-list-item-state {
+  .list__item--icon {
```
richmccartney: See line 14.

scss/_patterns_lists.scss
```diff
@@ -76,22 +89,47 @@
...
+      @extend %inline-list-item;
     }
+  }
+}
 
-    .last-item {
-      margin-right: 0;
+// Displays a list inline with items separated by dots
+@mixin vf-lists-inline-list-middot {
+  .inline-list--middot {
```
richmccartney: See line 14.

scss/_patterns_lists.scss
```diff
@@ -76,22 +89,47 @@
...
+    margin-left: 0;
+    padding-left: 0;
+
+    .inline-list__item {
```
richmccartney: You've used ```&__item``` on other lists items but here. Should we keep this consistent?

scss/_patterns_lists.scss
```diff
@@ -76,22 +89,47 @@
...
+        }
+      }
     }
   }
 }
 
 // Displays a step by step list of instructions
 @mixin vf-stepped-list {
-
   .list-step {
```
richmccartney: See line 14.

scss/_patterns_lists.scss
```diff
@@ -127,41 +165,47 @@
...
-    margin-bottom: 0;
-    padding: 10px 0;
+  &__item {
+    @extend %list-item;
   }
+}
 
-  li:last-of-type,
+// Placeholder for list items
+%list-item {
```
richmccartney: Shouldn't this also follow the BEM style so it would be ```%list__item```

deadlight: It looks like the standard that we're using is that all placeholder names just use a single "-"

scss/_patterns_lists.scss
```diff
@@ -73,25 +77,49 @@
...
+        }
+      }
     }
   }
 }
 
 // Displays a step by step list of instructions
-@mixin vf-stepped-list {
-
+@mixin vf-p-stepped-list {
```
richmccartney: Can we change this to be ```vf-p-list-stepped```

scss/_patterns_lists.scss
```diff
@@ -73,25 +77,49 @@
...
+      @extend %inline-list-item;
     }
+  }
+}
 
-    .last-item {
-      margin-right: 0;
+// Displays a list inline with items separated by dots
+@mixin vf-p-lists-inline-list-middot {
+  .p-inline-list--middot {
```
richmccartney: Should we rename this also to ```.p-list-inline--middot```?

scss/_patterns_lists.scss
```diff
@@ -73,25 +77,49 @@
 }
 
 // Displays a list inline with spacing
-@mixin vf-lists-inline-list {
-  .inline-list {
+@mixin vf-p-lists-inline-list {
+  .p-inline-list {
```
richmccartney: Should we rename this also to ```.p-list-inline```?

scss/_patterns_lists.scss
```diff
@@ -1,41 +1,45 @@
...
+    background-repeat: no-repeat;
+    background-position: 0 50%;
+    padding-left: 25px;
   }
 }
 
 // Displays split as one on small screens
-@mixin vf-lists-combined-list {
+@mixin vf-p-lists-combined-list {
   .combined-list {
```
richmccartney: Should this be ```.p-list-combined```?

scss/_patterns_lists.scss
```diff
@@ -73,25 +77,49 @@
...
+      }
     }
   }
 }
 
 // Displays a step by step list of instructions
-@mixin vf-stepped-list {
-
+@mixin vf-p-stepped-list {
   .list-step {
```
richmccartney: Think this should be ```.p-list--step```

scss/_patterns_lists.scss
deadlight: We're using the more natural language ordering with other classes - like "inline-list" - and it stops it being mistaken for a modifier this way.

scss/_patterns_lists.scss
deadlight: It doesn't share anything with the .list component. It's a top-level Block in its own right.

scss/_patterns_lists.scss
```diff
@@ -1,41 +1,45 @@
...
+    background-repeat: no-repeat;
+    background-position: 0 50%;
+    padding-left: 25px;
   }
 }
 
 // Displays split as one on small screens
-@mixin vf-lists-combined-list {
+@mixin vf-p-combined-list {
   .combined-list {
```
richmccartney: @deadlight  Should this be a pattern too?

richmccartney: @deadlight Needs a ```p-``` prefix.

scss/_patterns_lists.scss
```diff
@@ -73,25 +77,49 @@
 }
```
richmccartney: On line 54 is depricated with the new grid layout we don't use ```.last-col``` anymore.

deadlight:
@richmccartney: so can this be marked Code/QA approved?

richmccartney:
@deadlight yeah I dunno if there is any more design QA needed?

deadlight:
@richmccartney yeah, that's covered in the "design needed" label but I'll mark it as QA approved when design is signed off.

PR #564, opened 2016-09-02
Grid list audit

grahambancroft:
## Done

Rebuild grid list using flex box

## QA

View the grid list example in s, m & l

### DIFF
--- a/scss/_patterns_grid-list.scss
+++ b/scss/_patterns_grid-list.scss
@@ -50,7 +50,7 @@
 
     &__title {
       line-height: 1;
-      margin-bottom: .5em;
+      margin-bottom: .5rem;
       margin-top: 0;
     }
 
--- a/scss/_patterns_grid-list.scss
+++ b/scss/_patterns_grid-list.scss
@@ -1,43 +1,20 @@
 // List Grid
-
 @mixin vf-p-grid-list {
-  $grid-border-style: 1px dotted $color-warm-grey;
-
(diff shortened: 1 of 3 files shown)


scss/_patterns_grid-list.scss
```diff
@@ -1,54 +1,35 @@
...
-      max-width: 60px;
-    }
+  .p-grid-list {
+    align-items: flex-start;
+    display: flex;
+    flex-direction: row;
+    flex-wrap: wrap;
+    margin-left: 0;
 
     p {
```
richmccartney: @grahambancroft I would BEM this element so it fits inline with other patterns.

scss/_patterns_grid-list.scss
```diff
@@ -1,54 +1,35 @@
 // List Grid
-
-@mixin vf-grid-list {
+@mixin vf-p-grid-list {
   $grid-border-style: 1px dotted $color-warm-grey;
```
richmccartney: @grahambancroft I think we should move this into a settings file and then set it as a ```!default``` so it can be overwritten by themes if needed.

richmccartney:
@grahambancroft Made two comments and there are conflicts that need fixing.

barrymcgee:
@grahambancroft Can you please convert `px` and `em` values to `rem`?

Then just rebase to resolve conflicts and we're good to go 👍🏻

Issue #387, opened 2016-07-22
Audit grid


barrymcgee:
## Context

The current grid provides twelve columns and the ability to nest within each column. 

* Is the mixin used still the most efficient way to construct responsive grids? 
* Does it fulfil all requirements from very small screens upwards? 
* Does it need another breakpoint by default?
* Could flexbox improve it in any way?
* Do we currently provide sufficient settings?
* Is there a smarter way of emulating `.last-col` ?

- Should also include docs and examples for:
 - offset utils
 - pull/push utils

Naming conventions (`columns-X`) should be maintained.

## Actions

- [ ] Browser testing
- [ ] Accessibility testing
- [ ] Sass linting
- [ ] Update documentation
- [ ] Convert sizing units to rems

barrymcgee:
There seems to be a bug at certain resolutions scaling down, see:

https://www.dropbox.com/s/hz4h5g9lbzf9ybz/Screenshot%202016-07-27%2011.56.28.png?dl=0

Issue #344, opened 2016-04-20
Issue with Grid append-one


anthonydillon:
## Context
When working on jujucharms we have a row with eight-col / four-col and a row below which is seven-col append-one / four-col and they do not line up. The columns should align.

barrymcgee:
Will be addressed by #387

PR #565, opened 2016-09-02
Audit and overhaul grid

### LINKED ISSUE #293: Grid is lacking documentation
The grid component is missing documentation and examples in the docs. There should be a mention of the append and prepend functionality.Maybe even split them out into there own sub-mixins of the grid.

barrymcgee:
## Done

- Remove previous grid framework and drop in Shelves grid - http://shelvesgrid.org
- Add docs for new grid
- Refactor other docs to use new grid naming
- This new grid should result in about 300 lines less of un-minified code 

## QA

- Sync code and check grid docs in `vf.io`

## Details

Fixes #293 #344 #387


richmccartney:
@barrymcgee just to repeat what we spoke about, in ```_settings_grid.scss``` can we use the shelves variables instead and override their defaults. This would stop the grid breaking if we update the code to a new version as changes have been made in ```grid/shelves/mixins/_variables.scss```

Also should we make this a dev dependancy with npm and not have the code in our local framework?

PR #570, opened 2016-09-02
Navigation: Support alternative markup format

### LINKED ISSUE #486: We should support lists in navigation
There is far from a consensus in the community on [not using `<li>`s for navigation](https://css-tricks.com/navigation-in-lists-to-be-or-not-to-be/). I think therefore the responsible thing is to write our navigation styling to explicitly support multiple markup forms, based on the developer's semantic choice:

``` html
<nav>
   <ul>
      <li><a>item one</a></li>
      <li><a>item two</a></li>
  </ul>
</nav>
```

or

``` html
<nav>
  <a>item one</a>
  <a>item two</a>
</nav>
```

or

``` html
<nav>
  <div><a>item one</a></div>
  <div><a>item one</a></div>
</nav>
```

Presumably the fact that we use BEM classes instead of targeting elements should make it easier to support all three?

nottrobin:
Closes #486.

To support either of the possible alternative markup forms, either using a `<ul>` or wrapping each anchor in a `<div>`, all we have to do is remove all padding and margin from those elements and make the `display: inline-block`.

As this is a common feature that you might want to do on any block-level element, I've made a utility class, `u-inline-block`, to achieve this. Although I'm not sure if we might want to expand this class to remove more than just padding and margin - e.g. we might want to do `border: 0` or `list-style-type: none`. Or is that gold-plating for now?

I have only given an example of the `<ul>` form of the markup, because I think that's the more common one, and if anyone wanted to use the `<div>` form, it would be pretty obvious how to do it from the `<ul>` example provided.

QA
---

Run `vanillaframework.io` with this codebase and see <http://localhost:3000/patterns/navigation/>. Observe that the menu looks the same with the `<ul>` markup as it does without it.

### DIFF
--- a/docs/patterns/_navigation.md
+++ b/docs/patterns/_navigation.md
@@ -46,3 +46,43 @@ title: Navigation
         </div>
     </nav>
 </header>
+
+```html
+<header class="p-navigation" role="banner">
+    <div class="p-navigation__logo">
+        <a class="p-navigation__link" href="#">
+            VF
+        </a>
+    </div>
+    <nav class="p-navigation__nav" role="navigation">
+        <span class="u-off-screen">
+            <a href="#main-content">Jump to main content</a>
+        </span>
+        <ul class="p-navigation__links">
+            <li class="p-navigation__link"><a href="#">Link1</a></li>
+            <li class="p-navigation__link"><a href="#">Link2</a></li>
+            <li class="p-navigation__link"><a href="#">Link3</a></li>
+            <li class="p-navigation__link"><a href="#">Link4</a></li>
+        </ul>
+    </nav>
+</header>
+```
+
+<header class="p-navigation" role="banner">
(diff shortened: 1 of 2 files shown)


barrymcgee:
I'm really not convinced we need this approach - even moreso now that we need to add an extra utility to force this pattern to behave as the current markup already does natively.

I also don't feel that becuase there is divided opinion in the wider community, that is a reason that we should support both approaches - we should decide which one we think is correct and implement that.

Adding `WG:Proposal` label to discuss in Working Group.

barrymcgee:
Agreed in Working Group to incorporate inline-block style to current pattern, removing the need for extra utility.

nottrobin:
Ready to be reviewed again, @barrymcgee. There's no longer any trace of `u-inline-block`.

barrymcgee:
@nottrobin Fix conflicts then all good

PR #576, opened 2016-09-05
Pacify linter - pull quotes pattern

barrymcgee:
## Done
Addressed all warnings when linting pull quote pattern by re-ordering source order

## QA
Sanity check code, check I haven't broken pattern in vf.io


pmahnke:
LGTM 👍

PR #577, opened 2016-09-05
Linting notification pattern

barrymcgee:
## Done
- Re-ordered source order
- Swapped out modified alpha border color of cool grey to light grey as colors should only be defined in color settings file.

## QA
- Sanity check code
- QA in vf.io


pmahnke:
LGTM 👍

PR #578, opened 2016-09-05
Lint navigation pattern

barrymcgee:
## Done
Re-ordered source order to pacify linter

## QA
- Sanity check code
- Qa with vf.io


pmahnke:
LGTM 👍

PR #580, opened 2016-09-05
Lint lists pattern

barrymcgee:
## Done
- Lint lists pattern 
- change `em` units to `rem`
- change `px` units to `rem`
- change middot to HTML entity

## QA
- Check code
- QA with `vf.io`


pmahnke:
LGTM 👍

PR #587, opened 2016-09-07
Lint code numbered pattern

barrymcgee:
## Done

- Re-order selector order to pacify linter

## QA

- Check code
- Check pattern in vf.io


anthonydillon:
Looks good :+1: only needs a code review as if a reorder.

PR #588, opened 2016-09-07
Lint card pattern

barrymcgee:
## Done

- Re-order source order
- Refactor extend pattern to use placeholder
- Use $color-light-grey instead of alpha transparency of $color-mid-grey

## QA

- Check code
- Check pattern in vf.io


anthonydillon:
Good stuff, thanks :+1:

PR #589, opened 2016-09-07
Lint equal height

barrymcgee:
## Done

- Re-ordered source order

## QA

Check code


deadlight:
Linted code looks good and builds.

PR #590, opened 2016-09-07
Lint base typography

barrymcgee:
## Done

Re-ordered source order - there are three warnings here unresolved as they are exceptions to the rule;

```
9:3  warning  * (universal) selectors are not allowed  no-universal-selectors
  10:5  warning  Vendor prefixes should not be used       no-vendor-prefixes
  11:5  warning  Vendor prefixes should not be used       no-vendor-prefixes
```

## QA

Check code

### DIFF
--- a/scss/_base_typography.scss
+++ b/scss/_base_typography.scss
@@ -7,8 +7,8 @@
   }
 
   * {
-    -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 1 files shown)


deadlight:
The exceptions look reasonable in context. :+1:

PR #594, opened 2016-09-07
Lint base forms

barrymcgee:
## Done

- Re-order source order
- Also disabled  `no-universal-selectors` in sass-lint.yml as we use universal selectors in several places.

## QA

Check code


deadlight:
Can confirms that the rule change works and the files all lint.
Code looks good.

PR #601, opened 2016-09-08
Change travis IRC alert settings

nottrobin:
Change travis IRC alert settings

- The notifications are getting a little spammy in #vanilla-framework, so change to only notify when the build changes
- It would also be nice to get build alerts in #canonical-webteam, if they're not too spammy
- Remove ducksboard notifications - ducksboard is dead


anthonydillon:
Code looks good :+1:

PR #609, opened 2016-09-09
Add box-sizing: border-box rule

### LINKED ISSUE #603: Input field overflows fieldset
# 1. Bug template

## Process

1. Navigate Base > Form
2. Scroll to Fieldset section

## Expected result

The input field should not overflow outside the fieldset container.

## Screenshot

In Chrome.

screen shot 2016-09-08 at 17 08 14

barrymcgee:
## Done

Added box-sizing rule to make working with box model more consistent across the framework.

## QA

- Check code
- Run through vf.io patterns and look for breakage

## Details

Fix #603 

More info: https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/

### DIFF
--- a/scss/_layout_grid.scss
+++ b/scss/_layout_grid.scss
@@ -11,3 +11,14 @@ $shelves-column-name: $grid-col-name;
 @import
 'grid/shelves',
 'grid/shelves-grid';
+
+/* Box-sizing FTW!
+https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/
+*/
+html {
+  box-sizing: border-box;
+}
+
+*, *:before, *:after {
+  box-sizing: inherit;
+}
--- a/scss/_layout_grid.scss
+++ b/scss/_layout_grid.scss
@@ -11,14 +11,3 @@ $shelves-column-name: $grid-col-name;
 @import
 'grid/shelves',
 'grid/shelves-grid';
-
-/* Box-sizing FTW!
-https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/
-*/
-html {
-  box-sizing: border-box;
-}
-
-*, *:before, *:after {
-  box-sizing: inherit;
-}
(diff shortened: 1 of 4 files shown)


nottrobin:
I feel like this should be a proposal, to be discussed with the team.

scss/_layout_grid.scss
```diff
@@ -11,3 +11,14 @@ $shelves-column-name: $grid-col-name;
...
 'grid/shelves-grid';
+
+/* Box-sizing FTW!
+https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/
+*/
+html {
+  box-sizing: border-box;
+}
+
+*, *:before, *:after {
```
nottrobin: Shouldn't any `*` rules and rules applied to elements directly (especially `html`) be in a `_base` file?

barrymcgee:
The is not addition to Vanilla, the stable version already has: https://github.com/ubuntudesign/vanilla-framework/blob/master/scss/modules/_reset.scss#L11 - it simply got accidentally dropped during the reset refactor.

Good catch on `base` point.. will pick that up..

nottrobin:
Oh okay then @barrymcgee, apologies I didn't realise that.

nottrobin:
New file structure looks good. I'll QA this in a bit and :+1: if no errors.

nottrobin:
A couple of bugs:

- `<div class="row">` in vanillaframework.io has more margin than it did:
  old: old new: new
- `.grid-list` looks messed up:
  grid-list

barrymcgee:
@nottrobin I'm working on a rebuild on docs-vanilla-theme that already has border-box enabled until this is merged upstream.

`.grid-list` already has it's own work issue here: https://github.com/ubuntudesign/vanilla-framework/issues/615

nottrobin:
Okay so the `.row` problem is `docs-vanilla-theme`'s fault? That's fair enough.

I didn't see how #615 is directly related to `.grid-list`. In any case, this PR shouldn't breaks something vs what works upstream surely?

barrymcgee:
@nottrobin Grid list does not currently work upstream - there's currently 6 issues open against it - https://github.com/ubuntudesign/vanilla-framework/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20grid-list - it's high priority to confirm current designs for it, rename it so that it's not confused with being related to the actual grid and refactor the pattern accordingly. The first step of all of the above is to apply border-box to it and build upwards from there.

nottrobin:
@barrymcgee Okay. Let's merge this then.

PR #613, opened 2016-09-09
Revert font inclusion changes

WillMoggridge:
## Done

Revert font inclusion changes, as Ubuntu light does not display correctly.
We need to investigate our font options at a later time.

## QA

Check that Ubuntu font loads correctly, confirming Ubuntu light looks correct.


anthonydillon:
Nice one, thanks :+1:

PR #616, opened 2016-09-09
Add back rgba for drop shadow

yaili:
## QA

Go to Patterns > Card in docs, scroll to Card highlighted and check it has an rgba box-shadow


anthonydillon:
LGTM :+1:

PR #618, opened 2016-09-09
Remove border from placeholder

yaili:
## Done

Moved border from placeholder to card so it doesn't appear in card highlighted

## QA

Go to Patterns > Card and verify if both cards still have correct styling, and card highlighted has no border, only drop shadow.


anthonydillon:
Lgtm :+1:

PR #628, opened 2016-09-12
Added base lists to include

anthonydillon:
## Done
- Added the block namespace to the vf mixin
- Included into the vanilla package

## QA
- Load in vf.io
- Go to http://localhost:3000/base/lists/
- Check that the lists look so ... so nice!

## Details

- Fixes: https://github.com/ubuntudesign/vanilla-framework/issues/610


barrymcgee:
LGTM! :shipit:

PR #649, opened 2016-09-14
Updates docs for padding/margin collapse

### LINKED ISSUE #647: Should padding collapse work the same way as margin collapse
At the moment, padding collapse is based on breakpoint (no-padding--small, no-padding--medium, no-padding--large) whereas margin collapse is based on side (no-margin--top, etc.).

Should they both work the same way (ie based on side)?

barrymcgee:
## Done

- Fixed couple of errors in styling
- Included padding collapse in vanilla manifest as previously missed
- Added docs and examples for both utils
- Made spacing in utils consistent

## QA
- Sync up vf.io and test docs for each pattern

## Details

Fixes #647

### DIFF
--- a/scss/_utilities_vertically-center.scss
+++ b/scss/_utilities_vertically-center.scss
@@ -1,6 +1,7 @@
 // Vertically align
 
 @mixin vf-u-vertically-center {
+
   .u-vertically-center {
     align-items: center !important;
     display: flex !important;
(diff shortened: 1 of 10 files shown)


scss/_utilities_vertically-center.scss
```diff
@@ -1,6 +1,7 @@
 // Vertically align
 
 @mixin vf-u-vertically-center {
+
```
deadlight: This is causing a lint error.

deadlight:
QA'd and reviewed code. :+1:

PR #650, opened 2016-09-14
Remove redundant js file

barrymcgee:
## Done
This file is not being used so should not be in the codebase.


deadlight:
Confirm removed

PR #654, opened 2016-09-14
Remove base blockquote from docs

### LINKED ISSUE #614: [Proposal] Hide blockquote page from docs
# 2. Pattern proposal/amendment

- Amendment to docs

Blockquote is referenced in Base > Typography, which makes sense, but it is also on its own page in Base > Blockquote. I propose we remove this page from the docs.

barrymcgee:
## Done

Remove base blockquote from docs as base blockquote is already referenced in `base/typography`

## QA 

Check code

Fixes #614


deadlight:
I can confirm that it's gone ;)

PR #655, opened 2016-09-14
Remove role=navigation from header

barrymcgee:
## Done

Remove role=navigation from nav as it's implied.

## QA

Check code


deadlight:
It's gone!

PR #661, opened 2016-09-16
Remove top margin from headers and paragraphs

barrymcgee:
## Done

H1s and P tags should not have top margin

## QA

- Check code
- Sync up with vf.io


pmahnke:
LGTM 👍

PR #662, opened 2016-09-16
Fix up equal height docs to use theme outlines

### LINKED ISSUE #598: Equal heights missing code sample in docs
# 1. Bug template

## Process

1. Navigate to Layout > Equal height

## Expected result

There should be HTML blocks showing how to use equal height.

## Current result

There are no code examples right now.

barrymcgee:
## Done

- Make equal height docs use theme outline class
- Add flex 1 to make equal height items fill their parent by default 

## QA

- Check code
- Sync vf.io and test util

## Details

Fixes #598

### DIFF
--- a/docs/utilities/_equal-height.md
+++ b/docs/utilities/_equal-height.md
@@ -16,11 +16,11 @@ To ensure two or more elements have an equal height regardless of their content,
 </div>
 ```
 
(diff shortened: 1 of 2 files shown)


pmahnke:
LGTM 👍

PR #669, opened 2016-09-19
Remove padding-top from footer copy

### LINKED ISSUE #664: p-footer copy doesn't need padding top
# 1. Bug template

## Process

1. Navigate to Patterns > Footer

## Expected result

Same padding top, left and right on p-footer copy.

## Current result

There's too much space between horizontal line and top of footer text. If we remove the padding top it seems to fix it.

barrymcgee:
## Done

- Removed padding-top from footer copy

## QA

Check code, sync with vf.io

## Details

Fixes #664

### DIFF
--- a/scss/_patterns_footer.scss
+++ b/scss/_patterns_footer.scss
@@ -14,7 +14,6 @@
 
     &__copy {
       margin-bottom: 0;
-      padding-top: 1rem;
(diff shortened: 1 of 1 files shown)


grahambancroft:
Lgtm 👍

Issue #666, opened 2016-09-16
Add aria-disabled=true in elements that don't accept disabled attribute


yaili:
# 2. Pattern proposal/amendment

## Is this a proposal or an amendment?

- Proposal

Use aria-disabled="true" to indicate disabled elements that don't allow disabled attribute

## Context

On HTML elements that don't accept the disabled attribute. Elements that do are: button, input, select, textarea, optgroup, option, fieldset

Example: we use a disabled button style on link elements, not just button elements.

## State

Yes: disabled

PR #670, opened 2016-09-19
Change $color-mid-grey from #888 to #666

### LINKED ISSUE #660: Read-only textarea text colour is not accessible
# 1. Bug template

## Process

1. Navigate to Base > Forms

## Current result

Read-only textarea text doesn't have sufficient contrast with background. It should be readable as it is read-only, not disabled. Text colour should be changed to use our mid-grey, which will at a later stage be changed from #888 to #666

barrymcgee:
## Done

Change $color-mid-grey from `#888` to `#666` for accessibility reasons.

## QA

- Check code
- Qa with `vf.io`

## Details

Fixes #660

### DIFF
--- a/scss/_settings_colors.scss
+++ b/scss/_settings_colors.scss
@@ -6,7 +6,7 @@ $color-brand:            #333 !default;
 
 $color-light-grey:       #f7f7f7 !default;
 $color-warm-grey:        #cdcdcd !default;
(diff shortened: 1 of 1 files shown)


anthonydillon:
Looks good :+1: 

Might need to just close and reopen to rerun CI. I cant see why that failed.

PR #674, opened 2016-09-20
Refactor grid docs

### LINKED ISSUE #600: Grid should not appear in two sections
# 1. Bug template

## Process

1. Navigate to Settings > Grid
2. Navigate to Layout > Grid

## Expected result

There should only be one page called Grid.

## Current result

Grid is divided into two pages. 

It makes sense to leave the page under Layout so that it's not confusing to the user. But if we remove Grid from Layout, though, we should move Equal height to Patterns and Clearfix to Utilities or something, and remove the Layout section completely — without Grid, Layout becomes unnecessary.

barrymcgee:
## Done

- Move clearfix from layout section to utils
- Move grid from layout section to patterns
- Remove layout section from docs
- Add `.theme__outline` class to grid examples

## QA

Sync with `vf.io` and check docs make sense

## Details

Fixes #600


deadlight:
All good

Issue #605, opened 2016-09-09
Hide base links from "Base" section of docs


yaili:
# 2. Pattern proposal/amendment

## Is this a proposal or an amendment?

- Amendment

Proposal to remove base link from "Base" section (just in docs, not in CSS) and either a) hide it, or b) mention it in Patterns > Links.

I think it is confusing to have two pages that have the same title in two sections.

PR #677, opened 2016-09-20
Remove base links from docs

barrymcgee:
## Done

It's confusing to have two pages in vf.io named 'Links' and we decided in WG that `base/links` is superfluous.

## QA

Ccheck build still runs without errors and base/links disappears from side nav.

## Details

Fixes: #605


grahambancroft:
Lgtm 👍

PR #687, opened 2016-09-21
Remove release notes as it’s no longer needed

barrymcgee:


anthonydillon:
I think we need to turn this isn't a blog post or have it available on vf.io.

barrymcgee:
@anthonydillon It will be added to the v1.0.0 Github Release tab (https://github.com/ubuntudesign/vanilla-framework/releases) so it's linkable as a minimum and can then be added to a blog post at a later date if applicable.

PR #712, opened 2016-11-21
Remove explicit font sizes from docs

barrymcgee:
## Done

This value in the docs was incorrect but also, this value changes per breakpoint so cannot be represented via a single value. 

The range of font sizes across breakpoints is displayed in a table further down the page so I've removed this explicit static value.

## QA

Check code and ensure nothing has been removed that should not have bee removed.

## Details

Fixes: https://github.com/ubuntudesign/vanilla-framework/issues/708


barrymcgee:
[Remove explicit font sizes from docs](https://trello.com/c/0C67RRhG/2201-remove-explicit-font-sizes-from-docs)

PR #735, opened 2017-01-12
Add spacing between .p-strip pattern

barrymcgee:
## Done

This change will allow content to be sensibly spaced using just strips with the grid.

This change means that strips will have a bottom spacing of 2rem except when a strip pattern with a coloured background follows a strip pattern with a coloured background. This is because these strips have inherent padding and no bottom margin is needed.

To see this in action, see this Codepen: http://codepen.io/barrymcgee/pen/VPayrj

## QA

- Pull code
- Build Vanilla
- Include built CSS into a empty webpage 
- Copy markup from the Codpen above into it
- Ensure that none of the strips butt against each other.
- Switch and change the order of the strips and check webpage again.

## Details

Fixes: #734

### DIFF
--- a/scss/_patterns_strip.scss
+++ b/scss/_patterns_strip.scss
@@ -4,6 +4,7 @@
   %strip {
     background-color: transparent;
     clear: both;
+    padding: 2rem 0;
     width: 100%;
   }
 


scss/_patterns_strip.scss
```diff
@@ -1,9 +1,11 @@
 // Full width strip pattern
 @mixin vf-p-strip {
 
+  // strip
   %strip {
     background-color: transparent;
     clear: both;
+    margin-bottom: 2rem;
```
anthonydillon: Couldn't we just use `padding: 2rem 0;` for all strips? Then we wouldn't need to do the margin dance? Does the margin-bottom give us something I'm missing?

For example:
http://codepen.io/anthonydillon/pen/vgKKjw

barrymcgee: It was an attempt to get around the 'double spacing (margin + padding)' issue when two strips of the same type follow each other - but thinking again, perhaps that a subjective design answer to a question that has not yet been asked. Will change to your suggested approach.

anthonydillon:
@barrymcgee You have some lint issues in strips.

scss/_patterns_strip.scss
```diff
@@ -1,24 +1,26 @@
...
 
-    &--dark {
+    .p-strip {
       @extend %strip;
-      background-color: $color-dark;
-      color: $color-x-light;
+
+      &--light {
+        @extend %strip;
+        background-color: #f7f7f7;
```
anthonydillon: Please use `$color-light`

scss/_patterns_strip.scss
```diff
@@ -1,24 +1,26 @@
...
+      }
+
+      &--dark {
+        @extend %strip;
+        background-color: #333;
```
anthonydillon: Please use `$color-dark`

anthonydillon:
Another thought is that we want to override the padding amount for the brochure theme. Should this value be a variable? Or should we override the entire component in the theme? What do you think?

scss/_patterns_strip.scss
barrymcgee: Doh - copied from your codepen and forgot these were swapped out in my original 🙄

anthonydillon:
QA ok.

anthonydillon:
@barrymcgee You removed `--light` and `--dark`?

barrymcgee:
@anthonydillon No, just removed any changes to this so they're no longer visible in the changeset - as you suggested?

Spacing - I see this relating to standardising the spacing into variables https://github.com/ubuntudesign/vanilla-framework/issues/728

anthonydillon:
Ah, of course. Been a long week. Thanks :+1:

Issue #738, opened 2017-01-13
Footer pattern should have inherent bottom margin


barrymcgee:
The footer pattern should not require additional markup or styling to add a default bottom margin given it will nearly always be the last page element.

PR #739, opened 2017-01-13
Add inherent bottom margin to .p-footer

barrymcgee:
## Done

The footer [attern should not require extra markup or styling to work straight out of the box. This change gives it a default 1rem of bottom margin ensuring it does not butt off the edge of the viewport.

## QA

- pull down code
- run `gulp test` to lint code
- run docs.vf.io and navigate to 'footer' to test in-situ 

## Details

Fixes: #738

### DIFF
--- a/scss/_patterns_footer.scss
+++ b/scss/_patterns_footer.scss
@@ -4,11 +4,11 @@
   .p-footer {
     border-top: 1px dotted $color-x-dark;
     font-size: .75rem;
+    margin-bottom: 1rem;
     padding-top: .75rem;
 
     @media (min-width: $breakpoint-medium) {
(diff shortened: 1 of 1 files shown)


scss/_patterns_footer.scss
```diff
@@ -4,11 +4,11 @@
   .p-footer {
     border-top: 1px dotted $color-x-dark;
     font-size: .75rem;
+    margin: 0 0 1rem;
```
anthonydillon: Do we want to remove margin that does not exist at this level? This could cause issues setting margins locally on the site. We should be prescriptive with are rules. Maybe we just need `margin-bottom: 1rem;`.

Issue #726, opened 2016-12-19
p-list ticks should not be vertically centered


anthonydillon:
## Process

1. Navigate to lists pattern
2. Scroll to the lists with ticks
3. Add more text to the item so it wraps onto to two lines

## Expected result

When the list item content wraps only two or more lines the tick should remain at the base line of the first line of text.

## Current result

The tick is vertically centred between the lines of text. 

## Screenshot

![ubuntu in the cloud cloud ubuntu](https://cloud.githubusercontent.com/assets/1413534/21312493/3dc05e7a-c5e4-11e6-9c85-bd351588ab9a.png)

PR #742, opened 2017-01-17
Ensure list tick icon stays anchored to top left of item

barrymcgee:
## Done

Changed position of background image providing list tick icon so that it stays at the top left, regardless of how deep the list item expands.

## QA

- Pull down code
- Sync with docs.vf.io
- Navigate to /patterns/lists/
- Scroll down to 'List with icons and dividers' section
- Edit a list item to increase text content so it wraps over multiple lines
- Observe that the tick remains static

## Details

Fixes: #726 

### Current (incorrect)

screenshot 2017-01-17 16 26 53

### Expected (correct)

screenshot 2017-01-17 16 26 31

### DIFF
--- a/scss/_patterns_lists.scss
+++ b/scss/_patterns_lists.scss
@@ -44,7 +44,7 @@
 @mixin vf-p-list-item-state {
   .is-ticked {
     background-image: svg-tick($color-mid-dark);
-    background-position: 0 50%;
+    background-position: 0 1rem;
     background-repeat: no-repeat;
     padding-left: 25px;
   }


scss/_patterns_lists.scss
```diff
@@ -44,7 +44,7 @@
 @mixin vf-p-list-item-state {
   .is-ticked {
     background-image: svg-tick($color-mid-dark);
-    background-position: 0 50%;
+    background-position: 0 15px;
```
anthonydillon: Should this be `1rem`?

PR #743, opened 2017-01-17
Enable combining u-text-align--center with u-vertically-center

anthonydillon:
## Done
- Added the flex alignment styles to the text-align utilities so they work in combination with flex and without it.

## QA
- Pull down this code
- Build with `./run build`
- Create demo.html with the following markup

```html
<html>
<head>
  <title></title>
  <link rel="stylesheet" type="text/css" media="screen" href="build/css/build.css">
</head>
<body>
  <section class="p-strip--dark">
    <div class="row">
        <div class="u-equal-height">
            <div class="col-7">
                <h2>Buy a BootStack managed cloud</h2>
                <p>BootStack gives you a private OpenStack cloud managed by Canonical.  It runs on your hardware, in the datacenter of your choice with our experts responsible for design, deployment and 24x7 support.</p>
                <ul class="p-list">
                    <li class="p-list__item is-ticked">Built to your specifications</li>
                    <li class="p-list__item is-ticked">Your choice of certified hardware</li>
                    <li class="p-list__item is-ticked">Retain complete control of every tool and every machine</li>
                    <li class="p-list__item is-ticked">Guaranteed availability backed by a clear SLA</li>
                    <li class="p-list__item is-ticked">Transfer operations anytime you want</li>
                </ul>
                <p><a href="/cloud/openstack/managed-cloud" class="p-button--positive">Learn more about BootStack</a></p>
            </div>
            <div class="col-5 u-text-align--center u-vertically-center">
              <img src="https://assets.ubuntu.com/v1/466046c7-illustration-cloud-bootstack.svg?w=320" alt="" class="u-hidden--small" width="320">
            </div>
        </div>
    </div>
</section>
</body>
</html>
```

- Check that the image is horizontally and vertically centred.

## Details

I didn't think this needed to be added to documentation as there is now mixed documentation.


scss/_utilities_text-align.scss
```diff
@@ -10,19 +10,22 @@
 @mixin vf-u-text-align--center {
   .u-text-align--center {
     text-align: center !important;
+    justify-content: center !important;
```
barrymcgee: You have linting errors with the 'property-sort-order'

anthonydillon: Without setting `justify-content` the text will be left aligned also. I suggest the name of the utility is updated to `.u-content-align` as it is what its doing. CSS have named the attribute incorrectly as it does align any block elements.

Either way, this need to be applied to this utility or the renamed one as it keeps the behaviour consistent.

barrymcgee: Ambiguous rule names within selectors is ok so long as the selector name itself is clear. 

`.u-content-align` sounds like the best approach then.

richmccartney: @anthonydillon @barrymcgee I think if we're adding properties which are meant only for flex we should state that explicitly. I would have no clue that ```u-content-align``` would only get effected if used with in a flex content. 

I suggest we should keep the utility classes to fully explain what they're achieving e.g 

```u-text-align--left``` or ```u-text-align--justify``` and they both ONLY effect that. If justify content is for flex it should be a ```u-flex-content--center``` or something similar as trumps should only overwrite one property.

**Comment by richmccartney** [comment]
@anthonydillon @barrymcgee Why are all our text utilities just fixed to one screen size? Can we add an update to have mobile text alignment too?

**Comment by barrymcgee** [comment]
@richmccartney What do you mean? `.u-text-align` applies across all breakpoints?

**Comment by anthonydillon** [comment]
@richmccartney could you give an example? The content/text alignment should remain consistent across breakpoints. 

For example, if some text is centred within its container then on small screen it should also be centred. Until we find a use case to change the alignment on different breakpoints.

**Comment by richmccartney** [comment]
@barrymcgee @anthonydillon I am already finding a use case in MAAS where we have right aligned text on desktop and it needs to be left aligned on mobile.

Having classes like this:

```u-text-sm--left``` ```u-text-md--left``` ```u-text-lg--left``` which would correspond to small, medium and large breakpoints. If we just add a trump which doesn't think about the effects of all screen sizes or at least the break points we have then we're not giving the correct amount of flexibility. 

Being dogmatic to the 'we should only create it when we find a use case' I feel is wrong. We need to be thinking of all aspects of a utility class and what they can effect as they're too powerful.

**Comment by anthonydillon** [comment]
@richmccartney thanks for the explanation. Could you bring this up in the working group as I believe your requirement is out of scope for this branch. Good to discuss.

**Comment by barrymcgee** [comment]
@anthonydillon I think you forgot to remove `_utilities_text-align.scss` ?

**Comment by anthonydillon** [comment]
Sorry good spot. Ready for review again.

**Comment by barrymcgee** [comment]
@anthonydillon Just realised we forgot to update the docs after this rename

**Comment by barrymcgee** [comment]
@anthonydillon I'm confused here - did you only change the mixin but not the actual classname?

https://github.com/ubuntudesign/vanilla-framework/blob/master/scss/_utilities_content-align.scss

**Comment by anthonydillon** [comment]
It appears so. Will fix now.

**Comment by barrymcgee** [comment]
@anthonydillon Last week, @yaili mentioned that 'content' is kinda redundant in this context, so if we haven't yet changed it, we should change it to..

```css
.u-align--center
.u-align--left
.u-align--right
```

PR #744, opened 2017-01-18
Add .scss-lint.yml for linters

anthonydillon:
## Done
- Created a `.scss-lint.yml` from the `.sass-lint.yml`

## QA
- Pull down branch
- Run `scss-lint scss/**/*.scss` // if you need to install scss-lint. Run `gem install scss-lint`
- You should only get a handful of warnings as I didnt disallow everything because I believe some are useful

## Details
The reason for this change is that atoms scss-lint package use's scss-lint which needs a `.scss-lint.yml` to allow things like no-important.

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -1,0 +1,19 @@
+scss_files: 'scss/**/*.css.scss'
+
+exclude:
+   - 'scss/_base_reset.scss'
+   - 'scss/grid/*'
+   - 'scss/_layout_grid.scss'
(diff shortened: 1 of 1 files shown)


barrymcgee:
Rather than maintain two linting config files - can we not all use the [sass-lint](https://atom.io/packages/linter-sass-lint) plugin for Atom instead?

anthonydillon:
Your right, lets use linter-sass-lint.

PR #750, opened 2017-01-20
Update link styles

### LINKED ISSUE #717: Update default links, and _patterns_list-link
A new link style has been agreed, which @anthonydillon and @nottrobin have separately done some work towards.

### Basic links

After being [implemented by Ant](https://github.com/ubuntudesign/cloud-vanilla-theme/blob/master/scss/modules/_links.scss) in cloud-vanilla-theme, the concept was refined in https://github.com/ubuntudesign/vanilla-design/issues/1, to simplify the base link style down to blue colour, default font-weight, no underline, add underline on hover. As I've implement in [this commit to docs-vanilla-theme](https://github.com/ubuntudesign/docs-vanilla-theme/pull/42/commits/9817e1ce7e2d652b3c7cd77f68ed030b3e26d4e4):

``` scss
a {
  border-bottom: 0;
  color: $link-color;
  font-weight: inherit;

  &:visited {
    color: $link-color;
  }

  &:hover {
    text-decoration: underline;
  }
}

h1 a,
h2 a,
h3 a,
h4 a,
h5 a,
h6 a {
  color: $text-color;
  font-weight: normal;

  &:visited {
    color: $text-color;
  }

  &:hover {
    color: $link-color;
  }
}
```

### Links in navigation

As I understand links in `p-navigation` should follow the same rules as links in headings. Although it was agreed that we should add this to individual navigation patterns rather than adding it to `nav` in general:

``` scss
// patterns_navigation
.p-navigation__link {
  color: $text-color;

  &:hover {
    text-decoration: underline;
    color: $link-color;
  }

  &.is-active {
    font-weight: normal;
  }
}
```

### _patterns_list-link

As discussed in https://github.com/ubuntudesign/vanilla-design/issues/1, there is a need for a list of links, where the links will standout more than regular links. In this case, links should be normal weight (rather than light) and normal text colour, the same as navigation links and headings. However, this should not be the default for links in lists, as we should assume first and foremost that lists contain normal text.

Therefore this should be its own pattern:

``` scss
.p-list-link {
  color: $text-color;

  &:hover {
    text-decoration: underline;
    color: $link-color;
  }
}
```

barrymcgee:
Note: *Design QA required before merging*

## Done

- Update base link styles
- Added `p-link--strong`
- Added `p-link--soft`
- Added `p-link--inverted`

## QA

View demo here: https://vanilla-8faed.firebaseapp.com/

## Details

Fixes #717

### DIFF
--- a/scss/_patterns_links.scss
+++ b/scss/_patterns_links.scss
@@ -29,6 +29,7 @@
     &--inverted {
       border-bottom-color: $color-light;
       color: $color-light;
+      font-weight: 400;
 
       &:hover {
         color: $color-light;
--- a/scss/_patterns_links.scss
+++ b/scss/_patterns_links.scss
@@ -17,12 +17,12 @@
     }
 
     &--strong {
-      border-bottom-color: $color-highlight;
+      border-bottom-color: $color-accent;
       color: $color-dark;
       font-weight: 400;
 
       &:hover {
-        color: $color-highlight;
+        color: $color-accent;
       }
     }
 
--- a/scss/_patterns_links.scss
+++ b/scss/_patterns_links.scss
@@ -11,6 +11,35 @@
     &--no-underline {
       border: 0;
     }
+
+    &--soft {
+      color: $color-dark;
+    }
+
+    &--strong {
+      border-bottom-color: $color-highlight;
+      color: $color-dark;
+      font-weight: 400;
+
+      &:hover {
+        color: $color-highlight;
+      }
+    }
+
+    &--inverted {
+      border-bottom-color: $color-light;
+      color: $color-light;
+
+      &:hover {
+        color: $color-light;
+        border-bottom-color: $color-light;
+      }
+
+      &:visited {
+        color: darken($color-light, 10%);
+        border-bottom-color: darken($color-light, 10%);
+      }
+    }
   }
 
   .p-top {
--- a/scss/_patterns_links.scss
+++ b/scss/_patterns_links.scss
@@ -16,7 +16,7 @@
       color: $color-dark;
 
       &:hover {
-        color: $color-accent;
+        color: $color-link;
       }
 
       &:visited {
@@ -30,7 +30,7 @@
     }
 
     &--strong {
-      border-bottom-color: $color-accent;
+      border-bottom-color: $color-link;
       color: $color-dark;
       font-weight: 400;
 
@@ -39,8 +39,8 @@
       }
 
       &:hover {
-        border-bottom-color: $color-accent;
-        color: $color-accent;
+        border-bottom-color: $color-link;
+        color: $color-link;
       }
     }
 
--- a/scss/_patterns_links.scss
+++ b/scss/_patterns_links.scss
@@ -19,6 +19,11 @@
         color: $color-accent;
       }
 
+      &:visited {
+        border-bottom: 0;
+        color: $color-dark;
+      }
+
       &.is-selected {
         font-weight: 400;
       }
@@ -29,7 +34,12 @@
       color: $color-dark;
       font-weight: 400;
 
+      &:visited {
+        color: $color-dark;
+      }
+
       &:hover {
+        border-bottom-color: $color-accent;
         color: $color-accent;
       }
     }
--- a/scss/_patterns_links.scss
+++ b/scss/_patterns_links.scss
@@ -32,13 +32,13 @@
       font-weight: 400;
 
       &:hover {
-        color: $color-light;
         border-bottom-color: $color-light;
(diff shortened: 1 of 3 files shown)


yaili:
By looking at https://github.com/ubuntudesign/vanilla-design/blob/master/Links/Link%20styles%20Vanilla.png I think the inverted links are supposed to be font-weight: 400, otherwise they'd be invisible, can you confirm @spencerbygraves ?

spencerbygraves:
Yes, that's correct. Thanks

barrymcgee:
@yaili I've updated the [demo](https://vanilla-8faed.firebaseapp.com) with that inverted link font-weight change.

Does the `.p-link--soft` not need some differentiation also?

yaili:
I don't think so, but I don't think we should show the pattern like the others because it's not intended to be used in normal text but rather in patterns like navigation or sidebars. I'm worried that the inverted links aren't clear enough. Maybe we can leave them like this for now, but we might need to add underline later.

barrymcgee:
Ok - now ready for code review Ant

scss/_patterns_links.scss
```diff
@@ -11,6 +11,36 @@
...
+      color: $color-light;
+      font-weight: 400;
+
+      &:hover {
+        color: $color-light;
+        border-bottom-color: $color-light;
+      }
+
+      &:visited {
+        color: darken($color-light, 10%);
```
anthonydillon: This colour change is so subtle. It is almost impossible to tell the difference.

scss/_patterns_links.scss
```diff
@@ -11,6 +11,36 @@
     &--no-underline {
       border: 0;
     }
+
+    &--soft {
+      color: $color-dark;
```
anthonydillon: The design shows that the hover style should be the same as the standard link.

scss/_settings_colors.scss
```diff
@@ -14,7 +15,7 @@ $color-x-dark:           #000 !default;
 $color-negative:         #d73024 !default;
 $color-warning:          #f99b11 !default;
 $color-positive:         #0f8420 !default;
-$color-information:      #007aa6 !default;
+$color-information:      $color-highlight !default;
```
anthonydillon: After a discussion with @yaili it would be better to unlink `$color-information` from `$color-highlight`. If someone updates the `$color-highlight` as the colour for links then it would have the side effect of changing `$color-information`.

barrymcgee:
Ready for review @anthonydillon - I've haven't changed the subtle visited link as I've had no design direction to do so. I suggest @yaili review and create another issue if she feels appropriate.

scss/_settings_colors.scss
```diff
@@ -3,6 +3,7 @@
 $color-transparent:      transparent !default;
 
 $color-brand:            #333 !default;
+$color-accent:        #007aa6 !default;
```
anthonydillon: If I am not mistaken `$color-accent` was to add personality strip to the brochure theme. Why are we using that variable here? I thought that we were going to just uncouple `$color-highlight` from '$color-information' and renaming `$color-highlight` to `$color-link`

https://docs.google.com/document/d/1Tx9pbW_nKSNEtDB_hziDNIZZl-oJMKGZc-jOIUv_PxI/edit#heading=h.9n8tll52kpfr

scss/_patterns_links.scss
```diff
@@ -11,6 +11,40 @@
...
+    }
+
+    &--strong {
+      border-bottom-color: $color-accent;
+      color: $color-dark;
+      font-weight: 400;
+
+      &:hover {
+        color: $color-accent;
+      }
```
anthonydillon: You need to repeat the `&:visited` styles here because `.p-link--strong:hover` has higher specificity than `a:visited`.

vanilla-links-styles

scss/_patterns_links.scss
```diff
@@ -11,6 +11,40 @@
     &--no-underline {
       border: 0;
     }
+
+    &--soft {
+      color: $color-dark;
+
+      &:hover {
+        color: $color-accent;
+      }
```
anthonydillon: You are missing the `is-selected` styling. The design shows that is selected should be 400 font weight.
https://raw.githubusercontent.com/ubuntudesign/vanilla-design/master/Links/Link%20styles%20Vanilla.png

barrymcgee:
@anthonydillon Let's try again.. updated demo: https://vanilla-8faed.firebaseapp.com/

yaili:
It seems to be working fine. I don't want to show the subtle link in the docs in normal text, it has to be shown with correct usage, like a sidebar filter list.

PR #783, opened 2017-01-27
Add examples

anthonydillon:
## Done

Added examples of each base, pattern and utility for local QA'ing and to become available for the docs when using the new documentation builder.

When checking the code. It may be easier to step through commit at a time to have piecemeal reviews.

## QA

- Pull down this branch
- Run `gulp develop`
- Then check a handful of examples in the browser

## Details
Fixes https://github.com/ubuntudesign/vanilla-framework/issues/779

### DIFF
--- a/examples/_base_blockquotes.html
+++ b/examples/_base_blockquotes.html
@@ -1,0 +1,12 @@
+<html>
+<head>
+  <title>Vanilla framework - base - blockquotes</title>
+  <link rel="stylesheet" type="text/css" media="screen" href="../build/css/build.css">
+</head>
+<body>
+  <blockquote>
+    <p><span>“</span> Ubuntu is an ancient African word meaning 'humanity to others'. Ubuntu is an ancient African word meaning 'humanity to others'. <span>”</span></p>
+    <p><cite>Canonical</cite></p>
+  </blockquote>
+</body>
+</html>
--- a/examples/_base_button.html
+++ b/examples/_base_button.html
@@ -1,0 +1,9 @@
+<html>
+<head>
+  <title>Vanilla framework - base - button</title>
+  <link rel="stylesheet" type="text/css" media="screen" href="../build/css/build.css">
+</head>
+<body>
+  <button>Button</button>
+</body>
+</html>
--- a/examples/_base_code.html
+++ b/examples/_base_code.html
@@ -1,0 +1,17 @@
+<html>
+<head>
+  <title>Vanilla framework - base - code</title>
+  <link rel="stylesheet" type="text/css" media="screen" href="../build/css/build.css">
+</head>
+<body>
+  <pre>
+    <code>
+      this is a code sample line 1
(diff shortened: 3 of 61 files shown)


barrymcgee:
The code is fine but I have to say again I feel this approach is wrong. It felt wrong last week when mooted but even moreso now I'm looking at 61 static partials.

- There is no way to easily navigate between the examples
- These cannot be dynamically combined to create template/page examples
- As I mentioned last week, we should not be in a situation where we're having to maintain two separate instances of pattern markup as they will inevitable get out of date very quickly, despite very best intentions
- This makes any style-guide-driven-development needlessly laborious

There are a [wealth of tools](https://github.com/davidhund/styleguide-generators) available for exactly this purpose so why aren't we using any of them again?

anthonydillon:
The future vision here is to remove the examples and code blocks from the docs and add an iframe to the examples and use javascript to display the code. 

In this future;

> There is no way to easily navigate between the examples

There would you could use the documentation project to navigate the patterns or use the file directory for local development.

> These cannot be dynamically combined to create template/page examples

The template/page examples pages would need to be hand built at the moment. I am not sure having each pattern in an example file would hinder this work. Although I thought we descoped these as weren't sure where they would live.

> As I mentioned last week, we should not be in a situation where we're having to maintain two separate instances of pattern markup as they will inevitable get out of date very quickly, despite very best intentions

The only code for all patterns would live solely within the examples folder. The examples in the docs would iframe the content from the examples folder.

>  This makes any style-guide-driven-development needlessly laborious

I believe this is levitated by removing the duplication in which would become tedious.

I hope these points help to shine a light on why this branch could work. This method also gives us the bonus of having locally testable examples of check pattern. Therefore not needing to link the docs with this project locally.

barrymcgee:
I understand that's the grand vision but that work is not included in this PR so as soon as we merge this, there will be duplication which will need to be maintained.

The problem with using the docs project as a sandbox for local development is it requires the simultaneous running and linking of two separate projects which is needlessly laborious and prone to error. It also requires on-going maintenance of the vanilla-docs-theme to keep pace otherwise you'll be looking at a broken instance of docs, as is the case right now which is why we have to resort to including markup in PRs.

I'd much prefer using a simple Jekyll (or similar tool) to insert each pattern into a simple default template and host it on a local server, all of which can be actioned within vanilla-framework itself, making on-going development and (automated) testing frictionless.

I'll spike a demo to demonstrate the value..

anthonydillon:
It's not included in this PR because it seemed logical to brake the changes
up across a few PRs. I planned to replace the code blocks in the docs
directly after this branch lands. If you would feel better I can do that
within this branch so the change is one big one?

The only thing Jekyll can provide us with is the ability to have navigation
in the examples. Do you thinking navigation is required for local
development?

anthonydillon:
This PR is superseded by https://github.com/ubuntudesign/vanilla-framework/pull/793

PR #793, opened 2017-01-31
Use jekyll to build out local examples for style-guide-development

### LINKED ISSUE #789: Spike demo of hosting local example partials

barrymcgee:
## Done

- Building on previous work by @anthonydillon (#783), I've wrapped his example partials in a minimal Jekyll scaffold.
- This removed any duplication between the example partials
- This approach provides a foundation for Percy (visual regression testing) and/or Browserstack to run a headless browser over
- This approach means styleguide-driven-development can be done entirely within this repo without depending on any other project

## QA

- Pull code
- Run `rm -rf node_modules`
- Run `npm i`
- Run `bundle install`
- You should then be able to run `gulp develop` to kick off development watcher
- Navigate to `http://127.0.0.1:4000/`
- Make a change to your local scss, save and refresh
- Profit!

Fixes #789


anthonydillon:
This does look and function much better than a bunch of example pages. I have a few concerns, though;

### More dependencies 
We already have a lot of dependencies for Vanilla. Currently, we have 496 packages node modules. This results in maintenance overhead and a longer time to install. This method adds a whole new package manager (Bundler) which requires the user to enter their sudo password when setting up local development. Which I don't really like as another step to setting up.

### Integration into docs
One of the requirements of this task was to consolidate the example markup into a single location that can be used locally and in the docs to make it a living documentation. The other requirement was to leave Vanilla documentation using vanilla-docs-theme and divorce the styling from the examples. I don't see how we could pull these patterns into the docs pages as they are? How would we display the markup to recreate this pattern in the docs? I understood that we would iframe each pattern into the docs locally. And use a snippet of JavaScript to display the iframed body content in a code block.

### Visual regression testing
With single page patterns, VRT would be more accurate. If anything changes on the page Percy will report it as a true manipulation. Whereas, with this method, Percy would fail each time we added or removed a pattern as the navigation list would change? This also tightly couples `p-list` to all other patterns as we are using `p-list` to display the navigation. Therefore any changes to `p-list` would result in all patterns failing across the board.

### My two cents...
I don't think the benefit of having in page navigation to switch between pattern over using the file system is worth the cons that come with it. I think simple and clean testing is a better route for this project. Have I missed anything? I may be overlooking the final vision of this method and if so, please let me know.

barrymcgee:
- More dependencies
More dependencies don't particularly worry me so long as the dependency management is transparent, frictionless and the dependency adds value. This dependency has not been added flippantly and solves a problem that is currently hindering local development which to me justifies its inclusion. 

- sudo install
I do not have to `sudo` to `bundle install` which suggests that requirement may be an OS restriction, not limited to this project? As for adding bundler as an additional package manager, a Gem will also be required for Percy meaning it's somewhat unavoidable. Percy is also setup to work seamlessly 

- Visual regression testing
This a good point, it's not as clean as it could be as Percy doesn't seem to offer a way to test against a given page element such as `div#main` for example, which I thought it could. However, this is easily resolved by moving the nav out into the index page, leaving just the pattern on each page with a simple back link. This keeps all the benefits of dynamic templates while keeping the pattern partials true.

- Port forwarding
I forgot to mention this in my original PR but another benefit of this approach is being able to use something like localtunnel.me to quickly send a link to my localhost for UX/design review. This is increasingly important for me as I'm remote and can't just call someone to look over my shoulder.

- Docs integration
I just copied over the partials you created with the assumption the Javascript solution to keep docs in sync discussed last week is still a viable solution? Each pattern still has it's own partial as before simply swapping out html `<head>` with front matter and removing the closing `</body></html>`.

Aside: Aa a wider thought outside the scope of this conversation  - I'm also wondering if we should park the whole idea of hosting the documentation using the `vanilla-docs-theme` due to the circular dependency and also the fundamental difference between Vanilla docs and other docs within the company. All other docs are static and signpost instructions to a particular product. Vanilla docs are a living pattern library and as such point at themselves for examples of use.

gulpfile.js
```diff
@@ -17,12 +17,14 @@ wrench.readdirSyncRecursive('./gulp').filter(function(file) {
...
+  console.log('jekyll  - Run gulp develop with Jekyll instance to demo examples');
   console.log('test  - Lint Sass');
   console.log('build  - Lint Sass files and generate minified CSS for production');
 });
 
 /* Gulp default task list */
 gulp.task('default', ['help']);
-gulp.task('develop', ['watch', 'sass:develop', 'jekyll']);
+gulp.task('develop', ['watch', 'sass:develop']);
+gulp.task('jekyll', ['watch', 'sass:develop', 'jekyll:serve']);
```
anthonydillon: I thought we were calling the command server not jekyll?

PR #835, opened 2017-02-08
Update code block styles

anthonydillon:
## Done
Removed the dark code snippet styling and updated the code blocks to match the latest design.

## QA
- Run `gulp jekyll`
- Check the following patterns match the design
 - http://127.0.0.1:4000/vanilla-framework/examples/base/code/
 - http://127.0.0.1:4000/vanilla-framework/examples/patterns/code-numbered/
 - http://127.0.0.1:4000/vanilla-framework/examples/patterns/code-snippet/

## Details
Fixes https://github.com/ubuntudesign/vanilla-framework/issues/818

**Note** - Removal of `p-code-snippet--dark`, replace with `p-code-snippet`


scss/_base_code.scss
```diff
@@ -20,10 +21,10 @@
   }
 
   pre {
-    background: $color-x-dark;
-    border: 0;
-    border-radius: .5rem;
-    color: $color-mid-dark;
+    background-color: $color-light;
+    border: 1px solid $color-mid-dark;
+    border-radius: 2px;
```
anthonydillon: No, the border-radius should alway remain 2px.

anthonydillon: @yaili What do you think?

yaili: Yes, border radius should remain 2px

barrymcgee: It's frustrating to see declarations but no explanations. If all other units in a pattern are relative, why would this unit be static?

yaili: Not all units are relative, border is set to 1px. I think that if you zoom in a page where border radius are set in pixels the border radius is increased anyway. What else do you need to know?

barrymcgee: @yaili so, if there is no difference, why would not just stay consistent and use `rem` everywhere?

anthonydillon: @barrymcgee in my opinion, the border-radius of a component is a direct design attribute. Therefore should not change based on the font-size. A design attribute is like a signal of or style. If the style changes it no longer becomes instantly recognisable as "Vanilla".

Let defer this conversation to dev catch up if you don't mind.

PR #836, opened 2017-02-08
Add gutter to example template

barrymcgee:
## Done

Added a `2rem` margin inline style to the default example layout - without this, patterns sit too tight to the browser chrome to visually assess.

Example without, left border cannot be seen: https://cl.ly/1B2W3z2W3f1u

Example with margin, left border clearly visible: https://cl.ly/2U0511471C2s


barrymcgee:
If we gave it an id, can't be just be removed as part of the JS injection?

anthonydillon:
I would like to keep the `example-js` script simple. I also think we should keep our examples as untouched as possible, so we know truly what is being affected by the framework. 

The use of the examples should be only used by devs which should understand the border is touching the browser chrome.

barrymcgee:
Taking an overall view of the complexity of `example-js`, adding a simple [unwrap](http://stackoverflow.com/questions/19261197/how-can-i-remove-wrapper-parent-without-removing-the-child) function seems pretty trivial?

As for polluting the patterns, this is an inline style completely uncoupled from the framework, so it should never affect any patterns in any way.

It's not a question of understanding, I knew immediately what was happening but think that the developer experience is greatly enhanced by not having to fiddle with inspector every time because the top/left edge of every pattern is butting the viewport?

anthonydillon:
Ok, I feel it's over complicating things but can see it would help developers. I'll create an issue on example-js to add this functionality.

_jekyll/_layouts/default.html
```diff
@@ -12,6 +12,8 @@
     </head>
 
     <body>
-{{ content }}
+      <div style="margin:2rem;">
```
anthonydillon: Can you place an id on here to be targeted by example-js please.

Issue #737, opened 2017-01-12
Button should be able to sit next to inline text


barrymcgee:
This is needed to build out CTA straps downstream. If too complex to achieve here, this should be moved downstream to the brochure theme.

![button--inline](https://cloud.githubusercontent.com/assets/505570/21898925/6077e6ee-d8e6-11e6-8d82-0b5c6879f65e.jpg)

PR #846, opened 2017-02-13
Allow base button to sit inline with text

barrymcgee:
## Done

- Allow base button to sit inline with text

## QA

- Run `gulp jekyll`
- Edit `examples/button.html` to add a chunk of inline text before the button element, like so;
```html
<span>Lorem ipsum dolor sit amet, consectetur adipisicing elit.</span><button>Button</button>
```
- View in browser: `http://127.0.0.1:4000/vanilla-framework/examples/base/button/` 
    

## Details

Fixes: #737


scss/_base_button.scss
```diff
@@ -5,6 +5,10 @@
   button {
     @include button-pattern;
     line-height: 1rem;
+
+    span + & {
+      margin-left: .5rem;
```
anthonydillon: We may want an entire `1rem` here. Difficult without a spec.

anthonydillon:
Interesting idea to use a span to indicate "inlineness".

scss/_base_button.scss
anthonydillon: @yaili do you think it needs 16px or 8px on large screens?

barrymcgee: With `.5rem` of margin-left: https://cl.ly/1k1W0J330q2r

With `1rem` of margin-left: https://cl.ly/2a2S46252z3t

yaili: I think 1rem looks better. Should I create a design spec?

barrymcgee:
@yaili Yes, if we think we need a spec, we probably do..

PR #865, opened 2017-02-15
Updated lists.md

yaili:
## Done

- Added description to each list type
- Renamed definition list to description list (as it's now refered to in w3c standard)
- Added related section at the end

## QA

Go to docs > en > base > lists.md and check sections listed above.

### DIFF
--- a/docs/en/base/lists.md
+++ b/docs/en/base/lists.md
@@ -6,6 +6,8 @@ title: Lists
 
 ## Ordered list
 
+Use an ordered list when the order of the items is important.
+
 <a href="https://ubuntudesign.github.io/vanilla-framework/examples/base/lists/ordered-list/"
     class="js-example">
     View example of the base ordered list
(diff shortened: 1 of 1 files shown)


docs/en/base/lists.md
```diff
@@ -6,21 +6,32 @@ title: Lists
...
 
+Use an unordered list when the order of the items isn't important.
+
 <a href="https://ubuntudesign.github.io/vanilla-framework/examples/base/lists/unordered-list/"
     class="js-example">
     View example of the base unordered list
 </a>
 
-## Definition list
+## Description list
```
anthonydillon: Interesting, it seems there is no set name for them.

Some call them definition lists:
- https://www.w3.org/TR/html401/struct/lists.html#h-10.3

Some call them description lists:
- https://www.w3.org/TR/html5/grouping-content.html#the-dl-element
- https://html.spec.whatwg.org/multipage/semantics.html#the-dl-element

yaili: I think the name changed for HTML5

barrymcgee: Didn't realise that

PR #929, opened 2017-03-06
Move base lists docs to typography

### LINKED ISSUE #880: How can we avoid having two pages called List in docs
If someone is reading our documentation to know which types of lists are available in our framework, it's odd that there are lists in two different pages. In some pages, I've added a section at the bottom of the page with the title "Related" that links to other relevant pages — for example, link to Buttons in Forms. In the case of lists though I think we should only have one page. Is this something that can be done in the documentation? I don't mean to change the way we divide the actual sass files.

barrymcgee:
## Done

Move base lists docs to typography

## QA

- Pull code
- Check docs

## Details

Fixes #880


anthonydillon:
This one has some conflicts.

PR #935, opened 2017-03-08
Fix up breadcrumb links

### LINKED ISSUE #828: Arrow in breadcrumbs not correctly positioned in narrow screens
# 1. Bug template

## Process

1. Navigate to https://docs.vanillaframework.io/patterns/breadcrumbs/
2. In Inspector, replace example label text with labels that have more than one word
3. Narrow screen size until one element is wrapping over 2 lines
4. Arrow is incorrectly positioned

## Expected result

Arrow to keep its place between the two items, tested in Chrome.

## Screenshot

breadcrumbs-arrow

barrymcgee:
## Done

Amended breadcrumb styling to match spec

## QA

- Pull code
- Run `gulp jekyll`
- Compare [example](http://127.0.0.1:4000/vanilla-framework/examples/patterns/breadcrumbs/) to [spec](https://github.com/ubuntudesign/vanilla-design/tree/master/Breadcrumbs)

## Details

Fixes #828

### DIFF
--- a/scss/_patterns_breadcrumbs.scss
+++ b/scss/_patterns_breadcrumbs.scss
@@ -7,20 +7,36 @@
     width: 100%;
 
     &__item {
-      display: inline;
-      font-weight: normal;
-      margin-right: 1rem;
+      display: inline-block;
+      margin-bottom: .25rem;
+      margin-left: .25rem;
+      margin-right: .75rem;
       position: relative;
 
-      &::after {
-        content: '\203A';
-        position: absolute;
-        right: -.75rem;
-        top: -.15rem;
+      &:not(:first-of-type) {
+        margin-right: -.25rem;
+        text-indent: 1rem;
       }
 
-      &:last-child::after {
-        content: ' ';
+      &:first-of-type {
+        margin-right: -.25rem;
+      }
+
+      &:not(:first-of-type)::before {
+        content: '\203A';
+        left: -.75rem;
+        position: absolute;
+        top: 0;
+      }
+    }
+
+    &__link {
+      color: $color-dark;
+      font-weight: 400;
+      text-decoration: none;
+
+      &:hover {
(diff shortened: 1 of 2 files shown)


scss/_patterns_breadcrumbs.scss
```diff
@@ -7,20 +7,27 @@
     width: 100%;
 
     &__item {
-      display: inline;
+      display: inline-block;
       font-weight: normal;
```
anthonydillon: From the design is look like the links are `font-weight: 300` @yaili ?

scss/_patterns_breadcrumbs.scss
```diff
@@ -7,20 +7,27 @@
...
         position: absolute;
-        right: -.75rem;
-        top: -.15rem;
+        top: 0;
       }
+    }
+
+    &__link {
+      color: $color-dark;
+      text-decoration: none;
```
anthonydillon: `a` elements have this rule already. Let's try and avoid overrides in the CSS.

anthonydillon:
QA issue:
On small screens, the sub-links should wrap but be indented so the arrow is aligned with the current page.

barrymcgee:
@anthonydillon Identing doesn't work as you can't be sure how long each breadcrumb is, so you don't know when they will hoist back up inline leaving them with large left margin before the medium breakpoint.

The only alternative is to make them `display:block` on small screens so each item always breaks.

anthonydillon:
What about:
```css
@media only screen and (max-width: $breakpoint-medium) {
    .p-breadcrumbs__item:not(:first-of-type) {
        text-indent: 1rem;
        margin-right: -.25rem;
    }
    .p-breadcrumbs__item:first-of-type {
        margin-right: 0;
    }
}
```

Looks like this:

barrymcgee:
@yaili ?

Issue #878, opened 2017-02-17
Headings have mismatched elements and classes in docs


earnubs:
Visit: https://docs.vanillaframework.io/patterns/headings/

Headings have mismatched elements and classes, eg:

```<h6 class="p-heading--one">Heading six</h6>```

Suggest tag and class should match.

**Comment by barrymcgee** [comment]
@earnubs That's actually deliberate. The purpose of that pattern is to allow users to assign different visual heading styles based on their semantic equivalents. 

For example, in a sidebar, headings may be semantically `<h3>` according to page flow but visually the design dictates they should use a `.p-heading--four` style.

Do you find the explanation on that page ambiguous? If so, perhaps it could be improved?

**Comment by earnubs** [comment]
@barrymcgee Aha! Well that's a great feature that's kind of made to look like a mistake in the docs ;)

In the past I've used class names that didn't imply level (1-6) but style (heavy, mid, light), but maybe in this case a simple line would be enough to help the likes of me:

```
Heading classes are not tied to elements and can be freely mixed, for example you can apply `p-heading--one` style to an `h3` element if that better suits your document style and tree.
```

Perhaps with a regular h1-h6 list before that, and a mixed up example after?

But I'm also happy if you close this issue without change.

barrymcgee:
@earnubs We're aware there is room for improvement docs wise across all patterns and your suggestion sounds like a sensible enhancement to me. Anything which makes it clearer to the user the why & where is  👍🏻

yaili:
Hi, I only noticed this issue now. There are 2 open issues to improve heading documentation:

- https://github.com/vanilla-framework/vanilla-framework/issues/831
- https://github.com/vanilla-framework/vanilla-framework/issues/881

So hopefully this should be done soon. 👍

barrymcgee:
Closed by https://github.com/vanilla-framework/vanilla-framework/pull/936

Issue #831, opened 2017-02-08
Improve heading documentation


yaili:
Current documentation over two pages is confusing, should be merged and simplified:
- https://docs.vanillaframework.io/base/typography/
- https://docs.vanillaframework.io/patterns/headings/

yaili:
Issue/proposal #881 is related to this one.

barrymcgee:
Closed by https://github.com/vanilla-framework/vanilla-framework/pull/936

PR #936, opened 2017-03-08
Update heading docs

### LINKED ISSUE #881: Move the Heading docs to the Typography page
Currently we have the Typography page https://docs.vanillaframework.io/base/typography/ and the Headings page https://docs.vanillaframework.io/patterns/headings/ which I think is confusing.

I don't think we need more than what Boostrap does where they just say you can also use .h1 to .h6 classes: http://getbootstrap.com/css/#type

Thoughts?

barrymcgee:
## Done

- Remove heading examples from base typography
- Clarify headings pattern and explain and how heading classes can be used on different head elements

## QA

- Pull code
- Run `gulp jekyll`
- Review docs changes
- Review http://127.0.0.1:4000/vanilla-framework/examples/patterns/headings/default/
- Review http://127.0.0.1:4000/vanilla-framework/examples/patterns/headings/mixed/

## Details

Fixes #881 #831 #878

### DIFF
--- a/examples/patterns/headings/default.html
+++ b/examples/patterns/headings/default.html
@@ -1,0 +1,12 @@
+---
+layout: default
+title: Headings / Default
+category: _patterns
+---
+
+<h1 class="p-heading--one">Heading one</h1>
+<h2 class="p-heading--two">Heading two</h2>
+<h3 class="p-heading--three">Heading three</h3>
+<h4 class="p-heading--four">Heading four</h4>
+<h5 class="p-heading--five">Heading five</h5>
+<h6 class="p-heading--six">Heading six</h6>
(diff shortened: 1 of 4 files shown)


examples/patterns/headings/default.html
```diff
@@ -0,0 +1,12 @@
...
+title: Headings / Default
+category: _patterns
+---
+
+<h1 class="p-heading--one">Heading one</h1>
+<h2 class="p-heading--two">Heading two</h2>
+<h3 class="p-heading--three">Heading three</h3>
+<h4 class="p-heading--four">Heading four</h4>
+<h5 class="p-heading--five">Heading five</h5>
+<h6 class="p-heading--six">Heading six</h6>
```
anthonydillon: Default styling should have no classes.

barrymcgee: That already exists - http://localhost:4000/vanilla-framework/examples/base/headings/

This is to ensure the test case of semantic elements with semantic classes and semantic elements and mixed classes are covered

anthonydillon: Isn't this tested more thoroughly via the mixed example? We don't want me make people think they much use the classes to style headings.

barrymcgee: At the minute, these styles reflect base styles but that might not always be the case. There's always a fallback for people to use without any classes but this section of the docs is specifically patterns - what you are describing is base styles.

PR #950, opened 2017-03-10
Test change to trigger percy

barrymcgee:


barrymcgee:
Percy now reporting correctly - closing.

PR #951, opened 2017-03-10
Fix title typo

barrymcgee:
## Done

Fix title typo in docs

## QA

Sanity check code


yaili:
We don't use title case in titles, can you please make it Vertically center?

Issue #939, opened 2017-03-08
We should demo vertically centered better


anthonydillon:
We should update the example for vertically centred to demonstrate it better:
https://docs.vanillaframework.io/en/utilities/vertically-center

barrymcgee:
This should be fixed by #942

anthonydillon:
Fixed as part of #942

PR #958, opened 2017-03-14
Use .p-card pattern to demo utils in docs

### LINKED ISSUE #942: Docs need example helper
In the previous version of the docs, certain examples such as the equal height util had a helper to outline the borders of the containers. 

This was different to the markup that was displayed as an example but now that the same markup is used for both, these outlines no longer appear making examples slightly inadequate.

barrymcgee:
## Done

Updated example markup for utils with cards to demo their effects more visually.

Note: I had to alter the styles of `.p-card` itself to use it in this context:

- Add inherent bottom margin of 1rem to a card
- Remove bottom margin from nested cards

## QA

- Pull code
- Run `gulp jekyll`
- View each of the utils at http://127.0.0.1:4000/vanilla-framework/

## Details

Fixes #942 #939

### DIFF
--- a/examples/utilities/equal-height.html
+++ b/examples/utilities/equal-height.html
@@ -4,11 +4,11 @@ title: Equal height
 category: _utilities
 ---
 
-<div class="u-equal-height">
-  <div class="col-8">
+<div class="p-card u-equal-height">
+  <div class="p-card">
     <p>This is a long paragraph of text - Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec at elit augue. Maecenas eleifend varius leo id facilisis. Nunc sit amet hendrerit nisl. Fusce posuere bibendum mi dignissim venenatis. Ut ornare quis velit ac lobortis. Vestibulum faucibus tortor hendrerit viverra. Maecenas in molestie sapien.</p>
   </div>
-  <div class="col-4">
+  <div class="p-card">
     <p>This is a short piece of text</p>
   </div>
 </div>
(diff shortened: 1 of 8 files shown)


examples/utilities/equal-height.html
```diff
@@ -4,11 +4,11 @@
 category: _utilities
 ---
 
-<div class="u-equal-height">
-  <div class="col-8">
+<div class="p-card u-equal-height">
```
anthonydillon: Can you put `.row` here.

examples/utilities/equal-height.html
```diff
@@ -4,11 +4,11 @@
...
+  <div class="p-card">
```
anthonydillon: Add `.col-6` to demo the true use of equal-height

examples/utilities/equal-height.html
```diff
@@ -4,11 +4,11 @@
...
     <p>This is a long paragraph of text - Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec at elit augue. Maecenas eleifend varius leo id facilisis. Nunc sit amet hendrerit nisl. Fusce posuere bibendum mi dignissim venenatis. Ut ornare quis velit ac lobortis. Vestibulum faucibus tortor hendrerit viverra. Maecenas in molestie sapien.</p>
   </div>
-  <div class="col-4">
+  <div class="p-card">
```
anthonydillon: `col-6`

PR #988, opened 2017-03-31
Add ./run build scripts

### LINKED ISSUE #978: Add ./run scripts

barrymcgee:
@nottrobin When trying to run `./run jekyll` I get an error: `Command not recognised.` - any ideas?

## Done

Added run scripts

## QA

Check each command runs without error:

```
./run help
./run develop
./run jekyll
./run test
./run build
```

## Details

Fixes #978


nottrobin:
@barrymcgee please change the "jekyll" script to being called "start" instead - this is a more standard way of running the server.

Then we should tweak the ./run script slightly so that `./run` will default to running `./run start`, like with our django projects. I can do that.
Then we should tweak the ./run script slightly so that `./run` will default to running `./run start`, like with our django projects. I can do that.

barrymcgee:
@nottrobin Ready for re-review 👀

PR #990, opened 2017-03-31
Refactor styling to .p-code-numbered

### LINKED ISSUE #586: [Proposal] Grid-list and image-grid should be renamed
# Pattern amendment

- Amendment

`.p-grid-list` / `.p-image-grid`

## Context

in my opinion, the naming of these patterns is misleading as neither of them have anything to do with the actual grid which could easily lead to confusion given the underpin tenet of all CSS framework tends to be a grid.

## Proposed action

Rename both of these patterns so they don't reference 'grid'

### LINKED ISSUE #856: Update code numbered pattern

- Amendment

If amendment, what is the name of the pattern to be amended?
Code numbered

## Visual

Current build with right-aligned line numbers:
code numbered vanilla framework examples2

Proposal build with background-color behind the line numbers:
code numbered vanilla framework examples

## Context
A recent update to the numbered code block required the line numbers to right-align as to prevent the numbers with double digits pushing out single digit number lines. This requires the number area to be a set width to allow room for the numbers to grow to the left.

This styling makes it clear the numbers are separate from the code content. Giving a clear indication of the length of growth too.

The current line number growth with the attached design is 9999.

barrymcgee:
## Done

Refactor styling to .p-code-numbered

## QA

- Pull code
- Run `gulp jekyll`
- Check [example](http://127.0.0.1:4000/vanilla-framework/examples/patterns/code-numbered/) against the [spec](https://github.com/ubuntudesign/vanilla-design/blob/master/Code/code.md)

## Details

Fixes #856

### DIFF
--- a/scss/_patterns_code-numbered.scss
+++ b/scss/_patterns_code-numbered.scss
@@ -7,7 +7,7 @@
     background: $color-x-light;
     color: $color-dark;
     counter-reset: line-numbering;
-    padding: 1rem 0 0;
+    padding: $sp-medium 0 0;
     position: relative;
 
     &::before {
@@ -19,7 +19,7 @@
       background: $color-light;
       display: block;
       margin: -#{$sp-large} 0 0 0;
-      padding: $sp-x-small 0 $sp-x-small 5.5rem;
+      padding: $sp-x-small $sp-medium $sp-x-small #{$sidebar-width + $sp-medium};
       position: relative;
 
       &:first-child,
--- a/scss/_patterns_code-numbered.scss
+++ b/scss/_patterns_code-numbered.scss
@@ -19,7 +19,7 @@
       background: $color-light;
       display: block;
       margin: -#{$sp-large} 0 0 0;
-      padding: $sp-x-small $sp-medium $sp-x-small #{$sidebar-width + $sp-medium};
+      padding: $sp-x-small $sp-medium 0 #{$sidebar-width + $sp-medium};
       position: relative;
 
       &:first-child,
--- a/scss/_patterns_code-numbered.scss
+++ b/scss/_patterns_code-numbered.scss
@@ -1,21 +1,56 @@
 // Code and pre styles
 @mixin vf-p-code-numbered {
+
+  $sidebar-width: 4.5rem;
+
   .p-code-numbered {
-    background: $color-light;
+    background: $color-x-light;
     color: $color-dark;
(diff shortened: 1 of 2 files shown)


yaili:
Are you sure it's #586?

barrymcgee:
Oops, should be #856

scss/_patterns_code-numbered.scss
```diff
@@ -4,18 +4,44 @@
...
-      counter-increment: line-numbering;
-      display: inline-block;
-      padding-right: $sp-medium; // space after numbers
-      pointer-events: none;
-      text-align: right;
-      user-select: none;
-      width: $sp-xxx-large;
+    code {
+      padding: 0 0 0;
+      position: relative;
```
anthonydillon: It doesn't look like these styles change anything and we should not target element in patterns.

anthonydillon:
When the line wraps it should only wrap within the grey area and the white number area should be the full height. Also on small screens there is a small grey line between the numbers.

barrymcgee:
Ready for re-review 👀

scss/_patterns_code-numbered.scss
```diff
@@ -1,21 +1,56 @@
...
+    &::before {
+      background-color: $color-x-light;
+      width: $sidebar-width;
+    }
+
+    .code-line {
+      background: $color-light;
+      display: block;
+      margin: -#{$sp-large} 0 0 0;
+      padding: $sp-x-small 0 $sp-x-small 5.5rem;
```
anthonydillon: Shouldn't the left padding here be relative to `$sidebar-width + 1rem`

scss/_patterns_code-numbered.scss
```diff
@@ -1,21 +1,56 @@
...
+
+  $sidebar-width: 4.5rem;
+
   .p-code-numbered {
-    background: $color-light;
+    background: $color-x-light;
     color: $color-dark;
     counter-reset: line-numbering;
-    padding: $sp-medium;
+    padding: 1rem 0 0;
```
anthonydillon: Can this `1rem` be a spacing value? `$sp-???`

anthonydillon:
The line spacing between lines of code is too large:

anthonydillon:
The text gets close to the right hand side. I think it would be better to have equal padding.

anthonydillon:
There are small white lines between code on small screens:

PR #1013, opened 2017-04-07
Remove trailing whitespace from examples

barrymcgee:
## Done

Remove trailing whitespace from examples as this whitespace is being improted and respected within code blocks in docs

Note: I had to edit my Atom config to prevent it from adding this trailing newline on save.

## QA

Check only whitespace removed


grahambancroft:
Lgtm 👍

PR #1029, opened 2017-04-21
Fix fixed-width nav example so mobile nav works

barrymcgee:
## Done

Add id to parent element so hide/show works for mobile

## QA

Run `gulp jekyll`, check example in small screen

### DIFF
--- a/examples/patterns/navigation/fixed-width.html
+++ b/examples/patterns/navigation/fixed-width.html
@@ -4,7 +4,7 @@ title: Navigation / Fixed width
 category: _patterns
 ---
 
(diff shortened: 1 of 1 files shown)


yaili:
There is a bug at exactly 768px, see screenshot.

screen shot 2017-04-24 at 10 13 48

barrymcgee:
@yaili Good spot though it's best not to add bugs like this to closed/merged Pull Requests as they can be easily lost... I've added this as a separate issue here: https://github.com/vanilla-framework/vanilla-framework/issues/1038

PR #1030, opened 2017-04-21
Fix full-width nav example so mobile nav works

barrymcgee:
## Done

Add id to parent element so hide/show works for mobile

## QA

Run `gulp jekyll`, check example in small screen


yaili:
At exactly 768px there is a problem, see screenshot.

screen shot 2017-04-24 at 10 13 48

PR #1043, opened 2017-04-26
Rename warning to caution

anthonydillon:
## Done
Renamed warning to caution and deprecated the warning classes and colour.

## QA
- Pull down code and run `gulp jekyll`
- Go to http://127.0.0.1:4000/vanilla-framework/examples/patterns/notifications/caution/
- Check that the notification is now caution, not warning.
- Check the documentation is correct

## Details
Fixes https://github.com/vanilla-framework/vanilla-framework/issues/979

### DIFF
--- a/scss/_patterns_notifications.scss
+++ b/scss/_patterns_notifications.scss
@@ -78,8 +78,8 @@
 }
 
 // Warning notification styling
-@include deprecate('2.0.0', 'Use .vf-notifications-caution instead') {
-  @mixin vf-notifications-warning {
+@mixin vf-notifications-warning {
+  @include deprecate('2.0.0', 'Use .vf-notifications-caution instead') {
     .p-notification--warning {
       @include notification;
       border-color: $color-warning;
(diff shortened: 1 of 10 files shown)


scss/_patterns_notifications.scss
```diff
@@ -64,19 +64,34 @@
...
       padding-left: $sp-large;
     }
   }
 }
 
+// Warning notification styling
+@include deprecate('2.0.0', 'Use .vf-notifications-caution instead') {
```
barrymcgee: Shouldn't the deprecation notice reference the new classname, not the mixin?

anthonydillon:
@barrymcgee moved deprecate inside the warning mixin so you only get it if you cherrypick that mixin.

scss/_patterns_notifications.scss
```diff
@@ -64,19 +64,34 @@
...
       padding-left: $sp-large;
     }
   }
 }
 
+// Warning notification styling
+@mixin vf-notifications-warning {
+  @include deprecate('2.0.0', 'Use .vf-notifications-caution instead') {
```
barrymcgee: Whata i mean is `.vf-notifications-caution` is not a class, it's the name of the mixin. This line should be:

` @include deprecate('2.0.0', 'Use .p-notification-- instead') {`

anthonydillon:
Done.

anthonydillon:
@pmahnke mind approving the percy changes please

PR #1086, opened 2017-05-29
add margin-bottom to form labels + fix padding bug on fields

pmahnke:
## Done

* added margin-bottom: 8px per the [spec](https://github.com/ubuntudesign/vanilla-design/blob/master/Forms/forms.md]
* swapped the vertical and horizontal padding on fields as they were added in the incorrect order per the spec

## QA

1. build the css
2. look at [jekyll demo page](http://127.0.0.1:4000/vanilla-framework/examples/base/forms/input/)
3. see that the label has 8px margin on the bottom and the text field has Padding vertical: 8px and Padding horizontal: 12px per the spec

## Screenshots

image

### DIFF
--- a/scss/_base_forms.scss
+++ b/scss/_base_forms.scss
@@ -163,7 +163,7 @@
   font-size: $sp-medium;
   font-weight: 300;
   outline: none;
-  padding: .5333rem .8rem;
+  padding: $sp-x-small .8rem;
   vertical-align: baseline;
   width: 100%;
 
--- a/scss/_base_forms.scss
+++ b/scss/_base_forms.scss
@@ -5,6 +5,7 @@
     display: block;
     font-size: $sp-medium;
     line-height: 1.5;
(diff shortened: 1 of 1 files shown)


barrymcgee:
Demo: http://vanilla-framework-update-form-label-margin.demo.haus/vanilla-framework/

yaili:
There's a little jump when it goes between large and small screen and I can't figure out why, but there should be no jump, sizes should be the same. Can we fix this?

spencerbygraves:
Demo seems to be down

barrymcgee:
@spencerbygraves Demo should be back up

anthonydillon:
Sorry @pmahnke but the margins have been superseded by the vertical spacing PR. Also, it looks like we don't have a small screen variant of the input padding in the spec.

anthonydillon:
I think this PR can be closed but thank you for helping on Vanilla. If you disagree please reopen and we can talk.

PR #1087, opened 2017-05-29
add header to p-card

### LINKED ISSUE #1052: Document (and build?) card with heading
# 2. Pattern proposal/amendment

Extension of card pattern to have a heading if needed

- Amendment

## Visual

Design specs: https://github.com/ubuntudesign/vanilla-design/tree/master/Card

## Context

Original proposal: https://github.com/vanilla-framework/vanilla-framework/issues/906
Design issue: https://github.com/ubuntudesign/vanilla-design/issues/11

pmahnke:
## Done

* added a header to p-card per the [spec](https://github.com/ubuntudesign/vanilla-design/blob/master/Card/card.md)
* updated the example
* updated the documentation

## QA

1. run locally
2. see the [jekyll example](http://127.0.0.1:4000/vanilla-framework/examples/patterns/card/card/)
3. compare to the spec and [image](https://github.com/ubuntudesign/vanilla-design/blob/master/Card/cards.png)

## Details

Fixes #1052 

## Screenshots

image

### DIFF
--- a/docs/en/patterns/card.md
+++ b/docs/en/patterns/card.md
@@ -29,3 +29,11 @@ The highlighted card should be used when you can interact with the content.
     class="js-example">
     View example of the highlighted card pattern
 </a>
+
+## Optional header and footer
+
+Both cards have styles for a header and footer.
+
+The header will accept either text or an image with a max height of 32 pixels and has a border below.
+
+The footer adds a border above.
--- a/examples/patterns/card/card.html
+++ b/examples/patterns/card/card.html
@@ -5,6 +5,9 @@ category: _patterns
 ---
 
 <div class="p-card">
+  <header class="p-card__header">
+    <img src="http://placehold.it/200x32?text=header image" alt="header image" />
+  </header>
   <h3 class="p-card__title">Title</h3>
   <p class="p-card__content">Lorem ipsum dolor sit amet, consectetur adipisicing.</p>
   <footer class="p-card__footer">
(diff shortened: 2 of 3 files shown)


pmahnke:
@yaili I see the spec says 25px for the images, but the example has a 31px one... and the page that brought this up has a 65px one... perhaps we should remove the max-height on the image in the header to be safe?  what do you think?

yaili:
@pmahnke so the height is manually added case by case?

barrymcgee:
Demo: [http://vanilla-framework-1052-add-heading-to-p-card.demo.haus](http://vanilla-framework-1052-add-heading-to-p-card.demo.haus)

barrymcgee:
Demo: http://vanilla-framework-1052-add-heading-to-p-card.demo.haus/vanilla-framework/

pmahnke:
@yaili I just think that we would likely want to break the 32px limit and that would be an overwrite... we could just see how it goes?  what have we done in other situations?

yaili:
We can see how it goes.

- [ ] The header spec has no font size defined, it should not be constrained to 1.5rem as we might want smaller/larger headings for each case
- [ ] Similarly to the other issue, the card padding changes to 21.3px at smaller screens - it should stay the same

barrymcgee:
@yaili As far as I can see, the card padding is 20px through all breakpoints?

spencerbygraves:
@barrymcgee is the demo for this one down?

spencerbygraves:
@WillMoggridge the demo for this is down, could you help please?

Thanks

anthonydillon:
## Demo
http://vanilla-framework-1052-add-heading-to-p-card.demo.haus/vanilla-framework/examples/patterns/card/card/

PR #1092, opened 2017-05-31
Fix gutters and layout of inline images pattern

grahambancroft:
## Done

Replace wrapping div with and unordered list
Reduce gutters

## QA

- run locally
- see the change via jekyll

## Details

Fixes https://github.com/vanilla-framework/vanilla-framework/issues/1084

### DIFF
--- a/examples/patterns/inline-images.html
+++ b/examples/patterns/inline-images.html
@@ -5,7 +5,7 @@ category: _patterns
 ---
 
 
-<div class="p-inline-images">
+<ul class="p-inline-images">
   <li class="p-inline-images__item">
     <img src="http://placehold.it/150x200" alt="Placeholder image" />
   </li>
@@ -24,4 +24,4 @@ category: _patterns
   <li class="p-inline-images__item">
     <img src="http://placehold.it/175x200" alt="Placeholder image" />
   </li>
-</div>
+</ul>
--- a/scss/_patterns_inline-images.scss
+++ b/scss/_patterns_inline-images.scss
@@ -11,13 +11,15 @@
     &__item {
       display: inline-block;
       margin: 1.875rem;
-      max-width: 6rem;
+      max-height: 5.625rem;
+      max-width: 5.625rem;
+      overflow: hidden;
       text-align: center;
       vertical-align: middle;
-      width: 100%;
 
       @media (min-width: $breakpoint-medium) {
-        margin: $sp-medium;
+        margin: 1.875rem;
+        max-height: 11.25rem;
         max-width: 11.25rem;
       }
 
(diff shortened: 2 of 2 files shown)


yaili:
Will there be a demo link, or shall I wait until it's in develop?

barrymcgee:
Just tried to build a demo here but it doesn't seem to be working? http://vanilla-framework-inline-images.demo.haus/vanilla-framework/ 🤔

@nottrobin Can you help?

deadlight:
@barrymcgee - @WillMoggridge is very close to getting the new demo system working. I think it might be best to wait for that rather than fix the system that's being retired.

anthonydillon:
Testing locally. 

- The gutters are too small. The [pattern spec](https://github.com/ubuntudesign/vanilla-design/blob/master/Lists/list-inline-images.md) states the gutters should be 60px on large screens. This branch leaves the gutters at 32px.

- The spec also states "Items per line: 3" this is confusing. @yaili does this mean to increase the image size with 60px gutters to only display 3 per line on large screens? In that case. Setting the image max-widths should be in percentages or flex.

- On small screens, the images should be set to a max-height of 90px. In this branch, they are 96px. 

- On small screens, the gutter should be 30px between images. You currently have them at 64px.

yaili:
@anthonydillon I think maybe it's better that there is no limit for large screen, and that is determined by container, what do you think?

anthonydillon:
Sounds good. Would you mind updating the spec please.

yaili:
Done

anthonydillon:
Thanks

barrymcgee:
@anthonydillon I've made the QA changes you listed. 

@yaili On small screens, the spec says it should only be 2 per line on small screens.. small screen are all the way up to 768px, only 2 per line while constraining image size to 90x90 leaves acres of whitespace?

barrymcgee:
Demo: http://vanilla-framework-inline-images.demo.haus/vanilla-framework/examples/patterns/inline-images/

spencerbygraves:
On large screen the images are not equal in size:
image

On small screen the images are not centred:
image

I think we could have more that 2 items per line up to 768px but I think @yaili would need to agree.

barrymcgee:
@spencerbygraves Images should not be equal in size as logos, for example, might be different widths. 

These images will currently just fill the container and there is no limitation of two per small screens imposed. @yaili - can you review pls?

yaili:
🤔

Isn't the goal that images/logos are the same size if you use [this pattern](https://github.com/ubuntudesign/vanilla-design/blob/master/Lists/list-inline-images.md)? Or are they just being constrained to a max-width? How is this resizing the images?

Agree with no limitation if more items fit.

barrymcgee:
@yaili They're constrained by max width:

screen shot 2017-06-27 at 11 05 44

yaili:
Sizing of the images seems correct.

To fix @spencerbygraves's point about not being centred, I think we need to remove the default padding and margin left from the ul.

PR #1095, opened 2017-06-01
Implementation of the new vertical rhythm

anthonydillon:
## Done
Converted the vertical rhythm from being based on margin-bottom to margin-top. Updated some spacing and font sizes to match the design.

## QA
- Pull down this branch and run `./run serve`
- Go to http://0.0.0.0:8101/vanilla-framework/examples/base/typo-scale/
- Check that the spacing and font sizes match [the design](https://github.com/vanilla-framework/vanilla-framework/issues/960#issue-214417670)

_Once this branch has landed. I will follow up with the removal of the typo-scale page from examples._

## Details
Fixes: https://github.com/vanilla-framework/vanilla-framework/issues/960


barrymcgee:
@anthonydillon Percy is showing a lot of broken pattern, including the grid - can you review first?

anthonydillon:
It looks like this has caused a lot of regressions and now that my head is hurting, I may need to pick this one up again tomorrow with fresh eyes.

anthonydillon:
@richmccartney @barrymcgee @yaili I am finally happy with the percy output. The changes are expected. Here are a few notes:

- Headings spacing has been reversed. That means that the heading actually dictates the space to the next element. So when we have h1 followed by h2 the h2 will be as close as everything else would to the h1. This is not a bug and makes sense, it will be rare we would have to headings without content next to them. Example of headings here: http://0.0.0.0:8101/vanilla-framework/examples/base/typo-scale/
- I have no idea what is affecting the footer list. They have exactly the same computed styles on every element. 
- There have been some updates to the form spacing to match the latest design.
- I added `p-list--item` to the list of patterns. This has given them some spacing.
- The utilities have more space because the default spacing is 32px.

anthonydillon:
## Demo
http://vanilla-framework-vertical-rhythm.demo.haus/vanilla-framework/

yaili:
@anthonydillon I need to look more carefully, but first:
- What do you mean with your first point, could you explain again, I don't follow, sorry
- I don't understand what you mean is happening with the footer list, is it the one in the footer?

anthonydillon:
> What do you mean with your first point, could you explain again, I don't follow, sorry

I'll try and explain. We no longer simply apply margin-bottom to each heading. We now style anything after the heading with top margin. For example:
``` css
h1 + * {
  margin-top: 2rem;
}
```
Now this means that whether it's a p tag or an h2 directly after the h1. They will both have the same margin-top. I hope this is a little clearer.

> I don't understand what you mean is happening with the footer list, is it the one in the footer?

Yes, the inline list in the footer. You can see there is a diff which Percy is picking up and is there but I could find what was causing it.

spencerbygraves:
On medium screen, h1 is 40px and not 32 as requested:

https://github.com/vanilla-framework/vanilla-framework/issues/960#issuecomment-291483472

yaili:
Thank you @anthonydillon!

Some notes for review:
- [ ] The no bottom margin utility doesn't seem to be doing anything - do we still leave it there anyway, for completeness?
screen shot 2017-06-05 at 15 25 57

- [ ] The same between a label and its input seems about the same as between a label and the previous input, which makes it hard to visually connect labels and their inputs — @spencerbygraves what space do you think we should have in this case?
screen shot 2017-06-05 at 15 25 10

- [ ] At small screen, when there is validation in the form input, there isn't enough space between the validation message and the label that follows - can we have the same space as if the label was preceded but an input? (at large screen it's fine)
screen shot 2017-06-05 at 15 26 50

- [ ] Is there a reason why this h1 has top and bottom margin? (/examples/patterns/headings/default/)
screen shot 2017-06-05 at 15 31 44

- [ ] The class p-heading--one in /examples/patterns/headings/mixed/ and the h1 in /examples/base/headings/ don't match the h1 in the page mentioned above, and the h6 doesn't match p-heading--six either (maybe the rest doesn't match either)

- [ ] Lists items in inline lists shouldn't have margin top, e.g. breadcrumbs
screen shot 2017-06-05 at 15 36 08

- [ ] Figure seems to have some left/right margin of 40px, is this a default margin, can we remove?
screen shot 2017-06-05 at 15 58 10

- [ ] Figure should have the same default top spacing as other elements, it's smaller right now - the hr above is adding the margin bottom, and if you remove the hr below the space is only 16px instead of 20px default space at S screens
- [ ] Figcaption shouldn't have the default spacing from the image above it, it's too much - @spencerbygraves can you recommend a size?
screen shot 2017-06-05 at 16 01 19

We're nearly there! 💯

spencerbygraves:
I thought I'd added these to the issue, I may have only emailed them, apologies.

For the space between a label and it's input, I had 12 and 20px:
image

And the spacing for the caption is included here:
image

anthonydillon:
> On medium screen, h1 is 40px and not 32 as requested:

done.

anthonydillon:
> The no bottom margin utility doesn't seem to be doing anything - do we still leave it there anyway, for completeness?

It is in fact correct. The only margin we have now is margin-top so the margin top util is displaying touching.

anthonydillon:
> At small screen, when there is validation in the form input, there isn't enough space between the validation message and the label that follows - can we have the same space as if the label was preceded but an input? (at large screen it's fine)

Made the spacing above a form validation on small screens 20px as it is on large screens.
forms - validation - vanilla framework - examples 1

anthonydillon:
> Is there a reason why this h1 has top and bottom margin? (/examples/patterns/headings/default/)

*Nope*: my bad. Its gone.

anthonydillon:
> The class p-heading--one in /examples/patterns/headings/mixed/ and the h1 in /examples/base/headings/ don't match the h1 in the page mentioned above, and the h6 doesn't match p-heading--six either (maybe the rest doesn't match either)

They now match.

anthonydillon:
> Figcaption shouldn't have the default spacing from the image above it, it's too much - @spencerbygraves can you recommend a size?

I've gone for 8px margin-top as 10px isn't a multiple. If you feel strongly in the review then I will change to `.625rem`.

anthonydillon:
Updated the field and input spacing to be 12px and 20px.

anthonydillon:
@spencerbygraves @yaili I believe all the updates have been done. Demo has been updated: http://vanilla-framework-vertical-rhythm.demo.haus/vanilla-framework/

yaili:
- [ ] The validation errors in forms have larger font size at smaller screens than at large screens - the font size shouldn't change
- [ ] Should this new way of spacing things mean that the h1 shouldn't have margin top?
- [ ] There should only be two sets of font sizes - [large screen, and medium/small screen](https://github.com/vanilla-framework/vanilla-framework/issues/960#issue-214417670) - but the font sizes change twice as you reduce the viewport here /patterns/headings/default/ - at 1029px and 767px
- [ ] I don't think figure should have margin-bottom, the spacing will be added by the element that follows it, right?
screen shot 2017-06-09 at 14 48 40

spencerbygraves:
- [ ] At 768 there is a different set of margin top values than small screen, which is correct. It should be the same for both. This includes body text
- [ ] Can we use a white background for the base button section

Thanks

anthonydillon:
> Should this new way of spacing things mean that the h1 shouldn't have margin top?

If the H1 is the first element then it gets `16px` margin-top otherwise it gets larger spacing.

anthonydillon:
> There should only be two sets of font sizes - large screen, and medium/small screen - but the font sizes change twice as you reduce the viewport here /patterns/headings/default/ - at 1029px and 767px

I'm only getting a change of font-size at breakpoint `1030px`

anthonydillon:
> Can we use a white background for the base button section

Currently, the buttons have transparency. If we change it to white then if the button is in a different colours background. Then it will show as white. Is this ok?

The reason we changed to transparency was the use of the base button in a blue strip on jujucharms.

spencerbygraves:
@anthonydillon sorry, I meant the page background, so we can see the hover state, not the button background.

anthonydillon:
@yaili @spencerbygraves Updated with a few comments.

yaili:
Some notes:
- I don't follow why the h1 has margin top if it's the first element. If we put an h1 inside a card, it will have margin top?
- I don't see any changes in the demo, has it been updated, is there another link?
- I think I see what's happening, although I don't have a lot of time check if this is right: it seems the font size is changing at 1029px, but margins are changing at 767px - can someone confirm this please? Reduce the viewport slowly to see all the times things change

Thank you

spencerbygraves:
@yaili @anthonydillon I checked the margins earlier and I'm sure it was correct, but now it seems to have gone back.

The margins between 1030 and 768px are wrong, they should be the same as below 768. It is currently showing the medium size specs which have been removed.

anthonydillon:
Done and the demo has been updated: http://vanilla-framework-vertical-rhythm.demo.haus/vanilla-framework/

spencerbygraves:
Sorry @anthonydillon it still looks the same on that demo link. I've tried a hard refresh.

anthonydillon:
@spencerbygraves Could you provide some screens shots, please. Not sure which margins have reverted. Maybe we could catch up tomorrow.

anthonydillon:
Removed the medium margins on headings.

Demo: http://vanilla-framework-vertical-rhythm.demo.haus/vanilla-framework/

examples/patterns/headings/default.html
```diff
@@ -4,9 +4,9 @@
...
 ---
 
-<h1 class="p-heading--one">Heading one</h1>
-<h2 class="p-heading--two">Heading two</h2>
-<h3 class="p-heading--three">Heading three</h3>
-<h4 class="p-heading--four">Heading four</h4>
-<h5 class="p-heading--five">Heading five</h5>
-<h6 class="p-heading--six">Heading six</h6>
\ No newline at end of file
+<h1>Heading one</h1>
```
barrymcgee: If these do not have classes, they should be in /base/

scss/_base_media.scss
```diff
@@ -13,11 +13,21 @@
   }
 
   figure {
+    margin-bottom: 0;
+    margin-left: 0;
     width: 100%;
 
-    caption {
+    @at-root {
```
barrymcgee: Didn't know about this.. nice!

scss/_utilities_floats.scss
```diff
@@ -14,44 +14,52 @@
 
   .u-float-right {
     float: right !important;
+    margin-top: 0;
```
barrymcgee: This breaks our maxim of having utils do only one thing and for that thing is to concisely described in the util. Is this really needed?

barrymcgee:
If you check Percy, on page /vanilla-framework/examples/base/forms/form/index.html - checkboxes are no longer aligned correctly.

barrymcgee:
Radio buttons are now also misaligned - /vanilla-framework/examples/base/forms/radio-buttons/index.html

anthonydillon:
Thanks, fixed your points @barrymcgee.

barrymcgee:
@anthonydillon Great work - approved.. I'll let you push the big green merge, you deserve it! 🎉

anthonydillon:
@barrymcgee @yaili @spencerbygraves Thank you very much.

This will bump a major version in the next release as it will require careful integration into the themes.

PR #1120, opened 2017-06-20
Change grid-gutter-widths

grahambancroft:
## Done

Change layout example of $grid-gutter-widths to $grid-gutter-width as per the variable in _settings_grid.scss

## QA

Make sure I’m not lying and the variable in _settings_grid.scss is actually $grid-gutter-width and not $grid-gutter-widths


barrymcgee:
@grahambancroft Good spot, thanks!

PR #1132, opened 2017-06-26
Modular build

anthonydillon:
## Done
Created a base, settings modular build. Imported settings to the top of each pattern so they can be loaded independently of needing to import other parts of the framework.

## QA
- Run `./run build`
- Check that there are a base.css and settings.css files in the `build/css` directory
- Check there are no errors

## Details
Fixes https://github.com/vanilla-framework/vanilla-framework/issues/412


anthonydillon:
@barrymcgee done.

PR #1136, opened 2017-06-29
Update docs index

barrymcgee:
## Done

Update docs index file from https://docs.google.com/document/d/1OJJHIiTVHDKhXCZOStbsJpf2IXVdMkldsTjiq1tUUYw/edit#

## QA

- Pull code
- Run `./run serve --watch`
- Ensure no errors
- Run process to build docs
- Cross ref with Google doc

### DIFF
--- a/docs/en/index.md
+++ b/docs/en/index.md
@@ -8,7 +8,7 @@ You can use Vanilla in your projects in a few different ways.
 
 ### Install via npm
 
-The recommended way to get Vanilla is through npm:
+The recommended way to get Vanilla is through [npm](https://www.npmjs.com/):
 
 `npm install --save vanilla-framework`
 
--- a/docs/en/index.md
+++ b/docs/en/index.md
@@ -2,21 +2,57 @@
 title: Home
 ---
 
-## A simple extensible CSS framework, written in Sass.
+## Get started
 
-[![Build Status](https://travis-ci.org/vanilla-framework/vanilla-framework.svg?branch=master)](https://travis-ci.org/vanilla-framework/vanilla-framework)
-[![npm version](https://badge.fury.io/js/vanilla-framework.svg)](http://badge.fury.io/js/vanilla-framework)
-[![Downloads](https://img.shields.io/npm/dm/vanilla-framework.svg)](https://www.npmjs.com/package/vanilla-framework)
-[![devDependency Status](https://david-dm.org/vanilla-framework/vanilla-framework/dev-status.svg)](https://david-dm.org/vanilla-framework/vanilla-framework#info=devDependencies)
-[![Chat in #vanilla-framework on Freenode](https://img.shields.io/badge/chat-%23vanilla--framework-blue.svg)](http://webchat.freenode.net/?channels=vanilla-framework)
+You can use Vanilla in your projects in a few different ways.
 
-Vanilla Framework contains a basic CSS grid and pattern classes, and is designed to be extended either directly or by creating extension themes.
+### Install via npm
 
-Sites created within Canonical should follow this style guide closely, whereas external sites are free to adapt and expand the existing components to their needs.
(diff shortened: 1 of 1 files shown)


docs/en/index.md
```diff
@@ -2,21 +2,57 @@
...
-[![Downloads](https://img.shields.io/npm/dm/vanilla-framework.svg)](https://www.npmjs.com/package/vanilla-framework)
-[![devDependency Status](https://david-dm.org/vanilla-framework/vanilla-framework/dev-status.svg)](https://david-dm.org/vanilla-framework/vanilla-framework#info=devDependencies)
-[![Chat in #vanilla-framework on Freenode](https://img.shields.io/badge/chat-%23vanilla--framework-blue.svg)](http://webchat.freenode.net/?channels=vanilla-framework)
+You can use Vanilla in your projects in a few different ways.
 
-Vanilla Framework contains a basic CSS grid and pattern classes, and is designed to be extended either directly or by creating extension themes.
+### Install via npm
 
-Sites created within Canonical should follow this style guide closely, whereas external sites are free to adapt and expand the existing components to their needs.
+The recommended way to get Vanilla is through npm:
```
anthonydillon: `npm` should be a link.

docs/en/index.md
```diff
@@ -2,21 +2,57 @@
...
 
-For ways to use Vanilla framework, see [vanilla-framework/vanilla-framework on GitHub](https://github.com/vanilla-framework/vanilla-framework).
+`npm install --save vanilla-framework`
```
anthonydillon: --save-dev?

docs/en/index.md
```diff
@@ -2,21 +2,57 @@
...
 
-Code licensed [LGPLv3](http://opensource.org/licenses/lgpl-3.0.html) by [Canonical Ltd.](http://www.canonical.com/).
+This will pull down the latest version into your local `node_modules` folder and save it into your project's dependencies in `package.json`.
 
-With ♥ from Canonical.
+Now when you build Sass, make sure to include modules from node_modules. E.g. for node-sass:
```
anthonydillon: Node-sass should be a link.

docs/en/index.md
```diff
@@ -2,21 +2,57 @@
...
+
+```
+// Optionally override some settings
+$color-brand: #ffffff;
+
+// Import the theme
+@import 'vanilla-framework/scss/vanilla';
+```
+
+If you don't need the whole framework, you can just include specific parts, e.g. `@include vf-b-forms`.
```
anthonydillon: you can just `@include` specific parts.

docs/en/index.md
barrymcgee: I commented a long time back on the Google doc that this should be `--save` as VF is a runtime dependency, I seem to remember @nottrobin concurring but it seems to have been lost again in the Google doc.

docs/en/index.md
barrymcgee: I feel this is superfluous, the actual code is three words away - this is describing what the code does, it shouldn't reference the code required.

barrymcgee:
@anthonydillon Feedback addressed - re-review needed pls.

docs/en/index.md
nottrobin: I don't remember concurring. The difference between `--save` and `--save-dev` is that in the former case the dependency will be included in all *child project*. That is, if `my-project` `--save`s `vanilla-framework`, and then `ants-project` adds `my-project` as a dependency, `ants-project` will also get `vanilla-framework`.

I don't think this is at all relevant for the expected use-case. I think in normal websites, `vanilla-framework` is simply a build resource, only relevant for local building, not an actual dependency for the website project itself, which actually just needs some built CSS.

So I think we should use `--save-dev`.

I realise this has now been merged, but I think I'll file an issue about it now.

PR #1156, opened 2017-07-05
Update footer styling to match spec

### LINKED ISSUE #1151: Footer does not match spec
Font size is currently 13px, spec says it should be 14px across all breakpoints. 

Other specs should also be checked.

barrymcgee:
## Done

Update footer styling to match spec

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/footer/
- Cross ref with [spec](https://github.com/ubuntudesign/vanilla-design/blob/master/Footer) and verify example matches design

## Details

Fixes #1151

### DIFF
--- a/scss/_patterns_footer.scss
+++ b/scss/_patterns_footer.scss
@@ -3,10 +3,16 @@
 // Footer pattern styling
 @mixin vf-p-footer {
   .p-footer {
-    border-top: 1px solid $color-mid-dark;
+    border-top: 1px solid $color-mid-light;
+    color: $color-dark;
     font-size: .875rem;
+    line-height: 1.5;
     margin-bottom: $sp-medium;
-    padding-top: $sp-small;
+    padding-top: $sp-medium;
+
+    @media (min-width: $breakpoint-medium) {
+      padding-top: $sp-large;
+    }
 
     &__copy {
       margin-bottom: 0;
@@ -14,12 +20,16 @@
 
     &__links {
       margin: 0;
-      padding: $sp-x-small 0;
+      padding: $sp-small 0 $sp-medium;
 
       @media (min-width: $breakpoint-medium) {
         margin-top: 0;
-        padding: 0;
+        padding: $sp-medium 0 $sp-large;
       }
+    }
+
+    nav {
+      margin-top: 0;
     }
 
     &__item {
@@ -38,7 +48,6 @@
     &__link {
       border-bottom: 0;
       color: $color-dark;
-      font-size: $sp-small;
 
       &:visited {
         color: darken($color-dark, 10%);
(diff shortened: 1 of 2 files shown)


barrymcgee:
Demo: http://vanilla-framework-1151-footer-spec.demo.haus/vanilla-framework/

yaili:
Design review notes:
The bottom padding in L screen should be 24px, but right now it's 16px from the footer container plus 24px from the inline links. Can we remove the bottom padding on the links, and increase the footer container bottom padding to 24px (16px in S screen)?

scss/_patterns_footer.scss
```diff
@@ -3,13 +3,15 @@
 // Footer pattern styling
 @mixin vf-p-footer {
   .p-footer {
-    border-top: 1px dotted $color-x-dark;
-    font-size: $sp-small;
+    border-top: 1px solid $color-mid-light;
+    color: $color-dark;
```
anthonydillon: We need to be careful when setting this color as we are not setting the background color. As this is currently the body color anyway I would suggest removing it.

scss/_patterns_footer.scss
```diff
@@ -18,14 +20,18 @@
...
+      padding: $sp-small 0 $sp-medium;
 
       @media (min-width: $breakpoint-medium) {
         margin-top: 0;
-        padding: 0;
+        padding: $sp-medium 0 $sp-large;
       }
     }
 
+    nav {
```
anthonydillon: Should target only classname in patterns.

scss/_patterns_footer.scss
```diff
@@ -42,7 +48,6 @@
     &__link {
       border-bottom: 0;
       color: $color-dark;
```
anthonydillon: Same here with the color

scss/_patterns_footer.scss
barrymcgee: I only added it as the spec is explicit - if we remove it here, we should remove it there @yaili?

scss/_patterns_footer.scss
barrymcgee: Note, this will make the next release a minor.

scss/_patterns_footer.scss
barrymcgee: By not setting a color here, they will be the default link color, which is blue.

yaili:
The issue is still there:
- [ ] p-footer__links should have no bottom padding
- [ ] p-footer should have 24px bottom padding on L screen, 16px on S screen

barrymcgee:
@yaili Demo didn't update, it should now be updated.

scss/_patterns_footer.scss
anthonydillon: I see, @yaili could we set the background color of the footer pattern?

yaili: @anthonydillon I think the footer should work on a white or grey background without setting colour

PR #1157, opened 2017-07-05
Zero margin bottom on `.u-equal-height` util

### LINKED ISSUE #1152: .u-equal-height util wrapper adds extra margin
Due to vertical rhythm changes:

screen shot 2017-07-05 at 14 49 01

barrymcgee:
## Done

Zero margin bottom if `.u-equal-height` is the only class applied to an element

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/utilities/equal-height/
- Duplicate the code block and verify that `.u-equal-height` does not have bottom margin

## Details

Fixes #1152

### DIFF
--- a/scss/_utilities_equal-height.scss
+++ b/scss/_utilities_equal-height.scss
@@ -9,4 +9,9 @@
       display: flex;
     }
   }
+
+  // Zero margin bottom if .u-equal-height is the only class applied to an element
+  [class="u-equal-height"] {
+    margin-bottom: 0 !important;
+  }
 }


barrymcgee:
@anthonydillon Yes, I mulled it for a while myself but this util only works by adding a wrapper as a single element within a row as per our docs: https://docs.vanillaframework.io/en/utilities/equal-height

This is why I've added it as a specific class attribute selector so that it only applies when added in isolation.

anthonydillon:
True, maybe in this case the equal-height utility should just dictate
`margin: 0`. As it's what we expect from this utility. If it's used in
conjunction with a pattern this may cause issues as utilities trump
patterns.

barrymcgee:
@anthonydillon Wouldn't adding a trump which affects all margins increase the risk of intended consequence rather than just overriding the specific rule we know is added via vertical spacing?

yaili:
Hi, will there be a demo link? It has the design review label.

barrymcgee:
@yaili Demo: http://vanilla-framework-1151-footer-spec.demo.haus/vanilla-framework/examples/utilities/equal-height/

barrymcgee:
@yaili Demo: http://vanilla-framework-1152-equal-height-margin.demo.haus/vanilla-framework/

anthonydillon:
Doesn't this change effectivity add margin-bottom to all uses of equal
height?

barrymcgee:
@anthonydillon  Add margin-bottom? It removes any margin-bottom from any element which only has a class of `.u-equal-height`

anthonydillon:
@barrymcgee the problem with this fix is that someone may add a js hook class or a local class to the div which will result in the margin returning. This would be unexpected and difficult to diagnose.

I think this solution needs more thought. Maybe highlighting a fault in the global vertical spacing all together.

barrymcgee:
@anthonydillon That sounds like an edge case and people are responsible for consequence when adding their own styling regardless as it's beyond or control. 

This may be a systemic issue with vertical spacing as you say but this change fixes the live bug where `.u-equal-height` adds unexpected bottom spacing. 

I think the best course if you're unsure is to merge this and raise a separate issue to revisit addressing spacing issue upstream, prioritising accordingly?

nottrobin:
I really think we need a hard-and-fast rule that vanilla needs to be as self-contained as possible. And that we should never use this technique for that reason.

It's not just the community, it's also within our own team. If one of our own frontends happened to need to use the `u-equal-height` class on, say, a takeover div - they would get it working, then add their own class `my-hot-new-takeover` so they could make the background blue. Or worse yet, they might include a third-party JavaScript library that wants to track clicks on that div, so it adds its own control classes on the div.

It would take that developer hours if not days to work out the difference between their markup and the example pattern, because never ever in the history of CSS has it mattered if you add extra classes.

Vanilla-framework should never be breaking fundamental assumptions that the web is built on.

barrymcgee:
I'm just realising I misread @anthonydillon concern about inadvertently adding a class and the margin returning. Valid point.

anthonydillon:
@barrymcgee I think this solution needs some thought. Let's use the margin-collapse utility for now until we have a true solution.

PR #1173, opened 2017-07-12
Remove the site_title from the docs

anthonydillon:
## Done
Removed the site_title attr from the docs

## QA

- Check all site_titles have been removed.

### DIFF
--- a/docs/en/metadata.yaml
+++ b/docs/en/metadata.yaml
@@ -107,6 +107,9 @@ navigation:
     - location: utilities/floats.md
       title: Floats
 
+    - location: utilities/hide.md
+      title: Hide
+
     - location: utilities/margin-collapse.md
       title: Margin collapse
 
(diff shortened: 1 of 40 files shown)


docs/en/metadata.yaml
```diff
@@ -119,8 +119,8 @@ navigation:
     - location: utilities/vertically-center.md
       title: Vertically center
 
-    - location: utilities/visibility.md
-      title: Visibility
+    - location: utilities/show.md
```
nottrobin: Why is this change here?

nottrobin: Sorry, I discovered. But shouldn't you add "hide.md" too?

anthonydillon: Sorry, should have mentioned. It a drive-by fix. The pattern name was changed but the navigation wasn't.

anthonydillon:
Done.

PR #1182, opened 2017-07-18
Updated breadcrumbs link style

### LINKED ISSUE #1034: Update breadcrumbs link style based on new design spec
1. Go to a page that uses Vanilla breadcrumbs.
2. Try to tell which of the breadcrumbs are links.

Currently, breadcrumbs that are links look quite similar to breadcrumbs that are not. The only difference is in font weight.

Sometimes this difference in weight is fairly obvious, such as in Firefox on Ubuntu:
vanilla breadcrumbs ubuntu

But sometimes it is barely noticable, such as in Safari on iOS:
vanilla breadcrumbs safari

And sometimes, as on [build.snapcraft.io](https://build.snapcraft.io/) build pages, all the breadcrumbs are links. So it wouldn’t matter what weight is used, the breadcrumbs would still look _more similar_ to nearby non-link text than they do to nearby link text:
vanilla breadcrumbs

(Even if repeating the heading at the end of the breadcrumbs was a solution to this problem, that would be using text to solve a problem that could be more concisely solved with visual design.)

I understand that [in navigation menus and other closely bordered areas, it isn’t necessary for links to be coloured and/or underlined](https://www.nngroup.com/articles/guidelines-for-visualizing-links/): the context makes it clear. But breadcrumbs often are not in an area like that.

Linked breadcrumbs become coloured and underlined on hover, but on touch devices there is no hover.

**Pattern proposal:** Unless they are inside a [strip](https://docs.vanillaframework.io/en/patterns/strip), breadcrumb links should be the same colour as normal links.

[Resolving this is part of [build.snapcraft.io issue #653](https://github.com/canonical-ols/build.snapcraft.io/issues/653).]

bartaz:
## Done

Removed custom link style from breadcrumbs pattern, so breadcrumbs links now have regular link style (as per spec).

## QA

- Pull code
- Run `./run serve --watch`
- Go to breadcrumbs example
- Breadcrumbs link should have style as [in spec](https://github.com/ubuntudesign/vanilla-design/tree/master/Breadcrumbs)

## Details

Fixes #1034 

## Screenshots

screen shot 2017-07-18 at 11 44 56

### DIFF
--- a/examples/patterns/breadcrumbs.html
+++ b/examples/patterns/breadcrumbs.html
@@ -5,8 +5,8 @@ category: _patterns
 ---
 
 <ul class="p-breadcrumbs">
-  <li class="p-breadcrumbs__item"><a class="p-breadcrumbs__link" href="#">Breadcrumb One</a></li>
-  <li class="p-breadcrumbs__item"><a class="p-breadcrumbs__link" href="#">Breadcrumb Two</a></li>
-  <li class="p-breadcrumbs__item"><a class="p-breadcrumbs__link" href="#">Breadcrumb Three</a></li>
+  <li class="p-breadcrumbs__item"><a href="#">Breadcrumb One</a></li>
+  <li class="p-breadcrumbs__item"><a href="#">Breadcrumb Two</a></li>
+  <li class="p-breadcrumbs__item"><a href="#">Breadcrumb Three</a></li>
   <li class="p-breadcrumbs__item">Breadcrumb Four</li>
 </ul>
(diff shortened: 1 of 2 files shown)


bartaz:
@barrymcgee With this change the `.p-breadcrumbs__link` class is removed from styles, because breadcrumb links are just regular links.

Should example be updated as well to reflect the change? Or do we want to keep the `.p-breadcrumbs__link` in example (if at some point breadcrumbs links change again we may need it again in future).

barrymcgee:
@bartaz No, remove redundant class from example also please. Thanks!

PR #1187, opened 2017-07-18
Remove explicit color setting from strip mixin

barrymcgee:
## Done

Remove explicit color setting from strip mixin. 

This is because including this as part of the strip mixin menas it is overriding --color variants of strip further downstream.

## QA

- Verify there are no visual changes via Percy

## Related

https://github.com/vanilla-framework/vanilla-brochure-theme/issues/99


grahambancroft:
Lgtm 👍

PR #1191, opened 2017-07-19
Set text color where bg color is already set

barrymcgee:
## Done

Set text color where bg color is already set

## QA

- Check code
- Ensure there are no changes here from Percy

## Details

Fixes https://github.com/vanilla-framework/vanilla-brochure-theme/issues/82


grahambancroft:
Lgtm 👍

PR #1202, opened 2017-07-21
Add aria and accessibility improvements to navigation pattern

### LINKED ISSUE #971: ARIA navigation update
Here is a proposed option for making the navigation far more accessible. After speaking to @anthonydillon I've re-written the navigation in a codepen which we can use with JS to manipulate the ARIA controls to hide / show and display the current page.

https://codepen.io/anon/pen/evrwRB

barrymcgee:
## Done

Add aria and accessibility improvements to navigation pattern

Note: This does not include all the changes as recommended by @richmccartney in his Codepen as I don't think we should add state aria markup that will be invalid unless custom JS scripting is added. 
The reason for this is that many people will copy/paste the navigation snippet from the docs without ever writing the JS needed to make the aria state labels work correctly thus making sites more confusing for screen readers etc. to that end, I've only added aria labels that will always be correct, with or without JS.

## QA

- Pull code
- Review aria changes
- Run `./run serve --watch`
- Review http://0.0.0.0:8101/vanilla-framework/examples/patterns/navigation/dark/
- Review http://0.0.0.0:8101/vanilla-framework/examples/patterns/navigation/light/

## Details

Fixes #971

### DIFF
--- a/examples/patterns/navigation/dark.html
+++ b/examples/patterns/navigation/dark.html
@@ -13,16 +13,16 @@ category: _patterns
         </a>
       </div>
       <a href="#navigation" class="p-navigation__toggle--open" title="menu">Menu</a>
-      <a href="#navigation-closed" class="p-navigation__toggle--close" title="close menu">Menu</a>
+      <a href="#navigation-closed" class="p-navigation__toggle--close" title="close menu">Close Menu</a>
     </div>
-    <nav class="p-navigation__nav">
+    <nav class="p-navigation__nav" role="menubar">
       <span class="u-off-screen">
         <a href="#main-content">Jump to main content</a>
       </span>
-      <ul class="p-navigation__links">
-        <li class="p-navigation__link"><a href="#">Link1</a></li>
-        <li class="p-navigation__link"><a href="#">Link2</a></li>
-        <li class="p-navigation__link"><a href="#">Link3</a></li>
+      <ul class="p-navigation__links" role="menu">
+        <li class="p-navigation__link" role="menuitem"><a href="#">Link1</a></li>
+        <li class="p-navigation__link" role="menuitem"><a href="#">Link2</a></li>
+        <li class="p-navigation__link" role="menuitem"><a href="#">Link3</a></li>
       </ul>
     </nav>
   </div>
--- a/examples/patterns/navigation/dark.html
+++ b/examples/patterns/navigation/dark.html
@@ -13,7 +13,7 @@ category: _patterns
         </a>
       </div>
       <a href="#navigation" class="p-navigation__toggle--open" title="menu">Menu</a>
-      <a href="#navigation-closed" class="p-navigation__toggle--close" title="close menu">Close Menu</a>
+      <a href="#navigation-closed" class="p-navigation__toggle--close" title="close menu">Close menu</a>
     </div>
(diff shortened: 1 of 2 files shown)


examples/patterns/navigation/dark.html
```diff
@@ -13,16 +13,16 @@
         </a>
       </div>
       <a href="#navigation" class="p-navigation__toggle--open" title="menu">Menu</a>
-      <a href="#navigation-closed" class="p-navigation__toggle--close" title="close menu">Menu</a>
+      <a href="#navigation-closed" class="p-navigation__toggle--close" title="close menu">Close Menu</a>
```
anthonydillon: Thank you!!!

anthonydillon: Can you change to `Close menu` please

examples/patterns/navigation/light.html
```diff
@@ -13,16 +13,16 @@
         </a>
       </div>
       <a href="#navigation" class="p-navigation__toggle--open" title="menu">Menu</a>
-      <a href="#navigation-closed" class="p-navigation__toggle--close" title="close menu">Menu</a>
+      <a href="#navigation-closed" class="p-navigation__toggle--close" title="close menu">Close Menu</a>
```
anthonydillon: Close menu please.

examples/patterns/navigation/dark.html
anthonydillon: This should be a button. With `aria-controls="menu"`

examples/patterns/navigation/dark.html
```diff
@@ -13,16 +13,16 @@
...
-    <nav class="p-navigation__nav">
+    <nav class="p-navigation__nav" role="menubar">
       <span class="u-off-screen">
         <a href="#main-content">Jump to main content</a>
       </span>
-      <ul class="p-navigation__links">
-        <li class="p-navigation__link"><a href="#">Link1</a></li>
-        <li class="p-navigation__link"><a href="#">Link2</a></li>
-        <li class="p-navigation__link"><a href="#">Link3</a></li>
+      <ul class="p-navigation__links" role="menu">
```
anthonydillon: Should have an id that aria-controls can target.

examples/patterns/navigation/light.html
```diff
@@ -13,16 +13,16 @@
...
     </div>
-    <nav class="p-navigation__nav">
+    <nav class="p-navigation__nav" role="menubar">
```
anthonydillon: aria-hidden would help here.

barrymcgee: It'll not always be hidden though and requires JS to toggle this attribute?

examples/patterns/navigation/dark.html
barrymcgee: This breaks the :target functionality though?

anthonydillon: I see, good point. Let's leave it as a link for now but I believe we would be better off using JavaScript to toggle the menu in the future.

examples/patterns/navigation/light.html
anthonydillon: Yes as mentioned I believe we need is for this pattern. Ok to be a follow up.

PR #1251, opened 2017-08-07
Improve sup and sub positioning

anthonydillon:
## Done
Made the positioning of a `sup` and `sub` relative to the element not root.

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/base/sup/
- Check that both `sup`s in the title and paragraph text is correct

## Details
Fixes https://github.com/vanilla-framework/vanilla-framework/issues/1109

### DIFF
--- a/examples/base/sub.html
+++ b/examples/base/sub.html
@@ -1,0 +1,8 @@
+---
+layout: default
+title: Sub
+category: _base
+---
+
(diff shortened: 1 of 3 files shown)


barrymcgee:
Think we should also include an example of `<sub>` given we have sets styles for it?

anthonydillon:
@barrymcgee done.

PR #1253, opened 2017-08-08
Port the media-object pattern from brochure theme

anthonydillon:
## Done
Port the media-object pattern from brochure theme

## QA

- Pull code
- Run `./run serve --watch`
- Check that the media-object pattern looks ok

## Details

Fixes https://github.com/vanilla-framework/vanilla-framework/issues/1177

### DIFF
--- a/docs/en/metadata.yaml
+++ b/docs/en/metadata.yaml
@@ -77,6 +77,9 @@ navigation:
   - location: patterns/matrix.md
     title: Matrix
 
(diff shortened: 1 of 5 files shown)


yaili:
col-6 is creating too much left margin on medium/large screen, and it's generally making the pattern jump as you move between breakpoints. This shouldn't happen, do we need col-6 here?

anthonydillon:
@yaili the `col-6` is gone.

PR #1278, opened 2017-08-18
Remove card content spacing

### LINKED ISSUE #1277: Remove margin rules on card content
The card pattern shouldn't affect the spacing of any content within it, as dictated by the [spec](https://github.com/ubuntudesign/vanilla-design/tree/master/Card).

Removing this spacing means that the card footer will also need specific margin-top added.

barrymcgee:
## Done

Remove card content spacing

## QA

- Pull code
- Run `./run serve --watch`
- Check that card content reflects vertical rhythm 

## Details

Fixes #1277

### DIFF
--- a/docs/en/patterns/card.md
+++ b/docs/en/patterns/card.md
@@ -41,10 +41,8 @@ conjunction with a strip image.
   View example of the patterns card overlay
 </a>
 
-## Optional header and footer
+## Optional header
 
-Both cards have styles for a header and footer.
+Both cards have styles for an optional header.
 
 The header will accept either text or an image with a max height of 32 pixels and has a border below.
-
-The footer adds a border above.
--- a/examples/patterns/card/card.html
+++ b/examples/patterns/card/card.html
@@ -10,7 +10,4 @@ category: _patterns
   </header>
   <h3 class="p-card__title">Title</h3>
   <p class="p-card__content">Lorem ipsum dolor sit amet, consectetur adipisicing.</p>
-  <footer class="p-card__footer">
-    Card action or summary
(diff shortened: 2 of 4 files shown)


kwm14:
@barrymcgee 
- what is `p-card__footer` as I can't reference it in the spec for review?
- as `p-card__footer` has a border-top line it sits on the base of the card content which gives the impression it's broken

anthonydillon:
Waiting for design comments to be addressed before I review.

barrymcgee:
@kwm14 The card footer was originally included to allow for use cases where cards needed a CTA, such as on Snapweb where cards are used to display Snaps which then have an "Install" button. This was however, pre-vertical rhythm, so perhaps there is no need for it anymore. I don't know when they were removed from the spec. Should I remove it?

kwm14:
@barrymcgee ah ok that makes sense. I don't have a problem with `p-card__footer` class it just that the dividing line is sitting on the base of above content. We just need to remove margin top from that element for the design to look right. 

screen shot 2017-08-22 at 12 17 20

If we were to remove `p-card__footer` would our vertical rhythm allow users to add CTAs (link or button) to our current card pattern?

barrymcgee:
@kwm14 You mean _add_ margin, right?

I added a simple `<hr>` and `<button>` and it seems with vertical rhythm, that would work just as well. 

screen shot 2017-08-22 at 12 54 28

kwm14:
@barrymcgee yes sorry add not remove.

Your demo using vertical rhythm works very well. Yes let's remove the card-footer style.

Thanks.

PR #1332, opened 2017-09-12
Add aside pattern from docs

### LINKED ISSUE #1217: Combine docs theme aside pattern

barrymcgee:
## Done

Add aside pattern from docs

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/aside/
- Check pattern matches spec

Fixes #1217

### DIFF
--- a/examples/patterns/aside.html
+++ b/examples/patterns/aside.html
@@ -9,10 +9,10 @@ category: _patterns
     <h4 class="p-aside__header">Insert title here</h4>
     <nav class="p-aside__nav">
       <ul class="p-list">
-        <li><a class="nav__link" href="#link1">Link 1</a></li>
-        <li><a class="nav__link" href="#link2">Link 2</a></li>
-        <li><a class="nav__link is-active" href="#link3">Link 3 (Active)</a></li>
-        <li><a class="nav__link" href="#link4">Link 4</a></li>
+        <li><a class="p-aside__link" href="#link1">Link 1</a></li>
+        <li><a class="p-aside__link" href="#link2">Link 2</a></li>
+        <li><a class="p-aside__link is-active" href="#link3">Link 3 (Active)</a></li>
+        <li><a class="p-aside__link" href="#link4">Link 4</a></li>
       </ul>
     </nav>
   </div>
--- a/examples/patterns/aside.html
+++ b/examples/patterns/aside.html
@@ -4,9 +4,16 @@ title: Aside
 category: _patterns
 ---
 
-<aside id="side-content" class="p-aside">
+<aside class="p-aside">
   <div class="p-aside__section">
-    <h4 class="p-aside__header">Lorem Ipsum</h4>
-    <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit. In dolorum soluta, tempore quo laboriosam ipsam deserunt! Odit enim voluptas eum, aut facere. Ut amet, labore totam! Dicta voluptatum tempore, doloribus.</p>
+    <h4 class="p-aside__header">Insert title here</h4>
+    <nav class="p-aside__nav">
+      <ul class="p-list">
+        <li><a class="nav__link" href="#link1">Link 1</a></li>
+        <li><a class="nav__link" href="#link2">Link 2</a></li>
+        <li><a class="nav__link is-active" href="#link3">Link 3 (Active)</a></li>
+        <li><a class="nav__link" href="#link4">Link 4</a></li>
+      </ul>
+    </nav>
   </div>
 </aside>
--- a/examples/patterns/aside.html
+++ b/examples/patterns/aside.html
@@ -6,19 +6,7 @@ category: _patterns
 
 <aside id="side-content" class="p-aside">
   <div class="p-aside__section">
-    <h4 class="p-aside__header">Contents</h4>
-    <nav>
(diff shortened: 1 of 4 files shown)


anthonydillon:
@barrymcgee I can't see an `Aside` design spec. Could you link to it, please? Also, doesn't this PR fix an issue?

examples/patterns/aside.html
```diff
@@ -0,0 +1,24 @@
...
+layout: default
+title: Aside
+category: _patterns
+---
+
+<aside id="side-content" class="p-aside">
+  <div class="p-aside__section">
+    <h4 class="p-aside__header">Contents</h4>
+    <nav>
+      <ul class="p-toc">
```
anthonydillon: We should not use cross patterns in the examples as the user may not have `p-toc` and will be confused when this does not look like the example.

scss/_patterns_aside.scss
```diff
@@ -0,0 +1,58 @@
...
+    align-self: flex-start;
+    border-top: 1px solid $color-mid-light;
+    flex: 0 0 16rem;
+    font-size: .875rem;
+    padding: 0 $sp-medium;
+
+    @media (min-width: $breakpoint-medium) {
+      border-left: 1px solid $color-mid-light;
+      border-top: 0;
+      margin-top: $sp-xxx-large * 2;
```
anthonydillon: Why do we need such a large margin-top? This is probably going to have to be overridden in most other cases.

scss/_patterns_aside.scss
```diff
@@ -0,0 +1,58 @@
+@import 'settings';
+
+@mixin vf-p-aside {
+  .p-aside {
+    align-self: flex-start;
+    border-top: 1px solid $color-mid-light;
+    flex: 0 0 16rem;
```
anthonydillon: Shouldn't we use the grid to define the width of this pattern? All patterns should be fluid within its parent container.

scss/_patterns_aside.scss
```diff
@@ -0,0 +1,58 @@
...
+        border-bottom: 1px dotted $color-mid-light;
+      }
+    }
+
+    &__nav {
+      list-style: none;
+      margin: 0;
+      padding: 0;
+
+      .nav__link {
```
anthonydillon: Is this part of a separate pattern called `nav`?

scss/_patterns_aside.scss
barrymcgee: This was part of the pattern in vanilla-docs-theme - I didn't amend it.

scss/_patterns_aside.scss
barrymcgee: As lifted from vanilla-docs-theme, this is a pattern that is positioned outside of the page flow so it will be fixed to the top of the viewport as a user scrolls which is why I presume it is not in the grid to start with?

scss/_patterns_aside.scss
barrymcgee: No - but to nest further within this pattern would create grandchild nesting - `.p-aside__nav__link`

scss/_patterns_aside.scss
anthonydillon: I think this should be removed as the docs theme was created quickly and without the full process. I think navigation should be fluid with its container.

scss/_patterns_aside.scss
anthonydillon: Again I think this should also be removed. It limits the use of this pattern in other situations.

scss/_patterns_aside.scss
anthonydillon: You shouldn't continue to nest elements with BEM. Instead, modify the name of the element. In this case, I would suggest something like `p-aside__nav-link` or simply `p-aside__link`.

scss/_patterns_aside.scss
```diff
@@ -0,0 +1,55 @@
...
+        &:visited {
+          color: $color-dark;
+        }
+
+        &:hover {
+          color: $color-link;
+        }
+
+        &.is-active {
+          font-weight: bold;
```
anthonydillon: `.p-link--strong` is `font-weight: 400;`.

scss/_patterns_aside.scss
```diff
@@ -0,0 +1,55 @@
...
+    font-size: .875rem;
+    padding: 0 $sp-medium;
+
+    @media (min-width: $breakpoint-medium) {
+      border-left: 1px solid $color-mid-light;
+      border-top: 0;
+      padding: 0 $sp-x-large;
+    }
+
+    &__header {
```
anthonydillon: Should be `line-height: 1.5` (24px)

anthonydillon: Padding should also be `1rem`

scss/_patterns_aside.scss
```diff
@@ -0,0 +1,55 @@
+@import 'settings';
+
+@mixin vf-p-aside {
+  .p-aside {
+    border-top: 1px solid $color-mid-light;
+    font-size: .875rem;
+    padding: 0 $sp-medium;
```
anthonydillon: Padding should be `1.5rem`

kwm14:
LGTM 👍

PR #1333, opened 2017-09-12
Add image pattern

### LINKED ISSUE #1220: Combine docs theme image pattern

barrymcgee:
## Done

Add image pattern 

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/image/bordered/ and http://0.0.0.0:8101/vanilla-framework/examples/patterns/image/shadowed/
- Cross ref with the [spec](https://github.com/ubuntudesign/vanilla-design/blob/master/Image/image.md)

## Details

Fixes #1220

### DIFF
--- a/examples/patterns/image/bordered.html
+++ b/examples/patterns/image/bordered.html
@@ -4,5 +4,5 @@ title: Image / with border
 category: _patterns
 ---
 
-<img class="p-image--bordered" src="http://via.placeholder.com/350x350?text=*" alt="" class="p-image--bordered">
+<img class="p-image--bordered" src="http://via.placeholder.com/350x350?text=*" alt="">
 
--- a/examples/patterns/image/bordered.html
+++ b/examples/patterns/image/bordered.html
@@ -1,0 +1,8 @@
+---
+layout: default
+title: Image / with border
+category: _patterns
+---
+
+<img class="p-image--bordered" src="http://via.placeholder.com/350x350?text=*" alt="" class="p-image--bordered">
+
--- a/examples/patterns/image/shadowed.html
+++ b/examples/patterns/image/shadowed.html
@@ -1,0 +1,8 @@
+---
+layout: default
+title: Image / with shadow
+category: _patterns
+---
+
+<img class="p-image--shadowed" src="http://via.placeholder.com/350x350?text=*" alt="">
+
--- a/scss/_base_media.scss
+++ b/scss/_base_media.scss
@@ -5,6 +5,7 @@
 
   img {
     border: 0;
+    border-radius: 2px;
     height: auto;
     max-width: 100%;
   }
--- a/scss/_patterns_image.scss
+++ b/scss/_patterns_image.scss
@@ -1,0 +1,18 @@
+@import 'settings';
+
+@mixin vf-p-image {
+  .p-image {
+
+    &--bordered {
+      border: {
+        color: $color-mid-light;
+        style: solid;
+        width: 1px;
+      }
+    }
+
+    &--shadowed {
+      box-shadow: 0 1px 5px 1px rgba(51, 51, 51, .2); // sass-lint:disable-line no-color-literals
+    }
+  }
+}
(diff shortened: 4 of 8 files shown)


examples/patterns/image/bordered.html
```diff
@@ -0,0 +1,8 @@
+---
+layout: default
+title: Image / with border
+category: _patterns
+---
+
+<img class="p-image--bordered" src="http://via.placeholder.com/350x350?text=*" alt="" class="p-image--bordered">
```
anthonydillon: Double class.

scss/_patterns_image.scss
```diff
@@ -0,0 +1,18 @@
...
+    &--bordered {
+      border: {
+        color: $color-mid-light;
+        style: solid;
+        width: 1px;
+      }
+    }
+
+    &--shadowed {
+      box-shadow: 0 1px 5px 1px rgba(51, 51, 51, .2); // sass-lint:disable-line no-color-literals
```
anthonydillon: You should use transparentize here http://sass-lang.com/documentation/Sass/Script/Functions.html#transparentize-instance_method here

anthonydillon:
You are also missing the docs navigation update.

scss/_base_media.scss
```diff
@@ -5,6 +5,7 @@
 
   img {
     border: 0;
+    border-radius: 2px;
```
anthonydillon: @kwm14 I'm concerned we don't want border radius on every image on every site. I feel that is very opinionated and likely to lead to a lot of overrides. I think it's fine if to style image patterns with border radius.

examples/patterns/image/shadowed.html
```diff
@@ -0,0 +1,8 @@
+---
+layout: default
+title: Image / with shadow
+category: _patterns
+---
+
+<img class="p-image--shadowed" src="http://via.placeholder.com/350x350?text=*" alt="">
```
kwm14: Instead of using an asterisk (*) as text could we update to use the word 'Vanilla'

scss/_base_media.scss
kwm14: @anthonydillon originally me and Spencer wanted consistency as we use border-radius on buttons, cards etc so that was the reasoning behind it.

Speaking with Spencer this morning we're happy to have border-radius as a style to add onto a default un-styled image.

examples/patterns/image/shadowed.html
barrymcgee: I wanted to remove text entirely but couldn't - the reason for this is that these text is liable to minute changes (which are outside of our control) on a  regular basis which can knock off Percy

kwm14: Ok no worries let's leave it as is then

scss/_base_media.scss
barrymcgee: Given it's such a minimal visual tweak - Is anyone really gonna add a class to add 2px of radius to an image? I really doubt it. Either we should do it site-wide or not at all?

scss/_patterns_image.scss
```diff
@@ -0,0 +1,18 @@
...
+    &--bordered {
+      border: {
+        color: $color-mid-light;
+        style: solid;
+        width: 1px;
+      }
+    }
+
+    &--shadowed {
+      box-shadow: 0 1px 5px 1px transparentize(rgba($color-mid-light, .2)); // sass-lint:disable-line no-color-literals
```
anthonydillon: Transparentize needs two parameters. A colour (`$color-mid-light`) and a ratio to make transparent. The second works in reverse to how rgba works. Therefore, to achieve the same CSS you need to use `.8`.

https://www.sassmeister.com/gist/729f518a3c8bacfef71b64571c70f200

You can then remove the disable of the sass-lint.

barrymcgee: Didn't know this -thanks!

anthonydillon:
I don't understand why Percy is reporting so many changes

kwm14:
Looking good 👍

PR #1341, opened 2017-09-14
Remove !important from animations util

### LINKED ISSUE #1323: Investigate animations with Percy
Percy is picking up animated icon as a diff every time as the icon is rotated differently each snap shot. 

Is there a way to stop animations in examples for Percy or do we need to move the animated patterns to there own example and have Percy ignore it?

barrymcgee:
This is because Percy, the visual regression tool, needs to be able to override this rule to provide accurate test results.

## Done

Remove !important from animations util

## QA

- Check that Percy returns no regressions on this PR.

## Details

Fixes #1323

### DIFF
--- a/scss/_utilities_animations.scss
+++ b/scss/_utilities_animations.scss
@@ -1,7 +1,7 @@
 /// Animations
 @mixin vf-u-animations {
   .u-animation--spin {
-    animation: spin map-get($animation-duration, sleepy) infinite linear !important;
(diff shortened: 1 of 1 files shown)


barrymcgee:
This hasn't resolved the issue - have reached out to the Percy support team.

anthonydillon:
Maybe remove the animation from the spinner. That would limit the test and confirm it was the animation.

PR #1345, opened 2017-09-14
Removed margins from images using image-position util

### LINKED ISSUE #1339: image-position pattern needs to remove margin
There is no case where we want `margin` on an image that is positioned using the `image-position` pattern. It should be set to `0`.

Example:
https://codepen.io/anthonydillon/pen/JroErj

barrymcgee:
## Done

- Removed margins from images using image-position util
- Refactored code to remove duplication by using scoped %placeholder

## QA

- Sanity check code
- Verify no Percy regressions on this PR

## Details

## Screenshots

Fixes #1339

### DIFF
--- a/scss/_utilities_image-position.scss
+++ b/scss/_utilities_image-position.scss
@@ -1,31 +1,36 @@
 // Create utility class for positioning images at top or bottom of the section
-
 @mixin vf-u-image-position {
   .u-image-position {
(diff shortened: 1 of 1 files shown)


pmahnke:
@barrymcgee did you see there are some percy diffs? do you know if they are significant?

PR #1361, opened 2017-09-28
Remove !important from animations util

### LINKED ISSUE #1323: Investigate animations with Percy

barrymcgee:
## Done

This is becuase !important was overriding the Percy styles which pause the animations for accurate regression testing.

## QA

- Check that Percy reports no visual changes below.

## Details

Fixes #1323


barrymcgee:
As this has not resolved the issue with Percy reporting changes with the animating icon, I've raised a support issue with the Percy team.

PR #1374, opened 2017-10-10
Increase threshold for text color contrast in navigation pattern

### LINKED ISSUE #1373: Change the navigation color threshold
The current threshold is set to 50%. This results in Ubuntu orange have dark text which is not what we want. The threshold should be slightly adjusted to 70% instead.

barrymcgee:
## Done

Increase threshold for text color contrast in navigation pattern

## QA

- Pull code
- Run `./run serve --watch`
- Verify navigations pattern display correctly

## Details

Fixes #1373


grahambancroft:
Lgtm 👍

PR #1383, opened 2017-10-18
Update vanilla import from 'vanilla' to 'build'.

squidsoup:


barrymcgee:
As discussed with @squidsoup on the [Vanilla Framework Slack group](http://vanillaframework.slack.com), using:

`@import 'vanilla-framework/scss/build';` 

...is still the recommended way to use Vanilla with Sass.

PR #1390, opened 2017-10-20
Remove CSS mask, use icon mixin instead

### LINKED ISSUE #1380: .p-link still uses CSS mask despite limited browser support
CSS Mask should not be used as it is not supported in IE at all and has very limited support in Firefox.

barrymcgee:
## Done

Remove CSS mask, use icon mixin instead

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/links/links-external/
- See that external links now match the spec

## Screenshots

screen shot 2017-10-20 at 09 41 05

### DIFF
--- a/docs/en/patterns/links.md
+++ b/docs/en/patterns/links.md
@@ -14,9 +14,6 @@ The `.p-link--external` class should be used on hyperlinks that go to a differen
     View example of the external link pattern
 </a>
 
-!!! Note:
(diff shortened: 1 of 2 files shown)


anthonydillon:
@barrymcgee I think we need to spend more time on this solution as there are a number situations we need it to work in.

PR #1405, opened 2017-10-24
Decouple strips from external links

anthonydillon:
## Done
Removed the cross-linking of strips and external links and used currentColor to use the strip color to style the external link. Created a layout to test all combinations of the external link.

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/templates/external-links/
- Check that all external links are styled as expected.

_There is some issues with base button but that is out of scope for this issue_

## Details
Fixes https://github.com/vanilla-framework/vanilla-framework/issues/1378

### DIFF
--- a/docs/en/patterns/links.md
+++ b/docs/en/patterns/links.md
@@ -28,7 +28,7 @@ The `.p-link--soft` class should be used on hyperlinks where many links are grou
 
 ## Strong link
 
-The `.p-link--strong` class should be used on hyperlinks that require emphasis.
+The `.p-link--strong` class should be used on hyperlinks that require emphasis or on a dark background.
 
 <a href="https://vanilla-framework.github.io/vanilla-framework/examples/patterns/links/links-strong/"
     class="js-example">
--- a/examples/templates/external-links.html
+++ b/examples/templates/external-links.html
@@ -1,0 +1,96 @@
+---
+layout: default
+title: External links
+category: _templates
+---
+
+<div class="p-strip">
+  <div class="row">
+    <div class="col-12">
+      <p><a href="#">Plain link</a></p>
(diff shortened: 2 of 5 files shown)


kwm14:
## Visual review

- [ ] remove the word "plain" from first two example links so it just reads 'Link' and 'External link'

Some text links don't pass WCAG 2.0 color contrast standards. We require a contrast ratio of 4.5:1 to pass AA accessibility. So we need to set a rule that the links listed below can't be applied to darker backgrounds. 

### p-strip--dark
- [ ] remove '**plain_link**' not AA accessible on this background 
- [ ] remove '**p-link--external**' not AA accessible on this background
- [ ] remove '**p-button--base**' not visible or AA accessible on this background

### p-strip--accent
- [ ] remove '**plain_link**' not AA accessible on this background 
- [ ] remove '**p-link--external**' not AA accessible on this background
- [ ] remove '**p-button--base**' not visible or AA accessible on this background

### p-strip--image is-dark
- [ ] remove '**plain_link**' not AA accessible on this background 
- [ ] remove '**p-link--external**' not AA accessible on this background
- [ ] remove '**p-button--base**' not visible or AA accessible on this background

anthonydillon:
@kwm14 thanks for looking into the accessibility of the links but when using a dark strip your are suppose to use the strong link. I just added every combination to test they all work as expected.

kwm14:
@anthonydillon thanks for updating and your template works well in displaying all link examples, 

**Solutions from discussion:**

- Link (blue) if used on a light background
- Link (white) if used on a light background
- Now both link examples solve accessibility issues on either light/dark backgrounds

PR #1437, opened 2017-11-20
Divider pattern proposal

lyubomir-popov:
## Done

Divider pattern - move border from right edge of preceding column to left edge of current column  so that it can be placed inside the margin (closer to the midpoint between two columns), rather than on the left edge of the preceding column.

## QA

- Pull code
- Run `./run serve --watch`
http://0.0.0.0:8101/vanilla-framework/examples/patterns/divider/
- [Add additional steps]

## Details
Currently, due to the narrow grid gutters, divider lines are placed on the left of the gutter, and content in the preceding column is offset with a 1rem padding-right. With the proposed wider gutters, dividers can go inside the margin, closer to (or precisely in) the midpoint between the two columns. (Otherwise, the space between the divider and the left-aligned text becomes a bit wide.)

## Screenshots

Before: 
image

After:
image


lyubomir-popov:
Closing as this will be included in the fixes on the gutter-width PR

PR #1493, opened 2017-12-14
Add Flask app to serve docs

WillMoggridge:
## Done

Adding Flask app to serve docs.
An update to ./run will come later to allow it to be served through it as a secondary site.

## QA

- Pull code
(Can't use ./run just yet 😞)
- `yarn run build-docs`
- Install python from requirements.txt in your favourite Python env manager
- `FLASK_APP=docs/app.py flask run`
- Follow address in terminal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,4 @@ env[23]/
 js/app.js.map
 
 /docs/build
+/docs/templates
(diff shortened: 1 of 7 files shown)


anthonydillon:
Flask works but the examples are not working.

WillMoggridge:
@anthonydillon Updated with template copied from existing docs repo and updated documentation-builder.

Issue #1416, opened 2017-11-02
Modify ./run script to add "serve-docs" endpoint


nottrobin:
This has not been done before, so it might take a bit of discussion.

Basically, vanilla-framework already has a `./run` function, which shows the examples. We want to add one to serve the documentation.

Maybe the documentation should be the default one (`./run` or `./run serve`), and examples should be a separate one (`./run serve-examples`)? Or maybe to avoid confusion they should both be explicit (`./run serve-docs` and `./run serve-examples`)?

Then there's the question of how we deal with the fact that we had been trying to make the `run` script standard without local modifications. Is there any way that this could be done such that the `run` script could still remain unmodified?

Anyway, decide what to do and do it. Discuss it with people as need be.

PR #1494, opened 2017-12-14
Dockerfile for docs

WillMoggridge:
This should be merged after #1493 

## Done

Add a docker image for the docs app

## QA

- Pull code
- Install latest documentation-builder (`snap install documentation-builder`)
- `yarn run build-docs`
- `cd docs`
- `docker build . -t vanilla-docs --build-arg COMMIT_ID=$(git rev-parse HEAD)`
- `docker run --rm -p 8080:80 vanilla-docs`
- Visit http://localhost:8080/ and it should look beautiful!


nottrobin:
> ./run exec yarn run build-docs

Why isn't there a native ./run command for this?

nottrobin:
``` bash
$ ./run exec yarn run build-docs

yarn run v1.2.1
$ documentation-builder --base-directory docs --output-path docs/templates --template-path docs/template.html --tag-manager-code 'GTM-K92JCQ' --no-link-extensions --force
/bin/sh: 1: documentation-builder: not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I get this initially. This is because the `./run exec` command doesn't check for and install dependencies first. Which highlights the need for `build-docs` to have its own native endpoint.

nottrobin:
Givan that the Dockerfile is in the `docs/` folder, which seems like a good idea to me, I expected to be able to do:

``` bash
docker build docs -t vanilla-docs --build-arg COMMIT_ID=`git rev-parse HEAD`
```

But then I get:

``` bash
Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
```

Was there a reason why you put the `requirements.txt` at the top level instead of inside the `docs/` folder?

WillMoggridge:
> Why isn't there a native ./run command for this?
I am working to create one in #1416. I wanted to move the other items along while that gets created.

> Was there a reason why you put the `requirements.txt` at the top level instead of inside the `docs/` folder?
This is because the ./run command currently looks for it in the root (until the update) but I do think it would be better in the docs folder as well.

nottrobin:
> This is because the ./run command currently looks for it in the root (until the update) but I do think it would be better in the docs folder as well.

At the moment, the run script doesn't really help you run the docs flask app at all, so could you just move the `requirements.txt` file to where you think it should live and then craft QA steps independently of `./run` until `./run` supports it?

WillMoggridge:
Yeah sure, I will update this shortly.

PR #1498, opened 2017-12-19
1452: Change footer to move to bottom of page if not enough content

### LINKED ISSUE #1452: Layouts: Sticky footer
This is a proposal for a new pattern (sticky footer) or amendment to current footer pattern to optionally make it sticky to to bottom of the screen (when there is not enough content on the page).

## Visual

screen shot 2017-11-21 at 13 16 40
screen shot 2017-11-21 at 13 16 32

## Context

On pages that can have very little content it may be useful to have footer stick to the bottom.

It was implemented in snapcraft.io https://github.com/canonical-websites/snapcraft.io/pull/151
(demo at: https://staging.snapcraft.io/search)

## Progressive enhancement

Snapcraft.io implementation uses flexbox (based on https://css-tricks.com/couple-takes-sticky-footer/#article-header-id-3). For browsers not supporting flexbox it will simply show a footer right after the content without breaking the layout.

Caleb-Ellis:
## Done

- Changed default footer patter to be 'sticky' (i.e. it moves to the bottom of page if there's not enough content to fill a browser window)

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/footer/
- Check that the footer is at the bottom of the browser window
- Check it's compatible with the [browsers we support](https://github.com/canonical-webteam/practices/blob/master/coding/browser-support.md)
- **Note: All body tags in the examples have an inline style of `margin: 1rem` so the window height is actually 100%+2rem, meaning a scrollbar appears where it wouldn't normally**

## Details

Fixes #1452

### DIFF
--- a/_jekyll/_layouts/default.html
+++ b/_jekyll/_layouts/default.html
@@ -13,6 +13,6 @@
     </head>
 
     <body style="margin: 1rem;">
-{{ content }}
+      {{ content }}
     </body>
 </html>
--- a/scss/_patterns_footer.scss
+++ b/scss/_patterns_footer.scss
@@ -6,8 +6,21 @@
     border-top: 1px solid $color-mid-light;
     font-size: .875rem;
     line-height: 1.5;
+    margin-top: auto;
     padding-bottom: $sp-medium;
     padding-top: $sp-medium;
+
+    @at-root {
+      body,
+      html {
+        height: 100%;
+      }
+
+      body {
+        display: flex;
+        flex-direction: column;
+      }
+    }
 
     @media (min-width: $breakpoint-medium) {
       padding-bottom: $sp-large;


Caleb-Ellis:
I had a feeling changing the body to display: flex and height: 100% would cause a few random issues, but I didn't think it was going to break literally everything :sob:

bartaz:
It seems that in examples we have different stuff directly in body, which is affected by the flex layout.

If (most likely in real website examples) inside of the body are layout patters such as grid, strips, etc it shouldn't be an issue.

And in many cases Percy seems to show the diff on whole page while visually (comparing them manually) there doesn't seem any difference. But maybe flex is changing display a bit in a way that affects Percy.

_jekyll/_layouts/default.html
```diff
@@ -12,7 +12,9 @@
         {% endif %}
     </head>
 
-    <body style="margin: 1rem;">
-{{ content }}
+    <body>
+      <div style="margin: 0 1rem 1rem 1rem;">
```
anthonydillon: This mark up will show up in the example code snippet as `example-js` displays the contents of the body.

scss/_patterns_footer.scss
```diff
@@ -6,6 +6,7 @@
     border-top: 1px solid $color-mid-light;
     font-size: .875rem;
     line-height: 1.5;
+    margin-top: auto;
```
anthonydillon: This looks like a good use case for `@root`. Root allows you to add styling to parent elements from within a pattern. This means if someone doesn't include the footer pattern then it wont apply the height to the body.

https://www.sassmeister.com/gist/5cd04a820b838320f16a9ce29254e103

_jekyll/_layouts/default.html
```diff
@@ -12,7 +12,13 @@
         {% endif %}
     </head>
 
-    <body style="margin: 1rem;">
-{{ content }}
+    <body>
+      {% if page.title == 'Footer' %}
+      {{ content }}
+      {% else %}
+      <div style="margin: 1rem;">
```
anthonydillon: This mark up will be displayed in the foot example code snippet. Also by removing the margin from all over patterns the edges of the patterns will be hidden.

scss/_patterns_footer.scss
Caleb-Ellis: @anthonydillon I didn't know this was a thing. Thanks!

anthonydillon:
@Caleb-Ellis I think the movement is ok. I would like to see the output of Percy if you removed the margin div from the examples and added `style="margin: 0 1rem;"` to the body.

Caleb-Ellis:
@anthonydillon pushed that up now. I tried it earlier and for a few things it works, but for anything placed as a direct descendant of the `<body>` tag it gets pretty messed up. I could go through all the examples on a case-by-case basis but I thought it'd be easier to just have a single internal div for all the examples

anthonydillon:
@Caleb-Ellis yeah think your right. Think its worth going back to just `margin: 1rem` on the body. The problem with having a div in the body is that people will copy that as part of the pattern.

Caleb-Ellis:
@anthonydillon That still leaves the problem of all the patterns directly in body getting affected though. Should I go through them and add an internal div to them? Or is there a way to change example-js to include a div as well as the body?

anthonydillon:
We cant add divs to the direct patterns as they will become part of the pattern. I think if some examples have a small gap at the top its not the end of the world. Its more true to the pattern as it is. When used in reality this wont be an issue.

Caleb-Ellis:
Got it. I don't mind the extra margin, I was more referring to things like the image pattern. When the body is display: flex, flex-direction: column, it expands to fit the entire width of the window. At the moment every page has the whole built css file included, so the `@at-root` in the footer is still applying the body flex stuff.

Caleb-Ellis:
Wait I think you can just add `align-items: flex-start` and that should fix it

Caleb-Ellis:
Nevermind, now everything's broken in a slightly different way hahaha

anthonydillon:
Maybe base images need display block?

_jekyll/_layouts/default.html
```diff
@@ -13,6 +13,6 @@
     </head>
 
     <body style="margin: 1rem;">
-{{ content }}
+      {{ content }}
```
anthonydillon: Can you remove this spacing as it will be shown in the example-js.

anthonydillon:
@Caleb-Ellis Yeah the images are a problem ...

Caleb-Ellis:
## Findings
- Adding `display: flex` and `flex-direction: column` to the `<body>` tag seems to have far-reaching, unintended side-effects. Any block element directly within `<body>` without an explicit width (i.e. `width: auto`) is rendered as wide as the browser window.
- While as @bartaz said it's unlikely anything would be rendered as a direct child of the `<body>`, and would instead likely be in a strip or grid etc, it's still technically possible and it doesn't address how the examples pages and consequently Vanilla docs would be affected.

## Options
- Have an optional class for the `<body>` tag for having a sticky footer (e.g. `.p-has-sticky-footer`)
- Add explicit widths or `align-self: start` to every pattern we don't want to be affected by the flexbox format
- Investigate a method that doesn't target the `<body>` element

Closing this PR until a suitable method is decided.

PR #1513, opened 2018-01-10
1484: Add single direction pagination

### LINKED ISSUE #1484: Single direction pagination is not supported
When building pagination we need states from when you can't go in one direction. Also the previous arrow is added to the first type so it has both if its the only button.

https://codepen.io/anthonydillon/pen/NwVNxv

Caleb-Ellis:
## Done

- Added support for single direction pagination
- Changed chevron from `background-image` to `content`
- Added new modifiers to `p-pagination__link`: `--previous` and `--next`
- Removed `:first-child` and `:last-child` pseudo-selectors
- Updated pagination example with new classes

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/pagination/
- Remove `p-pagination--next` from the DOM
- Check that the previous button is still styled properly, and that resizing the screen to small will not hide the content
- Refresh
- Remove `p-pagination--previous` from the DOM
- Check that the next button is still styled properly, and that resizing the screen to small makes the next button scale to full-width
- Check that the pattern still matches the [design docs](https://github.com/ubuntudesign/vanilla-design/tree/master/Pagination)

## Details

Fixes #1484


examples/patterns/pagination.html
```diff
@@ -5,11 +5,11 @@
 ---
 
 <footer class="p-pagination">
-  <a class="p-pagination__link" href="#previous">
+  <a class="p-pagination__link p-pagination__previous" href="#previous">
```
anthonydillon: Can you make these a modification instead of a new class. For example: `p-pagination__link--previous`, `p-pagination__link--next` and a default `p-pagination__link`

kwm14:
## Design review
- [ ] On small screens - Article title font size: 1.125rem

scss/_patterns_pagination.scss
```diff
@@ -1,60 +1,72 @@
 @import 'settings';
 
 @mixin vf-p-pagination {
+  %chevron-icon {
+    color: $color-mid-dark;
+    content: '\203A';
+    font-size: 2rem;
+    position: absolute;
```
tbille: there was no way to use `flex` ?

PR #1532, opened 2018-01-23
1525: Remove display: inline-block from notification pattern

### LINKED ISSUE #1525: Notifications don't have width 100% anymore
## Process

Short notifications don't span across whole container width anymore.
Notifications used to have width 100%, but it seems to be removed in #1408.
This causes shorter notifications to look misaligned.

## Expected result

Notifications should have width 100% (or there should be a way to make them 100% width with modifier).

## Screenshot

Before:
screen shot 2018-01-19 at 12 08 43

After:
screen shot 2018-01-19 at 12 08 31

Caleb-Ellis:
## Done

- Removed `display: block` from notification pattern, so it is a block by default therefore giving it full width of its container

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/notifications/notifications/
- Check that the notification block (not the paragraph text) takes up the full width of its container (the visible screen)

## Details

Fixes #1525


kwm14:
Looking good @Caleb-Ellis 👍

PR #1534, opened 2018-01-23
1526: Fix content alignment utilities

### LINKED ISSUE #1478: Max-width breaks alignment
Setting the `max-width` on p's and heading's is breaking the central alignment.
https://codepen.io/anthonydillon/pen/XzGxeW

### LINKED ISSUE #1526: Centred headings are misaligned
## Process

Headings that use centred text are misaligned because of `max-width` style applied to them.

When heading is having centred text in it max-width limit makes it not really centered inside to container.

## Expected result

Centred headings should not have max-width limit or there should be a way to disable max-width limit.

## Current result

screen shot 2018-01-19 at 12 13 09

When it should be:

screen shot 2018-01-19 at 12 16 12

Caleb-Ellis:
## Done

- Added margin overrides to `.u-align` utilities to account for `max-width` inclusion
- Added a small mixin for base typography so that if a parent `<div>` has a `.u-align` utility it properly defines the margins

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/utilities/align/
- Check that the three paragraphs in the cards are aligned as expected
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/headings/default/
- Add the class `u-align--center` to the `<body>` tag
- Check that all the headings align to the centre, and that they keep their max-widths (so there should be margins left and right of the heading)

## Details

Fixes #1478, fixes #1526

### DIFF
--- a/scss/_base_typography.scss
+++ b/scss/_base_typography.scss
@@ -170,6 +170,7 @@
 
   p {
     @include p-max-width;
+    @include align-text;
     margin-bottom: 0;
   }
 
@@ -289,8 +290,24 @@
   max-width: 60em;
 }
 
+@mixin align-text {
+  .u-align--center & {
+    margin-left: auto;
+    margin-right: auto;
+  }
+
+  .u-align--left & {
+    margin-right: auto;
+  }
+
+  .u-align--right & {
+    margin-left: auto;
+  }
+}
+
 @mixin vf-heading-1 {
   @include heading-max-width--short;
+  @include align-text;
   font-size: 2rem;
   font-weight: 300;
   line-height: 1.3125;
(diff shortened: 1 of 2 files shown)


kwm14:
@Caleb-Ellis content alignment looks great:
- Utilities - `u-align` 👍 
- Headings - `u-align--center` 👍

scss/_base_typography.scss
```diff
@@ -289,12 +290,33 @@
...
+}
+
 @mixin vf-heading-1 {
   @include heading-max-width--short;
+  @include align-text;
   font-size: $sp-x-large;
   font-weight: 300;
   line-height: 1.2;
 
+  .u-align--center & {
```
WillMoggridge: Is this block needed? It looks duplicated from the mixin

PR #1586, opened 2018-02-09
Fix notifications.scss linting error

Caleb-Ellis:
## Done

- Disabled a single line of sass linting because the Jenkins sass lint throws a fit with negative value variables


barrymcgee:
:shipit:

PR #1600, opened 2018-02-22
Add file size check to CI

anthonydillon:
## Done
Add filesize check to CI with a threshold

## QA
- Run `./run clean`
- Run `tests/file-size-lint.sh` and see that it errors because it cant find the build file
- Run `./run build`
- Run `tests/file-size-lint.sh` and see it passes and outputs the current file size of the build with the threshold

_The threshold is set to 400kb which is way to high. Lets half that and more as soon as possible_

### DIFF
--- a/tests/file-size-lint.sh
+++ b/tests/file-size-lint.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 FILE='build/css/build.min.css'
-if [ ! -f $FILE ]
-then
+if [ ! -f $FILE ]; then
   echo 'ERROR: Build file could not be found or is not readable. Please build first with `./run build`'
   exit 1
 fi
@@ -11,8 +10,7 @@ CURRENTSIZE=$(du -b $FILE | cut -f 1)
 REPONSESTART='Your generated CSS is '
 REPONSEJOIN=' with a threshold of '
 MESSAGE=$REPONSESTART$CURRENTSIZE$REPONSEJOIN$THRESHOLD
-if [ $CURRENTSIZE -lt $THRESHOLD ]
-then
+if [ $CURRENTSIZE -lt $THRESHOLD ]; then
   STATUS='SUCCESS: '
   EXIT=0
 else
--- a/tests/file-size-lint.sh
+++ b/tests/file-size-lint.sh
@@ -7,16 +7,14 @@ fi
 
 THRESHOLD=400000
 CURRENTSIZE=$(du -b $FILE | cut -f 1)
-REPONSESTART='Your generated CSS is '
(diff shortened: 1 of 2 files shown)


anthonydillon:
@nottrobin so CI failed but not in a way I would like it too :smile:

nottrobin:
@anthonydillon I don't know why you're getting this suddenly, but try adding `gem install rake` to the install steps in `.travis.yaml`

WillMoggridge:
@anthonydillon I'm not entirely sure what started it, current develop branch seems to build just fine.

May be this helps: https://github.com/travis-ci/travis-ci/issues/8978#issuecomment-354035321

anthonydillon:
@nottrobin nope :(

anthonydillon:
@WillMoggridge @nottrobin I'm stumped

anthonydillon:
@nottrobin I have updated the test to be more resilient and should now work with travis.

tests/file-size-lint.sh
```diff
@@ -0,0 +1,20 @@
...
+if [ ! -f $FILE ]; then
+  echo 'ERROR: Build file could not be found or is not readable. Please build first with `./run build`'
+  exit 1
+fi
+
+THRESHOLD=400000
+CURRENTSIZE=$(du -b $FILE | cut -f 1)
+MESSAGE="Your generated CSS is ${CURRENTSIZE} with a threshold of ${THRESHOLD}"
+if [ $CURRENTSIZE -lt $THRESHOLD ]; then
+  STATUS='SUCCESS'
```
nottrobin: Why not simply `echo` and exit directly in each of these cases? I know you'll be duplicating the message, but arguably those messages should be slightly different, given the radically different results:

``` bash
if [ ${CURRENTSIZE} -gte ${THRESHOLD} ]; then
  echo "ERROR: Your CSS is over ${THRESHOLD} (it's ${CURRENTSIZE})"
  exit 1
fi

echo "Success: your generated CSS is ${CURRENTSIZE} (under ${THRESHOLD})"
```

tests/file-size-lint.sh
```diff
@@ -0,0 +1,20 @@
...
+  EXIT=0
+else
+  STATUS='ERROR'
+  EXIT=1
+fi
+
+echo "${STATUS}: ${MESSAGE}"
```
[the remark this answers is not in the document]
anthonydillon: Good idea.

tests/file-size-lint.sh
anthonydillon: Good idea

anthonydillon:
@nottrobin done and ready for another look.

PR #1601, opened 2018-02-27
Creates “Tables” category

### LINKED ISSUE #1590: Table patterns are hidden inside “Interactive elements”
1\. Go to [docs.vanillaframework.io](https://docs.vanillaframework.io.).
2\. Try to find the Table pattern.

What happens: It’s inside the “Interactive elements” category.
What should happen: It isn’t.

Sortable table columns and expanding cells are interactive elements, but tables in general are not.

Since there are four pages for different table patterns, “Tables” probably should have their own category.

matthewpaulthomas:
Fixes #1590 by creating a “Tables” category for “Table”, “Table sortable”, “Table expanding”, and “Table mobile card”.

### DIFF
--- a/docs/en/metadata.yaml
+++ b/docs/en/metadata.yaml
@@ -144,6 +144,21 @@ navigation:
   - location: patterns/pull-quote.md
     title: Quotes
 
+- title: Tables
+  children:
+
+  - location: base/table.md
+    title: Table
+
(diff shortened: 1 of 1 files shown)


anthonydillon:
Looks good to me :+1: . Makes more sense to have it as its own entity then an interactive element.

PR #1612, opened 2018-03-05
Expands “Media” to “Images and media”

matthewpaulthomas:
“Page structure” contains several image-related patterns:
- Images
- Inline images
- Image position
- Heading icon

And one _prima facie_ media-related pattern:
- Media object

Meanwhile, “One-off layout” contains one image-related pattern:
- Image position

And “Media” contains two more image-related patterns:
- Assets
- Icons

This change renames “Media” to “Images and media” and puts all these patterns there.

Of these, the most questionable are “Media object” and “Heading icon”.
- The name “Heading icon” reads as if it’s a kind of icon. If it’s more a kind of heading, so it should be left where it is, you can exclude the last commit and pull up to d1fbe91 instead.
- If “Heading icon” _and_ “Media object” would be better left where they are, you can exclude the last two commits and pull up to db75cb3 instead.


Caleb-Ellis:
Code and QA +1
@kwm14 do you have any thoughts from a UX perspective?

kwm14:
@Caleb-Ellis nope looks good to me 👍

PR #1656, opened 2018-03-28
Remove baseline grid from examples pages

Caleb-Ellis:
## Done

- Removed baseline grid and toggle until we figure out a way to do it that won't break the vanilla docs

## QA

- Pull code, `./run`, check that there is no unnecesary markup in the examples


anthonydillon:
It would also be nice to remove the styling for the baseline utility.

PR #1700, opened 2018-04-10
1658: Fix issue with navigation logo sizing

### LINKED ISSUE #1658: Header has lost some spacing
During the conversion of the cloud-init.io site. I noticed that the header has changed. 

It seems that the:
- logo size is set to `24px` but was `32px` before
- there was margin around the entire logo but it has no margin bottom now

Before:
screenshot-2018-4-2 cloud-init - the standard for customising cloud instances

After:
screenshot-2018-4-2 cloud-init - the standard for customising cloud instances 1

Caleb-Ellis:
## Done

- Set a height of 3rem on navigation logos
- Made it so logo can be between 1.5rem (24px) and 2rem (32px) in height
- Updated nav examples to depict 24px high Vanilla logo

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/navigation/default/
- Check that the nav looks the same
- Delete `<div class="p-navigation__banner">` from the DOM and check that the nav height is still 3rem
- Undo, now delete `<nav class="p-navigation__nav" role="menubar">` and check that the nav height is still 3rem
- Undo, now remove `width="95"` from the nav logo, and check that the logo reaches the max-height (32px) and is still vertically centered

## Details

Fixes #1658

### DIFF
--- a/examples/patterns/navigation/default.html
+++ b/examples/patterns/navigation/default.html
@@ -7,7 +7,7 @@ category: _patterns
   <div class="p-navigation__banner">
     <div class="p-navigation__logo">
       <a class="p-navigation__link" href="#">
-        <img src="https://assets.ubuntu.com/v1/d96d86b5-vanilla_black-orange_hex.svg" alt="" class="p-navigation__image" />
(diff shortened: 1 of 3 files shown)


kwm14:
Looking good @Caleb-Ellis. Love that we now have 2 logo heights to cater for different styled logo types 👍

PR #1702, opened 2018-04-10
Make u-vertically-center work with img elements

### LINKED ISSUE #1669: u-vertically-center class not working inside p-card_header
## cn.ubuntu.com upgrade to v1.7.0-alpha-1
Reviewing the beta test branch I found an example of `p-card_header` that uses `u-vertically-center` class but for some reason it isn't applying.

Inspecting the styles it seems that `display: flex` seems to be the issue. But I'll leave it in the hands of development to correct me 😄 

Link to issue: http://www.ubuntu.com-pr-2859.run.demo.haus/cloud/public-cloud

## Current result
screen shot 2018-04-03 at 14 19 57

Caleb-Ellis:
## Done

- Added an `img` child selector to `u-vertically-center` to override the default `align-self: flex-start` behaviour

## QA

- Go to http://www.ubuntu.com-pr-2859.run.demo.haus/cloud/public-cloud
- Open Chrome Dev Tools and change `align-self: flex-start` to `align-self: center` from the card img elements
- Check they are now vertically centered
- Check Percy that I haven't inadvertently broken anything

## Details

Fixes #1669 

## Screenshots
Fix applied on [vanilla test on ubuntu](http://www.ubuntu.com-pr-2859.run.demo.haus/cloud/public-cloud)
screenshot

### DIFF
--- a/scss/_utilities_vertically-center.scss
+++ b/scss/_utilities_vertically-center.scss
@@ -6,6 +6,10 @@
     .u-vertically-center {
       align-items: center !important;
       display: flex !important;
+
+      > img {
+        align-self: center !important;
+      }
     }
   }
 }


lyubomir-popov:
@Caleb-Ellis I think I added this to prevent images with margin:0 from stretching to cover flex containers. Would another value be safe to use - like center?

Caleb-Ellis:
@lyubomir-popov I see. In that case I think the default option should be as-is (`align-self: flex-start`) because that's how a non-flex, unstyled img would behave. And I will edit the vertical center utility to target imgs

lyubomir-popov:
@Caleb-Ellis this is assigned to me but not sure what for - review or fixing?

Issue #1665, opened 2018-04-03
Muted heading has margin-bottom by default


kwm14:
## insights.ubuntu.com upgrade to v1.7.0-alpha-1
Reviewing the beta test branch I noticed that `p-muted-heading` is applying margin-bottom: 15px; which is causing unnecessary padding to elements it's contained within.

Link to issue: http://insights.ubuntu.com-pr-290.run.demo.haus/

## Expected result
![screen shot 2018-04-03 at 10 33 47](https://user-images.githubusercontent.com/17748020/38241585-84ab409e-372a-11e8-84b6-504d8ee8d63e.png)

## Current result
![screen shot 2018-04-03 at 10 32 42](https://user-images.githubusercontent.com/17748020/38241566-78d2b3ec-372a-11e8-8332-434adfce2f18.png)

anthonydillon:
This should be a card level override as I believe we want the margin-bottom on a stand alone `p-muted-heading`.

kwm14:
Agreed to add this card-component to Vanilla.

kwm14:
I've added to Vanilla backlog for a new component/modifier of the current card component. 

+1 @anthonydillon comment that for now it will be a card level override.

lyubomir-popov:
@Caleb-Ellis could you leave this issue open please? I'd like to add this to the vanilla card component with slight adjustments - so that the spacing above/underneath matches other places the muted heading is "sandwiched" between borders, like table headers. 
As it is now, the space above/below is a bit too wide for such a small font-size, if you compare with the expected result above - 
![image](https://user-images.githubusercontent.com/2741678/38632204-ec69d80a-3db3-11e8-9a28-98c6166788d8.png)

Caleb-Ellis:
@lyubomir-popov No worries. Closing the original PR for now.

lyubomir-popov:
#1713 addresses part of this by adding a muted header card to vanilla. This pattern is currently used in only two places, so I've taken the liberty of altering the makrup a bit.

lyubomir-popov:
@Caleb-Ellis Why are the image and heading with different Google Analytics tracking code? 
Shouldn't a card acting as a link to somewhere be clickable as a whole, meaning all content under .p-card wrapped in one <a>?

PR #1713, opened 2018-04-12
Add muted-heading card as part of addressing #1665

lyubomir-popov:
## Done

## QA

- Pull code
- Run `./run serve --watch`
- [Add additional steps]

## Details

## Screenshots

### DIFF
--- a/scss/_patterns_card.scss
+++ b/scss/_patterns_card.scss
@@ -1,27 +1,5 @@
 @import 'settings';
 
-@mixin vf-p-card-default {
-  .p-card {
-    @extend %p-card;
-    border: 1px solid $color-mid-light;
-    padding: $spv-intra--scaleable - $px $spv-intra--scaleable $spv-intra--scaleable;
-  }
-}
-
-@mixin vf-p-card-highlighted {
-  .p-card--highlighted {
-    @extend %p-card--highlighted;
-  }
-}
-
-@mixin vf-p-card-overlay {
-  .p-card--overlay {
-    @extend %p-card;
-    background: transparentize($color-x-light, .1);
-    border-radius: 0;
-  }
-}
-
 @mixin vf-p-card {
   %p-card {
     background: $color-x-light;
@@ -30,16 +8,28 @@
     margin-bottom: $spv-inter--scaleable;
     overflow: auto; // prevent overflow of children margins
     padding: $spv-intra--scaleable;
+  }
 
-    .p-card__content {
-      margin-top: - $sp-unit * 2;
-    }
+  %highlight {
+    box-shadow: 0 1px 5px 1px transparentize($color-dark, .8);
+  }
+
+  p.p-card__content {
+    margin-top: - $sp-unit * 2;
+  }
+
+  .p-card img {
+    vertical-align: top;
   }
 
   .p-card__header {
     border-bottom: 1px solid $color-mid-light;
     margin-bottom: $spv-intra--scaleable;
     padding-bottom: $spv-intra--scaleable - $px; // border compensation
+
+   .p-muted-heading {
+      margin-bottom: map-get($nudges, nudge--muted);
+    }
 
     > img {
       max-height: 2rem;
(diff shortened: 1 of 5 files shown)


scss/_patterns_card.scss
```diff
@@ -31,15 +9,28 @@
     overflow: auto; // prevent overflow of children margins
     padding: $spv-intra--scaleable;
 
-    .p-card__content {
-      margin-top: - $sp-unit * 2;
+    img {
```
anthonydillon: All patterns should only target classes. This allowed to pattern to be weekend agnostic.

scss/_patterns_card.scss
```diff
@@ -31,15 +9,28 @@
...
+      margin-bottom: $spv-inter--condensed-scaleable;
+      vertical-align: top;
     }
   }
 
+  %highlight {
+    box-shadow: 0 1px 5px 1px transparentize($color-dark, .8);
+  }
+
+  p.p-card__content { // sass-lint:disable-line no-qualifying-elements - this is because the class is applied to both divs and ps. ps have special rules to ring them closer to other text elements.
```
anthonydillon: This is over selected. Please don't use elements when selecting.

scss/_patterns_card.scss
```diff
@@ -31,15 +9,28 @@
...
+    margin-top: - $sp-unit * 2;
+  }
+
   .p-card__header {
     border-bottom: 1px solid $color-mid-light;
     margin-bottom: $spv-intra--scaleable;
-    padding-bottom: $spv-intra--scaleable - $px; // border compensation
+    // padding-bottom: $spv-intra--scaleable - $px; // border compensation
```
anthonydillon: Please remove commented out code.

scss/_patterns_card.scss
```diff
@@ -51,12 +42,59 @@
...
+}
+
+@mixin vf-p-card-highlighted {
+  .p-card--highlighted {
+    @extend %highlight;
+    @extend %p-card;
+  }
+}
+
+@mixin vf-p-card-muted-header {
```
anthonydillon: Can you think of a name to describe the purpose/effect of the use of this pattern. Muted-heading is describing the styling but in the future that may change but the role of the pattern should not. Maybe even replace the highlight card?

scss/_patterns_card.scss
lyubomir-popov: the trouble with this is, there already is an image with no class inside the header card pattern. Moving this rule to a class would require editing all instances of markup on all sites. It is a simple search and replace, so not too difficult, but still - is that ok?

anthonydillon: ok

anthonydillon: Can we create a class for it too and update the example to use the class so we can phase out this issue.

scss/_patterns_card.scss
lyubomir-popov: fixed

scss/_patterns_card.scss
lyubomir-popov: fixed

scss/_patterns_card.scss
lyubomir-popov: I've added a class .p-card__thumbnail that sets max-height to the current 2rem. @kwm14 Would be good to audit these and see if we can reduce the number of different sizes. I'm putting them in a sass map, so far I've got:

  .75rem - accordion icons
  1rem - media object meta icons, notification icons
  2rem - card header thumbnail image
  2.5rem - social icons 
  3rem - matrix
  5rem - media object
  6rem - media object large

scss/_patterns_card.scss
lyubomir-popov: Just had a chat with @Caleb-Ellis and this led to a larger topic. I think we should discuss in the WG meeting, so I'll add a don't merge label on this pr. Caleb already merged a simpler fix that deals with the reported bug on those cards.

In short - the elements that are currently found in cards are miscellaneous. Their styling is also not specific to a card, and could be used in other layouts. The current card examples, together with the resources and insights cards, are made up of:

• muted heading followed by a rule (current resources card)
• 2rem thumbnail image followed by a rule
• h3 / h4 with anchor
• image
• p-list 
• subscribe form with p-list inside
• button
• any combination of the above

I think it would be simpler to limit the card to just a wrapper with padding, which has modifiers for border and box-shadow. And then compose cards as necessary. 

To avoid breaking things, the current .p-card__* classes could just point to atomic placeholders / classes that format them so they look as they do now.

PR #1744, opened 2018-04-18
Inline-images: Fix image distortion

lyubomir-popov:
## Done

## QA

- Pull code
- Run `./run serve --watch`
- [Add additional steps]

## Details

## Screenshots

### DIFF
--- a/scss/_patterns_inline-images.scss
+++ b/scss/_patterns_inline-images.scss
@@ -26,9 +26,8 @@
       }
 
       * {
-        max-height: $sp-xxx-large;
+        max-height: map-get($icon-sizes, thumb--small);
         max-width: 7rem;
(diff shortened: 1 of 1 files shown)


Caleb-Ellis:
Horizontal margins seem a bit large for small screens. Shouldn't we be setting these in terms of `$sph-inter`?

lyubomir-popov:
@Caleb-Ellis normally - yes, normally this should use $sph-inter. I've reverted the margin to what it was ($sp-xxx-large) as this pattern has a lot of hardcoded values and changing them without reiewing all instances is a bit risky. (The value from the map I had put in the above commit was equal to the old one).

So in the latest commit I've only replaced the max-height, leaving the margin as it was.

PR #1773, opened 2018-04-25
Remove negative margin from consecutive .p-matrix__desc

### LINKED ISSUE #1772: No padding in-between two paragraphs in matrix desc
## v1.7.0-beta-2 review on ubuntu.com
No padding in-between two paragraphs in matrix desc, you can see the example of with text link sitting under paragraph tightly.
- Issue can be found here: http://www.ubuntu.com-pr-3035.run.demo.haus/cloud/managed-cloud

### Current result
screen shot 2018-04-24 at 17 17 33

Caleb-Ellis:
## Done

- Removed negative margin on consecutive `.p-matrix__desc` so vertical spacing is like two normal paragraphs

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/matrix/
- Add a second `<p class="p-matrix__desc">Short description</p>` to any of the matrix cells
- Check that the space between the two paragraphs is the same as any two consecutive `<p>`

## Details

Fixes #1772 

### Before

screenshot_3

### After

screenshot_2

### DIFF
--- a/examples/patterns/matrix.html
+++ b/examples/patterns/matrix.html
@@ -9,63 +9,77 @@ category: _patterns
     <img class="p-matrix__img" src="https://via.placeholder.com/60x60" alt="icon">
     <div class="p-matrix__content">
       <h3 class="p-matrix__title"><a class="p-matrix__link" href="#">Title</a></h3>
-      <p class="p-matrix__desc">Short description</p>
+      <div class="p-matrix__desc">
+        <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit.</p>
+        <p>Quasi, labore at suscipit necessitatibus cumque commodi velit.</p>
+      </div>
     </div>
   </li>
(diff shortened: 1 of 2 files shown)


kwm14:
Tested and reviewed design 👍

anthonydillon:
I would prefer us to use raw content. For example:
```html
<div class="p-matrix__desc">
  <p>Short description</p>
  <p>Short description</p>
</div>
```

Caleb-Ellis:
@anthonydillon Updated to allow for single `<div class="p-matrix__desc>`.

WillMoggridge:
Should the previous format still work?
Would it be better to have the examples have mix used of the existing HTML structure and some using the new one? Maybe including:
- Items with just `p`
- Items with the `div` and one child `p`
- Items with the `div` and a few children `p`

PR #1793, opened 2018-05-02
Add Sass build/watch functionality to Vanilla Docs

Caleb-Ellis:
## Done

- **Note:** I've created a feature branch `docs-1.7.0` as this PR's base for all the docs-related work going into releasing 1.7.0. I will have a separate PR for fixing up the styling.
- Added `node-sass` and `vanilla-framework` to the /docs packages
- Updated `package.json` to build the css from scss, and build the docs when `./run build` is run
- Fixed the `watch` script so that it rebuilds when `.scss` files are edited
- Moved the inline styles in `template.html` to `/static/sass/etc`

## QA

- Pull code
- `cd docs`
- Run `./run` and check that static/sass/*.scss get built into static/css/styles.css
- Check that the docs are served correctly, with Vanilla 1.7.0-beta-2.1 styling (will update to 1.7.0 when released).
- Open another terminal and run `./run watch`
- Edit `static/sass/styles.scss` and check that the css is rebuilt and the changes occur


docs/package.json
```diff
@@ -1,12 +1,16 @@
...
+    "watch": "watch -p 'static/sass/**/*.scss' -p 'node_modules/vanilla-framework/scss/*.scss' -c 'yarn run build-css'",
+    "build": "yarn run build-css && yarn run build-docs",
+    "build-css": "node-sass static/sass --output static/css",
+    "build-docs": "documentation-builder --base-directory . --output-path templates --template-path template.html --tag-manager-code 'GTM-K92JCQ' --no-link-extensions --force",
+    "test": "",
+    "serve": "cd dist && FLASK_DEBUG=true FLASK_APP=app.py flask run -h 0.0.0.0 -p $PORT"
+  },
+  "dependencies": {
+    "node-sass": "^4.5.3",
+    "vanilla-framework": "1.7.0-beta-2.1",
```
anthonydillon: Why does vanilla-framework need a dependency on vanilla-framework? Couldn't we just use the local files?

Caleb-Ellis: Hahahaha oh yeah

WillMoggridge: I think the only decent option to manage that in the run script is to use `./run --node-module ../` because of the container sandboxing.
I think this would involve manually entering that each time, and it would have risks of being confused by running local dev changes that you don't expect.

When talking about recreating the run script, I want to make an optional config file which could have standard argument like this if we chose to.
Although if we changed it to use the parent run script with a `serve-docs` endpoint, the local vanilla files would be available in the container.

steverydz: From what @WillMoggridge says it sounds like this is something that will be possible at a later date? If that's the case I don't see any reason why this can't be merged?

PR #1808, opened 2018-05-10
Vanilla docs styling fixes

Caleb-Ellis:
## Done

- Removed custom table styling from Docs sass
- Added padding-left to colour column on Color page

## QA

- Pull code
- Run `cd docs && ./run`
- Open http://0.0.0.0:8104/en/settings/color-settings
- Check that the cells in the colour column have .5rem padding on the left
- Open any page with iframes and check that they have `margin-top: 1rem`

### DIFF
--- a/docs/static/sass/styles.scss
+++ b/docs/static/sass/styles.scss
@@ -6,10 +6,6 @@
 @import 'codehilite';
 @include codehilite;
 
-td, th {
-  padding: 1rem 0.75rem;
-}
-
 h2:nth-child(2) {
   display: inline-block;
   margin-bottom: $sp-medium;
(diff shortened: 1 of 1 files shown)


barrymcgee:
👆🏻...does a `<td>` not have any default padding?

docs/static/sass/styles.scss
```diff
@@ -6,10 +6,6 @@
 @import 'codehilite';
 @include codehilite;
 
-td, th {
-  padding: 1rem 0.75rem;
-}
-
```
steverydz: Why do we prefer inline styles for the padding? Wouldn't it be better to just have some default padding?

Caleb-Ellis:
@barrymcgee @steverydz  `<td>`s only have top and bottom padding by default, to keep the first cells horizontally aligned, e.g:
screenshot_2

I'll change the rule to only target the second-column onwards and make a proposal to include in Vanilla.

PR #1811, opened 2018-05-11
Fix button spacing inside paragraph tags

### LINKED ISSUE #1810: Buttons placed inline inside a `<p>` tag collapse margins on small screens
1. Navigate to https://codepen.io/barrymcgee/pen/vjjxee?editors=1000#0
2. Decrease screen size to emulate a small screen
3. Observe that buttons placed inside a `<p>` collapse margins but placed within a `<div>`, they maintain them.

## Expected result

Two inline buttons should preserve margin regardless of their parent container.

## Screenshot

screen shot 2018-05-11 at 15 06 13

Caleb-Ellis:
## Done

- Fixed button spacing inside paragraph tags - they should now be spaced identically whether in a `<div>`, `<p>` or not in a container.
- Fixed a bug where if several buttons are placed inline and wrap to a new line, the new line has `margin-left`
- Aligned button text to baseline grid

## QA

- Pull code
- Open `/_jekyll/_layouts/default.html` and change `VF_ENV` to `"DEV"`
- `./run`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/buttons/brand/
- Change any of the `<p>` containers with buttons to `<div>` and check that nothing changes visually (margins and paddings change behind the scenes)
- Add a `<p>` with text before all the button containers, and check again that nothing changes visually if you change them to/from `<div>`s
- Copy and paste a few buttons inside one of the button containers until they wrap, and check that the new line of buttons does not have any `margin-left`
- Change the container with several buttons to a `<p>`, resize window to small screen (<460px) and check that margins between buttons stay the same
- Turn on the baseline grid and check that the button text sits on the lines on small and large screens

## Details

Fixes #1810

### DIFF
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -8,7 +8,6 @@
   }
 
   %vf-button-base {
-    @extend %bordered-text-vertical-padding;
     @include vf-animation (#{background-color, border-color}, fast, in);
     @include vf-focus;
     border-radius: $border-radius;
@@ -19,8 +18,8 @@
     font-size: 1rem;
     font-weight: 300;
     line-height: map-get($line-heights, default-text);
-    padding-left: $sph-intra;
-    padding-right: $sph-intra;
+    margin-bottom: $spv-inter--scaleable + 2 * $spv-nudge-compensation;
+    padding: $spv-nudge - $px $sph-intra;
     text-align: center;
     text-decoration: none;
 
@@ -38,18 +37,13 @@
 
     @media only screen and (max-width: $breakpoint-x-small) {
       width: 100%;
-
-      p & + & {
-        margin-bottom: $spv-nudge-compensation;
-        margin-top: - $spv-inter--scaleable;
-      }
     }
 
-    @media only screen and (min-width: $breakpoint-x-small) {
+    @media only screen and (min-width: $breakpoint-x-small + 1) {
       width: auto;
 
-      & + & {
-        margin-left: $sph-inter;
+      &:not(:last-of-type):not(:only-of-type) {
+        margin-right: $sph-inter;
       }
     }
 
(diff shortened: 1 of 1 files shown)


lyubomir-popov:
@Caleb-Ellis Not sure all of this needs to be removed. I remember I was trying to ensure buttons inside paragraphs behave the same as buttons outside paragraphs, but maybe I didn't target the rules as precisely as I should have, causing #1810.

This bit specifically counters p+p negative margin top (which brngs paragraphs closer together), preventing the buttons from going too close to the text. IF this gets merged, they will look like this 
screen shot 2018-05-11 at 16 05 08

At the moment, they have more spacing https://codepen.io/lyubomir-popov/pen/mLLwdd

So it just needs tweaking, not removal I suspect.

As to the number of rules, it is difficult to decrease it without having a base bem class. At the moment I have to list all possible modifiers. If there was a p-button and then p-button--positive, I would just list p> .p-button. Or at least I can't think of a better way to target any of the variety of button flavours placed inside a paragraph.

lyubomir-popov:
@Caleb-Ellis Aside for #1810, we never solved the left margin problem. The second button is indented if the width is narrow enough. https://codepen.io/lyubomir-popov/pen/mLLwdd
I remember intending to move the margin from right on the second p to left on any p, but I must have missed it. Sorry! Will revisit this early next week.

lyubomir-popov:
@Caleb-Ellis An example of where removing the extra white space between p and button (which the snippet below ensures)
```
    p & {	+    p > &:first-child {
       margin-bottom: $spv-nudge-compensation;	       margin-bottom: $spv-nudge-compensation;
       margin-top: $spv-nudge-compensation;	       margin-top: $spv-nudge-compensation;
     }
```
 would be problematic:

screen shot 2018-05-11 at 16 36 49

Caleb-Ellis:
@lyubomir-popov my intention was not to remove everything, just understand why these rules were included (and to see what Percy shows when they're removed). In any case the only part that I'm not sure of is:
``` scss
p & + & {
  margin-top: - $spv-inter--scaleable;
}
```
which is actually the only part causing the problem in #1810. Can you remember a case where this is useful?

lyubomir-popov:
@Caleb-Ellis No, I think I got this part. What needs to happen is:

Inside a p, the first button needs to negate padding or padding + negative margin applied to the p, so it looks like it is outside a p. Same for the margin-bottom. There are two cases:

p not following another p:
– the nudge padding-top (nudge--p)
p following another p, i.e.: p+p>button:
- the combination of nudge--p + negative margin top ($spv-inter--scaleable) above

Obv I've messed up the intricate selector logic. Is this urgent? Can have a look tue next week? Wouldn't want you to go through this convoluted logic.

lyubomir-popov:
@Caleb-Ellis If it is urgent, a good test case is this:

```
<button class="p-button--brand">Brand button</button><button class="p-button--brand">Brand button</button>
<p><button class="p-button--brand">Brand button</button><button class="p-button--brand">Brand button</button></p>
```
These two sets of buttons should look exactly identical on both desktop and mobile.

lyubomir-popov:
And we should stop nesting block level elements like buttons inside p's :)
So we can get rid of all of this.

kwm14:
Tested and looks good @Caleb-Ellis 👍

PR #1820, opened 2018-05-16
Convert .p-switch from button to checkbox

### LINKED ISSUE #1453: Convert .p-switch to a checkbox
https://codepen.io/Lukewh/pen/vWzOKq

The class names aren't bem, sue me.

The problems with the current switch being button based is that Javascript is required this:
- prevents progressive enhancement (for example having normal forms, unless JS is enabled)
- means an uncaught error could cause the input to be unusable

This also provides a fallback for CSS failing to load/ CSS errors/ CSS being disabled.

### LINKED ISSUE #1759: Switches have different focus ring color and offset
The focus appearance for a Vanilla switch is:
```css
  outline: 1px solid #007aa6;
  outline-offset: 3px;
```
vanilla-focus-switch

The focus appearance for, apparently, all other Vanilla controls is
```css
  outline: 1px solid #19b6ee;
  outline-offset: 2px;
```
vanilla-focus-button

At least the color should be consistent. The offset probably should be consistent too.

Caleb-Ellis:
## Done

- Converted the `.p-switch` component to a checkbox so it can be used without JS
- Updated example page with new markup and removed JS
- Deprecated the old button version, to be removed in v2.0.0
- Changed the baseline grid switch to the new checkbox version
- Aligned the switch component to the baseline grid
- Fixed the focus colour
- Added `width: fit-content` to `<label>`s so you can't click way outside the label to get the input to trigger

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/switch/
- Check that it behaves exactly like https://vanilla-framework.github.io/vanilla-framework/examples/patterns/switch/ (there will be a small difference in vertical spacing)
- Test on Chrome, IE, Edge and Firefox < v58
- Turn on the baseline grid (`_jekyll/_layouts/default.html` change `VF_ENV` to `"DEV"`)
- Check that the switches and labels sit on the lines
- Check that the focus colour is the same as all the other form elements (`#19b6ee`)
- Edit the markup of the page so it uses the button version (see [here](https://github.com/vanilla-framework/vanilla-framework/blob/develop/examples/patterns/switch.html)) and check that it still looks and functions right

## Details

Fixes #1453, fixes #1759

### DIFF
--- a/_jekyll/_layouts/default.html
+++ b/_jekyll/_layouts/default.html
@@ -26,15 +26,14 @@
             }
             class baselineGrid {
                 constructor() {
-                    this.target = document.body,
+                    this.target = document.body;
                     this.baselineHeight = 8;
                     this.baselineContainerClassName = "u-baseline-container";
                     this.baselineRowClassName = "u-baseline";
                     this.toggleGridClass = "u-hide";
-                    this.showGrid = false;
                 }
                 addBaselineGridToggle() {
-                    var controls = fragmentFromString('<div class="controls"><button id="switch-on" class="p-switch js-baseline-toggle" type="button" role="switch" aria-checked="false" aria-labelledby="switch-on-label"><span>On</span><span>Off</span></button></div>');
+                    var controls = fragmentFromString('<div class="controls"><label><input type="checkbox" class="p-switch js-baseline-toggle" /><div class="p-switch__slider"></div></label></div>');
                     this.target.appendChild(controls);
                 }
                 appendBaselineRow(parent) {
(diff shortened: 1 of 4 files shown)


Caleb-Ellis:
@anthonydillon Yeaaahhhh looks like it's broken in IE and Edge, and older versions of Firefox because of `::after` and `::before` being used on a self-closing tag (`<input />`). I think the only way to make it work cross-browser is to wrap the whole thing in a `<div>` and put the pseudo-selectors on that.

PR #1838, opened 2018-05-25
Fix inline image aspect ratio and crop bug

barrymcgee:
## Done

This change ensures that images that are wider than the space allocated to them on small screens do not overflow.

Before this change: https://cl.ly/3r2B3z1E1i36

After this change: https://cl.ly/1d2x2g1Y1l0n

## QA

- Pull code
- Run `./run serve --watch`
- Ensure Inline Images pattern works as expected across breakpoints

### DIFF
--- a/scss/_patterns_inline-images.scss
+++ b/scss/_patterns_inline-images.scss
@@ -12,24 +12,21 @@
 
     &__item {
       display: inline-block;
-      margin: 1.875rem;
-      max-height: 5.625rem;
-      max-width: 5.625rem;
+      margin: 1rem;
       overflow: hidden;
       text-align: center;
       vertical-align: middle;
 
-      @media (min-width: $breakpoint-medium) {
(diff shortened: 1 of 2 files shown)


Caleb-Ellis:
New fix applied:

screenshot_1

scss/_patterns_inline-images.scss
```diff
@@ -12,24 +12,21 @@
...
 
-      @media (min-width: $breakpoint-medium) {
+      @media only screen and (min-width: $breakpoint-medium) {
         margin: 1.875rem;
-        max-height: 11.25rem;
-        max-width: 11.25rem;
       }
 
-      * {
+      img {
```
anthonydillon: This should be a class. We should never style elements in patterns.

Caleb-Ellis:
@anthonydillon Added `.p-inline-images__logo` class (currently used in ubuntu.com) and deprecated `.p-inline-images__item img`

PR #1853, opened 2018-06-06
Align anchor tags that follow a button properly

### LINKED ISSUE #1840: Button followed by link should align intelligently
## Process

1. Have any button be immediately followed by any link
2. No intelligent padding or alignment occurs

## Expected result

screenshot_2
screenshot_3

## Current result

screenshot
screenshot_1

Caleb-Ellis:
## Done

- Added a small bit of sass to the base button pattern that aligns any anchor tags that immediately follow it
- I'm not sure if it's considered bad practice to target `a`'s in the button sass file, but it was the easiest way to target every instance of button (`<button>`, `<a class="p-button--positive">`, etc)
- Reduced margin-bottom of buttons on small screens by .5rem

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/base/button/
- Add `<a>Link</a>` directly after the button
- Check that on large screens, the link is horizontally aligned, and there is a margin between it and the button
- On small screens (<460px), check that the link is centered and directly below the button
- On small screens, check that the margin-bottom on the button is 1.2rem

## Details

Fixes #1840 

## Screenshots

screenshot_5

screenshot_6

### DIFF
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -36,6 +36,7 @@
     }
 
     @media only screen and (max-width: $breakpoint-x-small) {
+      margin-bottom: $spv-inter--condensed-scaleable + 2 * $spv-nudge-compensation;
       width: 100%;
     }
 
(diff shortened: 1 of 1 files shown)


Caleb-Ellis:
## Outcome

After talking with @anthonydillon we've decided it's best not to target links in the button pattern, as it is likely to cause regressions. In future, to get the desired look we should wrap the elements in an inline list. The markup would then look something like:
``` html
<ul class="p-inline-list">
    <a class="p-button--positive">Button</a>
  </li>
    <a class="p-link--external">Link</a>
  </li>
</ul>
```

PR #1867, opened 2018-06-18
Make each icon pattern its own mixin

### LINKED ISSUE #1817: Icons on dark backgrounds
It's impossible (or at least very hard) to get icons work on dark backgrounds.

Vanilla icons have a variant prepared for a dark background but it only works inside of `p-strip--dark` pattern.

There are 2 issues with that:

### Dark icons only work if `--dark` is in the end of class name.

Because of CSS `[class$="--dark"]` selector icons for dark background will work ONLY if `--dark` is in the end of class attribute in HTML, not in the end of any class name.

So while `<div class="p-strip--dark">` will work, `<div class="p-strip--dark some-other-class">` will not. It's very hard to debug and not very intuitive unless someone looks into internals of Vanilla and understands how this selector works.

It should either be fixed in CSS or at least mentioned in the docs.

### There is no other way of getting dark variant of the icons

Currently the only way to get icon designed for dark background is to properly add `--dark` class in the end of some parent node.

It would be good if icon pattern itself would provide a modifier to get dark variant of the icon, so it can be used in any context (for example `p-button--positive` which has dark background colour).

### LINKED ISSUE #1866: Icon mixins should be able to be imported individually
Currently, you need to import the `vf-p-icons` mixin to get the icon styles, which includes every icon. You should be able to import only the icons you want, and have them styled correctly.

Caleb-Ellis:
## Done

- Made each icon pattern its own mixin, e.g. `vf-p-icon-plus` - this allows you to import each icon individually instead of being forced to import all 32.
- Fixed a bug in the sidebar navigation pattern that caused the minus icon to have `display: none` if included before `vf-p-navigation`
- Changed the selector for light icons from `[class$="--dark"] &`, to  `[class*="--dark"] &, &.is-light`. This means that icons inside a `p-strip--dark` will appear light regardless of where the `--dark` suffix appears in the class name (i.e. `p-strip--dark some-other-class` works now). Also you can give an icon a state class `is-light` which will turn the icon into its lighter variation

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/icons/icons-light/
- Give any of the icons an extra class `is-light` and check that it changes to the light version
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/icons/icons-dark/
- Add any class to `p-strip--dark` (e.g. `class="p-strip--dark some-other-class")
- Check that the icons are still light
- Open `scss/_vanilla.scss` and change line 91 to any of the individual icon mixins (e.g. `@include vf-p-icon-plus`)
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/icons/icons-light/ and check that the icon shows up alone

## Details

Fixes #1866, fixes #1817

### DIFF
--- a/scss/_patterns_accordion.scss
+++ b/scss/_patterns_accordion.scss
@@ -1,4 +1,7 @@
 @import 'settings';
+@import 'patterns_icons';
+@include vf-p-icon-minus;
+@include vf-p-icon-plus;
 
 @mixin vf-p-accordion {
   $icon-size: map-get($icon-sizes, accordion);
--- a/scss/_patterns_icons.scss
+++ b/scss/_patterns_icons.scss
@@ -1,7 +1,69 @@
+@import 'settings';
+
+$default-icon-size: map-get($icon-sizes, default);
+$social-icon-size: map-get($icon-sizes, social);
+
+@mixin vf-p-icons {
+  @include vf-p-icon-plus;
+  @include vf-p-icon-minus;
+  @include vf-p-icon-expand;
+  @include vf-p-icon-collapse;
+  @include vf-p-icon-chevron;
+  @include vf-p-icon-close;
+  @include vf-p-icon-help;
+  @include vf-p-icon-info;
+  @include vf-p-icon-delete;
+  @include vf-p-icon-error;
+  @include vf-p-icon-warning;
+  @include vf-p-icon-external-link;
+  @include vf-p-icon-contextual-menu;
+  @include vf-p-icon-code;
+  @include vf-p-icon-menu;
+  @include vf-p-icon-copy;
+  @include vf-p-icon-search;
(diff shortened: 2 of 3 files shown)


scss/_patterns_accordion.scss
```diff
@@ -38,12 +38,12 @@
 
       // aria-selected controls the open and closed state for the accordion tab
       &[aria-expanded='true'] {
-        @include vf-icon-minus(vf-url-friendly-color($color-mid-dark));
+        @include vf-icon-minus($color-mid-dark);
```
anthonydillon: If this pattern is using mixins from icons shouldn't we import icons at the top so people can include just this pattern?

PR #1877, opened 2018-06-19
Change h1 font-weight to 100

### LINKED ISSUE #1841: There should be a pattern to make headings thin
There is currently no way to change font-weights outside of custom/inline styling. We should have guidelines for when we should/shouldn't change font-weights, but we are using thin `h1`s in a few places so I think it should be a pattern.

screenshot_6

screenshot_5

Caleb-Ellis:
## Done

- Changed h1 font-weight to 100 (thin)

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/headings/default/
- Check that the top heading has `font-weight: 100`

## Details

Fixes #1841

### DIFF
--- a/scss/_base_typography.scss
+++ b/scss/_base_typography.scss
@@ -347,7 +347,7 @@
     @include heading-max-width--short;
     font-size: pow($ms-ratio, 8) * 1rem;
     font-style: normal;
-    font-weight: 300;
+    font-weight: 100;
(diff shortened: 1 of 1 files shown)


kwm14:
Looking good @Caleb-Ellis 🥇 will check Percy now

PR #1891, opened 2018-06-21
1879: Make media object spacing more reliable

### LINKED ISSUE #1879: Media object spacing is wrong if meta data is above description
Use Media object pattern, but instead of having description after heading place meta data list there and have description below:
```html
<div class="p-media-object">
  <img src="https://via.placeholder.com/72x72" class="p-media-object__image">
  <div class="p-media-object__details">
    <h3 class="p-media-object__title">
      <a href="#">Event title</a>
    </h3>
    
    <ul class="p-media-object__meta-list">
      <li class="p-media-object__meta-list-item--date">
        <span class="u-off-screen">Date: </span>21 - 23 February 2017
      </li>
      <li class="p-media-object__meta-list-item--venue">
        <span class="u-off-screen">Venue: </span>Venue 360
      </li>
      <li class="p-media-object__meta-list-item--location">
        <span class="u-off-screen">Location: </span>Santa Clara, CA
      </li>
    </ul>
    <p class="p-media-object__content">Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</p></div>
</div>
```
https://jsfiddle.net/zsfoq1xb/2/

On large screen there will be too much spacing between heading and meta data, while not enough spacing between meta data and description.

screen shot 2018-06-20 at 13 30 24

It seems to be caused by negative margin used on the description to move it closer to heading.

Surprisingly it seems to work quite nice on smaller screens:

screen shot 2018-06-20 at 13 32 25

## Expected result

Media object should not assume description will be always right after the heading. Spacing should work correctly regardless of description even being there at all.

Caleb-Ellis:
## Done

- Moved placeholders outside media object mixin so they can be invoked without including the whole pattern
- Updated the Media object - large example with a meta-list-item
- Made the media object spacing more flexible and reliable. You should now get nice spacing whether the content comes before or after the metadata, or whether either is present at all.

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/media-object/media-object/
- Turn on the baseline grid and make sure all text sits on it nicely (check small + large screens)
- Move `<p class="p-media-object__content">` below the metadata and check that the spacing is good, and the text still sits on the baseline (check small + large screens)
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/media-object/media-object/
- Do the same checks as with the default media object

## Details

Fixes #1879 

## Screenshots

screenshot_2

### DIFF
--- a/examples/patterns/media-object/media-object-large.html
+++ b/examples/patterns/media-object/media-object-large.html
@@ -9,5 +9,10 @@ category: _patterns
   <div class="p-media-object__details">
     <h1 class="p-media-object__title">Title</h1>
     <p class="p-media-object__content">Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</p>
+    <ul class="p-media-object__meta-list">
+      <li class="p-media-object__meta-list-item--location">
+        <span class="u-off-screen">Location: </span>London, UK
+      </li>
+    </ul>
   </div>
 </div>
--- a/scss/_patterns_media-object.scss
+++ b/scss/_patterns_media-object.scss
@@ -1,22 +1,25 @@
+@import 'settings';
+
+%p-media-object {
+  display: flex;
+  flex-shrink: 0;
+  margin-bottom: $spv-inter--scaleable;
+}
+
+%vf-meta-list-item {
+  @extend %small-text;
+  color: $color-dark;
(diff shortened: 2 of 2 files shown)


kwm14:
Nice fix @Caleb-Ellis design looking good 👍 will check Percy now

PR #1910, opened 2018-07-04
Fix accordion toggle so it opens & closes

### LINKED ISSUE #1909: Accordion pattern does not close panels when already opened
**Describe the bug**

The accordion pattern does not close panels when already opened

**To Reproduce**

- View accordion pattern
- Open a panel
- Try to close that same panel
- Panel does not close

**Expected behaviour**

- Panel should close

barrymcgee:
## Done

Fix accordion toggle so it opens & closes

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/accordion/
- Check that accordion opens and closes when individually clicked

## Details

Fixes #1909

### DIFF
--- a/examples/patterns/accordion.html
+++ b/examples/patterns/accordion.html
@@ -52,13 +52,15 @@ function setupAccordionListener(accordionContainerSelector) {
     .querySelector(accordionContainerSelector)
     .addEventListener('click', e => {
       const target = e.target;
-      // Find any open panels within the container and close them.
-      e.currentTarget
-        .querySelectorAll('[aria-expanded=true]')
-        .forEach(element => toggle(element, false));
-      // Toggle the target.
       const isTargetOpen = target.getAttribute('aria-expanded') === 'true';
-      toggle(target, !isTargetOpen);
+      // Find any open panels within the container and close them.
+      if (target.classList.contains('p-accordion__tab')) {
+        e.currentTarget
+          .querySelectorAll('[aria-expanded=true]')
+          .forEach(element => toggle(element, false));
+        // Toggle the target.
+        toggle(target, !isTargetOpen);
+      }
     });
 }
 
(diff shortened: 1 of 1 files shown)


Caleb-Ellis:
@barrymcgee I'm not sure we want to change the pattern to no longer act like an accordion, i.e. a true accordion only allows one panel open at a time. If we just want to fix #1909 it might be easier to just toggle the target as opposed to setting it to open, e.g.

``` javascript
// Toggle the target.
const isTargetOpen = target.getAttribute('aria-expanded') === 'true';
toggle(target, !isTargetOpen);
```

barrymcgee:
@Caleb-Ellis Good call - PR now amended.

examples/patterns/accordion.html
```diff
@@ -52,14 +52,13 @@
     .querySelector(accordionContainerSelector)
     .addEventListener('click', e => {
       const target = e.target;
-      if (target.classList.contains('p-accordion__tab')) {
```
Caleb-Ellis: We still need this conditional otherwise the function will run when you click the accordion panel too (which doesn't have an aria-expanded attribute)

examples/patterns/accordion.html
```diff
@@ -52,14 +52,13 @@
...
-        e.currentTarget
-          .querySelectorAll('[aria-expanded=true]')
-          .forEach(element => toggle(element, false));
-        // Open the target.
-        toggle(target, true);
-      }
+      // Find any open panels within the container and close them.
+      e.currentTarget
+        .querySelectorAll('[aria-expanded=true]')
+        .forEach(element => toggle(element, false));
```
Caleb-Ellis: This should only be run on tabs that aren't the target, otherwise it toggles off and immediately on again, keeping the panel open
``` javascript
if (element !== target) {
  toggle(element, false);
}
```

kwm14:
### Design review
- [ ] I'm unable to close a row when individually clicked

close-accordion

kwm14:
@barrymcgee do you have any time this week to fix these updates? 😺

barrymcgee:
@kwm14 @Caleb-Ellis Ready for re-review.

PR #1913, opened 2018-07-09
p-cards: Apply fixed padding

lyubomir-popov:
## Done

Card padding should not be scaleable. I've added a variable ($spv-intra--regular) to handle cases like this, and I've applied it to all the card flavours.

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/card/highlighted/ for example
- Verify that at $multi: 2; nothing changes; at $multi: 1 nothing changes as well. Previously, at $multi: 1 the padding would become .5 rem, which is not useful even in the context of high-density ui like maas.

## Details

## Screenshots
Before: ($multi: 1)
image

After: ($multi: 1)
image


scss/_settings_spacing.scss
```diff
@@ -53,6 +53,7 @@ $spv-nudge-compensation: $sp-unit - $spv-nudge !default;
 $spv-intra--condensed: $sp-unit * .5 !default; // buttons margins inside table,
 $spv-intra--condensed--scaleable: $spv-intra--condensed * $multi !default;
 $spv-intra: $sp-unit !default;
+$spv-intra--regular: $sp-unit * 2; //card padding
```
anthonydillon: Please remove this comment. We may use this variable for other patterns in the future.

lyubomir-popov:
@anthonydillon I've only removed the comment for now. As to the naming, everything that is not --scaleable is static. If we're adding it, I should rename everything that isn't scaleable to say 'static', but this makes you type more. 

Maybe I could rename it to $spv-intra--double? As it is two units?

.5 units    $spv-intra--condensed 
1 unit       $spv-intra 
1.5 units  $spv-intra--expanded 
2 units    $spv-intra--regular / --double (*) 


Or I can rename the others, so we have
.5 units    $spv-intra--ultra-condensed 
1 unit       $spv-intra--condensed 
1.5 units  $spv-intra
2 units    $spv-intra--expanded (*) 

The asterisk denotes the current one.

The condensed/expanded/regular notation is borrowed from typefaces, it gives a flexible number of levels (up to 7) - ultracondensed / condensed / semi-condensed / regular (or no suffix, as in $spv-intra) / semi-expanded / expanded / ultra-expanded. Might be worth thinking of another naming schemd if this one is confusing.

kwm14:
@lyubomir-popov to pick up and fix once he has completed work transitioning MAAS to v1.8.0 👍

PR #1938, opened 2018-07-24
Remove the inline-block whitespace between tabs

anthonydillon:
## Done
Remove the inline-block whitespace between tabs. Make the scrolling auto so it only shows when its needed.

## QA
- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/tabs/
- Check there are no spaces between the tab items on all screen sizes

## Details
Fixes https://github.com/vanilla-framework/vanilla-framework/issues/1898
Fixes https://github.com/vanilla-framework/vanilla-framework/issues/1899


Caleb-Ellis:
Tabs still wrap to a second line when there's a lot.

screenshot

You could probably get rid of `overflow: hidden` on `p-tabs__list` and give it `display: flex` and it should work?

kwm14:
LGTM 👍

PR #1956, opened 2018-08-09
Entrypoint for talisker gevent workers

nottrobin:
Fixes https://github.com/ubuntudesign/base-squad/issues/108

QA
--

``` bash
./run
```

See that it mentions "Booting worker" 3 times. Go to http://localhost:8104 - browse around, check the site works. Then:

``` bash
docker build --tag vf-docs --build-arg COMMIT_ID=`git rev-parse HEAD` .
docker run -ti -p 8099:80 vf-docs
```

See that it mentions "Booting worker" 3 times. Go to http://localhost:8099 - check the site works.

Also, check the `X-VCS-Revision` shows correctly:

``` bash
$ curl -Is localhost:8099 | grep VCS
X-VCS-Revision: {commit_hash}
```


pmahnke:
@nottrobin I am getting an error

```
docker build --tag vf-docs --build-arg COMMIT_ID=`git rev-parse HEAD` .
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /home/peter/src/vanilla-framework/Dockerfile: no such file or directory
```

also ./run works on port 8101, but not 8104

PR #1958, opened 2018-08-13
Add canonial url to context

tbille:
# Summary 

Serve canonical url to context 

# QA

- With this version of the documentation builder https://github.com/canonical-webteam/documentation-builder/pull/131
- `yarn build`
- `./run`
- The header of every page should have a canonical url


tbille:
We need to find another solution

PR #1963, opened 2018-08-20
Add sub-heading example to documentation

kwm14:
## Done

- Approved in WG group [here](https://github.com/ubuntudesign/vanilla-squad/issues/279
)
- Document sub-headings

## QA

- Pull code
- Run `./run serve --watch`
- [Add additional steps]

## Details

- Fixes issue here: https://github.com/ubuntudesign/vanilla-squad/issues/279

### DIFF
--- a/docs/en/base/typography.md
+++ b/docs/en/base/typography.md
@@ -63,6 +63,15 @@ modified to look like a particular heading size.
     View example of the heading pattern
 </a>
 
+### Sub-headings
+
+Sub-headings are used to visually convey importance beneath a heading or as a title when inserted as a divider between sections.
+
+<a href="https://vanilla-framework.github.io/vanilla-framework/examples/base/sub-headings/"
+    class="js-example">
+    View example of the heading pattern
+</a>
+
 ### Mixed heading classes
 
 It is also possible to apply heading classes directly to heading elements if that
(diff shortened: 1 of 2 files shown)


docs/en/base/typography.md
```diff
@@ -63,6 +63,15 @@ modified to look like a particular heading size.
     View example of the heading pattern
 </a>
 
+### Sub-headings
+
+Sub-headings are used to visually convey importance beneath a heading or as a title when inserted as a divider between sections.
```
matthewpaulthomas: “a title when inserted as a divider between sections” sounds very much like just a subsection heading `<h2>`.

I think what we mean by subheading, here, is “a line of text that expands on the meaning of the heading immediately before it, rather than being a standalone heading for the text after it”.

examples/base/sub-headings.html
```diff
@@ -0,0 +1,8 @@
+---
+layout: default
+title: Sub-headings
+category: _base
+---
+<h1>Title</h1>
+<p class="p-heading--three">Sub-heading</p>
```
matthewpaulthomas: So if an `<h2>` has a subheading, should that be `<p class="p-heading--four">`? Maybe be explicit about using whichever size is two levels below the heading.

kwm14: @matthewpaulthomas you make a good point. Will update 😄 thanks

matthewpaulthomas:
“Convey importance beneath a heading” is a little unclear — convey importance of what? But that‘s a quibble.

PR #1994, opened 2018-09-18
Button styling and state updates

kwm14:
## Done

Updates to button styling and a new active (pressed) state.
- Update `base` and `neutral` button from normal text by using link blue `$color-link`
- Change the border radius to `4px` to make them look different to text fields
- Update border colour on `neutral` button to `$color-mid`
- Added an `active` state to all buttons

## QA

- Pull code
- Run `./run serve --watch`
- Check buttons look good on styles:
  - Base, Neutral, Brand, Positive and Negative

## Details

Fixes issues here: https://github.com/vanilla-framework/vanilla-framework/issues/1932

## Visual
buttons

### DIFF
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -10,7 +10,7 @@
   %vf-button-base {
     @include vf-animation (#{background-color, border-color}, fast, in);
     @include vf-focus;
-    border-radius: $border-radius-button;
+    border-radius: $border-radius;
     border-style: solid;
     border-width: 1px;
     cursor: pointer;
@@ -81,7 +81,7 @@
 /// override colors as desired
 @mixin vf-button-pattern (
   $button-background-color: $color-x-light,
-  $button-text-color: $color-link,
+  $button-text-color: $color-x-dark,
   $button-disabled-background-color: $color-transparent,
   $button-disabled-border-color: $color-mid,
   $button-border-color: $color-mid,
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -87,7 +87,7 @@
   $button-border-color: $color-mid,
   $button-hover-background-color: $color-light,
   $button-hover-border-color: $color-mid,
-  $button-active-background-color: darken($color-mid, 20%),
+  $button-active-background-color: darken($color-light, 20%),
   $button-active-border-color: $color-mid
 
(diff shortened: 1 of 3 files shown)


kwm14:
@deadlight currently having an issue with `active` state on `base` and `neutral` buttons not displaying, and I can't seem to work out why. Your help will be greatly appreciated 👍

spencerbygraves:
@kwm14 it looks like the hover is using the press colour, as there is currently no difference between the two. Also, can we put the 4px border on the press outline please, then we'll decide on that. Cheers.

kwm14:
@spencerbygraves thanks for your feedback.

- The hover state is using 10% of the button background colour and the pressed state is using 20% of the button background colour, but as they're very close it's difficult to see the difference. Maybe we need to increase the % more on pressed state to make it more visible?
- In regards to the outline on the pressed state that is not possible to do, I asked @deadlight and @Caleb-Ellis but as it's outside the element your unable to add any CSS styling to it

kwm14:
## Design review
_with @spencerbygraves_ 
- tested pressed outline but agreed it doesn't look great so happy to remove
- pressed state and hover state background color look the same - investigate

Caleb-Ellis:
@kwm14 it's actually possible to have a radius on the outline, we would just have to change the structure a little bit from an `outline` to an `::after` pseudo-element.

scss/_base_button.scss
```diff
@@ -97,6 +100,11 @@
   }
 
   &:active,
+  &:active:hover {
+    background-color: $button-active-background-color;
+    border-color: $button-active-border-color;
```
Caleb-Ellis: Have you tried adding `transition-duration: 0s;`? It seems to fade to a darker colour pretty slowly

kwm14: Nope, but I can do 👍 thanks

deadlight:
@kwm14 Go through the Percy diff because this introduces a lot of changes. We can fix up this branch to exclude things that we don't want to update.

kwm14:
Reviewed with @spencerbygraves and agreed to revert `base` and `neutral` text color back to `$color-x-dark` and border-radius as it affects other components in the framework. Which will need more time to investigate and solve 👍

PR #2012, opened 2018-10-04
Add Prettier ✨

barrymcgee:
## Done

- Add Prettier to the codebase
- Edit `.sass-lint` to remove indentation rule as Prettier takes care of that
- Add a `precommit` hook

## QA

- Pull code
- Run `./run serve --watch`
- Ensure no tests fail
- Ensure no visual regressions


barrymcgee:
@anthonydillon Big up Percy which caught a very hard to spot anomaly first time round, namely - Prettier will make `- map-get($foo, bar);`, `-map-get($foo, bar);` which appears to then not parse correctly in Sass as a negative value?

Having set Percy to ignore those negative functions, it's reporting [one single regression](https://percy.io/vanilla-framework/vanilla-framework/builds/1062297?utm_campaign=vanilla-framework&utm_content=vanilla-framework&utm_source=github_status) with the code-numbered pattern. 

It seems to be the font-size in that pattern has reduced slightly even though it didn't actually make any changes to that specific `.scss` file. Any ideas? 🤨

barrymcgee:
👆🏻Actually, not the fault of Prettier - the functions weren't interpolated properly.

`- map-get($foo, bar)` should be `-#{map-get($foo, bar)};`

Have fixed that up and running Percy again.

deadlight:
@barrymcgee can you check out the things that Percy has flagged? It looks like there are some regressions.

barrymcgee:
@deadlight This should be now ready to roll ✨

PR #2027, opened 2018-10-16
WIP: Action feedback demo

Caleb-Ellis:
http://0.0.0.0:8101/vanilla-framework/examples/templates/action-feedback/

### DIFF
--- a/examples/templates/action-feedback.html
+++ b/examples/templates/action-feedback.html
@@ -1,0 +1,303 @@
+---
+layout: default
+title: Action feedback
+category: _templates
+---
+
+<section class="p-strip">
+  <div class="row">
+    <div class="col-3">
+      <h3>Base</h3>
+      <h4 class="p-muted-heading">Indeterminate</h4>
+      <button class="p-button--base p-action-feedback is-indeterminate is-cancellable js-indeterminate">
+        <span class="p-action-feedback__text">Loading...</span>
+      </button>
+      <h4 class="p-muted-heading">Determinate</h4>
+      <button class="p-button--base p-action-feedback is-determinate is-cancellable js-determinate">
+        <svg class="p-action-feedback__svg">
+          <circle class="p-action-feedback__circle" r="7" cx="8" cy="8"></circle>
+        </svg>
+        <span class="p-action-feedback__text">75%</span>
+      </button>
+      <h4 class="p-muted-heading">Determinate steps</h4>
+      <button class="p-button--base p-action-feedback is-stepped is-cancellable js-stepped">
+        <svg class="p-action-feedback__svg">
+          <circle class="p-action-feedback__circle" r="7" cx="8" cy="8"></circle>
+        </svg>
+        <span class="p-action-feedback__text">2 of 5</span>
+      </button>
+      <h4 class="p-muted-heading">Non-cancellable</h4>
+      <button class="p-button--base p-action-feedback js-noncancel">
+        <span class="p-action-feedback__text">Action</span>
+      </button>
+      <h4 class="p-muted-heading">Done</h4>
+      <button class="p-button--base p-action-feedback is-done js-indeterminate">
+        <span class="p-action-feedback__text">Done</span>
+      </button>
+    </div>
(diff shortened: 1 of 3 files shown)


kwm14:
💯 💯 💯

examples/templates/action-feedback.html
```diff
@@ -0,0 +1,303 @@
...
+      <button class="p-button--base p-action-feedback is-stepped is-cancellable js-stepped">
+        <svg class="p-action-feedback__svg">
+          <circle class="p-action-feedback__circle" r="7" cx="8" cy="8"></circle>
+        </svg>
+        <span class="p-action-feedback__text">2 of 5</span>
+      </button>
+      <h4 class="p-muted-heading">Non-cancellable</h4>
+      <button class="p-button--base p-action-feedback js-noncancel">
+        <span class="p-action-feedback__text">Action</span>
+      </button>
```
matthewpaulthomas: Encouraging developers to add cancel behaviour to a commit button has several drawbacks:

1. It makes how to cancel (and even the _possibility_ of cancelling!) much less obvious to users, because the “Cancel” label is not visible except on hover.

2. Quite often, a commit button is in a dialog with an existing “Cancel” button right next door. This pattern would then result in two “Cancel“ buttons alongside each other.

3. If you double-clicked the button accidentally, the action would cancel and you wouldn’t know why.

4. If you clicked “Cancel” a fraction of a second too late, the action _wouldn’t_ cancel and you wouldn’t know why. Unless the developer had been extremely thoughtful about including a timeout, the action might even repeat from scratch!

For these reasons, when I’ve specified in-dialog progress (as in the two examples above), I’ve specified that everything in the dialog except the existing Cancel button should be greyed out. That way, you can cancel during the action exactly the same way that you cancel before the action: by clicking the existing Cancel button. And it’s super obvious, because it’s the only enabled element in the dialog.

I don’t think it’s necessary to implement the “disable everything else” behavior in Vanilla. But I do think we should _discourage_ the behavior shown here of an action button becoming a Cancel button during an action. That would also simplify this pattern further: you wouldn’t need this “non-cancellable” variation.

examples/templates/action-feedback.html
```diff
@@ -0,0 +1,303 @@
...
+      <h4 class="p-muted-heading">Done</h4>
+      <button class="p-button--base p-action-feedback is-done js-indeterminate">
+        <span class="p-action-feedback__text">Done</span>
+      </button>
```
matthewpaulthomas: Changing the button label to “Done” has the same drawbacks as changing it to progress text: it alters the button width, and causes adjacent elements to jump horizontally. Instead, I specified that the label should be _replaced by_ a checkmark for a couple of seconds until the dialog closes.

Caleb-Ellis:
Thanks, @matthewpaulthomas! Those are some great insights. I'll add this to the next Vanilla working group to discuss more and get the ball rolling on an "official" implementation.

lilyvidenova:
Thanks for all your comments @matthewpaulthomas! 

I agree the best way to discuss these will be in the Vanilla working group as all our propositions above make sense and are useful in the context of MAAS, however may not necessarily apply to other applications/sites.

kwm14:
## WG
- Not ready to be proposed to Vanilla
- Agreed further conversation needs to happen with @lilyvidenova @cassiocassio
- New discussion that buttons could have fixed widths to stop jumping as mentioned by @matthewpaulthomas above
- Do we need determinate steps?

matthewpaulthomas:
@lilyvidenova For each of the five states, I guess there’s three possibilities:
- multiple Vanilla sites should be using it → add to Vanilla
- “may not necessarily apply to other applications/sites” → keep specific to MAAS
- even MAAS would benefit from changing to something else → change to something else

Screenshots or videos would help here. For starters, here’s a screencast of the “Add repos” dialog in build.snapcraft.io:

progress-button

If it was desired (and possible) to show more granular progress feedback, I wouldn’t put it inside the button, I’d replace the selection count alongside the Cancel button, for example “1 of 3 added…”.

PR #2031, opened 2018-10-19
2016 - Update button paddings with and without icon

### LINKED ISSUE #2016: Button height in unreleased latest vanilla not matching height of other for elements any more
**Describe the bug**
The height of the button has ben altered by this commit:
git blame scss/_base_button.scss
e48d3409 2018-08-06 14:38:52 +0100  22   padding: $sp-unit * 1.5;

This makes the button height 44px, instead of the previous 37. This results in the button being higher than all other input elements.

**To Reproduce**
1 Pull latest develop
2 Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/forms/form-inline/
3 Observe mismatch of heights:
image
(Not specific to the example above, as the padding haas been applied to the base pattern).

**Expected behavior**
Button matching other input element heights

**Screenshots**
(Pls see above)

Caleb-Ellis:
## Done

- Reverted top and bottom button padding to match form input elements
- Changed left and right padding to `.75rem/12px`

## QA

- Pull code
- Run `./run serve --watch`
- Go to http://0.0.0.0:8101/vanilla-framework/examples/patterns/buttons/icon/
- Check that the button with only icon is square (note it's not a *perfect* square - 38 x 36.78)
- Go to http://0.0.0.0:8101/vanilla-framework/examples/patterns/forms/form-inline/
- Check that the buttons are the same height as the form inputs
- Check both in Chrome and Firefox because there's some weirdness with how each determine inline-block spacing

## Details

Fixes #2016 

## Screenshots
screenshot

screenshot_1

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -61,7 +61,6 @@ PLATFORMS
 DEPENDENCIES
   jekyll (~> 3.8.4)
   jekyll-redirect-from
-  jekyll-watch (= 2.0.0)
 
 BUNDLED WITH
    1.16.6
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -17,9 +17,9 @@
     display: inline-block;
     font-size: 1rem;
     font-weight: 300;
-    line-height: 1;
+    line-height: map-get($line-heights, default-text);
     margin-bottom: $spv-inter--scaleable + 2 * $spv-nudge-compensation;
-    padding: $sp-unit * 1.5;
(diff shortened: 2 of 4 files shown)


kwm14:
Looking good @Caleb-Ellis just a small tweak 😉 

- Update left/right margin from `9px` to `0.75rem/12px`
- Original build design spec here: https://github.com/vanilla-framework/vanilla-framework/issues/1902

button-with-icon-and-text-003

kwm14:
Thanks @Caleb-Ellis LGTM 👍

Issue #2028, opened 2018-10-17
Remove version dropdown as we don't support versioning


kwm14:
**Describe the bug**
Remove version dropdown from Docs homepage as we don't support versioning so users can look back at previous releases.

**To Reproduce**
1. Go to https://docs.vanillaframework.io/en/
2. Click on v1.8.0
3. Nothing happens

![screenshot 2018-10-17 at 16 24 03](https://user-images.githubusercontent.com/17748020/47097610-a9db1700-d229-11e8-93db-693c5e0f4fba.png)

**Expected behavior**
Remove version dropdown and allow search to move up.

![screenshot 2018-10-17 at 16 24 51](https://user-images.githubusercontent.com/17748020/47097617-ae9fcb00-d229-11e8-8c6b-397e4a4f8f65.png)

PR #2035, opened 2018-10-23
Remove versioning dropdown from Docs homepage

kwm14:
## Done
Remove version dropdown from Docs homepage as we don't support versioning so users can look back at previous releases.
- Side note mentioned in Dev catch up this week:
> Robin: Something to consider is the fact that we are moving away from versioned documentation (via discourse).

## QA

- Pull code
- Run `./run serve --watch`
- Check Docs homepage to see that version dropdown has been removed

## Details
- Fixes issue #2028 

## Screenshots
screenshot 2018-10-23 at 12 16 14

### DIFF
--- a/docs/templates/en/index.html
+++ b/docs/templates/en/index.html
@@ -1,0 +1,1094 @@
+<!doctype html>
+<html lang="en">
+
+  <head>
+
+    <!-- Google Tag Manager -->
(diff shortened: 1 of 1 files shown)


kwm14:
Conflicts with this PR. Will close and redo 👍

PR #2063, opened 2018-11-26
Fix strip borders

### LINKED ISSUE #2045: Bordered, coloured p-strips have no border
Because of a -1px margin-bottom applied to `p-strip is-bordered`, the border is actually hidden behind the following strip's background (if it has one).
To reproduce, write the following markup:
``` html
<section class="p-strip is-bordered"></section>
<section class="p-strip--light is-bordered"></section>
```

The first strip does not have a visible border

deadlight:
## Done

Changed the borders on strips - which caused issues with the baseline grid - by making p-strip position: relative by default and use %vf-pseudo-border--bottom

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/strips/is-bordered/
- See that the strip is bordered
- Add another strip to the markup with the "--light" modifier and see that the border is visible
- High five me

## Details

Fixes #2045

### DIFF
--- a/scss/_patterns_strip.scss
+++ b/scss/_patterns_strip.scss
@@ -37,6 +37,7 @@
   %vf-strip {
     @extend %section-padding--default;
     clear: both;
+    position: relative;
     width: 100%;
   }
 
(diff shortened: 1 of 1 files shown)


kwm14:
LGTM. Will check Percy now 👍

PR #2082, opened 2018-12-14
Use 100vh instead of height 100% on body with sticky footer

bartaz:
## Done

Updates `height: 100%` into `min-height: 100vh` for sticky footer.
Using `height: 100%` was causing layout issues (forcing body to always have the height of the viewport). For example https://github.com/canonical-webteam/global-nav/issues/62

Using `min-height: 100%` didn't do the trick, but it seems new viewport units can do what we need.
Now `html` and `body` will have minimum height of the viewport when there is little content, and will grow larger when content doesn't fit on the screen.

## QA

- Pull code
- Enable $sticky-footer
- Create an example page with sticky footer
- Run `./run serve --watch`
- Check if sticky footer works

### DIFF
--- a/scss/_utilities_sticky-footer.scss
+++ b/scss/_utilities_sticky-footer.scss
@@ -1,7 +1,7 @@
 @mixin vf-u-sticky-footer {
   html,
   body {
-    height: 100%;
+    min-height: 100vh;
   }
 
   body {
--- a/scss/_utilities_sticky-footer.scss
+++ b/scss/_utilities_sticky-footer.scss
@@ -1,15 +1,17 @@
 @mixin vf-u-sticky-footer {
-  html,
-  body {
-    min-height: 100vh;
-  }
+  @media screen and (min-width: $breakpoint-small) {
+    html,
+    body {
+      min-height: 100vh;
+    }
 
-  body {
-    display: flex;
-    flex-direction: column;
-  }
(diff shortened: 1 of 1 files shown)


lyubomir-popov:
@bartaz one thing to be aware of with vh units is, they cause a 1-3 second freeze on older mobile devices like iphone 4 or 5. This is because it can take that long to recalculate the position of every element on the page for the duration of the sliding browser chrome animation. This happens every time you change scroll direction=, which triggers the chrome (address bar etc) to appear / disappear

deadlight:
The only way we could target the older iOS devices that have this issue would be with JavaScript, which isn't an option at a Vanilla-wide level.

It would be great if we could test with an older iOS device to see if this is still an issue with software updates, etc.

I'll take a look and see what portion of our visitors are on apple devices from this era to inform our decision.

bartaz:
@deadlight As these styles are not turned on by default, and require separate build of Vanilla I guess the usage may be quite limited. So also worth checking who is using it.

Also, if only mobile devices are affected we could wrap it in media query. Very likely 'sticky' footer would not be needed on mobile (there will be enough content to push footer below the fold). So we could turn it on above the mobile threshold only.

deadlight:
@bartaz: can you make the above changes, please? We can then merge this.

PR #2120, opened 2019-01-22
Updates to 'Icons' layout on docs

### LINKED ISSUE #2058: The tables of icons don’t include the icons
_Original comment by @matthewpaulthomas posted [here](https://github.com/canonical-websites/vanillaframework.io/issues/183)_ 

[The “Icons” page](https://docs.vanillaframework.io/en/patterns/icons) contains two tables of icons: the first for all other icons, and the second for social icons.

Each table has two columns: the first gives the “CLASS NAME” for each icon, and the second … is completely empty. Neither column actually contains the icon. The icons are shown, unlabelled, in a box below. This makes it gratuitously difficult to tell which is which.

icons-page-table

What should happen: Each icon is shown before its name in the table.

kwm14:
## Done

- Updated the icon table to include classes
- New visual layout

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/icons/
- Check that all three of the examples in the list are updated correctly on documentation
  - Light / Dark / Social

## Details
- Fixes #2058 

## Screenshots
screenshot 2019-01-22 at 11 24 36
screenshot 2019-01-22 at 11 24 45

### DIFF
--- a/docs/en/patterns/icons.md
+++ b/docs/en/patterns/icons.md
@@ -109,7 +109,7 @@ Icons can be added via an `<i>` element like so: `<i class="p-icon--{name}"></i>
 When placed within `.p-strip--dark`, icon colors are reverted to ensure legibility.
 
 <section>
-  <div class="p-strip--dark is-shallow">
+<div class="p-strip--dark is-shallow" style="background-color: transparent;">
     <div class="row u-equal-height">
       <div class="p-card col-4 u-vertically-center" style="background-color:#111;">
         <i class="p-icon--plus" style="margin-right: 1rem;"></i><code style="color: #fff">p-icon--plus</code>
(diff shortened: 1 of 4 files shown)


kwm14:
## Desk review
_with @deadlight_ 
- drag icon not displaying - investigate 🤓 
- chevron icon not displaying - investigate 🤓 
- add `u-equal-height` class to row so when the page scales each row are aligned 
- remove JS examples as they're not needed with new table 😉 
- add dark version of new icon layout

deadlight:
@kwm small issue with the social icons at smaller screens:
image

kwm14:
@deadlight thanks for spotting. I've tried multiple options but can't seem to resolve 😭 
- `u-vertically--center` only affects medium and large screens so that's why it's occurring
- I've tried adding `line-height` to the text so it matches the height of the icon
- I've tried mixed heading classes as `h3` line height is the same as the icon at `2rem`

It would be great if we can fix up but I'm not going to lose sleep over it, as stats show 11% of users are viewing documentation on their mobile devices.

PR #2126, opened 2019-01-23
move base typography rules at beginning of base includes

lyubomir-popov:
## Done

## QA

- Pull code
- Run `./run serve --watch`

### DIFF
--- a/scss/_base.scss
+++ b/scss/_base.scss
@@ -1,4 +1,6 @@
 // Base
+@import 'base_placeholders';
+@import 'base_typography';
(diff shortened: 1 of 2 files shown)


lyubomir-popov:
@Caleb-Ellis do you remember why we include _typography.scss after tables? Wondering if it is just an overlooked thing or if there was actually a reason that requires that include order

deadlight:
@lyubomir-popov, @Caleb-Ellis - it looks like the ordering is just alphabetical so it's fine to impose a specific order.

PR #2162, opened 2019-02-19
Add max-width to <label> elements

### LINKED ISSUE #2154: Labels lacking max-width
**Describe the bug**
Since optimum line length applies to all types of text, I think it makes sense to apply this as the default behaviour. We already have a utility that removes the max-width as an override, should this be required somewhere. [feedback](https://zpl.io/br81Y3L).

**To Reproduce**
1. Go to 'https://vanilla-framework.github.io/vanilla-framework/examples/base/forms/labels/'
2. Replace label content with 'List of external networks (not previously known), that will be allowed to use MAAS for DNS resolution.'
3. Observe lack of wrapping

**Expected behavior**
Max-width applying to forms elements.

**Screenshots**
Currently: (the lack of wrapping is visible on the first and third label - "Upstream DNS...")
image

Proposed change:
image

deadlight:
## Done

Added the paragraph max width to labels for consistency in typographic elements.

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/forms/form-help-text/
- Inspect the element and add enough words for the label to wrap
- See that it's now wrapping sooner

## Details
Fixes #2154

### DIFF
--- a/scss/_base_typography.scss
+++ b/scss/_base_typography.scss
@@ -158,4 +158,8 @@
     cursor: help;
     text-decoration: none;
   }
+
+  label {
+    @include p-max-width;
+  }
 }


kwm14:
LGTM 👍

PR #2164, opened 2019-02-21
Remove  $px  var in preparation for font-size change

lyubomir-popov:
## Done

To be able to change the :root font-size on a breakpoint, we need to remove any variables that assume 1 rem is a constant value. This is because any calculations will be wrong once that value changes from 16 to e.g. 18.

In our case, the only such variable is $px, which is used where the presence of a bordr shouldn't ypset the vertical rhythm. E.g. the padding on a button is calculated as
```{padding top in rems} -  $px, where $px is .0625rem (1/16 of a rem)```

This pr changes these calculations to:
```calc(#{rem-value} - 1px)```

In some places as in the checkbox, px values are converted to rem values, so they scale properly as well.

## QA

- Pull code
- Run `./run serve --watch`
- Verify nothing looks broken

## Details

This takes care of https://github.com/ubuntudesign/vanilla-squad/issues/510 and a small subset of https://github.com/ubuntudesign/vanilla-squad/issues/509

### DIFF
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -19,7 +19,7 @@
     font-weight: 300;
     line-height: map-get($line-heights, default-text);
     margin-bottom: $spv-outer--scaleable + 2 * $spv-nudge-compensation;
-    padding: $spv-nudge - $px $sph-inner--small * 1.5;
+    padding: calc(#{$spv-nudge} - 1px) $sph-inner--small * 1.5;
     text-align: center;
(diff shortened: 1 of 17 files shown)


lyubomir-popov:
@kwm14 visual diff shows differences due to accidental fixes :)
The plus was not perfectly centered vertically. Now it is:
image

The text under the context menu buttons wasn't sitting on the baseline. It is now:
image

PR #2171, opened 2019-02-22
Add u-no-print utility to exclude elements from printed pages

### LINKED ISSUE #1982: Add a print stylesheet with a .u-no-print class
We should have a print.css file in Vanilla.

We should borrow ideas from the https://www.ubuntu.com/static/css/print.css file and making printing with Vanilla great!

deadlight:
## Done

Added `u-no-print` utility

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/utilities/no-print/
- Pretend to print the page and see that one of the paragraphs doesn't show in the preview (don't really print, think of the :deciduous_tree:!)
- `cd docs`
- `./run`
- See that the [documentation exists](http://0.0.0.0:8104/en/utilities/no-print)

## Details

Fixes #1982

### DIFF
--- a/docs/en/metadata.yaml
+++ b/docs/en/metadata.yaml
@@ -64,6 +64,9 @@ navigation:
       - location: utilities/hide.md
         title: Hide
 
+      - location: utilities/no-print.md
+        title: No print
+
       - location: utilities/margin-collapse.md
         title: Margin collapse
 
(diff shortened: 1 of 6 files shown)


kwm14:
Nice. LGTM 👍

PR #2177, opened 2019-02-26
Add local font of ‘Ubuntu’ to all font-face calls

barrymcgee:
https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization

## Done

Add local font of ‘Ubuntu’ to all font-face calls - many of our users will be running Ubuntu as their OS so will have Ubuntu already available to them.

## QA

- Pull code
- Run `./run serve --watch`
- Ensure no Percy changes

### DIFF
--- a/scss/_base_fontfaces.scss
+++ b/scss/_base_fontfaces.scss
@@ -4,49 +4,49 @@
       font-family: 'Ubuntu';
       font-style: normal;
       font-weight: 300;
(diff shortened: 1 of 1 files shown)


deadlight:
The local version of the ubuntu font doesn't have the thin variant so we're unable to rely on it for our websites

deadlight:
`develop` branch:
image

This PR's demo:
image

Both are FF on Ubuntu 18:10

barrymcgee:
Annoying - I wonder if we should try to get Google fonts to host the correct font stack? Anyway, not the issue for here.

PR #2183, opened 2019-03-05
Add message on end build

tbille:
- `./run`
- should have a message at the end: `Serve at http://0.0.0.0:8104`


tbille:
I just rememebered something, will do a small update

PR #2222, opened 2019-03-19
The trouble with unitless line-heights

lyubomir-popov:
## Done

I've been asked a few times why we're not using unitless line-heights, so I made this pr to demonstrate. When using fractional line-heights, they are rounded to 1.0 1.33 1.66 2.0. Values in between are snapped to one of these. Since the vertical rhythm we use relies on multiples of .5rem, we can't get the exact values as specced.

## Details

- Pull code
- Run `./run`
- Look at the typographic spacing example, try changing line-heights to see the peculiar behaviour.

### DIFF
--- a/_jekyll/_layouts/default.html
+++ b/_jekyll/_layouts/default.html
@@ -12,7 +12,7 @@
         {% endif %}
     </head>
 
-    <body style="margin: 1rem;">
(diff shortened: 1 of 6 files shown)


lyubomir-popov:
This was just to demonstrate why unitless line-heights don't work. Closing.

PR #2228, opened 2019-03-21
Remove sidebar navigation and docs example

### LINKED ISSUE #1623: Deprecate sidebar navigation component from framework
## Process

Remove tagline from sidebar navigation component as it isn't in the Design spec.

1. Component: https://docs.vanillaframework.io/en/patterns/navigation#sidebar-navigation
2. Design spec: https://github.com/ubuntudesign/vanilla-design/tree/master/Sidebar%20navigation

## Expected result
sidebar-navigation

## Current result
screen shot 2018-03-14 at 17 13 47

deadlight:
## Done

Removed the sidebar navigation component.
Also removed the "docs" layout example since it's only purpose was showcasing this component.

## QA

- Pull code
- Run `./run serve --watch`
- See that the component no longer exists

## Details

Fixes #1623


docs/sass/_settings.scss
```diff
@@ -2,7 +2,6 @@ $breakpoint-medium: 900px;
 $grid-max-width: none;
 
 $content-padding: 0 5%;
-$sidebar-width: 18rem;
```
b-m-f: This is still used in https://github.com/vanilla-framework/vanilla-framework/blob/1fcc0471ff0dccbaeded9cef47f72351c74f638f/docs/sass/styles.scss#L83 and gets redefined in https://github.com/vanilla-framework/vanilla-framework/blob/16048244cdea109353b4d912baf084aef284f5eb/scss/_patterns_code-numbered.scss#L2 . 

Second one shouldnt be a problem, but maybe another name at a later point?

b-m-f:
CI is breaking

docs/sass/_settings.scss
deadlight: Explained in-person:
 - /docs has it's own nested, NPM-installed version of Vanilla so it can continue using the variable until it is upgraded to v2.0
 - the same-named variable is scoped to that component and is OK to stay

PR #2231, opened 2019-03-25
Grid: Add classes to reset margin of first column on rows other than first, per breakpoint

lyubomir-popov:
## Done

Shelves grid lacks the ability to specify first-column per breakpoint. 
This gets in the way of making responsive layouts. For example:

Say you want to create a logo strip containing 12 logos, 2 rows x 6 logos on desktop, 3 rows x 4 logos on mobile.

If you put 12 logos in one row, you have the following problem:
on desktop: the first logo on the second row is offset by a margin-left, because the grid is not aware you're in column 1.

If you apply u-no-margin--left on that logo to fix the alignment, you're breaking the mobile view, because the utility removes margins across all breakpoints.

To solve this, we need what is commonly implemented as a first-col class - a reset of the margin that only affects mobile/tablet/desktop.

This pr provides those classes.

Example of this working can e found on the maas homepage in this pr: 
https://maas-io-canonical-websites-pr-316.run.demo.haus/

## QA

- Pull code
- Run `./run serve --watch`
- [Add additional steps]

## Details

## Screenshots

### DIFF
--- a/scss/_utilities_grid.scss
+++ b/scss/_utilities_grid.scss
@@ -1,17 +1,17 @@
 @mixin grid-first-col {
-  .u-first-col--mobile {
+  .u-first-col--small {
     @media (max-width: $breakpoint-small) {
       margin-left: 0;
     }
   }
 
-  .u-first-col--tablet {
+  .u-first-col--medium {
     @media (min-width: $breakpoint-small) and (max-width: $breakpoint-medium) {
       margin-left: 0;
     }
   }
 
-  .u-first-col--desktop {
+  .u-first-col--large {
     @media (min-width: $breakpoint-medium) {
       margin-left: 0;
     }
--- a/scss/_utilities_grid.scss
+++ b/scss/_utilities_grid.scss
@@ -1,18 +1,21 @@
 @mixin grid-first-col {
   .u-first-col--small {
     @media (max-width: $breakpoint-small) {
+      clear: left;
       margin-left: 0;
     }
   }
 
   .u-first-col--medium {
     @media (min-width: $breakpoint-small) and (max-width: $breakpoint-medium) {
+      clear: left;
       margin-left: 0;
     }
   }
 
   .u-first-col--large {
     @media (min-width: $breakpoint-medium) {
+      clear: left;
       margin-left: 0;
     }
   }
--- a/scss/_utilities_grid.scss
+++ b/scss/_utilities_grid.scss
@@ -1,0 +1,19 @@
+@mixin grid-first-col {
+  .u-first-col--mobile {
+    @media (max-width: $breakpoint-small) {
+      margin-left: 0;
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


deadlight:
@lyubomir-popov this is failing CI. Can you check that it runs locally.

kwm14:
@lyubomir-popov comment above by @deadlight 👍

lyubomir-popov:
@kwm14 @deadlight this one is ready for another review as well

scss/_utilities_grid.scss
```diff
@@ -0,0 +1,19 @@
...
+    }
+  }
+
+  .u-first-col--tablet {
+    @media (min-width: $breakpoint-small) and (max-width: $breakpoint-medium) {
+      margin-left: 0;
+    }
+  }
+
+  .u-first-col--desktop {
```
anthonydillon: We don't use device names in Vanilla. We use small, medium and large.

scss/_utilities_grid.scss
```diff
@@ -0,0 +1,19 @@
+@mixin grid-first-col {
+  .u-first-col--mobile {
```
anthonydillon: What is the use case for these?

scss/_utilities_grid.scss
```diff
@@ -0,0 +1,19 @@
+@mixin grid-first-col {
```
anthonydillon: Couldn't we just add sizes to `u-no-margin` and `u-no-padding`. These classes are very specific.

deadlight: I agree that they're very specific to the pattern but is that a bad thing. 

Using utility classes to fix a pattern isn't a good plan, the pattern should work with its own BEM classes. We should discourage the use of utilities unless they're absolutely necessary. This is a tricky use-case as they're workaround classes for a tricky problem, so it's not quite so black-and-white but I, personally, prefer this to using utilities.

scss/_utilities_grid.scss
deadlight: They are for selecting the first column of wrapping items in lists, from what I understand.

i.e., if you have 8 items on a large viewport and two rows of 4 items wrapping in a small viewport then the 5th item needs the `.u-first-col--mobile` (I agree with your other comment, it should be `.u-first-col--small`) to fix the first-item specific styling.

lyubomir-popov:
@anthonydillon @deadlight I'm closing this. Turns out shelves grid actually has margin reset classes for first columns, they're just not documented anywhere ([actually, the docs explicitly say there aren't .first classes](http://shelvesgrid.org/docs/css#columns)). But in the generated css you can search for  e.g. .first-mobile-col, which I traced to https://github.com/vanilla-framework/vanilla-framework/blob/48b54155b5e3f1a16af4348357929bb26360c503/scss/grid/shelves/_variables.scss#L61

PR #2232, opened 2019-03-25
New example: Responsive logo strip using the proposed u-first-col--* classes

lyubomir-popov:
## Done

This is meant to replace the inline images pattern. In order to reduce the amount of rewriting necessary to move a site to vanilla 2.0, I'd suggest deprecating the inline images pattern on a point release rather than 2.0.

Also unlike inline-images, this doesn't need to be a pattern - it can be built with vanilla plain grid classes. Still useful to have as a reference implementation we can point people to.

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/logo-strip.html
- Verify logo strip looks ok on all breakpoints.

## Details

## Screenshots

### DIFF
--- a/examples/patterns/logo-strip.html
+++ b/examples/patterns/logo-strip.html
@@ -1,0 +1,44 @@
+---
+layout: default
+title: Logo strip
+category: _patterns
+---
+
+<div class="row">
+    <div class="mobile-col-1 tablet-col-1 col-2 u-first-col--tablet u-first-col--desktop">
+        <img src="https://assets.ubuntu.com/v1/4c6fb640-logo-verizon.png" alt="verizon logo">
+    </div>
+    <div class="mobile-col-1 tablet-col-1 col-2">
+        <img src="https://assets.ubuntu.com/v1/bf709df8-logo-db.png" alt="db logo">
+    </div>
+    <div class="mobile-col-1 tablet-col-1 col-2">
+        <img src="https://assets.ubuntu.com/v1/22fe3811-logo-telefonica.png" alt="telefonica logo">
+    </div>
+    <div class="mobile-col-1 tablet-col-1 col-2">
+        <img src="https://assets.ubuntu.com/v1/01158ac3-logo-riot-games.png" alt="riot games logo">
+    </div>
+    <div class="mobile-col-1 tablet-col-1 col-2 u-first-col--mobile">
+        <img src="https://assets.ubuntu.com/v1/ab9074df-logo-tele2.png" alt="tele2 logo">
+    </div>
+    <div class="mobile-col-1 tablet-col-1 col-2">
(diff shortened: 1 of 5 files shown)


kwm14:
@lyubomir-popov can you estimate please?

kwm14:
👍  LGTM @lyubomir-popov nice update to the forever broken/issues/history of `p-inline-images` 👍

examples/patterns/logo-strip.html
```diff
@@ -0,0 +1,44 @@
+---
```
deadlight: This file should be `examples/templates/logo-strip.html` as it's a use-case, not a pattern example.

kwm14:
@lyubomir-popov any update on this PR so @deadlight can review?

lyubomir-popov:
@kwm14 @deadlight ready for another review

examples/templates/logo-strip.html
```diff
@@ -0,0 +1,44 @@
+---
+layout: default
+title: Logo strip
+category: _templates
+---
+
+<div class="row">
+    <div class="mobile-col-1 tablet-col-1 col-2 u-first-col--tablet u-first-col--desktop">
```
anthonydillon: This is extremely fiddly. Why not create a pattern that encapsulates all this single style classes?

scss/_utilities_grid.scss
```diff
@@ -0,0 +1,19 @@
+@mixin grid-first-col {
+  .u-first-col--mobile {
```
anthonydillon: I disagree with this approach in your other PR.

examples/templates/logo-strip.html
lyubomir-popov: @anthonydillon 

> Why not create a pattern that encapsulates all this single style classes?

I've proposed this separately as "layout presets", but it isn't on the roadmap yet.
Common scenarios should indeed be invoked much easier. In this case, a layout preset could be: .layout--4-6-6 where the digits indicate the number of elements per row per breakpoint.

But lacking a set of presets as we do at the moment, we need flexibility. A logo strip on with a different number of logos might require 2-2-4 (2 logos on mobile, 2 on tablet, 4 on desktop).

What is common between those is the need to wrap logos in the same .row, so we don't get a gap. What will differ depending on the content is the exact number of columns required per breakpoint.

lyubomir-popov:
> Also, all patterns should have an example and documentation.
@anthonydillon I've updated the documentation, but the only way to qa is to copy paste the changed section in the browser. This is because the documentation links to the released version on https://vanilla-framework.github.io/vanilla-framework, which at the moment is 1.8.

lyubomir-popov:
@anthonydillon @deadlight Closing this as well. Once we transition to css grid, this will work in a slightly different (simpler) way, with no need to explicitly specify first columns. Since this is going to ship together with the css grid I don't see any reason to continue with the floats implementation.

PR #2240, opened 2019-03-26
Change name of 'aside' component to 'table of contents'

### LINKED ISSUE #1827: “Aside” pattern is misnamed/ambiguous
[The description of the “Aside” pattern](https://docs.vanillaframework.io/en/patterns/aside) is, in full:
> An aside can be used to display a contextual menu within a page, such as a table of contents.

This is confusing in several ways.

First, “contextual menu” is a macOS-specific synonym for “context menu”. Anyone familiar with either term will know it as a menu of items related to an element that has just been right-clicked or long-pressed (references: [Windows](https://msdn.microsoft.com/en-gb/library/windows/desktop/dn742392.aspx), [macOS](https://developer.apple.com/macos/human-interface-guidelines/menus/contextual-menus/), [Gnome](https://developer.gnome.org/hig/stable/menus.html.en), [KDE](https://community.kde.org/KDE_Visual_Design_Group/HIG/ContextMenu)). Anything that appears “within a page” is, therefore, not a contextual menu.

Second, this pattern both evokes and uses [HTML’s `<aside>` element](https://html.spec.whatwg.org/multipage/sections.html#the-aside-element), which “represents a section of a page that consists of content that is tangentially related to the content around the aside element, and which could be considered separate from that content.” But a table of contents is not, in any sense, “tangentially related to the content around the aside element”.

Third, tables of contents are presented as an _example_ of when you might use this pattern. But it seems to provide markup specific to tables of contents:
```html
<li><a class="p-aside__link is-active" href="#link3">Link 3 (Active)</a></li>
```
It’s not clear whether the pattern is useful for anything else, and if so, what benefit you get from using it.

I suggest:
- If tables of contents are a legitimate use for this pattern, it should not be called “Aside”.
- If tables of contents are not a legitimate use for this pattern, the reference shouldn’t give one as an example.
- If anything else is a legitimate use for this pattern, the reference should provide an example.

deadlight:
## Done

Renamed the "aside" component "table of contents" and updated the docs and example

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/table-of-contents/
- See the name change
- `cd docs`
- `./run`
- http://0.0.0.0:8104/en/patterns/table-of-contents
- See that the content and name are updated

## Details

Fixes #1827


docs/en/patterns/table-of-contents.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Table of contents
+table_of_contents: false
+---
+
+## Table of contents
+
+<hr>
+
+An table of contents can be used to display supplementary links to a page.
```
barrymcgee:
```suggestion
A table of contents can be used to display supplementary links to a page.
```

docs/en/patterns/table-of-contents.md
```diff
@@ -0,0 +1,19 @@
...
+
+<a href="https://vanilla-framework.github.io/vanilla-framework/examples/patterns/table-of-contents/"
+    class="js-example">
+View example of the table of contents pattern
+</a>
+
+### Design
+
+For more information [view the table of contents design spec](https://github.com/ubuntudesign/vanilla-design/tree/master/Table of contents), which includes the specification in markdown format and a PNG image.
```
barrymcgee:
```suggestion
For more information [view the table of contents design spec](https://github.com/ubuntudesign/vanilla-design/tree/master/Table%20of%20contents), which includes the specification in markdown format and a PNG image.
```

docs/en/patterns/table-of-contents.md
deadlight: hah, good catch!

deadlight:
> I've always preferred that examples were not specific, but generic. Like the design, example should be "link 1", "link 2" etc
> 
> https://github.com/ubuntudesign/design-vanilla-framework/blob/master/Table%20of%20contents/table-of-contents.png
> 
> Thoughts?

@barrymcgee all examples in the docs have been updated to use real-world examples rather than the generic ones - this is to keep consistent with them.

barrymcgee:
LGTM 👌🏻

kwm14:
LGTM 👍 awaiting Percy to finish and will review there too

PR #2267, opened 2019-04-09
Add override to remove motion when the user has set prefer-reduced-motion

### LINKED ISSUE #2116: Add accessibility override to reduce motion
Some people with vertigo, epilepsy etc can change a setting in their OS to reduce or remove visual motions.

We can now specify this via a media query:

```css
@media (prefers-reduced-motion: reduce) { 
 * { 
   animation: none !important;
   transition: none !important; 
  } 
 }
```

Browser support is [patchy](https://caniuse.com/#feat=prefers-reduced-motion) but iOS supports it and everything else will just ignore it, progressive enhancement FTW.

I think we should add this to the base of Vanilla?

deadlight:
## Done

Added overrides to stop animations if the user has set the prefer-reduced-motion flag.

## QA

- Set the setting in your operating system - [Mac OS instructions](http://osxdaily.com/2018/12/17/how-reduce-motion-mac-disable-animations/)
- Check that you are using a [supported browser](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion#Browser_compatibility)
- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/icons/icons-light/

## Details

Fixes #2116

### DIFF
--- a/docs/en/settings/animation-settings.md
+++ b/docs/en/settings/animation-settings.md
@@ -46,6 +46,12 @@ include animation by including the animation mixin.
 @include animation(property: all, duration: brisk, easing: out);
 ```
 
+### Reduced motion
+
+Vanilla implements the [prefers-reduced-motion CSS media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) so that - in supported browsers and operating systems - users with an accessibility need or preference see reduced animation applied to page elements.
+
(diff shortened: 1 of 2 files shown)


nottrobin:
This is cool

kwm14:
Nice 👍

PR #2271, opened 2019-04-10
Add a highlight to inline code elements

### LINKED ISSUE #2040: Update inline code component
## Pattern to amend

https://docs.vanillaframework.io/en/base/code#inline-code

## Context
Inline code blocks aren't as visible as they should be, proposed at bi-weekly WG meeting and validated here https://github.com/vanilla-framework/vanilla-framework/issues/1995
- Design issue here - https://github.com/ubuntudesign/vanilla-design/issues/329
- Updated design spec here for build reference - https://github.com/ubuntudesign/vanilla-design/blob/master/Code/code.md

## Visual
inline-code-001

deadlight:
## Done

Added a highlight to inline code elements

## QA

- Pull code
- Run `./run serve --watch`
- http://0.0.0.0:8101/vanilla-framework/examples/base/code/
- See that the lower example matches the [spec](https://github.com/ubuntudesign/design-vanilla-framework/blob/master/Code/code.md)

## Details

Fixes #2040

### DIFF
--- a/examples/base/code.html
+++ b/examples/base/code.html
@@ -8,3 +8,5 @@ category: _base
 Settings:
 maas_url=http://192.168.122.1:5240/MAAS</code>
 </pre>
+
+<p>The quick brown <code>fox&nbsp;jumps</code> over the lazy dog</p>
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -5,8 +5,12 @@
   code,
   samp,
   kbd {
+    background-color: $color-mid-x-light;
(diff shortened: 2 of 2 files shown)


kwm14:
Looks good @deadlight but the update seems to have added a grey border on the left side of the `code-numbered` component, not sure if the styles are linked?

**Percy.io pickup**
Screenshot 2019-04-10 at 15 12 16

deadlight:
@kwm14 good catch, I looked but didn't see that. I had to reset the margins too. Should be good now.

kwm14:
Nice, thanks for updating 👍

PR #2273, opened 2019-04-11
Cssgrid

lyubomir-popov:
## Done

Implement grid functionality using css grid.
Rename shelves grid .row to .shelves-row and shelves columns from .col to .column, so they can be compared

Due to the way css grid works, there's no direct analog to push / pull / prefix / suffix.
This has been replaced with breakpoint-scoped grid-column-start classes and order classes.

## QA

- Pull code
- Run `./run serve --watch`
- [Add additional steps]

## Details

## Screenshots

### DIFF
--- a/scss/_patterns_grid.scss
+++ b/scss/_patterns_grid.scss
@@ -1,31 +1,202 @@
 @import 'settings';
 
-// This file imports Shelves Grid framework
-// http://shelvesgrid.org/
-// The settings below can be changed in _settings_grid.scss
-$shelves-max-width: $grid-max-width;
-$shelves-columns: $grid-columns;
-$shelves-margin: $grid-margin-width;
-$shelves-gutter: $grid-gutter-width;
-$shelves-tablet-breakpoint: $breakpoint-medium;
-$shelves-mobile-breakpoint: $breakpoint-small;
-$shelves-column-name: $grid-col-name;
-
-@import 'grid/shelves-grid';
-
 @mixin vf-p-grid-modifications {
   .row {
     width: 100%;
   }
+}
 
-  // Demo helper
-  [grid-demo] [class*='col-'] {
-    background: $color-mid-light;
-    margin-bottom: $spv-inner--small;
+@mixin row-reset-margin {
+  margin-left: 0;
+  margin-right: 0;
+}
+
+@mixin row-reset-padding {
+  padding-left: 0;
+  padding-right: 0;
+}
+
+@mixin row-reset {
+  @include row-reset-margin;
+  @include row-reset-padding;
+}
+
+@mixin vf-p-grid {
+  %span-full-grid-on-mobile {
+    @media (max-width: $threshold-4-6-col) {
+      grid-column: 1 / span $grid-columns-small;
+    }
   }
 
-  [grid-outline] [class*='col-'] {
-    outline: 1px solid $color-x-light;
-    padding: $spv-inner--small $sph-inner--small;
+  %span-full-grid-on-tablet {
+    @media (min-width: $threshold-4-6-col) and (max-width: $threshold-6-12-col) {
+      grid-column: 1 / span $grid-columns-medium;
+    }
+  }
+
+  %span-full-grid-on-desktop {
+    @media (min-width: $threshold-6-12-col) {
+      grid-column: 1 / span $grid-columns;
+    }
+  }
+
+  .grid-demo [class*='#{$grid-column-prefix}'] {
+    background: transparentize($color-negative, 0.9);
+    margin-bottom: $spv-outer--small;
+  }
+
+  %row {
+    display: grid;
+    grid-template-rows: auto;
+    margin-left: auto;
+    margin-right: auto;
+    max-width: $grid-max-width;
+
+    [class*='#{$grid-column-prefix}'] {
+      grid-column-start: auto;
+    }
+
+    & & {
+      @include row-reset;
+    }
+
+    // set static gutter width per breakpoint
+    @media (max-width: $threshold-4-6-col) {
+      grid-gap: 0 map-get($grid-gutter-widths, small);
+      grid-template-columns: repeat($grid-columns-small, minmax(0, 1fr));
+      padding-left: map-get($grid-margin-widths, small);
+      padding-right: map-get($grid-margin-widths, small);
+
+      * {
+        grid-column-end: span $grid-columns-small;
+      }
+    }
+
+    @media (min-width: $threshold-4-6-col) and (max-width: $threshold-6-12-col) {
+      grid-gap: 0 map-get($grid-gutter-widths, medium);
+      grid-template-columns: repeat($grid-columns-medium, minmax(0, 1fr));
+      padding-left: map-get($grid-margin-widths, medium);
+      padding-right: map-get($grid-margin-widths, medium);
+
+      * {
+        grid-column-end: span $grid-columns-medium;
+      }
+    }
+
+    @media (min-width: $threshold-6-12-col) {
+      grid-gap: 0 map-get($grid-gutter-widths, large);
+      grid-template-columns: repeat($grid-columns, minmax(0, 1fr));
+      padding-left: map-get($grid-margin-widths, large);
+      padding-right: map-get($grid-margin-widths, large);
+
+      * {
+        grid-column-end: span $grid-columns;
+      }
+    }
+  }
+
+  .row {
+    @extend %row;
+  }
+
+  // mobile grid
+  @media (max-width: $threshold-4-6-col) {
+    @for $i from 1 through $grid-columns-small {
+      .#{$grid-small-col-prefix}#{$i} {
+        grid-column-end: span #{$i};
+
+        //nesting
+        @if $i > 1 {
+          & > .row {
+            grid-template-columns: repeat($i, minmax(0, 1fr));
+          }
+        }
+      }
+    }
+  }
+
+  // tablet grid
+  @media (min-width: $threshold-4-6-col) and (max-width: $threshold-6-12-col) {
+    @for $i from 1 through $grid-columns-medium {
+      .#{$grid-col-medium-prefix}#{$i} {
+        grid-column-end: span #{$i};
+
+        //nesting
+        @if $i > 1 {
+          & > .row {
+            grid-template-columns: repeat($i, minmax(0, 1fr));
+          }
+        }
+      }
+    }
+  }
+
+  // desktop grid
+  @media (min-width: $threshold-6-12-col) {
+    @for $i from $grid-columns through 1 {
+      // set col-* to span respective number of columns on desktop
+      .#{$grid-desktop-col-prefix}#{$i} {
+        grid-column-end: span #{$i};
+
+        //nesting
+        @if $i > 1 {
+          & > .row {
+            grid-template-columns: repeat($i, minmax(0, 1fr));
+          }
+        }
+      }
+    }
+  }
+
+  // span full grid on other breakpoints; (to match block level element behaviour)
+  @for $i from 1 through $grid-columns-small {
+    .#{$grid-small-col-prefix}#{$i} {
+      @extend %span-full-grid-on-tablet;
+      @extend %span-full-grid-on-desktop;
+    }
+  }
+
+  @for $i from 1 through $grid-columns-medium {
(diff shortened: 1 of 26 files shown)


lyubomir-popov:
@anthonydillon @deadlight I think this is ready for an initial review. Not looking to land this yet, but would be nice to hear what you think of the implementation. 

One thing to note, percy refuses to render css grid, so all screenshots look broken. Not sure why, as everything works as expected in chrome and firefox locally.

deadlight:
I really need to see a Percy diff working to review this properly.

examples/patterns/card/overlay.html
```diff
@@ -6,7 +6,7 @@
 
 <section class="p-strip--image is-light" style="background-image:url('https://assets.ubuntu.com/v1/0a98afcd-screenshot_desktop.jpg')">
   <div class="row">
-    <div class="col-12 prefix-6">
+    <div class="col-6 col-start-large-6">
```
deadlight: I think that the "col-start" class could be better named. Perhaps make it a little more like BEM?

I can't think of an ideal solution now but something like `col-start-6--large` might be more in-line with the rest of Vanillia.

examples/patterns/grid/default.html
```diff
@@ -3,98 +3,109 @@
...
-        <div class="col-1">
-            <span>.col-1</span>
-        </div>
-        <div class="col-11">
-            <span>.col-11</span>
-        </div>
-    </div>
-</div>
\ No newline at end of file
+
```
deadlight: Drive by: can you make this file 2-space indented rather than 4

examples/patterns/grid/empty-columns.html
```diff
@@ -3,7 +3,38 @@
 title: Grid / Empty columns
 category: _patterns
 ---
+<h2>Old grid</h2>
```
anthonydillon: Why do we need the old grid?

examples/patterns/grid/default.html
```diff
@@ -3,98 +3,109 @@
...
+<div class="p-strip">
+    <grid css-grid-demo>
```
anthonydillon: Grid isn't a supported element is it?

examples/patterns/grid/empty-columns.html
```diff
@@ -3,7 +3,38 @@
...
 <div grid-demo>
```
anthonydillon: Should `grid-demo` be a class?

examples/patterns/grid/empty-columns.html
```diff
@@ -3,7 +3,38 @@
...
+        .column-7
+      </div>
+      <div class="column-4 suffix-1">
+        .column-4.suffix-1
+      </div>
+  </div>
+</div>
+
+<p-strip class="p-strip">
+  <h2>Css grid</h2>
```
anthonydillon: Should be CSS grid

examples/patterns/grid/empty-columns.html
```diff
@@ -3,7 +3,38 @@
...
+      <div class="column-7">
+        .column-7
+      </div>
+      <div class="column-4 suffix-1">
+        .column-4.suffix-1
+      </div>
+  </div>
+</div>
+
+<p-strip class="p-strip">
```
anthonydillon: Please separate the examples so documentation can separate them.

examples/patterns/grid/empty-columns.html
```diff
@@ -3,7 +3,38 @@
...
+  <h2>Css grid</h2>
+</p-strip>
+<div css-grid-demo>
```
anthonydillon: Should be a class?

examples/patterns/grid/empty-columns.html
```diff
@@ -16,16 +47,15 @@
       <div class="col-7">
         .col-7
       </div>
-      <div class="col-4 prefix-1">
-        .col-4.prefix-1
+      <div class="col-4 col-start-large-8">
```
anthonydillon: I don't understand how `col-start-8` would result in the col-4 being pushed over one column. Or is this just a new example?

examples/patterns/grid/empty-columns.html
```diff
@@ -16,16 +47,15 @@
...
+        .col-4.col-start-large-8
       </div>
   </div>
   <div class="row">
       <div class="col-7">
         .col-7
       </div>
-      <div class="col-4 suffix-1">
-        .col-4.suffix-1
+      <div class="col-4">
```
anthonydillon: You have removed the empty column and the contents of this grid element

examples/patterns/grid/nested.html
```diff
@@ -3,29 +3,265 @@
 title: Grid / Nested columns
 category: _patterns
 ---
-<div grid-demo grid-outline>
+
+<div class="p-strip">
```
anthonydillon: We shouldn't add p-strip to examples as they are for copy and pasting for the component. There is no guarantee that the project has included strips.

examples/patterns/grid/nested.html
```diff
@@ -3,29 +3,265 @@
...
   <div class="row">
-    <div class="col-6">
-      <span>.col-6</span>
+    <div class="col-small-4 col-medium-6 col-12">
+      <h2>Nesting at every breakpoint</h2>
+    </div>
+  </div>
+
+  <div class="row" css-grid-demo>
+    <div class="col-small-4 col-medium-6 css-desktop-col-12">
```
anthonydillon: What is the difference between `col-12` and `css-desktop-col-12`

examples/patterns/grid/nested.html
```diff
@@ -3,29 +3,265 @@
...
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
+
+<div class="p-strip">
+  <div class="row">
+    <div class="col-small-4 col-medium-6 col-12">
+      <h2>Nesting at breakpoints (resize to reveal)</h2>
```
anthonydillon: These should be separate examples

examples/patterns/grid/nested.html
```diff
@@ -3,29 +3,265 @@
...
+          <div class="col-small-1">
+            col-small-1
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+
+
+  <div class="u-hide--small u-hide--large">
```
anthonydillon: Shouldn't this just be a single example demoing the change in grid based on screen size

examples/patterns/grid/offsets.html
```diff
@@ -0,0 +1,29 @@
+---
+layout: default
+title: Grid / Offset columns
+category: _patterns
+---
+
+<div class="p-strip">
+  <div class="row">
+    <div class="col-small-4 col-medium-6 col-12">
+      <h2>Column reordering</h2>
```
anthonydillon: Please don't title the examples. They will be in the documentation.

examples/patterns/grid/offsets.html
```diff
@@ -0,0 +1,29 @@
...
+    </div>
+  </div>
+
+  <div class="row" css-grid-demo>
+      <div class="col-small-3 col-medium-4 col-9
```
anthonydillon: I hope we never have to do this and this is a pattern of its a common thing.

scss/_patterns_grid.scss
```diff
@@ -19,13 +19,193 @@ $shelves-column-name: $grid-col-name;
...
+    }
+  }
+
+  %span-full-grid-on-desktop {
+    @media (min-width: $breakpoint-large) {
+      grid-column: 1 / span $grid-columns;
+    }
+  }
+
+  [css-grid-demo] [class*='#{$grid-column-prefix}'] {
```
anthonydillon: The example markup are to be used as example usage. Therefore we need to find another way to style them or then bespoke attrs which are in the packaged version of Vanilla.

scss/_patterns_grid.scss
```diff
@@ -19,13 +19,193 @@ $shelves-column-name: $grid-col-name;
...
+    background: transparentize($color-negative, 0.9);
+    margin-bottom: $spv-outer--small;
+  }
+
+  .row {
+    display: grid;
```
anthonydillon: We often use equal-height for example of rows which sets the row to `display: flex`. Does this pattern need reworking?

examples/patterns/grid/empty-columns.html
lyubomir-popov: Good spot, I actually hadn't left an empty column there. Fixed. it should start at 9, not 8.

As I understand it, the css grid layout doesn't provide a direct equivalent to the prefixes/sufixes used by the shelves grid. So instead, for each div, you specify start/end position, or one of them together with the number of columns you want it to span. So to achieve the same effect, the div needs to replace ```col-4 prefix-1 ``` with ```col-3 col-start-large-8```.

lyubomir-popov: Actually, to achieve the effect of a background wrappping the empty column, we need .col4 and inside, .row>.col-3.col-start-2:
image
I've updated the example.

examples/patterns/grid/empty-columns.html
lyubomir-popov: Updated as above:

image

examples/patterns/grid/nested.html
lyubomir-popov: Fixed

examples/patterns/grid/nested.html
lyubomir-popov: just a left over from when I had to keep shelves and css grid classes different. fixed

deadlight:
@lyubomir-popov what's the status of this PR? Are you ready for a full review of the code to be merged.

examples/patterns/grid/default.html
lyubomir-popov: Done

scss/_patterns_grid.scss
lyubomir-popov: In my tests it doesn't change that behaviour, because align-items behaves the same way in flex and css grid. Can you point me to an example where it doesn't work?

lyubomir-popov: The equal heights example seems works fine, even when the clas is on the row element: image

lyubomir-popov:
@deadlight I think I addressed all comments, please re-review.

deadlight:
Visual issues from Percy:

- [/vanilla-framework/examples/patterns/lists/lists-stepped-detailed/index.html](https://percy.io/vanilla-framework/vanilla-framework/builds/1781935/view/116013065/1280?mode=diff&browser=firefox&snapshot=116013065)
- [/vanilla-framework/examples/patterns/strips/accent/index.html](https://percy.io/vanilla-framework/vanilla-framework/builds/1781935/view/116013059/1280?mode=diff&browser=firefox&snapshot=116013059)
- []()
- []()
- []()
- []()

lyubomir-popov:
@deadlight I've addressed all but:
• /vanilla-framework/examples/patterns/tabs/index.html (can't replicate)
•  /vanilla-framework/examples/templates/tick-element-comparison/index.html this one seems a ff bug on resize as demonstrated.

deadlight:
@lyubomir-popov The images in the typographic spacing example are [wrapping](https://percy.io/vanilla-framework/vanilla-framework/builds/1801395/view/117244062/1280?mode=diff&browser=firefox&snapshot=117244062)

lyubomir-popov:
@deadlight fixed the typographic-examples page by removing an unnecessary utility.

deadlight:
@lyubomir-popov you've changed the stepped-list bullet colour so that it no longer matches the [specification](https://github.com/canonical-web-and-design/design-vanilla-framework/blob/master/Lists/lists.md)

deadlight:
@lyubomir-popov we're getting close to landing this but you need to update the docs to include the new utility classes and to remove the old ones.

lyubomir-popov:
Ready for re-review.

lyubomir-popov:
@kwm14 @deadlight I've squashed and merged with the tables fixed. @deadlight can you replicate tabstrip? I can't.

lyubomir-popov:
@kwm14 it stopped showing the tab diff, can you please look at the last remaining 2 items, it's just a small shift due to grid width.

kwm14:
@lyubomir-popov done and looking good 👍

deadlight:
@lyubomir-popov I checked the tabstrip and it looks fine in browser.

PR #2283, opened 2019-04-17
Upped the default max-width for typography

### LINKED ISSUE #1794: Remove or revise 'max-width' from type elements
## Process

Proposal to remove or revise 'max-width' from type elements as it's been highlighted a few times as an issue, it would be great to discuss further with the rest of the team.
- Was there a issue with type before max-width?
- Do we need a rule when to use?
- Can be be applied when needed?

### Notifications
screen shot 2018-05-02 at 15 07 00

### Docs content
docs-screenshot

deadlight:
## Done

Upped 38em to 40em

## QA

- Pull code
- Run `./run serve --watch`
- Look through components but it's probably better to use Percy to see all of the impact

## Details

Fixes #1794

### DIFF
--- a/scss/_settings_spacing.scss
+++ b/scss/_settings_spacing.scss
@@ -141,7 +141,7 @@ $sp-after: (
 
 $max-widths: (
   long: 60em,
-  default: 38em,
+  default: 40em,
   heading--long: 30em,
   heading--short: 20em
 ) !default;


deadlight:
Removed the design review component cosidering that there's no visual change in any demos and the specific values have been agreed in advance.

PR #2286, opened 2019-04-23
Update the tabs example to include nesting within a row and update docs

### LINKED ISSUE #1894: Tabstrip has max-width, but it doesn't take grid margins into account
## Process
• Open [Maas layout example ](https://vanilla-framework.github.io/vanilla-framework/examples/templates/maas-layout/)
• Inspect tabstrip
• Remove the row between p-tabs and p-tabs__list;
• Observe how the tabstrip expands into the 24px grid margins

## Expected result
I believe it should either:
• have max-width and the regular row padding-left and right, so it aligns with rows
• or have neither max-width nor padding-left/right, if the idea was to let its parent container deal with the grid.

Both of the above would work with the current makrup, whereas at the moment, if you want it to align with other rows, you need to insert a div classed "row" in between  .p-tabs and .p-tabs__list.

deadlight:
## Done

Updated the tabs example to put it in a row to show how to have it sit properly with other content.

I also updated the docs to mention it.

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/patterns/tabs/
- See that it now has gutters

## Details

Fixes #1894

### DIFF
--- a/docs/en/patterns/tabs.md
+++ b/docs/en/patterns/tabs.md
@@ -11,7 +11,7 @@ Tabs organise and allow navigation between groups of content that are related an
 
 To select the active tab add the attribute `aria-selected="true"` and that list item will have the correct styling.
 
-To horizontally align the tablist with other content, the whole tabset can be contained within a `.row` element to provide correct gutters.
+To horizontally align the tab list with other content, the whole tab set can be contained within a `.row` element to provide correct gutters.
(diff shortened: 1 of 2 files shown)


kwm14:
Percy approved and visual update LGTM 👍

PR #2309, opened 2019-05-15
Introduces new variable to change the font-weight for bold fonts

b-m-f:
## Done

[- adds new variable `$font-weight-bold`
- uses variable where appropriate in code
## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/
- Confirm that the heading the navbar still have `font-weight: 400`

### DIFF
--- a/scss/_settings_font.scss
+++ b/scss/_settings_font.scss
@@ -11,4 +11,4 @@ $font-sizes: (
   base: 1rem,
   big: $font-size-largescreen
 ) !default;
-$font-weight-bold: 400;
+$font-weight-bold: 400 !default;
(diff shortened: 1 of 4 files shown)


scss/_settings_font.scss
```diff
@@ -11,3 +11,4 @@ $font-sizes: (
   base: 1rem,
   big: $font-size-largescreen
 ) !default;
+$font-weight-bold: 400;
```
lyubomir-popov: Shouldn't this have  !default at the end?

deadlight:
```suggestion
$font-weight-bold: 400 !default;
```

PR #2314, opened 2019-05-17
Fixed width class

lyubomir-popov:
## Done

Often content doesn't need to align to grid columns, but needs to be restrained by the fixed-width used by the grid. To provide that functionality separately, I've extracted fixed-width behaviour from .row into a placeholder, then .row and a new .u-fixed-width class extend that. 

## QA

- Pull code
- Run `./run serve --watch`
- Change row to u-fixed-width in any example. Observe how content within stays aligned to the fixed-width of the site.

### DIFF
--- a/scss/_patterns_grid.scss
+++ b/scss/_patterns_grid.scss
@@ -46,6 +46,8 @@
   }
 
   %row {
+    @extend %fixed-width-container;
+
     display: grid;
     grid-template-rows: auto;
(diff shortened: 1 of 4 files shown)


lyubomir-popov:
@deadlight just pushed a fix to the stepped detailed list (an unrelated issue that I spotted due to the percy diff). This is now readty for review.

PR #2318, opened 2019-05-17
Use run script v3.4.1

nottrobin:
QA
--

Check commands work:

``` bash
./run clean
./run build
./run yarn build-docs
./run
```

Browse the docs at http://127.0.0.1:8101.

``` bash
docker build --build-arg REVISION_ID=nope --tag docs docs
docker run -ti -p 8301:80 docs
```

Browse the docs at http://127.0.0.1:8301.


jpmartinspt:
Not sure this is related to your changes, but when using the Dockerfile to build the image, image assets don't seem to be served properly, at least in the `Heading Icon` section.

Apart from this everything else seems to work fine.

nottrobin:
@jpmartinspt this is a problem with the content of whole heading icon page, which is unrelated to this PR. It fails differently in local development:

image

than it does in production:

image

but both are broken for the same reason - the examples files aren't where they should be.

I've created issue #2349.

PR #2323, opened 2019-05-21
Add new variable - with default values - for navigation border and text colours

### LINKED ISSUE #2046: Provide variables to override navigation pattern colours
## Pattern to amend

p-navigation

## Context

Some colours are automatically generated within the p-navigation pattern from the colour variables passed in. There should be an option to override these generated colours with specific variables.

e.g. The background colour of hovered nav items is `background-color: darken($color-navigation-background, 3%);` with no way of avoiding this behaviour with a variable

deadlight:
## Done

There should be no visual changes to the navigation, the code is just restructured

## QA

- Pull code
- Run `./run serve --watch`
- http://0.0.0.0:8101/examples/patterns/navigation/default/
- See that the navigation looks good
- Change $color-navigation-background in `_settings_colors.scss` to #000
- http://0.0.0.0:8101/examples/patterns/navigation/default/
- See that the colours still work together

## Details

Fixes #2046

### DIFF
--- a/scss/_settings_colors.scss
+++ b/scss/_settings_colors.scss
@@ -9,6 +9,7 @@ $color-light: #f7f7f7 !default;
 $color-mid-x-light: #e5e5e5 !default;
 $color-mid-light: #cdcdcd !default;
 $color-mid: #999 !default;
+$color-navigation-selected: #757575 !default;
 $color-mid-dark: #666 !default;
 $color-dark: #111 !default;
 $color-x-dark: #000 !default;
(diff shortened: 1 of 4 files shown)


scss/_settings_colors.scss
```diff
@@ -31,6 +31,9 @@ $grid-border-style: 1px dotted $color-mid-light !default;
 
 // for _patterns_navigation.scss
 $color-navigation-background: $color-x-light !default;
+$color-navigation-border: if(lightness($color-navigation-background) > 50, $color-mid-light, transparent) !default;
+$color-navigation-text: if(lightness($color-navigation-background) > 70, $color-dark, $color-light) !default;
+
 $color-navigation-selected: #757575 !default;
```
lyubomir-popov: • I'm a bit concerned that doing this automatically based on lightness doesn't guarantee a good result. If the point is to provide a reasonable default for uses not covered by the default component, maybe no border is a better default? E.g. the ubuntu nav (and maas.io, jaas.ai, snapcraft.io all have to override this because they don't need a border). 
• Is there a reason why these are moved (including the conditionals) to a settings file? Shouldn't settings be for more universal things? Like optimal border on white background, which then gets used by tabs, hrs, menus, anything that needs to match. Otherwise all of these could clutter the settings file, when in fact they all refer to the same grey value.

deadlight:
I'm going to move the variables back into the component but keep them overrideable.

Re: the other comments on the suitability of the colour choices: that's a larger design conversation outside the scope of this code restructure.

lyubomir-popov:
I've moved variables that reference colours defined in ```scss/_settings_colors.scss``` into respective files. Only new shades or colours should live in the settings file.

deadlight:
I've fixed a missed reference to the removed variable.

lyubomir-popov:
@deadlight lgtm

PR #2348, opened 2019-05-28
Stacked form: replace hardcoded layout with grid classes

### LINKED ISSUE #1921: Forms stacked have hardcoded widths, which get in the way when used in a 6-column container
## Pattern to amend

.p-form__stacked
.p-form__label & .p-form__control have hardcoded widths (25% / 75%)
My proposal is to come up with a way to allow col- classes to override the hardcoded values.
This is a tricky thing to change, because simply taking the hardcoded values out, or reducing the specificity (currently 2 levels deep, e.g. .p-form__stacked .p-form__label, which overrides the 1 level specificity of col-* classes) will break pages that use it as it is. So not sure what the best course of action is.

## Visual

image

## Context

In what context does your amendment solve a problem?

There are a number of issues with this:
1) It doesn't use the grid, as it deosn't take gutters into account

2) has required overriding before (e.g. in maas [here](https://github.com/maas/maas/blob/7f7bcb050cfa796dd70d41a6cf4a5d1a5be4f6fb/src/maasserver/static/scss/_patterns_forms.scss#L7) and [here ](https://github.com/maas/maas/blob/7f7bcb050cfa796dd70d41a6cf4a5d1a5be4f6fb/src/maasserver/static/partials/dashboard.html#L116)). 

3) Any classes you add in an attempt to control the width of the label and control don't work, because the hardcoded values are further down in the cascade. This has resulted in a bit of a mess in maas, where we now have both col- classes that do nothing, and hardcoded override for vanilla's hardcoded values.
Provide a screenshot or link to a prototype of your pattern amendment.

image

lyubomir-popov:
## Done

Remove hardcoded css widths; update example to use the grid

## QA

- Pull code
- Run `./run serve --watch`
- QA /examples/patterns/forms/form-stacked/

## Details

Fixes #1921

## Screenshots

### DIFF
--- a/docs/examples/patterns/forms/form-stacked.html
+++ b/docs/examples/patterns/forms/form-stacked.html
@@ -5,41 +5,76 @@ category: _patterns
 ---
 
 <form class="p-form p-form--stacked">
-  <div class="p-form__group">
-    <label for="full-name-stacked" class="p-form__label">Full name</label>
-    <div class="p-form__control">
-      <input type="text" id="full-name-stacked" required>
+  <div class="p-form__group row">
+
+    <div class="col-4">
+      <label for="full-name-stacked" class="p-form__label">Full name</label>
     </div>
-  </div>
-  <div class="p-form__group">
-    <label for="username-stacked" class="p-form__label">Username</label>
-    <div class="p-form__control">
-        <input type="text" id="username-stacked">
-        <p class="p-form-help-text">
-            30 characters or fewer.
-        </p>
-    </div>
-  </div>
-  <div class="p-form__group p-form-validation is-error">
-      <label for="username-stacked-error" class="p-form__label">Email address</label>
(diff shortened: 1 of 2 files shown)


kwm14:
Design LGTM 👍

deadlight:
@lyubomir-popov see the Percy diff. Inline forms are now broken.

kwm14:
- /examples/patterns/forms/form-inline/index.html

**Before**
Screenshot 2019-05-29 at 09 14 55

**After**
Screenshot 2019-05-29 at 09 15 02

lyubomir-popov:
@deadlight @kwm14 fixed the inline form and removed align: baseline because this is already handled by the vertical rhythm.

kwm14:
Thanks for fixing @lyubomir-popov

PR #2350, opened 2019-05-29
DEMO: Make tooltip messages copyable

Caleb-Ellis:
Demo of copyable tooltip messages


deadlight:
I had a quick look, it's good stuff. Just two things from me:
1. the tooltips still disappear before the user can mouse over it when positioned left, right or above the button.
2. clicking the tooltip to select the text triggers a click even on the parent element

kwm14:
## WG meeting:
- @Caleb-Ellis to create an issue based on this PR proposal
- Proposal agreed 👍 
- Make sure it works for all directions of tooltip
- Keep the markup the same
- Investigate ways of getting around focus issues

PR #2373, opened 2019-06-06
line-height fix

lyubomir-popov:
## Done
Some components (e.g. navigation) inherit the increased line-height from html, which shouldn't happen.
It is only needed on html because increasing the font-size doesn't affect the computed line-height (it is still calculated against the default rem size, 16px).

This pr explicitly sets the line-height of anything not html to 1.5rem.

### DIFF
--- a/scss/_base_typography.scss
+++ b/scss/_base_typography.scss
@@ -40,6 +40,14 @@
     }
   }
 
+  @if ($increase-font-size-on-larger-screens) {
+    @media screen and (min-width: $breakpoint-x-large) {
+      *:not(html) {
+        line-height: map-get($line-heights, default-text);
+      }
+    }
+  }
+
   // headings
   h1 {
     @extend %vf-heading-1;
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_script:
 script:
   - yarn run test
   - yarn run build-docs
-  - percy snapshot --widths "375,1280" docs/_site/
+  - percy snapshot --widths "375,1280,1600" docs/_site/
 notifications:
   irc:
     channels:


scss/_base_typography.scss
```diff
@@ -40,6 +40,14 @@
     }
   }
 
+  @if ($increase-font-size-on-larger-screens) {
+    @media screen and (min-width: $breakpoint-x-large) {
+      *:not(html) {
+        line-height: map-get($line-heights, default-text);
```
anthonydillon: Will this not set all line-heights to the same value on desktop screens? Don't we want headings for example to have more line height?

deadlight:
No visual changes

scss/_base_typography.scss
lyubomir-popov: @anthonydillon Things that require a different line-height have that already set with a higher specificity rule. This is meant to handle elements that inherit from html. Percy hasn't generated a diff, so I think it's safe to say there there haven't been any side effects.

This just gets around the fact that setting 
```
html {
    font-size: 1.125rem; 
    line-height: 1.5rem
}
``` 
results in two things:
• for the html element, line-height iscomputed as if the font-size hasn't changed, so 24px (16 * 1.5)
• anything that inherits line-height from html however uses the font-size increase, so its line-height becomes 1.125rem * 1.5.

This begs the question why are we setting line-height as 1.125rem instead of 18px, can't remember off the top of my head, but there it was causing issues. I can look into it and provide a clearer answer if needed.

So all this pr does is set line-height for things that inherit from html to the expected value, 1.5rem.

There might be a better way to do this, but I haven't found one so far.

lyubomir-popov: @anthonydillon this is so we don't have to put explicit rules on things like .p-navigation__link etc. they have no line-height specifically set and become taller than they should be on screens > 1680px.

deadlight:
@lyubomir-popov I'm rebuilding the Percy tests with the large screen sizes

lyubomir-popov:
I wasn't running the watch task, sorry about this. The specificity of the :not rule has indeed overridenthings it shouldn't, I'm going to close the pr and find another solution.

PR #2376, opened 2019-06-10
Fix nav styling at smaller viewports so that it's full-width

deadlight:
## Done

Made the navigation elements full width at smaller screens

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/examples/patterns/navigation/default/
- Resize the browser to the smallest viewport
- Interact with the nav and see that it's full-width


kwm14:
Nice. LGTM 👍

PR #2381, opened 2019-06-11
Make grid col names overrideable

lyubomir-popov:
## Done

Add !default at the end of grid col variables, so they can be overriden. Useful for something like the migration temp scss file, the goal of which is to produce a pr that is mergeable with the least amount of changes.

### DIFF
--- a/scss/_patterns_grid.scss
+++ b/scss/_patterns_grid.scss
@@ -137,7 +137,7 @@
   @media (min-width: $threshold-6-12-col) {
     @for $i from $grid-columns through 1 {
       // set col-* to span respective number of columns on desktop
-      .#{$grid-big-col-prefix}#{$i} {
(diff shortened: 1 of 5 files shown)


scss/_patterns_grid.scss
```diff
@@ -137,7 +137,7 @@
   @media (min-width: $threshold-6-12-col) {
     @for $i from $grid-columns through 1 {
       // set col-* to span respective number of columns on desktop
-      .#{$grid-desktop-col-prefix}#{$i} {
+      .#{$grid-big-col-prefix}#{$i} {
```
deadlight: We usually use "large" for screen viewports (yes, "large" and "big" are two words for _almost_ the same thing and English is stupid)

lyubomir-popov: large sounds a lot better, thanks. Updated here and in a few other places.

PR #2384, opened 2019-06-13
configurable table layout

### LINKED ISSUE #2374: Tables probably shouldn't default to fixed width
I'm not sure if this is baked in to the style or part of the implementation where we use it, but using `table-layout: fixed;` doesn't produce great results for all but the simplest of tables (and even then has limitations).

Using a fixed width for the columns:

 - doesn't aid legibility
 - often causes table data to run over more lines than it should
 - is inefficient at using space (especially in reactive)
 - in extreme cases causes avoidable rendering errors

An example:

tables-001

This table pretty much illustrates all of the problems mentioned above.  However, with standard, flexible column widths it becomes more readable, more efficient and less broken

tables-002

Yes, we can still try harder to make things in tables less wordy, but still a huge improvement

lyubomir-popov:
Update vanilla dep to v2.0.1

## Done
1) Makes table-layout configurable via a variable ($table-layout)
2) Provides utility to switch to the alternative, i.e. :
- if $table-layout is set to 'auto', adds a utility class called .u-table-layout--fixed
- if $table-layout is set to 'fixed' (which is the default), adds a utility class called .u-table-layout--auto

## QA

- Pull code
- Run `./run serve --watch`
- Open /examples/base/table/
- add class 'u-table-layout--auto', verify it applies an override
- change $table-layout to 'auto' in a repo that imports vanilla;
- copy the table example to your repo, and add class 'u-table-layout--fixed' to the table
- verify the utility overrides as expected

## Details

Fixes #2374


lyubomir-popov:
@deadlight I've added documentation.

docs/_layouts/default.html
```diff
@@ -89,6 +89,7 @@
                   <li class="p-sidebar-nav__item"><a class="p-link--soft {% if page.url == '/patterns/grid/' %}is-active{% endif %}" href="/patterns/grid/">Grid</a></li>
                   <li class="p-sidebar-nav__item"><a class="p-link--soft {% if page.url == '/patterns/strip/' %}is-active{% endif %}" href="/patterns/strip/">Strip</a></li>
                   <li class="p-sidebar-nav__item"><a class="p-link--soft {% if page.url == '/settings/layout-settings/' %}is-active{% endif %}" href="/settings/layout-settings/">Structure</a></li>
+                  <li class="p-sidebar-nav__item"><a class="p-link--soft {% if page.url == '/settings/spacing-settings/' %}is-active{% endif %}" href="/settings/table-layout/">Table layout</a></li>
```
anthonydillon: The if statement here should be: `settings/table-layout/`

docs/settings/table-layout.md
```diff
@@ -0,0 +1,14 @@
+---
+layout: default
+---
+
+## Table layout
+
+By default, tables in vanilla use table-layout: fixed. It is faster, and ensures more predictable and controllable table behaviour, but requires setting widths explicitly on the first row of the table.
```
anthonydillon: - Please capitalise Vanilla.
- Please add back ticks around "table-layout: fixed"

scss/_settings_tables.scss
```diff
@@ -0,0 +1 @@
+$table-layout: fixed !default;
```
anthonydillon: This requires the user to learn the values where as if we are just switching between two it is much better to give them a switch. So this could be called `$table-layout-fixed: true !default;`. This makes it much easier to document.

scss/_utilities_layout.scss
```diff
@@ -4,4 +4,14 @@
   .u-fixed-width {
     @extend %fixed-width-container;
   }
+
+  @if ($table-layout != 'fixed') {
```
anthonydillon: This conditional is very difficult to document and can lead to unexpected regressions. If someone isn't aware of all the uses of these utilities and changes the variable in settings. This utility will disappear and things will regress.

scss/_utilities_layout.scss
```diff
@@ -4,4 +4,14 @@
...
+    .u-table-layout--fixed {
```
anthonydillon: I really dislike single style utilities. If the user has to apply this to each table then they might as well use `style` attribute.

scss/_utilities_layout.scss
```diff
@@ -4,4 +4,14 @@
...
+      table-layout: fixed;
```
anthonydillon: Utiliities should always have `!important` this allows then to override any pattern styling

docs/settings/table-layout.md
anthonydillon: I would prefer if the default table behavior was auto and if the user wants more control and is willing to add widths to the table columns then they switch `fixed`.

lyubomir-popov:
> Will there be cases that parts of the site/app will want fixed tables and in others they will want auto?

I think this is likely in any docs repo. Table column width is part of the design process, and any table that has been designed would should include widths.

docs/settings/table-layout.md
nottrobin: I do feel the `$table-layout-fixed` variable is unnecessary and overengineered. Defaulting to `auto` with a utility class to override should be quite sufficient and is more declarative and less surprising.

lyubomir-popov: The default has been 'fixed' for a while. Changing this may lead to regressions, especially in maas.

nottrobin:
> @lyubomir-popov I think this is likely in any docs repo. Table column width is part of the design process, and any table that has been designed would should include widths.

Surely explicit widths will always overwrite either `fixed` or `auto`, so this has nothing to do with explicitly set widths. This is about what happens when you leave it up to the browser to choose the width.

scss/_utilities_layout.scss
lyubomir-popov: My idea here was to keep the generated css as minimal as possible. What this rule does is, if the base rule is set to 
```table-layout: auto;```

then you don't need a utility that overrides this with the same value. You might however need a utility that overrides it with 'fixed', and viceversa.

No if you change $table-layout from `auto` to `fixed`, any `.u-table-layout--fixed` utilities in the markup are no longer necessary because this is the default. So we don't generate rules for that class. Any references to this utility in the markup would simply be ignored.

So it won't lead to a regression, just cleaner css and less cascade overrides.

nottrobin:
@deadlight Just curious - did you merge this because @lyubomir-popov wasn't available? Or was it just a mistake?

https://canonical-web-and-design.github.io/practices/workflow/github.html#merging-process

deadlight:
@nottrobin we merge passing PRs, I thought that that practice document had been updated to reflect this.

lyubomir-popov:
@nottrobin with table-layout:auto, column widths are ignored. For example, in this [pen](https://codepen.io/lyubomir-popov/pen/WqrbrE) the second cell is set to 10rem, and if longer should ellipsize (a common behaviour in maas).

Change table-layout to auto and you'll see that not only the width is ignored, but the table breaks out of the container (e.g. a site's fixed width).

AFAIK table-layout is considered the best practice, this is why we changed that a long time ago in vanilla. Here's one article on the subject going into more detail on why it makes tables a lot more predictable to work with: https://css-tricks.com/fixing-tables-long-strings/

docs/_layouts/default.html
lyubomir-popov: fixed

docs/settings/table-layout.md
lyubomir-popov: @anthonydillon @nottrobin table-layout has been set to fix since 1.7.1 and ensures tables behave predictably. This can potentially cause lots of regressions. Are you sure you want to change this?

This pr was simply meant to give flexibility to use auto for automatically generated tables from discourse on docs, in response to this [issue](https://github.com/canonical-web-and-design/vanilla-framework/blame/22d753c8a0cbce3481dc00bfd6c82ed84785a553/scss/_base_tables.scss).

I just realised that the doc wording makes it sound a lot more ominous than it actually is. When I wrote ```requires setting widths explicitly``` I meant it requires it if you're not happy with the default browser behaviour in this case, which is:

> Table and column widths are set by the widths of table and col elements or by the width of the first row of cells. Cells in subsequent rows do not affect column widths. [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout)

I've taken that part out.

scss/_settings_tables.scss
lyubomir-popov: fixed in #2387

scss/_utilities_layout.scss
lyubomir-popov: fixed in #2387

scss/_utilities_layout.scss
lyubomir-popov: there is a subtle difference - an inline style always causes a redefinition of the rule it overrides. A conditionally generated utility overrides only if the value it sets is different from the default.

PR #2422, opened 2019-07-25
fix table spacing

lyubomir-popov:
## Done
Changes padding-bottom to transparent border-bottom, as padding doesn't work with border-collapse and margin-bottom triggers a strange bug in older ff (the one used by percy).

## QA

- Pull code
- Run `./run serve --watch`
- Open /examples/base/table/
- Add h2 under table
- Observe h2 being correctly pushed down by margin-bottom on the table

## Details

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/2420

## Screenshots

image

### DIFF
--- a/scss/_base_tables.scss
+++ b/scss/_base_tables.scss
@@ -6,9 +6,14 @@
     border: 0;
     border-collapse: collapse;
     line-height: map-get($line-heights, default-text);
+    margin-bottom: $spv-outer--scaleable;
     overflow-x: auto;
-    padding-bottom: $spv-outer--scaleable;
     width: 100%;
+
+    &::after {
+      content: '';
+      margin-bottom: $spv-outer--scaleable;
+    }
 
     @if ($table-layout-fixed) {
       table-layout: fixed;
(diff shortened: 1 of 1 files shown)


lyubomir-popov:
Percy alerted me to the fact that some older versions of firefox add white space between table cells, and that space increases endlessly when resizing. I guess that's why we went for padding-bottom instead of a margin bottom. Will have to find another solutions, adding a don't merge label for now.

lyubomir-popov:
@deadlight I've converted the non-working padding-bottom to a transparent border bottom, as padding doesn't work with border-collapse and margin triggers an endless increase of the margin on resize in older versions of firefox. 

Border seems to be the only thing that works that I can think of, if anyone has ideas of a better fix please comment.

deadlight:
I'm not too happy with the using border here.

I think we need to find out why the whitespace issue is happening in FF, it would be nicer to use more intuitive CSS attributes.

lyubomir-popov:
I'm out of ideas regarding the margin, especially since it doesn't break in the latest ff. Closing this.

PR #2440, opened 2019-08-01
Hide second navigation list in docs side nav on page

deadlight:
## Done

Fix the navigation bug in docs where the duplicate nav is visible on page load

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/
- See that the docs sidebar ends with the "resources" section
- Test the search box still works


kwm14:
LGTM 👍

PR #2445, opened 2019-08-05
decrease scope of selector so only direct children of .row  span full width

lyubomir-popov:
Done

Decrease scope of .row * to .row > *, so that it only affects direct children of .row

## QA

- Pull code
- Run `./run serve --watch`
- Copy the markup and .grid-component style into an example
- Verify Left and right appear side by side instead of stacked.

## Details

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/2444


bartaz:
Seems to solve the issue, thanks!

Screenshot 2019-08-05 at 17 35 17

PR #2486, opened 2019-08-14
JS notifications

gozalig:
Added JS warning notifications

For pages Accordion, Code, Tables, Baseline grid, and Spacing settings.

## QA

- Pull code
- Run `./run serve --watch`
- [Add additional steps]

## Details

## Screenshots

### DIFF
--- a/docs/base/code.md
+++ b/docs/base/code.md
@@ -30,6 +30,13 @@ Code copyable should be used when presenting the user with a small snippet of co
 View example of the code copyable pattern
 </a>
 
+
+<div class="p-notification--caution">
+  <p class="p-notification__response">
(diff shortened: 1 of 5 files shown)


kwm14:
Nice work @gozalig 👍 

Question for @anthonydillon it's great we now highlight which components require JS, but do we supply an example to use.

As well as informing users 'This pattern requires JS to be functional.' could we also link to an example? Just an idea...

anthonydillon:
@kwm14 Yeah I did think of this but the best we could do is link to the examples source which could also be confusing. I think there are two tasks after this one:
- Elaborate on the notification with the classes/aria attributes that trigger the behaviours
- Update example-js to display the JS in another tab.

This is a good first step.

PR #2507, opened 2019-08-21
Theming spike

### LINKED ISSUE #2449: Contrast issues with hover in dark navigation
Default hover text color in navigation doesn't have enough contrast on dark background:
(see Forum on the screenshots:
Screenshot 2019-08-06 at 10 08 44

**Expected behavior**
It should be possible to configure nav hover color to adapt it to custom background. Or hover colour should be more universal.
Especially that most of Canonical sites now use dark navigation.

lyubomir-popov:
## Done

Introduces theming infrastructure and themes hr, nav, p-search-box. Note the subnav not themed in this spike.

- theming settings file _settings_theming.scss
- theming color maps in _settings_colors
- separates colors from rest of rules

Drive by fixes:

1. Removes button styling applied using the [type='submit'] selector, as it has very high specificity and is difficult to override. Example: green border  on hover on the current ubuntu.com search in nav:
image

2. p-search-box improvements:
 - focus states of button, reset, and p-search-box (the outer container) no longer overlap, and are same height
 - norder of magnifier button removed after consultation with @kwm14 
 - margin-bottom matches margin-bottom on select, other inputs, etc.
 - makes input absolutely positioned, while the reset and search buttons become relatively positioned on top. This allows removal of a lot of hacks and hardcoded vsalues around the positioning of those elements

3. Button improvements:
- Realigns horizontal padding on buttons with other components using $sph-inter With 1.7, left/right button padding was aligned with other elements using the $sph-inter variable. At some point later this changed to $sph-inter--small * 1.5, which doesn't match anything else.
- Improves icons inside buttons - these used display: inline-block, and hacks like top: 2px to counteract the extra white space inline-block causes. This has been replaced with flex, which behaves more predictably and doesn't require hacks. It should be thoroughly tested though.

Also fixes #2449

## QA

- Pull code
- Run `./run serve --watch`
- Open /examples/patterns/search-box/navigation/ and QA the example.
The default is currently set to light. Inspect on all screen sizes; add is-dark to both nav and search box. Type something into the searchbox, press tab 3 times to see focus outline not overlapping between input, reset, button
- Change the value for keys nav, p-search-box in _settings_themes.scss to true. Repeat the above, adding is-white instead of is-dark to see the alternative to the default theme.

Inspect hr example too.

Documentation forthcoming.

## Details

## Screenshots


lyubomir-popov:
@anthonydillon @deadlight This is ready for a first look I think. Let me know if somethign doesn't make sense from the above instructions.

scss/_patterns_icons.scss
```diff
@@ -203,7 +204,7 @@ $color-social-icon-foreground: $color-mid-x-light !default;
     @include vf-icon-anchor($color-mid-dark);
 
     [class*='--dark'] &,
-    &.is-light {
+    &.is-white {
```
lyubomir-popov: these renames  are a mistake, should be is-light.

anthonydillon:
@lyubomir-popov I've added comments to Percy. There seems to be a few issues stemming from the drive by's. Can we try and keep branches focused on a single feature/improvement please.

docs/examples/patterns/navigation/default-dark.html
```diff
@@ -0,0 +1,35 @@
+---
+layout: examples
+title: Navigation / Default - dark
+category: _patterns
+---
+<header id="navigation" class="p-navigation is-dark">
```
anthonydillon: Why didn't you make dark a modifier?

scss/_patterns_search-box.scss
```diff
@@ -41,13 +48,92 @@
...
+    &,
+    &:active,
+    &:focus,
+    &:hover {
+      background-color: transparent !important; //XXX: remove important once buttons have been themified
+      border-width: 0;
+    }
+  }
+
+  %search-box-input--white {
```
anthonydillon: light?

scss/_patterns_strip.scss
```diff
@@ -81,7 +81,7 @@
     background-repeat: no-repeat;
     background-size: cover;
 
-    &.is-light {
+    &.is-white {
```
anthonydillon: why this change?

scss/_settings_colors.scss
```diff
@@ -33,3 +32,24 @@ $states: (
...
+  background-highlighted: #f7f7f7,
+  border-default: #cdcdcd,
+  border-high-contrast: #999,
+  text-hover: #757575,
+  text-disabled: #666,
+  text-default: #111
+);
+
+//text-hover: minimum contrast on primary that satisfies contrast checker AA
+$colors--dark-theme: (
```
anthonydillon: I like this. Its simple and clear

scss/_settings_themes.scss
```diff
@@ -0,0 +1,5 @@
+$theme-default--dark: (
```
anthonydillon: I think this will be difficult to theme

scss/_patterns_strip.scss
lyubomir-popov: Reserving the name 'light' for a theme based on f7f7f7 background, as in the p-strip--light.

docs/examples/patterns/navigation/default-dark.html
lyubomir-popov: Just because the p-strip is called is-dark. WIl lchange it to --dark.

lyubomir-popov:
@anthonydillon just discovered [this](https://canonical-web-and-design.github.io/practices/coding/stylesheets.html) bem practices document, and according to it state classes is-/has- are not appropriate here as the theme is not in response to a state. However the document doesn't  provide guidance on concatenating modifiers, which is the issue I have in this pr as mentioned above.

scss/_patterns_icons.scss
lyubomir-popov: @anthonydillon do you agree with treating f7f7f7 as its own theme and reserving the name for it?

anthonydillon:
@lyubomir-popov that's right. We have misused state classes to help give use cross modifications. There has never been a solution for this. What would you propose? I feel that if we have multi modifications on a pattern we are failing to keep the framework simple to use and maintain. Do you agree?

lyubomir-popov:
This is difficult to answer in a succinct manner, so please bear with me :)


> We have misused state classes to help give use cross modifications. There has never been a solution for this. What would you propose?

It's just a consequence of not following BEM precisely, isn't it?

- we merge the modifier into the block class, i.e. p-strip--light instead of p-strip p-strip--light
- we make a distinction between states and appearance (not sure why?), whereas BEM handles everything - appearance, state, and behaviours - using the [modifier syntax](https://en.bem.info/methodology/quick-start/#modifier). 
- as soon as we find ourselves in the common scenario of needing two modifiers, we break the rules about state classes, which compounds the issue by introducing inconsistency in naming (e.g. p-strip is-dark).

Granted, we're saving a few characters by typing: 
```p-strip--light is-shallow```

instead of: 
```p-strip p-strip--light p-strip--shallow```

But there's a heavy price to pay for this:
- We have a practices doc which is at odds with what we're doing, and there's no clear way of fixing this without causing even more confusion
- This pursuit of single classes is followed so rigorously, that people happily drop functionality (like deleting mobile / tablet classes, which results in visual regressions) and consider this an improvement
- Departing from something like BEM, which most devs are familiar with, also adds a learning curve as you learn to adapt to the customised approach we use. 
- Every time someone tries to write an is-class to work around the limitations of single modifiers, a they get a Code:-1 label and a discussion like this ensues.

> I feel that if we have multi modifications on a pattern we are failing to keep the framework simple to use and maintain. Do you agree?

I think things should be made as simple as possible, but not simpler. Being able to specify both behaviour and appearance on the same element is not a sign of over-complication, it's just a very common requirement that every framework has to fullfill. A couple of examples:

- material design (all of this for just an input on a login form):
```mdc-text-field username mdc-ripple-upgraded mdc-text-field--invalid```
image


- bootstrap:
```navbar navbar-expand-md navbar-dark fixed-top bg-dark```
image

I wouldn't call these failures. My guess is that once you get to a certain scale, it is the only viable approach. Not pretty, but it works and you don't have to ever think about it.

---
In short, I think it is worth reconsidering the pros and cons of the single modifier rule. In the meantime, is-classes are short, and impose no limit on the amount of modifications you can specify, so they seem as a choice that involves fewer headaches.

docs/examples/patterns/navigation/default-dark.html
lyubomir-popov: agreed to leave like this

scss/_patterns_search-box.scss
lyubomir-popov: √ changed to light

scss/_patterns_strip.scss
lyubomir-popov: changed to light

scss/_settings_themes.scss
lyubomir-popov: agreed this is not actionable

lyubomir-popov:
I think this is ready for another look. I've fixed the actionable feedback above.

lyubomir-popov:
@deadlight how do we get around the !Important keyword limit?

lyubomir-popov:
@kwm14 percy shows external links as misaligned, but locally on ff it looks ok. Please test locally as I think the percy diff is not working correctly in this case:
image

kwm14:
Nice work @lyubomir-popov 👍 

Just noticed one thing on **/examples/patterns/navigation/default-dark/** is this part of this PR?

- [ ] Logo and menu on mobile that paddings are removed and snap to edges

Screenshot 2019-09-02 at 09 59 09
Screenshot 2019-09-02 at 09 59 22

lyubomir-popov:
@kwm14 the markup needs updating, will fix.

kwm14:
Thanks, @lyubomir-popov update looks good 👍

lyubomir-popov:
@kwm14 thanks, but there is some sort of merge regression in the pagination, I'll pushe once more and then merge even though everything is green.

lyubomir-popov:
@kwm14  Ithink I've fixed it. Can you please take one final look at percy and approve.

PR #2515, opened 2019-08-28
Fix positive buttons with icons

### LINKED ISSUE #2485: Button `has-icon` only works on unmodified `p-button`
**Describe the bug**
Button `has-icon` only works on unmodified `p-button` because of usage of `~=` attribute selector that matches whole words.
If button has any other button class, like `p-button--positive` `has-icon` has no effect.

**To Reproduce**
- have a `--positive` button with an icon
- check small screen
Example: https://jsfiddle.net/bartaz/Lcu29e78/

**Expected behavior**
`has-icon` should work on all buttons.

**Screenshots**
Screenshot 2019-08-14 at 12 44 06

lyubomir-popov:
## Done

- Fixes #2485.
- Changes icon color in positive buttons from grey to white

## QA

- Pull code
- Run `./run serve --watch`
- Open examples/patterns/buttons/icon/
- Verify positive button does not go full-width on mobile

### DIFF
--- a/docs/examples/patterns/buttons/icon.html
+++ b/docs/examples/patterns/buttons/icon.html
@@ -3,5 +3,5 @@ layout: examples
 title: Buttons / Icon
 category: _patterns
 ---
-<button class="p-button has-icon"><i class="p-icon--plus"></i></button>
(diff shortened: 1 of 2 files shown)


kwm14:
Design LGTM 👍

PR #2516, opened 2019-08-28
Add styling for detail element

### LINKED ISSUE #2455: `details` element doesn't have a bottom margin
While not used extensively, it seems the discourse docs produce a `details` element. Which when closed, do not have any margin (and when open rely on the margin of the inner elements).

https://codepen.io/Lukewh/pen/ymvqPa

It'd be nice to have the default margin applied as with `p` and `dd` elements :+1:

lyubomir-popov:
## Done

Fixes #2455

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/vanilla-framework/examples/detail
- Verify detail element has margin bottom


lyubomir-popov:
image

I've added basic styling, but I think we should also replace the triangle indicator with a chevron and indent the inside text to align wiht the summary. What do you think @kwm14?

kwm14:
@lyubomir-popov is 'Details' a new component?

kwm14:
> I've added basic styling, but I think we should also replace the triangle indicator with a chevron and indent the inside text to align with the summary. What do you think @kwm14?

@lyubomir-popov agree to update triangle icon with our chevron for consistency 👍

lyubomir-popov:
@kwm14 I wouldn't call it a new component, it is a standard html element that hasn't been covered by vanilla up until now. SO it is a new entry in the base category. One thing I'm not sure about is, should we add a documentaiton page? If yes, what can we say about it that doesn't repeat https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

kwm14:
> @kwm14 I wouldn't call it a new component, it is a standard HTML element that hasn't been covered by vanilla up until now. So it is a new entry in the base category.

Ok cool 
> 
> One thing I'm not sure about is, should we add a documentation page? If yes, what can we say about it that doesn't repeat https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

I'd recommend if it's in the framework we have it documented so a docs page would be good, even if it's a couple of lines with the example, similarly to [Reset](https://docs.vanillaframework.io/base/reset/) 👍then at least users can see it in the side-bar navigation.

lyubomir-popov:
Turns out stryling the triangle is a task in itself https://github.com/whatwg/html/issues/722. WIll have to do this as a separate issue.

lyubomir-popov:
@kwm14 I'll need to fix the labels to add it to the docs as this happens now:
I need to make them smaller that ok with you?

image

currnetly one is pushing the other

kwm14:
> @kwm14 I'll need to fix the labels to add it to the docs as this happens now:
> I need to make them smaller that ok with you?

Hmmm that is pretty tight and annoying it pushes the label down, happy to go smaller and match what has been implemented on JAAS support labels - https://jaas.ai/search?type=bundle I believe the font-size is `0.875rem`

lyubomir-popov:
@deadlight @kwm14 ready for another look

lyubomir-popov:
@kwm14 percy still shows adjacent label pushing each other, but that's because we need a vanilla release for the new label styling to become available to docs. The code is already in.

scss/_patterns_label.scss
```diff
@@ -10,11 +10,12 @@
...
 
 %vf-label {
-  border-radius: 0.125rem;
+  @extend %x-small-text;
+  border-radius: $border-radius;
   display: inline-block;
-  font-size: 0.875rem;
-  line-height: 1.3125rem;
-  padding: calc(#{$sp-unit * 0.625} - 1px) $sph-inner--small * 1;
+  font-size: 100px !important;
```
deadlight: I think that this is a mistake

kwm14:
@lyubomir-popov as well as fixing full-width labels just noticed another updated needed.

You're just missing the title from 'Details' docs page 😉 

Screenshot 2019-08-29 at 16 09 53

lyubomir-popov:
@kwm14 fixed the title

kwm14:
Thanks, @lyubomir-popov LGTM 👍

PR #2556, opened 2019-09-18
Update nav in dark theme to use #333

lyubomir-popov:
## Done

Fix https://github.com/canonical-web-and-design/vanilla-framework/issues/2554
## QA

- Pull code
- Run `./run serve --watch`
- Open dark nav example
- Verify background color is #333

### DIFF
--- a/scss/_settings_colors.scss
+++ b/scss/_settings_colors.scss
@@ -43,15 +43,14 @@ $colors--light-theme: (
   text-hover: #757575,
   text-disabled: #666,
   text-default: #111
-);
+) !default;
 
 //text-hover: minimum contrast on primary that satisfies contrast checker AA
 $colors--dark-theme: (
   background: hsl(0, 0%, 15%),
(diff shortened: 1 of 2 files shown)


scss/_settings_colors.scss
```diff
@@ -43,15 +43,14 @@ $colors--light-theme: (
   text-hover: #757575,
   text-disabled: #666,
   text-default: #111
-);
+) !default;
 
 //text-hover: minimum contrast on primary that satisfies contrast checker AA
 $colors--dark-theme: (
   background: hsl(0, 0%, 15%),
```
bartaz: Out of curiosity - why not changing background color (as it is already used by the nav)?
Do we want to keep `background` the same, because it's used elsewhere?

deadlight:
@lyubomir-popov Can this land or does @bartaz's comment need action?

lyubomir-popov:
@deadlight it can land once it gets a review +1

PR #2570, opened 2019-10-07
Add subset latin fonts

pmahnke:
## Done

- created a set of subset fonts for the Ubuntu typeface
    - they can be optionally added via a `$font-use-subset-latin: true;` setting. but will give users a ~66% font size savings to download.
    - the default is exactly want it is today
    - I also created non-google hosted subset fonts for cyrillic, greek, and latin (not sure why we need latin)
    - Removed the `$font-import: 'http://fonts.gstatic.com/s/ubuntu/v9/' !default;` setting as it is no longer used
- added a `$font-display-option: false !default;` setting and added the logic to allow users to easily add that option in `_base_typography.scss`
- Updated the `typography.md` documentation to reflect these changes
- preloading the subset fonts in `examples.md`, but this can be removed or extended

## QA

- Pull code
- Set combinations of the above variables
- Run `./run serve --watch`
- Look at the iframes in examples and their performance
- Read http://0.0.0.0:8101/base/typography/ to see if the new sections make sense

### DIFF
--- a/scss/_base_fontfaces.scss
+++ b/scss/_base_fontfaces.scss
@@ -2,6 +2,9 @@
   @if str-index($font-base-family, 'Ubuntu') {
     @if $font-use-subset-latin {
       @font-face {
+        @if $font-display-option {
+          font-display: $font-display-option;
+        }
         font-family: 'Ubuntu';
         font-style: normal;
         font-weight: 300;
@@ -10,6 +13,9 @@
       }
 
       @font-face {
+        @if $font-display-option {
+          font-display: $font-display-option;
+        }
         font-family: 'Ubuntu';
         font-style: normal;
         font-weight: 400;
@@ -18,6 +24,9 @@
       }
 
       @font-face {
+        @if $font-display-option {
+          font-display: $font-display-option;
+        }
         font-family: 'Ubuntu';
         font-style: italic;
         font-weight: 300;
@@ -26,6 +35,9 @@
       }
 
       @font-face {
+        @if $font-display-option {
+          font-display: $font-display-option;
+        }
         font-family: 'Ubuntu';
         font-style: italic;
         font-weight: 400;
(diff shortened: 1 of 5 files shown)


scss/_base_typography.scss
```diff
@@ -14,6 +14,9 @@
     -webkit-font-smoothing: antialiased;
     // sass-lint:enable no-vendor-prefixes
     color: $color-dark;
+    @if $font-display-option {
+      font-display: $font-display-option;
```
anthonydillon: Font display needs to be defined in the `@font-face` decoration.

anthonydillon:
Still need to QA

scss/_base_typography.scss
pmahnke: moved

scss/_base_fontfaces.scss
```diff
@@ -1,106 +1,224 @@
...
-      font-family: 'Ubuntu';
-      font-style: normal;
-      font-weight: 300;
-      src: url('#{$assets-path}e8c07df6-Ubuntu-L_W.woff2') format('woff2'), url('#{$assets-path}8619add2-Ubuntu-L_W.woff') format('woff');
-    }
+    @if $font-use-subset-latin {
+      @font-face {
+        @if $font-display-option {
+          font-display: $font-display-option;
+        }
```
anthonydillon: This is a lot of duplication that need to be maintained if there is a change. You could include all the common styles in a mixin and include it. For example, https://www.sassmeister.com/gist/839dc206787ad1545d3db1d9744cbb9b

anthonydillon:
Tested both new options. 

- `$font-use-subset-latin` reduces the font-weight from 411kb to 132kb. 
- `$font-display-option` was visible on a throttled network

scss/_base_fontfaces.scss
pmahnke: or is the default of `auto` a cleaner option?

pmahnke: I moved to the auto option, made that the default and updated the docs.  Also removed the preload on examples as it was giving a console warning for being after the actual resource was loaded, which I think is good.

anthonydillon:
Do not require design as there are not visual changes.

PR #2583, opened 2019-10-17
#1960 - Add padding to search box so text doesn't go underneath.

### LINKED ISSUE #1960: Search box reset button has unnecessary margin-right + strange hover/focus states
1. There should not be any margin on the reset button. It was introduced in the `_base_button.scss` file when trying to add margin to inline buttons (`button + button`).
2. The reset button appears over top of the focus outline
3. There should probably be extra padding-right on the input because at the moment the text can be hidden underneath the reset button

screenshot_1
screenshot_2

Caleb-Ellis:
## Done

- Added padding to search box so text doesn't go underneath.

## QA

- Pull code
- Run `./run serve --watch`
- Open 0.0.0.0:8101/examples/patterns/search-box/default/
- Add enough text to the search box input so it reaches the reset button
- Check that the text doesn't go underneath

## Details

Fixes #1960

### DIFF
--- a/scss/_patterns_search-box.scss
+++ b/scss/_patterns_search-box.scss
@@ -36,6 +36,7 @@
   .p-search-box__input {
     flex: 1 1 100%;
     margin-bottom: 0;
+    padding-right: calc(2 * #{2 * $spv-nudge + map-get($line-heights, default-text)});
     position: absolute;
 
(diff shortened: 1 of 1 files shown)


kwm14:
Nice one @Caleb-Ellis I didn't even realise the text went under buttons 👍nice fix 💯

PR #2592, opened 2019-10-22
doc page for theming

lyubomir-popov:
## Done

- Expand the color page to include information about the newly introduced theming mechanism
- Replace color theming map with variables for easier override

## QA

- Pull code
- In color-settings.md, check that the copy makes sense
- In color-settings.md, check that the described way of altering a theme works
- In color-settings.md, check that the described way of applying is-dark / is-light classes works
- If the above seem agreeable, build the documentation and check if it looks ok

## Details

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/2331

### DIFF
--- a/docs/patterns/navigation.md
+++ b/docs/patterns/navigation.md
@@ -20,7 +20,7 @@ displayed horizontally on larger screens.
 
 The navigation pattern is one of the first patterns to implement the new theming architecture in Vanilla. The default is light. But, to switch to a dark navigation, you can either:
 
-- Override the value of the `nav` key in the `$theme-default--dark` map (initially set in `_settings_themes.scss`) to `true`: `$theme-default--dark: map-merge($theme-default--dark, ("nav": true));`
+- Override the value of the `$theme-default-nav` in `_settings_themes.scss` to `dark`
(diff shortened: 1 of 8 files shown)


kwm14:
@lyubomir-popov docs page looking good but adding 'In progress' label inside contents menu isn't working. I think it strips out the CSS...think I recall that happening to me hence me adding next to the page title as you have.

Screenshot 2019-10-23 at 09 59 05

Also, all the default examples that can have to theme applied have been overridden to dark?

Screenshot 2019-10-23 at 09 57 33
Screenshot 2019-10-23 at 09 57 57
Screenshot 2019-10-23 at 09 58 08
Screenshot 2019-10-23 at 09 58 17
Screenshot 2019-10-23 at 09 58 25

lyubomir-popov:
@kwm14
- label -  fixed by removing it.
- navs - set to light

kwm14:
> @kwm14
> 
> * label -  fixed by removing it.
> * navs - set to light

Nice. Will have a look 👍

deadlight:
@lyubomir-popov the spell check is failing. I noticed "bersides" (which should be "besides") but run `./run test` to catch them all.

lyubomir-popov:
Thanks for spotting, didn't know about ./run test. It returned 'Tests succeeded' after fixing 'bersides'.

PR #2598, opened 2019-10-24
Removes <span> elements from <blockquote> example

### LINKED ISSUE #2597: Unexplained <span>s in blockquote example
1\. Go to [the blockquote reference](https://docs.vanillaframework.io/base/typography/#blockquotes-and-citations).
2\. Try to understand the example code.

image

What you see:
- The purpose of the `<span>`s is not explained.
- No specific style is applied to them, and (as far as I can tell) no script is attached to them.

What you should see: either
- the quote marks are not in `<span>` elements, or
- the purpose of the elements is explained.

In addition, there’s a mistaken space after the starting quote mark, and a mistaken space before the ending quote mark. Those mistakes probably wouldn’t have happened if the quote marks weren’t in separate elements.

matthewpaulthomas:
Fixes #2597.

## Done

- Removes `<span>` elements from `<blockquote>` example, and trailing+leading spaces.
- Fixes nested quote marks.

### DIFF
--- a/docs/examples/base/blockquotes.html
+++ b/docs/examples/base/blockquotes.html
@@ -4,6 +4,6 @@ title: Blockquote
 category: _base
 ---
 <blockquote>
-  <p><span>“</span> Ubuntu is an ancient African word meaning 'humanity to others'. Ubuntu is an ancient African word meaning 'humanity to others'. <span>”</span></p>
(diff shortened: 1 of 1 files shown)


kwm14:
Thanks, @matthewpaulthomas LGTM 👍

PR #2636, opened 2019-11-11
Provide utility to overlap table cell padding

lyubomir-popov:
## Done

Provide utility to overlap table cell padding

## QA

- Pull code
- Run `./run serve --watch`
- Open examples/utilities/table-row-padding-overlap/, verify example is correct and classes work as expected

## Details

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/2504
Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/2463


docs/examples/utilities/table-cell-padding-overlap.html
```diff
@@ -0,0 +1,22 @@
...
+        <td><button class="u-table-cell-padding-overlap">Select</button></td>
+        <td>Text</td>
+    </tr>
+    <tr>
+      <td><button class="is-dense u-table-cell-padding-overlap">Select</button></td>
+      <td>Text not wrapped in a tag</td>
+    </tr>
+    <tr>
+      <td><button class="u-table-cell-padding-overlap">Select</button></td>
+      <td><p class='u-no-margin--bottom u-table-cell-padding-overlap'>text in a paragraph</p></td>
```
kwm14: Capitalise 'Text' so it's consistent with other table row examples

kwm14:
Design LGTM 👍 just once comment for QA above 😄

lyubomir-popov:
@kwm14 well spotted, thanks! Fixed.

scss/_settings_spacing.scss
```diff
@@ -147,6 +147,9 @@ $icon-sizes: (
   thumb--large: $sp-unit * 12
 ) !default;
 
+$table-cell-padding-top: $spv-inner--x-small--scaleable !default;
+$table-cell-padding-bottom: $spv-inner--x-small--scaleable !default;
```
anthonydillon: Wouldn't we want these to always be the same value? If these are override able (!default) then they should be documented?

lyubomir-popov: Good point. The only purpose of these is to ensure table cell padding and the overlap utility always match, so no need for `!default` and no need for separate top/bottom vars. I've updated the code.

lyubomir-popov:
This is ready for another look.

PR #2651, opened 2019-11-14
Increase input margin bottom by .5rem

lyubomir-popov:
## Done

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/2497

## QA

- Pull code
- Run `./run serve --watch`
- set $multi to 1
- Open http://192.168.64.2:8101/examples/templates/tick-element-comparison/
- Verify margin-bottom on the inputs is .7rem 

## Screenshots

image

### DIFF
--- a/scss/_settings_spacing.scss
+++ b/scss/_settings_spacing.scss
@@ -1,3 +1,6 @@
+// Density multiplier
+$multi: 2 !default;
+
 // Options
 $pad-lists: true !default; // Adds padding to list items
 
@@ -70,9 +73,6 @@ $browser-rounding-compensations: (
   h1: 0.001rem
 ) !default;
 
-//scaling factor
-$multi: 2 !default;
-
 ////////////////////////////////////////////////////////
 // Variables connecting spacings that should be identical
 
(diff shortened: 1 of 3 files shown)


scss/_settings_spacing.scss
```diff
@@ -1,3 +1,6 @@
+// Density multiplier
+$multi: 2 !default;
```
anthonydillon: This is an unclear variable name

anthonydillon: Also looks unused

lyubomir-popov: @anthonydillon This is not new - the density multiplier got introduced with v 1.7. I just moved it to the top of the file as it is the most important setting in this file.

lyubomir-popov: Happy to rename but it will be a breaking change so I suggest doing that with v 3.0

kwm14:
Design LGTM 👍

scss/_settings_spacing.scss
anthonydillon: I couldn't find documentation for is use. I guess that is something I missed out should be added in a follow up.

lyubomir-popov: @anthonydillon True, I've made an [issue](https://github.com/canonical-web-and-design/vanilla-framework/issues/2656) to document it.

PR #2654, opened 2019-11-15
Add truncate util; increase size limit to pass tests

lyubomir-popov:
## Done

Add utility called u-truncate that truncates text.

## QA

- Pull code
- Run `./run serve --watch`
- Open http://192.168.64.2:8101/utilities/truncate/
- QA copy and example

## Details

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/2625

## Screenshots

image

### DIFF
--- a/docs/utilities/truncate.md
+++ b/docs/utilities/truncate.md
@@ -1,0 +1,27 @@
+---
+layout: default
+---
+
+## Truncation
+
+<hr>
+
+To truncate text, use the class `u-truncate`.
+
+When applied to a parent container, it will truncate any elements placed inside, so long as the parent's width is set.
+When applied to individual elements within a parent, e.g. a table cell, only those elements will be truncated.
+
+<a href="/examples/utilities/truncate/"
+    class="js-example">
+View example of the truncate utility
+</a>
+
+### Import
+
+To import just this utility into your project, copy the snippet below and include it in your main Sass file.
+
+```scss
+@import 'utilities_truncate';
+```
+
+For more information see [Customising Vanilla](/customising-vanilla/) in your projects, which includes overrides and importing instructions.
--- a/docs/utilities/truncate.md
+++ b/docs/utilities/truncate.md
@@ -16,6 +16,8 @@ When applied to individual elements within a parent, e.g. a table cell, only tho
 View example of the truncate utility
 </a>
 
+Note: `u-truncate` will affect any tooltips placed inside.
+
 ### Import
 
 To import just this utility into your project, copy the snippet below and include it in your main Sass file.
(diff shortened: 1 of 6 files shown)


spencerbygraves:
@lyubomir-popov could we add title attribute here, as [suggested](https://github.com/canonical-web-and-design/jaas-dashboard/issues/212) by @anthonydillon for the [jaas dashboard](https://github.com/canonical-web-and-design/jaas-dashboard/pull/214).

Or perhaps it would be a separate task.

tests/parker.js
```diff
@@ -12,7 +12,7 @@ function generateMetrics(file, metricsArray) {
     {
       name: 'Stylesheet size',
       benchmark: 150000,
-      threshold: 200000,
+      threshold: 250000,
```
Caleb-Ellis: If you're going to do this, I'd urge you to add a card and plan some time to refactor Vanilla so it's back below 200000. Typically we shouldn't change the tests so that the code passes, but rather change the code so that the test passes, otherwise the tests serve no purpose. i.e. Why stop at 250000 if whenever the threshold is reached its simply raised?

lyubomir-popov: The problem is, these are tiny commits, so there's nothing here to refactor, and all open issues would be blocked if we stick to the current limit. We need an issue to see if it is organic growth, or something that created an unreasonable amount of output. I've created an issue [here](https://github.com/canonical-web-and-design/vanilla-framework/issues/2657):

Caleb-Ellis: Nice! Yep, wasn't suggesting this should block merging work, just disagreed with increasing the threshold without a plan to investigate or improve.

docs/utilities/truncate.md
```diff
@@ -0,0 +1,27 @@
...
+---
+
+## Truncation
+
+<hr>
+
+To truncate text, use the class `u-truncate`.
+
+When applied to a parent container, it will truncate any elements placed inside, so long as the parent's width is set.
+When applied to individual elements within a parent, e.g. a table cell, only those elements will be truncated.
```
anthonydillon: Maybe warn that this will hide tooltips and if you want tooltips you should truncate inside the tooltip.

tests/parker.js
anthonydillon: Did you not need this in the end?

lyubomir-popov: We need it until we optimise or drop newly added icons that are mainly causing the file size increase. I've made an [issue](https://github.com/canonical-web-and-design/vanilla-framework/issues/2657) for it.

docs/utilities/truncate.md
lyubomir-popov: I've added a note of warning. I've also experimented a bit and I don't think placing a tooltip inside the truncated element can work. The only way I can think of would be to use js to position an independent tooltip to coincide with the truncate element's position.

kwm14:
@lyubomir-popov Percy approved 👍

lyubomir-popov:
thanks @karlos!

PR #2665, opened 2019-11-21
Borders exploration demo [wip] 

lyubomir-popov:
## Done

[ Not for review. Just a demo for internal visual design discussion ]

Adjust contrast of borders and creates a noticeable difference between inputs and buttons. Experiments with alternative table headings. Increases space under tables and other block elements using $spv-outer--scaleable.

## QA

- Pull code
- Run `./run serve --watch`
- Open /examples/templates/tick-element-comparison/ to see the demo

Proposal:

image

Current state:

image

### DIFF
--- a/docs/examples/patterns/tables/table-mobile-card.html
+++ b/docs/examples/patterns/tables/table-mobile-card.html
@@ -3,7 +3,7 @@ layout: examples
 title: Table / Mobile card
 category: _patterns
 ---
-<table class="p-table--mobile-card" role="grid">
+<table class="p-table--mobile-card u-table-layout--auto" role="grid">
   <thead>
     <tr role="row">
       <th>FQDN</th>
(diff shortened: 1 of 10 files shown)


lyubomir-popov:
@kwm14 @deadlight

matthewpaulthomas:
> Adjust contrast of borders

Hard to avoid the problem that a light grey border makes controls look disabled. You can’t rely on their labels being greyed out, because they might not have labels: for example, a navigation bar search field, an icon-only button, or a checkbox or radio button alone in a MAAS table cell.

> creates a noticeable difference between inputs and buttons

I realise this is very similar to Material, but it’s weird to me that `<select>` ends up looking much more like a text field than like a menubutton, when it behaves much more like a menubutton. What would/will a Vanilla `<datalist>` look like?

lyubomir-popov:
Thanks for the feedback @matthewpaulthomas, I'll make some adjustments to address these concerns.

cassiocassio:
I know that @lyubomir-popov is keen to explore input fields a la Material Design :-)

There's pros and cons of the look and feel of the underlined grey box vs the complete (round)rectangle

- write on the line like a handwriting exercise
vs
- click in the box

I guess one theory is that without a grey holding background for white boxes,  figure-ground relationship is destroyed.

We can play with this issue, but I feel that

(1) **material has a complete set of controls in both styles, and they claim it's a matter of visual taste **
Screenshot 2019-12-13 at 12 50 41

(2) **what's being proposed here is mix and match**

Screenshot 2019-12-13 at 12 56 49

bartaz:
@lyubomir-popov Is this PR still needed?

lyubomir-popov:
@bartaz No, closing.

PR #2666, opened 2019-11-21
Prevent table header border from disappearing when tbody contains no rows

lyubomir-popov:
## QA

- Pull code
- Run `./run serve --watch`
- Open any table example, e.g. /examples/patterns/tables/table-sortable/
- Delete everything after the thead
- Verify border on thead tr does not disappear

## Details

Fix https://github.com/canonical-web-and-design/vanilla-framework/issues/2662

## Screenshots

image

### DIFF
--- a/scss/_base_tables.scss
+++ b/scss/_base_tables.scss
@@ -31,8 +31,6 @@
 
   thead {
     &::after {
-      // prevent border on header tr from disappearing when there is no tbody
-      // https://github.com/canonical-web-and-design/vanilla-framework/issues/2662
       content: '';
     }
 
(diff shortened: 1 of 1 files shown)


scss/_base_tables.scss
```diff
@@ -30,6 +30,12 @@
   }
 
   thead {
+    &::after {
+      // prevent border on header tr from disappearing when there is no tbody
+      // https://github.com/canonical-web-and-design/vanilla-framework/issues/2662
```
anthonydillon: This does not need the comment here.

PR #2670, opened 2019-11-22
Fix middot list

lyubomir-popov:
## Done

Refactors the middot element:
- replaces the small dot made bigger by a font size increase with the bullet unicode character, which is what the typeface designer has deemed correct dot size relative to rest of text at that font-size
- compensates for the width to which white space is collapsed between inline elements - in my experiments, this width is around 3.6px, which is then rounded to 3 or 4 depending on whether it starts on a round pixel or not; therefore, I've created a new variable called $strut-width, which is constant and shouldn't depend on font, browser, or font size. (tested with ubuntu light, bold, thin, in ff and chrome). It will need to be documented with a subsequent pr

(Strut in red):
image

-removes some patches (e.g. text color) as the current implementation doesn't need it (I've expanded the example to include a list of links to demonstrate):
image
- removes some other unneeded styles and replaces them with a better targeted selector

- improve wrapping of dot (it wraps as soon as there's enough width for the dot to wrap, as opposed to staying on the preceding line)

## QA

- Pull code
- Run `./run serve --watch`
- Open examples/patterns/lists/lists-mid-dot/
- QA

## Details

https://github.com/canonical-web-and-design/vanilla-framework/issues/2557

## Screenshots

image

### DIFF
--- a/docs/patterns/lists.md
+++ b/docs/patterns/lists.md
@@ -76,6 +76,8 @@ View example of the inline list pattern
 Apply the class `.p-inline-list--middot` to add a middot character between
 inline list items.
 
+As with any inline/inline-block element, any amount of white space (one or more spaces, tabs, etc.) in the markup is rendered as a single white space character. Vanilla uses the <a href="https://css-tricks.com/fighting-the-space-between-inline-block-elements/#article-header-id-3">font size method</a> to elliminate white space within a `p-inline-list--middot`, that is not inside a `p-inline-list__item`. White space within the `p-inline-list__item` however will displace the dot.
+
 <a href="/examples/patterns/lists/lists-mid-dot/"
     class="js-example">
 View example of the middot list pattern
--- a/docs/patterns/lists.md
+++ b/docs/patterns/lists.md
@@ -76,7 +76,7 @@ View example of the inline list pattern
 Apply the class `.p-inline-list--middot` to add a middot character between
 inline list items.
 
-As with any inline/inline-block element, any amount of white space (one or more spaces, tabs, etc.) in the markup is rendered as a single white space character. Vanilla uses the <a href="https://css-tricks.com/fighting-the-space-between-inline-block-elements/">font size method</a> to eliminate white space within a `p-inline-list--middot`, that is not inside a `p-inline-list__item`. White space within the `p-inline-list__item` however will displace the dot.
+As with any inline/inline-block element, any amount of white space (one or more spaces, tabs, etc.) in the markup is rendered as a single white space character. Vanilla uses the <a href="https://css-tricks.com/fighting-the-space-between-inline-block-elements/" target="_blank">font size method</a> to eliminate white space within a `p-inline-list--middot`, that is not inside a `p-inline-list__item`. White space within the `p-inline-list__item` however will displace the dot.
 
 <a href="/examples/patterns/lists/lists-mid-dot/"
     class="js-example">
--- a/docs/patterns/lists.md
+++ b/docs/patterns/lists.md
@@ -76,7 +76,7 @@ View example of the inline list pattern
 Apply the class `.p-inline-list--middot` to add a middot character between
 inline list items.
 
-As with any inline/inline-block element, any amount of white space (one or more spaces, tabs, etc.) in the markup is rendered as a single white space character. Vanilla uses the <a href="https://css-tricks.com/fighting-the-space-between-inline-block-elements/#article-header-id-3">font size method</a> to elliminate white space within a `p-inline-list--middot`, that is not inside a `p-inline-list__item`. White space within the `p-inline-list__item` however will displace the dot.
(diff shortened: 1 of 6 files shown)


lyubomir-popov:
Adding label do not merge as I need to add something to the documentation.

docs/patterns/lists.md
```diff
@@ -76,6 +76,8 @@ View example of the inline list pattern
 Apply the class `.p-inline-list--middot` to add a middot character between
 inline list items.
 
+As with any inline/inline-block element, any amount of white space (one or more spaces, tabs, etc.) in the markup is rendered as a single white space character. Vanilla uses the <a href="https://css-tricks.com/fighting-the-space-between-inline-block-elements/#article-header-id-3">font size method</a> to eliminate white space within a `p-inline-list--middot`, that is not inside a `p-inline-list__item`. White space within the `p-inline-list__item` however will displace the dot.
```
bartaz: You refer to CSS Tricks article by linking to "Set the font size to zero" section, while it seems what is actually use is [Negative margin](https://css-tricks.com/fighting-the-space-between-inline-block-elements/#article-header-id-1).

Also, a markdown notation to link could be used instead of HTML, unless there is some reason to use HTML here in the docs?
`[negative margin](https://css-tricks.com/fighting-the-space-between-inline-block-elements/#article-header-id-1)`.

bartaz:
Negative margin makes list elements stick out when wrapping:
Screenshot 2019-12-03 at 12 32 49

bartaz:
This pattern worked just fine for years, it's still fine in 2.3.0 that we use on snapcraft.io.
Example 2.3.0: https://jsfiddle.net/bartaz/0st1wjyg/

Vanilla 2.4.0 introduced some changes that caused a regression.
Why can't we just revert them back to 2.3.0 state?

lyubomir-popov:
Thanks for the thorough QA @bartaz, I'll think of a way to fix these. 

For the record, centering the dot in a way that is agnostic of the use of white space in the markup has always been problematic, as the `<li>`'s have display:inline. It wasn't visible in screenshots, because the dot was centered for the case where the markup is indented on multiple rows, i.e.:

```
<ul class="p-inline-list--middot">
    Legal information
  </li>
    Data privacy
  </li>
    Report a bug on this site
  </li>
</ul>
```

If you checkout v2.3.0 and change the makrup to this, the dot goes off-center.

```<ul class="p-inline-list--middot">
  <li class="p-inline-list__item">Legal information</li>
  <li class="p-inline-list__item">Data privacy</li>
  <li class="p-inline-list__item">Report a bug on this site</li>
</ul>
```

Here's a screenshot of 2.3.0 when :
image

I did briefly explore making the ul flex, which would solve this, but it leads to other unexpected issues like this:

image

lyubomir-popov:
I've pushed un update:

- remove the negative margin, subtract the same value from padding-right
- update the documentation to link to the beginning of the css tricks article, as the anchor doesn't work properly

docs/patterns/lists.md
lyubomir-popov: didn't know about the markup notation, I'll update, thanks

lyubomir-popov: @bartaz ok tried markdown, but how do you specify target="_blank"?

bartaz:
There still seems to be some whitespace dependancy, for example the MAAS template in the example has still misaligned dots:

Screenshot 2019-12-04 at 10 13 33

It's because of spacing inside `<li>`:

Screenshot 2019-12-04 at 10 13 55

Middot example is fine, because it doesn't use whitespace inside `<li>`:

Screenshot 2019-12-04 at 10 12 55
Screenshot 2019-12-04 at 10 14 19


So this change to the pattern doesn't really solve the issue with whitespace - it just changes it.
Previous version (2.3.0) required whitespace to make the dot aligned - and our sites are already having markup that supports that.
This version of pattern requires to have no whitespace in the list items, so still is a breaking change for existing markup.

If we can't fix the pattern to make it independent from the whitespace in HTML, do we need to change it at all?
Does this version solve anything that 2.3.0 didn't (apart from whitespace which is still an issue)?

kwm14:
> If we can't fix the pattern to make it independent from the whitespace in HTML, do we need to change it at all?

> Does this version solve anything that 2.3.0 didn't (apart from whitespace which is still an issue)?

Agreed if this new proposal doesn't fix the component, and in v2.3.0 it wasn't broken. I think we should revert back to how we had it 👍

lyubomir-popov:
Fair enough. Closing this and I've opened a [new pr]( #2687) which reverts the breaking change in 2.4.

There are still a few issues with the old implementation, like hardcoded values:

These hardcoded values cause a couple of issues:

- when the rem value changes at bigger screens, the .25rem is no longer precise - this is because the strut value of the empty space incerted between inline elements does not scale with font size

- the hardcoded values also cause the spacing between inline list items to be ~5px wider than between other similar elements, but I guess this should be resolved separately:
image

bartaz:
@lyubomir-popov 
> There are still a few issues with the old implementation

Yes, there are issues with this pattern code (like hardcoded values) and behaviour (when wrapping in extreme small widths, etc).
The purpose of this immediate fix was to remove the regression bug introduced in 2.4.0 (that prevented for example snapcraft from updating).

Once we have this fixed in #2687, we can separately update and fix the problems with this pattern (while hopefully keeping it backwards compatible).

PR #2683, opened 2019-12-02
Remove Ubuntu and Canonical social icons

kwm14:
## Done

- Removed Ubuntu and Canonical icons from the social section, as they aren't actually social icons so aren't relevant in this section. If we were to drive users to the social content of Ubuntu or Canonical it would be via Facebook, Twitter, etc...

## QA
- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/patterns/icons/#social and http://0.0.0.0:8101/examples/patterns/icons/icons-social/ 
- You should see that Ubuntu and Canonical are no longer visible

## Details

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/2574

## Screenshots

Screenshot 2019-12-02 at 16 15 59


anthonydillon:
This will be a breaking change for sites using canonical and Ubuntu icons.

bartaz:
@anthonydillon 

> This will be a breaking change for sites using canonical and Ubuntu icons.

Exactly, I wanted to ask about that.
Do we know who may be using this? It seems that these icons are there in Vanilla for a while.
How do we keep track of changes like this to add them to release notes?

Should we first deprecate it (hide from docs, but keep in code)?
And remove when major version is released?

anthonydillon:
You can search GitHub for example: https://github.com/search?q=org%3Acanonical-web-and-design+p-icon--canonical&type=Code

I would suggest these icons should be deprecated for now.

scss/_patterns_icons.scss
```diff
@@ -32,8 +32,6 @@ $social-icon-size: map-get($icon-sizes, heading-icon--small);
   @include vf-p-icon-instagram;
   @include vf-p-icon-linkedin;
   @include vf-p-icon-youtube;
-  @include vf-p-icon-canonical;
```
bartaz: We need to keep `@include vf-p-icon-canonical; @include vf-p-icon-ubuntu;` to make sure these icons are still part of Vanilla for backwards compatibility.

bartaz:
While adding `deprecate` to SCSS makes sure we get warnings when building Vanilla, we should probably also make it visible in docs that these classes are deprecated?

If someone is using vanilla directly from CSS (not build it on their own), they may not notice the deprecation from the code. It should also be reflected in the docs.

Unless we treat removing them from the docs a direct indication that they shouldn't be used?

scss/_patterns_icons.scss
kwm14: Ah yes. Will add back in 👍

kwm14:
> While adding `deprecate` to SCSS makes sure we get warnings when building Vanilla, we should probably also make it visible in docs that these classes are deprecated?

Good idea. I can add them to our component status table - https://docs.vanillaframework.io/component-status#current-status

kwm14:
@bartaz added <code>depreacted</code> label to social icons heading and also added to our current status table.

- Social icons - https://vanilla-framework-canonical-web-and-design-pr-2683.run.demo.haus/patterns/icons/#social
- Component status - https://vanilla-framework-canonical-web-and-design-pr-2683.run.demo.haus/component-status/#current-status

bartaz:
@kwm14 Thanks.

I think that deprecation label in social icon docs looks a bit misleading - like the whole set of social icons is deprecated:

Screenshot 2019-12-03 at 12 48 47

Instead of showing "Deprecated" label right after the title, could we just add a note after the icons example that would say something like "Deprecated: `p-icon-canonical` and `p-icon-ubuntu` are deprecated and will be removed in future version of Vanilla" - similar to component status table?

kwm14:
> @kwm14 Thanks.
> 
> I think that deprecation label in social icon docs looks a bit misleading - like the whole set of social icons is deprecated:

Valid point 😉 
> 
> Instead of showing "Deprecated" label right after the title, could we just add a note after the icons example that would say something like "Deprecated: `p-icon-canonical` and `p-icon-ubuntu` are deprecated and will be removed in future version of Vanilla" - similar to component status table?

Will update 👍

docs/patterns/icons.md
```diff
@@ -174,23 +174,19 @@ Our social icons are used to drive users to social content.
...
-      <div class="p-card col-4 u-vertically-center" style="display:flex; align-items:center;">
-        <i class="p-icon--rss" style="margin-right: 1rem; flex-shrink: 0;"></i>p-icon--rss
-      </div>
       <div class="p-card col-4 u-vertically-center" style="display:flex; align-items:center;">
         <i class="p-icon--email" style="margin-right: 1rem; flex-shrink: 0;"></i>p-icon--email
       </div>
     </div>
     </div>
 
+Removing <code>p-icon--canonical</code> and <code>p-icon--ubuntu</code> from our social icon set, they will no longer be available from our future release v3.0. <span class="p-label--deprecated">Deprecated</span>
```
bartaz: I'd suggest rewording this a little bit:

```suggestion
<span class="p-label--deprecated">Deprecated</span> We will be removing <code>p-icon--canonical</code> and <code>p-icon--ubuntu</code> from our social icon set, they will no longer be available from our future release v3.0. 
```

kwm14: I actually had the deprecate label before but then moved it to the end 🤣

bartaz: We need a pattern for that 🤣

bartaz:
@kwm14 It doesn't seem related to this PR, but `p-icon-rss` and `p-icon-email` are empty in the docs, is this expected?

Screenshot 2019-12-04 at 09 53 41

kwm14:
> @kwm14 It doesn't seem related to this PR, but `p-icon-rss` and `p-icon-email` are empty in the docs, is this expected?
> 
> Screenshot 2019-12-04 at 09 53 41

Yes, it's from the previous PR. New icons won't show it the Docs as it's not running on the latest version

kwm14:
> > @kwm14 It doesn't seem related to this PR, but `p-icon-rss` and `p-icon-email` are empty in the docs, is this expected?
> > Screenshot 2019-12-04 at 09 53 41
> 
> Yes, it's from the previous PR. New icons won't show it the Docs as it's not running on the latest version 👎

PR #2684, opened 2019-12-03
Blue buttons outlines test

lyubomir-popov:
## Done

## QA

- Pull code
- Run `./run serve --watch`
- [Add additional steps]

## Details

## Screenshots

### DIFF
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -77,14 +77,14 @@
 /// override colors as desired
 @mixin vf-button-pattern(
   $button-background-color: $color-x-light,
-  $button-text-color: $color-x-dark,
+  $button-text-color: $colors--light-theme--text-link,
   $button-disabled-background-color: $color-transparent,
-  $button-disabled-border-color: $color-mid,
-  $button-border-color: $color-mid,
-  $button-hover-background-color: darken($color-x-light, 10%),
-  $button-hover-border-color: $color-mid,
-  $button-active-background-color: darken($color-x-light, 15%),
-  $button-active-border-color: $color-mid
+  $button-disabled-border-color: opacify($colors--light-theme--text-link, 0.5),
+  $button-border-color: $colors--light-theme--text-link,
+  $button-hover-background-color: transparentize($colors--light-theme--text-link, 0.95),
+  $button-hover-border-color: $colors--light-theme--text-link,
+  $button-active-background-color: transparentize($colors--light-theme--text-link, 0.9),
+  $button-active-border-color: $colors--light-theme--text-link
 ) {
   background-color: $button-background-color;
   border-color: $button-border-color;
--- a/scss/_patterns_buttons.scss
+++ b/scss/_patterns_buttons.scss
@@ -32,15 +32,6 @@
 @mixin vf-button-neutral {
   %p-button--neutral {
     @extend %vf-button-base;
-
-    @include vf-button-pattern(
-      $button-disabled-border-color: $color-mid,
-      $button-border-color: $color-mid,
-      $button-hover-background-color: darken($color-x-light, 10%),
-      $button-hover-border-color: $color-mid,
-      $button-active-background-color: darken($color-x-light, 15%),
-      $button-active-border-color: $color-mid
-    );
   }
 
   .p-button--neutral {
@@ -121,11 +112,9 @@
     @include vf-button-pattern(
       $button-background-color: $color-transparent,
       $button-border-color: $color-transparent,
-      $button-text-color: $color-x-dark,
-      $button-hover-background-color: darken($color-x-light, 10%),
       $button-hover-border-color: $color-transparent,
-      $button-active-background-color: darken($color-light, 15%),
-      $button-active-border-color: $color-transparent
+      $button-active-border-color: $color-transparent,
+      $button-disabled-border-color: $color-transparent
     );
   }
 
--- a/scss/_settings_colors.scss
+++ b/scss/_settings_colors.scss
@@ -42,6 +42,7 @@ $colors--light-theme--border-high-contrast: #999 !default;
 $colors--light-theme--text-hover: #757575 !default;
 $colors--light-theme--text-disabled: #666 !default;
 $colors--light-theme--text-default: #111 !default;
+$colors--light-theme--text-link: $color-link !default;
 
 //text-hover: minimum contrast on primary that satisfies contrast checker AA
 $colors--dark-theme--background: hsl(0, 0%, 15%) !default;


cassiocassio:
Can we write some guidance for how and why to use a blue-button?

e.g. 

- when a secondary Call to Action
- when next to other blue links
- when too many green Primary Calls to Action would be heavy
- when two different CTAs are equally valid options

when not to use?

e.g.
- as part of a dense field set that is not blue
- 

Can we propose a (semantic)  name for the blue button which describes it's meaning and purpose, rather than it's purpose?

matthewpaulthomas:
[Before](https://docs.vanillaframework.io/patterns/buttons/#base): One of these three bits of black text is a button, and the other two are not.

> image

[After](https://vanilla-framework-canonical-web-and-design-pr-2684.run.demo.haus/patterns/buttons/#base): One of these four bits of blue text is a button, and the other three are not.

> image

Separately, if you go ahead with this, I suggest drawing a set diagram of which controls should be blue and which should not be. The current colouring implies that neutral buttons and menubuttons behave more similarly to each other, since they’re both blue,

image image

while Copy buttons and radio menus (`<select>`) behave more similarly to each other, since they’re both grey.

image image

I think the opposite is true: Copy buttons are much more similar in behaviour to neutral buttons, while menubuttons are much more similar in behaviour to radio menus.

lyubomir-popov:
From the conversation in [Design review](https://github.com/canonical-web-and-design/meeting-notes/issues/468), I believe we tentatively agreed this should be a vanilla setting which changes the appearance of the neutral button form grey to blue, rather than a new flavour of button. This way, it can be turned on for brochure sites, but off for applications like maas and jaas for example. 

The set diagram is a good idea, I'll make one.

If the option is turned on, we still need to draw the line somewhere and decide what else (if anything) should turn blue. The general guidance that Vanilla follows seems to be: If an elements can be identified as clickable from its context (placement on screen, purpose, being part of a set, etc.), then no need to make it blue., e.g. tabstrip, menu, accordion. This would apply to selects, buttons embedded within inputs (like the code copyable button), etc.

If that logic sounds plausible, then all the blue neutral button is doing is raising the prominence of the neutral button compared to anything else. This way, it makes the neutral button closer in prominence to the primary button, and a good candidate for scenarios where one might use multiple primary buttons.

I'll clean up the pr and update you once it is ready for another look.

bartaz:
> If that logic sounds plausible, then all the blue neutral button is doing is raising the prominence of the neutral button compared to anything else. This way, it makes the neutral button closer in prominence to the primary button, and a good candidate for scenarios where one might use multiple primary buttons.

The name of 'neutral' button suggests it was not intended to have a prominence compared to primary button. We probably can verify its use on our sites, but I thing it's probably quite commonly used as a 'secondary' button, either next to other primary button, or simply as an action that doesn't need to draw users attention. 

Making the neutral button more prominent seems to be opposite to its original purpose.

cassiocassio:
I think we discussed a range of scenarios and contexts where the standard
appearance of the neutral button was weak:

e.g.
- a pair of call to actions where making them both green was too heavy
- a call to action that meant to stand out, *when used together* with set
of other blue links - otherwise the blue links were very strong and the
button rather weak

how important is the blue button?
- semantically, we call the green button 'primary' because it's the most
important button, and I don't think this blue button is meant to compete
with that

does it supplant the neutral button ?
- no. there's lots of fieldsets where the neutral button needs to appear
alongside a range of radios and inputs and selects, none of which is blue
- application-like situations already have enough going on, without making
the buttons blue...

when do we imagine using this most frequently?
- some editorial situations, where the button really has the meaning of
"strongly emphasised link", but is neither the primary (ie. default) button
of a fieldset, nor is it the "primary call to action" of a hero strip.

does that help decide between "new flavour of neutral button" or "new kind
of button" ?

lyubomir-popov:
@bartaz I can see what you mean about the naming, but it's a matter of relative - not absolute - prominence. As @cassiocassio mentioned, we identified cases where making the button blue creates a smoother hierarchy (content / clickable text-links / regular buttons / primary buttons), a few examples below:

image

@cassiocassio my original intention was for this to be a toggle, because I was conscious of how much work an additional flavour might generate - sorting through thousands of buttons on all websites and deciding which ones should be blue. I don't mind a flavour if that's what everyone prefers.

But I think it's worth considering the following fact - any form is only grey until you interact with it. As soon as you tick a checkbox, it is tinted in the information blue. Click a field - you get a different blue (the focus outline). If you have a terms and conditions link for example, that's a third blue.

If we made these match the link blue, then we have something that feels quite consistent. Current blues on the left, matching blues on the right:

image

On a related note, we often talk how the focus state color came from the platform team, who conducted research and determined it to be accessible. It turns out that by WCAG standards, this is not true. The [requirements](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) for non-text contrast is 3.0, whereas our focus color's contrast against white is 2.34:
image

image

For comparison, the link blue's colour contrast is well above the minimum of 3.0, so the matching blues in the example above would actually improve accessibility.

kwm14:
> On a related note, we often talk about how the focus state color came from the platform team, who conducted research and determined it to be accessible. It turns out that by WCAG standards, this is not true. The [requirement](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) for non-text contrast is 3.0, whereas our focus color's contrast against white is 2.34 (second image below):
> 
Game changer! I wasn't aware that it wasn't accessible... I've done some digging looking for the blue colorwork by the platform team but unfortunately, I couldn't find it.

But I did find this section on a color toolkit slide deck which relays your suggestion, but rather than using our link blue I'd propose using `$color-information` `#335280` blue.

**Semantics values of the SDK colours**
Screenshot 2019-12-17 at 09 57 57

Screenshot 2019-12-17 at 10 03 32

lyubomir-popov:
@kwm14 where do you find these :)

We seem to have moved away from all but the blue colour - we use a darker red, yellow instead of orange, and a grassier shade of green, my guess is because the original ones in the table above weren't entirely satisfactory when put to heavy use. Maybe that would explain why we do not use the blue in the table as originally specified for interaction?

I don't really mind which blue gets chosen, but it would be great it we can stick to one shade per `Semantic value`. This way, a blue button wouldn't feel foreign to a form, as interacting with the form in any way would tint anything you touch in the same blue.

This seems to have been the original intention -  the `Usage` column for the `Awareness` blue includes `selection, interaction` in addition to `focus`.

This can be achieved by sticking to the current link blue, which already satisfies all requirements, or nudging the lightness and saturation on the information blue until it satisfies all WCAG requirements, but also stands next to black body text so links are clearly identifiable. 

A quick test in the colour checker, adjusting lightness and saturation of the $color-information (the Awareness blue from the table above):
image

kwm14:
> @kwm14 where do you find these :)

Deep deep searching in the archives 🤣 

> We seem to have moved away from all but the blue colour - we use a darker red, yellow instead of orange, and a grassier shade of green, my guess is because the original ones in the table above weren't entirely satisfactory when putting to heavy use. Maybe that would explain why we do not use the blue in the table as originally specified for interaction?

Yeah exactly, and also Vanilla moved away from those values as the green, orange and red as they aren't fully accessible.

> This seems to have been the original intention - the `Usage` column for the `Awareness` blue includes `selection, interaction` in addition to `focus`.

👍 I'm easy on which color we chose just as long as we're not introducing a new blue...

> A quick test in the colour checker, adjusting lightness and saturation of the $color-information (the Awareness blue from the table above):

Getting confused by the 50 shades of blue now? 🤔 
- Are you suggesting a new blue value?
- Updating `$color-information` to a new shade?
- Or are you suggesting both `$color-link` and `$color-information` are deprecated and we use this new blue for both use-cases?

lyubomir-popov:
> 👍 I'm easy on which color we chose just as long as we're not introducing a new blue...

The link blue ticks that requirement, along with all others, and I favour that choice. No "new blue" discussion needed.

> * Are you suggesting a new blue value?
> * Updating `$color-information` to a new shade?

Only if you're not happy with the link blue for focus, then I was suggesting to work with the information blue until we make it "compliant".

kwm14:
> Only if you're not happy with the link blue for focus, then I was suggesting to work with the information blue until we make it "compliant".

Ah, my confusion! The new blue proposed it to replace the existing non-accessible one. Happy with the new blue if this is the case 👍 

Current focus blue| Proposed focus blue
------------ | -------------
`#19b6ee` | `#0a52bf`
Screenshot 2019-12-17 at 12 11 37 | Screenshot 2019-12-17 at 12 11 50

lyubomir-popov:
IMG_7940

Photo from our discussion @cassiocassio @kwm14

Next steps:
- Agreed to make replace the focus blue and the information blue (used inside checkboxes etc) with the current link blue, reducing the set of blues form 3 to 1
- Pick a set of highly complex forms pages from brochure sites, and test the blue buttons on them
- Decide on whether to proceed with the blue buttons proposals based on the outcome of the forms pages test

matthewpaulthomas:
> This way, it can be turned on for brochure sites, but off for applications like maas and jaas for example.

Is there any evidence that it’s a good idea to distinguish brochure sites and applications in the first place? I think the reverse is true — it’s better both for credibility, and for user flow, if they are tightly integrated — which is why I integrated them on ubuntu.com/advantage.

> I think we discussed a range of scenarios and contexts where the standard appearance of the neutral button was weak

The standard appearance is weak whenever it’s on a white background, because it’s white with no bevel, so it appears as a “[ghost button](https://designmodo.com/ghost-buttons/)”, which usually performs worse than a solid colour (A/B test evidence [1](https://cxl.com/blog/ghost-buttons/), [2](https://www.elevatedthird.com/news-insights/ghost-buttons-good-bad-spooky-part-2), [3](http://togetherincredible.com/should-ghost-buttons-be-busted/)). Changing from monochrome ghost buttons to blue ghost buttons [might help](https://www.kaylaheffernan.com/blog/2017/10/15/ghost-buttonsnot-as-bad-as-wethought), but probably not as much as a solid colour would.

> - some editorial situations, where the button really has the meaning of "strongly emphasized link", but is neither the primary (ie. default) button of a fieldset, nor is it the "primary call to action" of a hero strip.

I do think it’s unfortunate that we use button styles for things that are really just links, and it would be nice if we could find a CTA link style that had similar/better click rates. But many of the places we do that are primary buttons, not just neutral buttons.

> [ Suscribe now ]

(That … that can’t be real, can it? Cripes, it is real. Submitted [charmed-osm.com#23](https://github.com/canonical-web-and-design/charmed-osm.com/pull/23).)

> As soon as you tick a checkbox, it is tinted in the information blue. Click a field - you get a different blue (the focus outline). If you have a terms and conditions link for example, that's a third blue. If we made these match the link blue, then we have something that feels quite consistent.

> But I did find this section on a color toolkit slide deck which relays your suggestion

It was a mistake to conflate “Focus, selection, interaction” into one thing. They are very different! If you used the same blue for focus and selection, it would be more difficult to see whether a checked checkbox, checked radio button, switch, or link was focused; and it would be more difficult to scan a complex form to see what is currently focused.

anthonydillon:
I must say I absolutely agree with @matthewpaulthomas here. 

The neutral button is meant to be subtle and by making it more prominate it will be less clear which one is the main CTA. Why can't we use a link instead of creating a new "link styled" button? The more buttons we have the more inconsistences we will end up with across the websites and applications.

"Brochure" sites and applications should not be styled differently. We have a set of UI patterns/components which work in all situations and should be extended together to improve them both at the same time. Otherwise, we will end up forking the framework to create an application design system and the "other one".

> (That … that can’t be real, can it? Cripes, it is real. Submitted charmed-osm.com#23.)

Sorry about that, merged.

The focus blue is different from all other blues as it's usage needs to stand out in all situations.

spencerbygraves:
I originally thought the blue button was a good idea, but looking back at the design review conversation, the blue button was floated as an idea to help differentiate buttons from inputs. We then agreed in the same meeting to remove the corner-radius from inputs and increase the corner-radius from 2px to 4px on buttons, to achieve the same aim.

So the blue isn't needed to solve that issue.

If I remember correctly, when we designed these buttons these were the general rules (which may not be perfect):

- Ideally only one green button per page
- When another button is needed e.g. "Cancel", use the neutral button next to the primary one
- If a primary button is too strong and you need two buttons, use neutral and base
- If you need 3, use primary, neutral and base
- Too strong?, use neutral, base, base
- If you need 4, maybe you're doing it wrong 😃 

So maybe we don't need an extra style?

And @matthewpaulthomas raises some good points I think...

>Is there any evidence that it’s a good idea to distinguish brochure sites and applications in the first place? I think the reverse is true — it’s better both for credibility, and for user flow, if they are tightly integrated

>It was a mistake to conflate “Focus, selection, interaction” into one thing. They are very different! If you used the same blue for focus and selection, it would be more difficult to see whether a checked checkbox, checked radio button, switch, or link was focused; and it would be more difficult to scan a complex form to see what is currently focused.

This discussion has become quite long. Shall we sit down in together in the new year to make a decision?

Thanks.

lyubomir-popov:
Happy to drop this as there are strong arguments against. As @spencerbygraves said, we resolved the original issue in a different manner.

bartaz:
@lyubomir-popov Can this be closed, or is there still a discussion around it?

lyubomir-popov:
@bartaz there's still a discussion, to be picked up in the next design review meeting this Thursday. I will close once we're done discussing.

cristinadresch:
Closing but still discoverable.

PR #2702, opened 2019-12-10
POC: replace example-js with codepen embeds

bartaz:
## Done

Proof of concept to explore replacing examples with docs with CodePen embeds.

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/ or demo https://vanilla-framework-canonical-web-and-design-pr-2702.run.demo.haus/
- All examples in the docs should be replaced with equivalent CodePen embeds

## Screenshots
Screenshot 2019-12-11 at 11 41 57

Screenshot 2019-12-11 at 12 03 20

### Pros of using CodePen

- we can simplify example-js as we won't need to build iframe and code views (especially when we want to include JS tab, maybe React in future)
  - simplified example-js is still needed, because we need to fetch and parse example HTML and send it to codepen embed
- we can easily add JS code tab without implementing it ourselves
  - possibly in future we can use CodePen for React examples
- codepen is known and respected tool in webdev community
- free 'Open in CodePen' button that makes it super easy to open any example in CodePen and change/update it to play with vanilla patterns
  - would also make super easy to create small test case examples for issues reporting
- built in syntax highlighting for code
- built in support for preprocessors for HTML (like HAML etc) CSS (SCSS, etc) JS (Babel) 
- build in x1,x0.5,x0.25 scaling option, to make it easier to test responsive components

### Cons of CodePen (vs custom solution)

- we can't automatically adjust height of the embed like we do with iframe
  - we can define the height of the embed we want
  - if we want per-example adjustment, we would need to store height somewhere along the example (some metadata) and use it when creating the embed
- we have limited control on styling (some theming possible in terms of adjusting colors)
  - see below for theming details
- we have limited control on layout of the embed - for example code (HTML/JS) can be only on a side of 'Result', we can't show code under the rendered example like we do currently
- we rely on external resource to render our examples
  - because examples are rendered on codepen it may take longer for them to load
- no 'copy to clipboard' button
- need to inject some margin on body to prevent examples from sticking to the borders

## Themeimg

There are 2 built in themes dark and light:

Screenshot 2019-12-11 at 12 10 26

Screenshot 2019-12-11 at 12 09 32

If we want more adjustment having an account on CodePen allows to create one custom theme (we would probably want to have vanilla account in such case):

Screenshot 2019-12-11 at 12 14 01

Theming options are quite limited, they mostly allow to choose colors for different parts of the UI (inlcuding choosing one of preset syntax highlighting themes).
PRO (paid) users have more control with custom CSS.

Screenshot 2019-12-11 at 12 10 55

## Docs

https://blog.codepen.io/documentation/prefill-embeds/
https://blog.codepen.io/documentation/pro-features/unlimited-embed-themes/


kwm14:
@bartaz examples looking good with CodePen 🔥 

Is the height pre-defined or can make it automatic so it scales depending on the height of the component?

Screenshot 2019-12-11 at 10 19 34

bartaz:
@kwm14 

> Is the height pre-defined or can make it automatic so it scales depending on the height of the component?

We can't make it fully automatic. We don't control the height of the embed like we can with our own iframe.
We can only predefine height of the embed. This can be the same height for every example, on we can try to keep some height metadata with the example, to extract it when we create the embed.
But this means manually defining height for the examples.

kwm14:
> We can't make it fully automatic. We don't control the height of the embed like we can with our own iframe.

Damnnnn...
> We can only predefine the height of the embed. This can be the same height for every example, we can try to keep some height metadata with the example, to extract it when we create the embed.
> But this means manually defining a height for the examples.

Ah ok, that would be difficult to define a height for all the examples, I think we have that problem now trying to identify one value that works across the board. 

Happy to leave as is then. Love the HTML/JS toggles 👍

deadlight:
I love it. It's so much more maintainable than developing an in-house version. Especially when the brief is "do what codepen does".

barrymcgee:
Love it also!

anthonydillon:
I'm in favour! Good POC @bartaz

bartaz:
We consider POC done, we will continue with migration to CodePen in https://github.com/canonical-web-and-design/vanilla-squad/issues/678

PR #2705, opened 2019-12-13
Use Codepen for Vanilla docs examples

bartaz:
Fixes https://github.com/canonical-web-and-design/example-js/issues/51

## Done

- removed references to previous example-js implementation
- added script to replace example links with CodePen embeds
- cleaned up js-example links markup in docs

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/ or https://vanilla-framework-canonical-web-and-design-pr-2705.run.demo.haus/
- Browse documentation (patterns, utilities), see examples rendered via CodePen
- Make sure examples render correctly

## Details

We are using default dark CodePen theme.

CodePen embed uses hotlinked latest published version of Vanilla CSS.
Additionally has small CSS linked to add 1rem margin around body, to avoid examples sticking to embed borders.

All embeds have default height of 400px (as it may be more than needed for smaller examples, the embed needs to fit the code as well).
It's possible to change the height of individual examples via `data-height` property on the example link (see [Embedded media utility](https://vanilla-framework-canonical-web-and-design-pr-2705.run.demo.haus/utilities/embedded-media/))

Because CodePen embeds rely on external script provided by CodePen if that would fail we will fallback to displaying code of the example in standard Vanilla code blocks.

## Screenshots

Example pattern:
Screenshot 2019-12-13 at 14 11 16

HTML code open with result:
Screenshot 2019-12-13 at 14 11 25

JS open:
Screenshot 2019-12-13 at 14 11 32


docs/js/example.js
```diff
@@ -1,14 +1,125 @@
-var x = new XMLHttpRequest();
+/* global VANILLA_VERSION */
+(function() {
+  if (!window.VANILLA_VERSION) {
+    throw Error('VANILLA_VERSION not specified.');
```
bartaz: @anthonydillon One of the not ideal solutions used is the need for latest Vanilla version. I pass it via global variable (which seems the easiest and most cross browser friendly).

Optionally we could:
-  just hardcode it in URL in the script, but that would be one more place to update on releases
- pass it via attribute on `<script src="example.js" data-version="2.5.0">` and access via `document.currentScript` - but that would require polyfill for older browsers

Any other ideas?

docs/js/example.js
```diff
@@ -1,14 +1,125 @@
...
+  }
+
+  var CODEPEN_PREFILL_CONFIG = {
+    title: 'Vanilla example',
+    head: "<meta name='viewport' content='width=device-width, initial-scale=1'>",
+    stylesheets: [
+      // link to latest Vanilla CSS
+      'https://assets.ubuntu.com/v1/vanilla-framework-version-' + VANILLA_VERSION + '.min.css',
+      // link to example stylesheet (to set margin on body)
+      'https://assets.ubuntu.com/v1/4653d9ba-example.css'
```
bartaz: @anthonydillon Second tricky part was to pass custom CSS to embed to add `body {margin:1rem}`. I just created a file with this one rule and serve it from assets.

Other possibilities:
- have a utility class in Vanilla that does that, as CodePen allows passing class names to `html` element, but probably this utility would be very specific to Vanilla examples
- instead of passing it as stylesheet, pass `<style>` to head of CodePen (as in POC), but because head params are rendered before stylesheets it requires `!important`. And head params in CodePen are not really supposed to be used for styling.
- add CSS to CodePen, but this would add a CSS tab to example embeds with could be unnecessarily confusing probably

If you want to discuss other possibilities here are the docs of prefilled embeds we use here:
https://blog.codepen.io/documentation/prefill-embeds/

kwm14:
Yaaaaaas @bartaz implementation looking really good 👍 LGTM

kwm14:
Only one thing to note. Whether to do in this PR or one after is to remove the temporary JS notifications we put on docs pages, as we now include the CodePen snippet:

- /base/code/#copyable
- base/tables/#sortable
- base/tables/#expanding
- patterns/accordion/
- patterns/contextual-menu/
- patterns/list-tree/
- patterns/modal/
- patterns/navigation/#sub-navigation

docs/js/example.js
anthonydillon: I think this solution is better than the two hardcoded versions. The less we have to change per release the better.

docs/js/example.js
```diff
@@ -1,14 +1,125 @@
-var x = new XMLHttpRequest();
+/* global VANILLA_VERSION */
+(function() {
+  if (!window.VANILLA_VERSION) {
+    throw Error('VANILLA_VERSION not specified.');
+  }
+
+  var CODEPEN_PREFILL_CONFIG = {
+    title: 'Vanilla example',
```
anthonydillon: "Vanilla framework example" would be better because ... Vanilla is a bad name

docs/js/example.js
```diff
@@ -1,14 +1,125 @@
...
+    head: "<meta name='viewport' content='width=device-width, initial-scale=1'>",
+    stylesheets: [
+      // link to latest Vanilla CSS
+      'https://assets.ubuntu.com/v1/vanilla-framework-version-' + VANILLA_VERSION + '.min.css',
```
anthonydillon: This would mean there was no way to check the docs locally with local examples and framework?

docs/js/example.js
anthonydillon: This is ok as it gets out of the way of users. The only downfall is it is not clear to other developers what that is for. As the example does not display the body tag could you add the margin as an inline style?

docs/js/example.js
bartaz: This applies to CodePen embeds only, so yes, inside the docs the CodePen examples will always use latest released version of Vanilla, not the local one.

The example pages (https://vanilla-framework-canonical-web-and-design-pr-2705.run.demo.haus/examples/) use the /build.css as before, so local changes can still be viewed on them, Percy makes screenshots of them, etc.

So the only "downside" is that the examples embedded in docs use released version, not development one.
This on the other hand may be seen as an advantage, thanks to that we can deploy the docs site in between Vanilla releases, while being sure embedded examples are not affected by any local changes to SCSS. (But of course the example pages themselves would be still affected).

We probably could try to pass current /build.css instead of released version.
But, there may be a problem with local development, as passing localhost URL may have some issues when embedded in CodePen.

docs/js/example.js
bartaz: @anthonydillon  Add the margin as inline style on what? I'm not sure what you mean.

Body tag is generated by CodePen and we have no access to it at all. So we can't add inline style to that.
CodePen API allows to add some class names to `<html>` tag, but we would still need a CSS that would define those styles.
The other way I managed to inject the margin (in POC) is adding `<style>body { margin: 1rem !important }</style> to the `<head>` (which CodePen API allows), but because it is injected before the stylesheets it needs `important` and it's a bit of a hack on CodePen API (as it's intended for some meta data, not style injections).

Issue #2696, opened 2019-12-06
Navigation has a mid breakpoint stylinf failure


anthonydillon:
At 772px width the navigation wraps before snapping beween the small screen and large screen views.
![Screenshot_2019-12-06 Navigation Default Vanilla framework documentation](https://user-images.githubusercontent.com/1413534/70363702-5c809580-1881-11ea-8dc1-049cacf0100e.png)

PR #2712, opened 2020-01-06
Adjust breakpoints to fix midbreakpoint failure (#2696)

lyubomir-popov:
## Done
Subtract a pixel from nav breakpoints that are not overriden by a min-width query further down.

## QA
- Pull code
- Run `./run serve --watch`
- Open a navigation example, e.g. examples/patterns/navigation/default-dark/
- Resize window to 772px, verify nav displays like this:
image

## Details
Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/2696

### DIFF
--- a/scss/_patterns_navigation.scss
+++ b/scss/_patterns_navigation.scss
@@ -57,12 +57,7 @@ $navigation-hover-opacity: 0.58;
           top: 0;
 
           @media (min-width: $breakpoint-navigation-threshold) {
-            content: initial;
-            height: initial;
-            left: initial;
-            position: initial;
-            right: initial;
-            top: initial;
+            content: none;
           }
         }
 
--- a/scss/_patterns_navigation.scss
+++ b/scss/_patterns_navigation.scss
@@ -57,12 +57,12 @@ $navigation-hover-opacity: 0.58;
           top: 0;
 
           @media (min-width: $breakpoint-navigation-threshold) {
-            content: unset;
-            height: unset;
-            left: unset;
-            position: unset;
-            right: unset;
-            top: unset;
+            content: initial;
+            height: initial;
+            left: initial;
+            position: initial;
+            right: initial;
+            top: initial;
           }
         }
 
@@ -123,7 +123,7 @@ $navigation-hover-opacity: 0.58;
         flex: 1 1 auto;
         margin: $input-gap-top 0 auto auto;
         max-width: 20rem;
-        min-width: unset;
+        min-width: initial;
         order: 1;
       }
     }
@@ -207,7 +207,7 @@ $navigation-hover-opacity: 0.58;
       // XXX: patch - override utility, as it positions absolutely, which leads to overlap on smaller screens.
       // Should use flex instead; But that will require changing the markup
       position: relative;
-      right: unset;
+      right: initial;
     }
   }
 
--- a/scss/_patterns_navigation.scss
+++ b/scss/_patterns_navigation.scss
@@ -81,9 +81,9 @@ $navigation-hover-opacity: 0.58;
       padding: 0;
 
       @media (min-width: $breakpoint-navigation-threshold) {
-        margin-top: 0; // prevents bottom border of nav from moving 1px
         display: flex;
         flex-wrap: wrap;
+        margin-top: 0; // prevents bottom border of nav from moving 1px
       }
     }
 
@@ -272,10 +272,8 @@ $navigation-hover-opacity: 0.58;
   }
 
   .p-navigation__link > a::before {
-    @media (max-width: #{$breakpoint-navigation-threshold - 1px}) {
-      // separator color on small screens
-      background: $colors--light-theme--border-default;
-    }
+    // separator color on small screens
+    background: $colors--light-theme--border-default;
   }
 }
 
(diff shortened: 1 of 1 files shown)


bartaz:
Please rebase from master, this should fix travis build.

lyubomir-popov:
Rebased.

bartaz:
Using both `mix-width` and `max-width` breakpoints is a bit confusing and error prone. Devs need to know and remember to substract 1px for min-width breakpoints.

Would it be possible to rewrite the breakpoints to use only one type (min- or max-width)?
If we want to try be "mobile first" we should use only `min-width` to change styles on larger screens.
For now in Vanilla it seems that we are more "desktop first", and apply some mobile specific styles in `max-width` breakpoints. In such case we would use `max-width` only.

lyubomir-popov:
Changing to only use min or max width is a bigger change that probably needs discussion as it affects hundreds of queries.

Currently, the strategy is to have the smallest amount of overlap of rules:
- mobile rules wrapped in max-width queries 
- tablet queries using min width and max width
- desktop queries using min width

This way things overlap at the specific pixel of the query.

That still leaves these ways of doing the current fix:
1) keep css minimal, and subtract pixel to avoid the 1px overlap mentioned above; This is what I've done in this pr, the shortcomign being you need to remember the -1px if you need to sync any responsive changes with this code.
2) add a min-width query that follows the max-width one, so it redefines the behaviour on bigger screens for that specific pixel.
3) switch strategy and use min-widths only for this case, and in the future revise other breakpoints.

I'd go with as it has the least potential to break things. What do you think?

bartaz:
@lyubomir-popov 

I only meant this one file for that change. I'm aware that changing it everywhere would be huge amount of work.

If that seems too much for that fix, I would at least suggest to create additional local variable inside of this pattern that would hold the value for max breakpoint `$breakpoint-navigation-threshold-max: $breakpoint-navigation-threshold - 1px;`. This would at least allow to avoid repeating the substraction and `#{}` interpolation in many places.

bartaz:
@lyubomir-popov From your suggested approaches (if I understand them well) I would go with 3).

For this pattern (navigation) I would write all the breakpoints as min-width (or, if that's simpler max-width).

For future it would be good to have an approach that we want to keep, so if any pattern that uses media queries will be changed we can adjust them, without making a huge task of rewriting them all at once.

lyubomir-popov:
Fair enough. I'll do 3).

One question though. What would we do when something only needs to exist on mobile, then disappear?
To cleanly remove it (rather than just hide it) you need to unset everything you've set, e.g. for the border under menu items:
```
        &::before {
          content: '';
          height: 1px;
          left: 0;
          position: absolute;
          right: 0;
          top: 0;

          @media (min-width: $breakpoint-navigation-threshold) {
            content: unset;
            height: unset;
            left: unset;
            position: unset;
            right: unset;
            top: unset;
          }
        }
```

lyubomir-popov:
Refactored as requested:
- max-width rules moved outside queries
- properties altered outside of queries unset where needed
- some minor clean up

scss/_patterns_navigation.scss
```diff
@@ -269,7 +272,7 @@ $navigation-hover-opacity: 0.58;
   }
 
   .p-navigation__link > a::before {
-    @media (max-width: $breakpoint-navigation-threshold) {
+    @media (max-width: #{$breakpoint-navigation-threshold - 1px}) {
```
bartaz: Is this media query needed at all? Border is only visible on small screens, so having background being set on all widths wont matter.

scss/_patterns_navigation.scss
```diff
@@ -306,7 +309,7 @@ $navigation-hover-opacity: 0.58;
 
   // separator color on small screens
   .p-navigation__link > a::before {
-    @media (max-width: $breakpoint-navigation-threshold) {
+    @media (max-width: #{$breakpoint-navigation-threshold - 1px}) {
```
bartaz: Is this media query needed at all? Border is only visible on small screens, so having background being set on all widths wont matter.

scss/_patterns_navigation.scss
```diff
@@ -77,12 +77,11 @@ $navigation-hover-opacity: 0.58;
...
       list-style: none;
-      margin: 0;
+      margin: -1px 0 0 0;
       padding: 0;
-      @media (max-width: $breakpoint-navigation-threshold) {
-        margin-top: -1px; // prevents bottom border of nav from moving 1px
-      }
+
       @media (min-width: $breakpoint-navigation-threshold) {
+        margin-top: 0; // prevents bottom border of nav from moving 1px
```
bartaz: Order the properties alphabetically. I think this causes build to fail because of linting errors.

scss/_patterns_navigation.scss
```diff
@@ -64,6 +54,16 @@ $navigation-hover-opacity: 0.58;
           left: 0;
           position: absolute;
           right: 0;
+          top: 0;
+
+          @media (min-width: $breakpoint-navigation-threshold) {
+            content: unset;
```
bartaz: I think in theory having `content: unset` would be enough to make the border not visible even if other properties are set.

Of course, to be sure all styles are reset on desktop your approach is better. That's one of the drawbacks of mobile first styles. If we would ever need to unset/override something more complicated we could probably consider other solution.

lyubomir-popov:
One more change to support (hopefully) ie, but I will need to wait for the demo url to test it.

lyubomir-popov:
Turns out 'content: initial' is wrong, what's needed ids content: none.

Removed the remaining properties properties as they become redundant once content has been set to 'none'.

lyubomir-popov:
@bartaz could you please approve the percy as well, this page keeps turning up in diffs and noone knows why.

PR #2719, opened 2020-01-08
WIP: Encapsulate accordion

bartaz:
## Done

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/
- [Add additional steps]

## Details

## Screenshots


docs/_layouts/examples.html
```diff
@@ -28,6 +33,7 @@
       }
     </style>
 
+    {% unless page.individual %}
```
bartaz: Styling for `u-baseline-grid` utility and the baseline grid toggle (positioning, switch, etc) are included in full vanilla but not in individual pattern CSS, so I don't render the switch on individual pattern example page for now.

We can consider bringing it on this page once we have this utility styles encapsulated, so we can include just these styles here.

scss/individual/base.scss
```diff
@@ -0,0 +1,2 @@
+@import '../base';
```
bartaz: We are not using this file in any example directly, but we can keep it to make sure base styles can be built on their own.

scss/_base_forms.scss
```diff
@@ -1,5 +1,6 @@
 @import 'settings';
 @import 'base_forms-tick-elements';
+@import 'patterns_icons';
```
bartaz: This feels a bit weird. Base styles depend on a "pattern". 
Maybe we should extract icon placeholders and make them part of "base" styles.

docs/_layouts/examples.html
lyubomir-popov: what do you mean by encapsulated? wouldn't it work by just adding:

```
@import 'utilities_baseline-grid';
@include vf-u-baseline-grid;
```

I'm a bit concerned about the baseline grid being seen as a design whim with debatable value.
There are, for example, 2 open issues to remove it.

From a design perspective, it is the only way to notice when something is a couple of pixels off.  For example, the pagination buttons were  ~ 3px taller than other buttons, something that wasn't noticed for months because in the real world you would never put them on the same row as a button. We only caught it because the grid made it obvious.

I realise it's a tall order to ask developers to actively QA for this, but it is something designers should IMO. Otherwise small discrepancies can silently accumulate.

bartaz: @lyubomir-popov Thanks for the comment.

In this example we don't want to add baseline grid styles into accordion pattern CSS, because we want to check if accordion styles work and build on their own.

From your comment it seems that CSS for baseline grid should be quite easy to build separately, so I'll try to do that.

bartaz: I added the baseline toggle on individual example page as well.

It doesn't use `p-switch` styles (because it's a separate pattern that is not included), but apart from that is fully functional.

docs/examples/index.html
```diff
@@ -85,3 +85,22 @@ <h3>Templates</h3>
     </nav>
   </div>
 </div>
+<div class="row">
+  <div class="col-12">
```
anthonydillon: We normally limit text blocks to `col-8` could you change this to help if edited in the future.

docs/examples/index.html
```diff
@@ -85,3 +85,22 @@ <h3>Templates</h3>
...
+    <h3>Examples with individual CSS</h3>
+    <p>Examples below use CSS built with single pattern included instead of whole Vanilla framework CSS file.</p>
+  </div>
+  <div class="col-3">
```
anthonydillon: This should be inside a new `row`

scss/_base_forms.scss
anthonydillon: I agree or see what icon it requires. Could that to moved local to the base_forms?

scss/individual/example.scss
```diff
@@ -0,0 +1,4 @@
+// styles specific to example page
```
anthonydillon: Don't think this comment is required

anthonydillon:
## QA
- The example navigation item is not selected when clicked
- Example navigation is missing from the marketing site
- Example pages are missing favicons

anthonydillon:
Feel free to convert QA issues to issues.

kwm14:
Design LGTM. Will approve Percy now 👍

scss/_base_forms.scss
bartaz: @anthonydillon Base forms use chevron icon.

There are also other patterns that include "patterns_icons" (accortion for plus/minus, code copyable for copy icon). So it seems like at least the icon mixins (not `p-icon` pattern itself) should be part of base/global Vanilla, to be accessible from any pattern independently.

scss/_patterns_accordion.scss
```diff
@@ -1,5 +1,4 @@
 @import 'settings';
```
bartaz: @anthonydillon Why are settings imported in every single pattern separately, instead of being part of the base?

We require base to be imported (and `@include vf-base`) before any pattern code, can't settings be part of that?
This `@import settings` gives a fake impression that pattern needs only settings, while they quite often use other mixins/patterns from base.
I guess we reached the point when we know we can't easily make patterns built completely independently from rest of Vanilla, so maybe we can make `@import settings` part of this already required base?

anthonydillon: This was from a point when we wanted all patterns to build without base.

PR #2731, opened 2020-01-10
Add strong text example to typography

solazio:
## Done

Add strong text example to typography using Codepen

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/base/typography/
- Make sure the strong text example in Codepen is visible

### DIFF
--- a/docs/examples/base/strong.html
+++ b/docs/examples/base/strong.html
@@ -1,0 +1,7 @@
+---
+layout: examples
+title: Strong text
+category: _base
+---
+
+<p>The pastebin is <strong>blocked</strong>.</p>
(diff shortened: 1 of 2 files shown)


docs/examples/base/strong.html
```diff
@@ -0,0 +1,7 @@
+---
+layout: examples
+title: Strong text
+category: _base
+---
+
+<strong>This is some &lt;strong&gt; text.</strong>
```
anthonydillon: I think it would be a better example if the strong post of a sentence. Can you try make the sentence more real. Like: The pastebin is **blocked**

PR #2738, opened 2020-01-14
Remove unnecessary margin

bartaz:
## Done

Follow up to #2733
Removes inline margin from another table as well that was forgot during merge.

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/ or [demo](https://vanilla-framework-canonical-web-and-design-pr-2738.run.demo.haus/)
- Go to [components status page](https://vanilla-framework-canonical-web-and-design-pr-2738.run.demo.haus/component-status/)
- There should be no inline margin on tables

### DIFF
--- a/docs/component-status.md
+++ b/docs/component-status.md
@@ -36,7 +36,7 @@ When we add, make significant updates, or deprecate a component we update their 
 
 #### Previously in Vanilla 2.5
 
-<table style="margin-bottom: 1rem;">
+<table>
   <thead>
     <tr>
       <th style="width: 25%">Component</th>


nottrobin:
@bartaz could you illustrate the change you're expecting here a little more? I'm looking at the difference between https://vanilla-framework-canonical-web-and-design-pr-2738.run.demo.haus/component-status/ and https://docs.vanillaframework.io/component-status and a bit confused.

bartaz:
@nottrobin Sorry, the page had some additions since it went live during last release, so comparing to production could be confusing.
It was explained a bit better in original PR (#2733), this one was just same fix in another place.

PR #2740, opened 2020-01-14
Enable multistage builds, update pretty-quick

nottrobin:
This will allow us to use node10 in production builds, so we no longer need to pin pretty-quick at an old version.

QA
--

``` bash
./run clean  # To ensure we're not relying on locally built assets
DOCKER_BUILDKIT=1 docker build --build-arg BUILD_ID=mybuildid --tag vf .
docker run -ti -p 8999:80 vf
```

Now go to http://localhost:8999, check it works.

Also, run:

``` bash
$ curl -I http://localhost:8999
x-vcs-revision: mybuildid
```

^ check you see `mybuildid` in the output


bartaz:
@nottrobin Would be good to update travis config as well, to make sure it uses same version of node as we have on jenkins: https://github.com/canonical-web-and-design/vanilla-framework/blob/master/.travis.yml#L2-L5

nottrobin:
@bartaz fixed

PR #2761, opened 2020-01-20
Dark sub navigation fixes

### LINKED ISSUE #2752: Subnav background colour is darker than the nav bg colour
## Pattern to amend

p-subnav

## Visual

Provide a screenshot or link to a prototype of your pattern amendment.
image

## Context

When the navigation theme is set to dark, the subnav background colour is slightly darker.

### LINKED ISSUE #2753: Subnav items are indented to the right - we should have the option to remove the indentation
## Pattern to amend
p-subnav

## Visual

Provide a screenshot or link to a prototype of your pattern amendment.
image

## Context

Open the subnav on small screens and notice the items are indented. We should have the option of removing that indentation.

bartaz:
## Done

Fixes #2752 
Fixes #2753 

## QA

- Pull code
- Run `./run serve --watch`
- Go to [dark subnavigation example](https://vanilla-framework-canonical-web-and-design-pr-2761.run.demo.haus/examples/patterns/navigation/subnav-dark/)
- Make sure subnav background is the same as main nav
- Make sure there is no light border around dark subnav dropdown
- On small screen there should be no additional indent and no border around subnav items

## Screenshots

Screenshot 2020-01-20 at 16 49 48
Screenshot 2020-01-20 at 16 49 33

### DIFF
--- a/scss/_patterns_subnav.scss
+++ b/scss/_patterns_subnav.scss
@@ -115,7 +115,7 @@
     white-space: nowrap;
 
     @media (max-width: $breakpoint-navigation-threshold) {
-      padding: $spv-inner--small $grid-margin-width;
+      padding: $spv-inner--small 0;
     }
 
     @media (min-width: $breakpoint-navigation-threshold) {
@@ -154,7 +154,7 @@
   }
 
   %subnav-item--dark-theme {
-    background-color: $colors--dark-theme--background;
+    background-color: $colors--dark-theme--background-highlighted;
 
     &,
     &:active,
--- a/scss/_patterns_subnav.scss
+++ b/scss/_patterns_subnav.scss
@@ -50,9 +50,15 @@
       @extend %subnav-item--dark-theme;
     }
 
-    .p-subnav.is-light {
+    .p-subnav.is-light,
+    .p-navigation.is-light .p-subnav {
       &::after {
         @include vf-icon-chevron;
+      }
+
+      .p-subnav__items,
+      .p-subnav__items--right {
+        @extend %vf-is-bordered;
       }
 
       .p-subnav__item {
@@ -73,9 +79,15 @@
       @extend %subnav-item--light-theme;
     }
 
-    .p-subnav.is-dark {
+    .p-subnav.is-dark,
+    .p-navigation.is-dark .p-subnav {
       &::after {
         @include vf-icon-chevron;
+      }
+
+      .p-subnav__items,
+      .p-subnav__items--right {
+        border: 0;
       }
 
       .p-subnav__item {
@@ -101,7 +113,7 @@
     }
 
     @media (max-width: $breakpoint-navigation-threshold) {
-      border: 0;
+      border: 0 !important; // fight specificity of `.p-navigation.is-dark .p-subnav .p-subnav__items` from theme support
       box-shadow: none;
     }
   }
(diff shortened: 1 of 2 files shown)


lyubomir-popov:
@bartaz may I suggest a couple of driveby's:

- For the dark examples, the logo has black text on dark grey background, I've made a version with white text:
https://assets.ubuntu.com/v1/20673d9f-lxd_primary--for-dark-bg.svg
- It would be nice if the subnav items could use the default vf-focus focus state (currently they use the browser default dotted version)

scss/_patterns_subnav.scss
```diff
@@ -50,11 +50,17 @@
...
-    .p-subnav.is-light {
+    .p-subnav.is-light,
+    .p-navigation.is-light .p-subnav {
       &::after {
         @include vf-icon-chevron;
       }
 
+      .p-subnav__items,
+      .p-subnav__items--right {
+        @extend %vf-is-bordered;
```
lyubomir-popov: Not sure %vf-is-bordered is a good idea. The goal of the theming is to set only colours, keeping things like border-width in the rules defined above. This way, it wouldn't need to override the colour set in this placeholder.

lyubomir-popov: besides, the colours of these lines for both themese should already be set by `.p-navigation.is-dark .p-navigation__link>a::before` defined in p-navigation. So not sure why the borders are set here at all?

scss/_patterns_subnav.scss
```diff
@@ -73,11 +79,17 @@
...
-    .p-subnav.is-dark {
+    .p-subnav.is-dark,
+    .p-navigation.is-dark .p-subnav {
       &::after {
         @include vf-icon-chevron;
       }
 
+      .p-subnav__items,
+      .p-subnav__items--right {
+        border: 0;
```
lyubomir-popov: if the rule at row 61 didn't set the borders, this would also be unnecessary

lyubomir-popov:
@bartaz I expereimented removing some more stuff that I think isn't needed in this [branch](https://github.com/lyubomir-popov/vanilla-framework/tree/subnav-further-fixes).

Also, there's a problem with chevron colour - it should the variables I've added in the branch above from the theme colours. But it doesn't work.

scss/_patterns_subnav.scss
bartaz: @lyubomir-popov 
I think the current implementations is so confusing because we have border around subnav in light theme, but don't have the border in the dark theme. This causes issues when `is-dark` is applied and border is not removed.

I'll try to change it, so that the border is used in dark theme, but has same colour as background. To make it consistent and easier to theme.

bartaz:
@lyubomir-popov I updated this PR.

I removed %vf-is-bordered from theme code replacing it with changing just the color of the border. For consistency dark theme now also has border (previously only light had).

I also updated the chevron colors as in your branch, it seems to work for me (percy detected the change of chevron color).

I updated the logo in dark example - thanks for the image.

As for the focus - I left it as it is for now. Both p-navigation and p-subnav now use default browser focus consistently, if we want to change it we should do it in both places probably. Let's do it separately from this.

lyubomir-popov:
Thanks @bartaz looks much better, adding +1. 

One last question - and this is completely optional, but why not simply delete the code in red?

Screenshot 2020-01-21 at 09 25 41


The only borders we need are the pseudo elements in the screenshot below, and they are added in the p-navigation file:

Screenshot 2020-01-21 at 09 32 54

bartaz:
@lyubomir-popov Created https://github.com/canonical-web-and-design/vanilla-framework/issues/2765 if we want to get rid of the border. Feel free to comment there.

PR #2763, opened 2020-01-21
Add individual examples for layout patterns

### LINKED ISSUE #2751: Add example pages for individually built layout components
Build example pages for rest of individually built patterns:

card
matrix
modal
strip

bartaz:
## Done

Makes sure "layout" related patterns can be built separately and add individual examples for them.

Fixes #2751

## QA

NOTE: Visual side of the examples listed below was verified by CI on Percy, so QA steps are to make sure examples work as expected and detailed visual review of those is not needed.

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/ or [demo](https://vanilla-framework-canonical-web-and-design-pr-2763.run.demo.haus/)
- Go to [examples page](https://vanilla-framework-canonical-web-and-design-pr-2763.run.demo.haus/)
- Make sure that existing examples (full Vanilla) for components below work as before (check different variants for each pattern):
cards,
strips,
matrix,
modal
- Make sure that individual examples for these components work the same as full ones


nottrobin:
@bartaz is this difference expected?

image

nottrobin:
So... this new "Individual" section is really just for your own testing? Obviously the user doesn't care, they just want to see what the pattern looks like, and it should simply work the same in both cases. Is the plan for the "individual" version to become the canonical one at some point in the future? Why aren't you simply switching them all over to the "individual" version right away (testing as you go against live / master), so that you don't need to create the include and duplication etc.?

bartaz:
@nottrobin Good catch with the modal, I haven't noticed it uses another pattern in the example.

> So... this new "Individual" section is really just for your own testing?

Yes. It's to make sure all patterns build individually and that we don't introduce any dependencies in the future.
I think it would be good to keep both versions (full and individual) so we can verify (via Percy) that any change in given pattern is reflected in the same way in both places (when using full vanilla and individual pattern styles).

I agree that from user perspective it seems duplicated and unnecessary, and also makes the "Examples" page bit more confusing. Maybe we can make the individual examples section bit less prominent (or as a separate page).

lyubomir-popov:
@bartaz if both need to be there, maybe a tabstrip at the top to allow you to move between the two?

bartaz:
@lyubomir-popov I'll create separate issue to see how to best do it.

I was thinking about just adding a link at the bottom to a separate page (to make it even less prominent than tabs), but using tabs may be good as well. I'll let you know when I have something for design review.

nottrobin:
I agree it would be good to be testing both cases look identical with percy, but it would be much better if we could find a way to keep testing to a testing space, rather than making changes to user-space.

@lyubomir-popov's idea is interesting, but might be hard to achieve. I'd be interested if anyone else has any ideas.

bartaz:
@nottrobin I think technically speaking, if we want individual examples just for Percy, we don't need to link to them at all. Our current percy configurations just get a directory of static HTML and renders them all one by one. So as long as jekyll renders these into HTML files, Percy will see them.

But it still would be good for developers to be able to easily get to these pages without having to know their hidden URL.

PR #2768, opened 2020-01-22
Update specificity of navigation styles to fix issue with full width nav

### LINKED ISSUE #2767: Navigation full width is not full width anymore
Seems that full width example of navigation is not full width anymore:
https://docs.vanillaframework.io/examples/patterns/navigation/full-width/

Screenshot 2020-01-21 at 19 52 11

While `p-navigation__row--full-width` set max-width to 100% it's overridden by specificity of `.p-navigation .p-navigation__row--full-width` that extends `%row`.

bartaz:
## Done

Fixes #2767

Updates specificity of navigation style selectors to fix issue with full width navigation not taking whole 100% of screen.

## QA

- Pull code
- Run `./run serve --watch`
- Go to [full width navigation example](https://vanilla-framework-canonical-web-and-design-pr-2768.run.demo.haus/examples/patterns/navigation/full-width/) on a wide screen
- Make sure nav takes whole width
- Make sure other nav examples are constrained by grid width

### DIFF
--- a/scss/_patterns_navigation.scss
+++ b/scss/_patterns_navigation.scss
@@ -124,7 +124,6 @@ $navigation-hover-opacity: 0.58;
     }
 
     %navigation-row {
-      @extend %fixed-width-container;
       display: flex;
       flex-direction: column;
 
@@ -134,11 +133,13 @@ $navigation-hover-opacity: 0.58;
     }
 
     &__row {
+      @extend %fixed-width-container;
       @extend %navigation-row;
 
       &--full-width {
+        @extend %vf-grid-container-padding;
         @extend %navigation-row;
-        max-width: 100%;
+        width: 100%;
       }
     }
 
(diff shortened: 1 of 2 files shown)


scss/_patterns_navigation.scss
```diff
@@ -3,6 +3,16 @@
 $navigation-hover-opacity: 0.58;
 
 @mixin vf-p-navigation {
+  %navigation-row {
+    @extend %fixed-width-container;
```
lyubomir-popov: do we need to override the max width on line 141? Why not take `@extend %fixed-width-container;` outside of the %navigaiton-row placeholder, and add it on line 138? this way there will be no redefinition of max-width

bartaz: @lyubomir-popov Good point. Will try that.

bartaz: @lyubomir-popov Unfortunately, wont work. %fixed-width-container defines more than just max-width. It comes with margins, etc.
So if we don't extent it in `--full-width` variant, we would need to redefine width: 100%, margin: auto, etc that it provides.

PR #2769, opened 2020-01-22
Add separate page with individual CSS examples

bartaz:
## Done

Follow up on discussion on https://github.com/canonical-web-and-design/vanilla-framework/pull/2763#issuecomment-577161459

This moves individual CSS examples to separate page (tab) not to clutter original examples page.

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/ or [demo](https://vanilla-framework-canonical-web-and-design-pr-2769.run.demo.haus/)
- Go to [examples page](https://vanilla-framework-canonical-web-and-design-pr-2769.run.demo.haus/examples/)
- Examples page should show only full Vanilla CSS examples
- Click on a tab to go to individual examples page

## Screenshots

Screenshot 2020-01-23 at 08 58 26
Screenshot 2020-01-23 at 08 58 39

### DIFF
--- a/docs/examples/index.html
+++ b/docs/examples/index.html
@@ -4,11 +4,9 @@ layout: default
 
 <h2>Component examples</h2>
 
-<hr>
-
 {% capture pages %}
 {% for page in site.pages %}
-{% if page.url contains '/examples/' %}
+{% if page.title and page.url contains '/examples/' %}
 |{{ page.title }}#{{ page.url }}
 {% endif %}
 {% endfor %}
@@ -16,24 +14,15 @@ layout: default
 {% assign sortedpages = pages | split: '|' | sort %}
 
 <div class="row">
-  <div class="col-8">
-    <h3>Examples with full Vanilla framework CSS</h3>
-    <p>Examples below use CSS built full Vanilla framework including base styles, all components and utilities.</p>
-  </div>
-</div>
-
-<div class="row">
   <div class="col-3">
-    <h4>Base elements</h4>
+    <h3>Base elements</h3>
     <nav>
       <ul class="p-list">
         {% for page in sortedpages %}
         {% assign pageitems = page | split: '#' %}
         {% if pageitems[1] contains '/examples/base/' %}
         <li class="p-list__item">
-          <a href="{{ pageitems[1] | prepend:site.baseurl }}" class="p-link">
-            {{ pageitems[0] }}
-          </a>
+          <a href="{{ pageitems[1] | prepend:site.baseurl | strip }}">{{ pageitems[0] }}</a>
         </li>
         {% endif %}
         {% endfor %}
@@ -41,16 +30,14 @@ layout: default
     </nav>
   </div>
   <div class="col-3">
-    <h4>Components</h4>
+    <h3>Components</h3>
     <nav>
       <ul class="p-list">
         {% for page in sortedpages %}
         {% assign pageitems = page | split: '#' %}
         {% if pageitems[1] contains '/examples/patterns/' %}
         <li class="p-list__item">
-          <a href="{{ pageitems[1] | prepend:site.baseurl }}" class="p-link">
-            {{ pageitems[0] }}
-          </a>
+          <a href="{{ pageitems[1] | prepend:site.baseurl | strip }}">{{ pageitems[0] }}</a>
         </li>
         {% endif %}
         {% endfor %}
@@ -58,16 +45,14 @@ layout: default
     </nav>
   </div>
   <div class="col-3">
-    <h4>Utilities</h4>
+    <h3>Utilities</h3>
     <nav>
       <ul class="p-list">
         {% for page in sortedpages %}
         {% assign pageitems = page | split: '#' %}
         {% if pageitems[1] contains '/examples/utilities/' %}
         <li class="p-list__item">
-          <a href="{{ pageitems[1] | prepend:site.baseurl }}" class="p-link">
-            {{ pageitems[0] }}
-          </a>
+          <a href="{{ pageitems[1] | prepend:site.baseurl | strip }}">{{ pageitems[0] }}</a>
         </li>
         {% endif %}
         {% endfor %}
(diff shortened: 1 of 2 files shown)


squidsoup:
@bartaz there's a weird animation reset in the header every time one of these tabs is clicked, is that normal?

bartaz:
@squidsoup I don't get it every time (not sure how it depends on loads, cache, etc), but it's not related to this PR. It happens when you navigate around the docs as well.

I'll try to look into it separately.

bartaz:
@squidsoup  Reported #2770 about the animation

bartaz:
@spencerbygraves Oh, good catch. I'll have a look. The pages seem to have exactly the same structure, so I'm not sure where the difference comes from.

lyubomir-popov:
A couple of suggestions:

1) Update the typography to h2, h4, h5as in the screenshot below:
image


2) I think the wording can be improved - my attempt below - but probably best for a native speaker to reword it. @matthewpaulthomas could you help?

- Tab labels:  Full examples, Individual examples - is this the best way to put it?

---
Current text: 

[ Full examples ]

### Examples with full Vanilla framework CSS
Examples below use CSS built full Vanilla framework including base styles, all components and utilities.

[ Individual examples ]
### Examples with individual component CSS
Examples below use CSS built with single pattern included with base styles instead of the whole Vanilla framework CSS file.

---
My suggestion:

[ ??? Complete ]
### Examples built using Vanilla framework's complete css output

[ Individual ]
### Examples built using the bare minimum of required css 
(Vanilla's base + the component's own css)

lyubomir-popov:
@bartaz the Components column has an empty `<li>` at the top

lyubomir-popov:
@bartaz please ignore the typography comment above, @spencerbygraves made a good point that this will make it inconsistent with all other pages which use h3.

I might suggest it in the future asa whole site change, similar to how we did it in snapcraft a while ago (h2>h4>h5)

bartaz:
@spencerbygraves @lyubomir-popov The empty `<li>` making the pages misaligned is fixed now.

spencerbygraves:
Thanks @bartaz

matthewpaulthomas:
> I think the wording can be improved - my attempt below - but probably best for a native speaker to reword it. @matthewpaulthomas could you help?

Not as much as I’d like, because I’m confused at the use case for this tab. How does someone end up wondering “What is the list of all Vanilla patterns that I can use via individual CSS files”?

Separately, this change would result in the IA becoming even worse: instead of two mostly-overlapping lists of Vanilla patterns (sidebar + “Component examples”), we would now have three (sidebar + “Full examples” + “Individual examples”).

If there’s really a use case for scanning the list of patterns to see which of them can be used with an individual CSS file, that could be achieved with an icon next to the pattern name, with that icon linking to the standalone example. Give the icon an explanation above/below the list.

(And ideally, instead of having a page of “Examples” at all, we’d embed and link to each example from within the page about that pattern.)

With that out of the way…

> Examples below

These examples

> CSS built full Vanilla framework

CSS built from the full Vanilla framework [? I think]

> Vanilla framework including base styles

> single pattern included with base styles

What is a “base style”? Is it defined anywhere? If so, could we link to it?

> instead of the whole Vanilla framework CSS file

Is this the only place on the site where it’s mentioned that you don’t have to use the whole Vanilla framework CSS file? ([“Customising Vanilla”](https://vanilla-framework-canonical-web-and-design-pr-2769.run.demo.haus/customising-vanilla/) includes [“Importing individual components”](https://vanilla-framework-canonical-web-and-design-pr-2769.run.demo.haus/customising-vanilla/#importing-individual-components), but the code shown there is quite different from the CSS linked to from these examples.)

matthewpaulthomas:
> I’m confused at the use case for this tab

Sorry, I missed this comment from #2763:
> > So... this new "Individual" section is really just for your own testing?
>
> Yes. It's to make sure all patterns build individually and that we don't introduce any dependencies in the future.

So if it’s for Vanilla developers, rather than end users, again I suggest just linking each individual example from an icon after the pattern name, rather than having tabs with separate pattern lists and intro paragraphs. Then footnote the symbol, something like:

✤ = Example using standalone CSS for testing/debugging

bartaz:
@lyubomir-popov Any suggestions on a symbol/icon to use for individual examples links?

Screenshot 2020-01-23 at 14 45 14

lyubomir-popov:
@bartaz we had a chat with @matthewpaulthomas and @spencerbygraves and the following question emerged:

if this is solely for QA purposes for people working on vanilla, why should it have any footprint in the public visible pages? Could it be a page on a hidden url?

bartaz:
@lyubomir-popov It can be a separate hidden page. The problem with that is that to get to this page you need to know the URL, so not having any link to it makes it harder, even for developers during testing/debugging to find it and get to it.

Would it be acceptable to at least have a link to it on the bottom of examples page?
It's gonna be far from view, while still usable for those who need it.

Screenshot 2020-01-23 at 15 30 36

lyubomir-popov:
@bartaz That seems acceptable to me. 

BTW, We do have an [environment variable](https://github.com/canonical-web-and-design/vanilla-framework/blob/df5dafea8f5616191528baaec7d24bd2f900f66e/.env#L2), maybe that could be used to turn this on / off so it is only visible while developing?

Then it wouldn't need to be hidden and could be placed where it would be easiest for you to access.

spencerbygraves:
@bartaz @lyubomir-popov let's go for a link at the bottom of the page for now, we can see how that works for people.

bartaz:
@spencerbygraves @lyubomir-popov Thanks for comments.
Updated to remove tabs and go back to previous Examples pages design, just adding a link on the bottom to separate "Standalone" examples page.

Could you have another look at these pages (and Percy)? Thanks!

@lyubomir-popov Yes, we could use the env variable to hide it on production, but considering our move to Flask, the environment and how templates are rendered will change quite significantly, so it's probably not worth adding such logic now.

PR #2773, opened 2020-01-23
WIP: Convert to Flask

nottrobin:
You should be able to see the progress so far with:

``` bash
./run
```

And go to http://localhost:8101.


docs/examples/individual/patterns/_table-of-contents.html
```diff
@@ -1,8 +1,6 @@
----
-layout: examples
-title: Table of contents
-category: _patterns
-individual: patterns_table-of-contents
```
bartaz: Can examples have some kind of meta data (as in this front matter)?

For example `individual` fields was used to know which standalone css file to include in rendered example.
If we drop it, we have no way of giving the example CSS file it needs.

bartaz: Also the title was used on Examples page when listing all the examples, not just to render title in HTML. Will we know what title this page has when generating links to all examples?

urls.txt
```diff
@@ -0,0 +1,215 @@
+http://localhost:8101
```
bartaz: Why would we want this in the repo?

docs/examples/individual/patterns/_table-of-contents.html
nottrobin: Using the title when listing the pages is going to be very tricky. Can we just make do with the filename? We could rename some of the files to be better (`abbr.html` -> `abbreviation.html`).

HTML files can't have "metadata" in jinja2, but there we can set variables which can be used in includes (`{% with %}). But I feel like passing variable up the chain to the layout wrapper might be tricky... I can't remember.

nottrobin: I suppose we could solve both problems by maintaining an `example.yaml` or something in the root of `docs/`, which the app will read. But that's a pain because it means you have to keep it up-to-date whenever you add or change examples.

urls.txt
nottrobin: Why would we want this in the repo?

docs/examples/individual/patterns/_table-of-contents.html
bartaz: But `template-finder` supports markdown frontmatter, right? Does it also allow for some custom meta data? Could we use that?

So the example files would be also treated as markdown files, rendered via template finder instead of being regular flask templates?

bartaz: But I guess this does not support includes which we use to include actual example HTML content, like we did in jekyll...

bartaz: Other way would be to set some template variable (assuming you can set variable in the HTML file and the layout will be able to access it).
Or use a "stylesheet" block in the template to list name of the stylesheet to be used.

nottrobin: I think it does support includes in the HTML... I might be wrong. Give it a go. Yeah that's true, that does support frontmatter which gets passed to the wrapper template.

nottrobin:
Demo here - https://vanilla-framework-canonical-web-and-design-pr-2773.run.demo.haus/ - even though it didn't mention it

nottrobin:
Superseded by https://github.com/canonical-web-and-design/vanilla-framework/pull/2779

PR #2774, opened 2020-01-23
Update Travis and Percy configs to run on Flask docs

bartaz:
Updates travis configuration to build and run docs as flask based app.
Adds a script to get all the links from docs and snapshot them to Percy (using PercyScript).

### QA

- make sure travis build passes
- make sure snapshots of all pages (all docs and examples) are submitted to Percy

NOTE: 
Percy shows 383 missing snapshots agains 381 new snapshots.
Missing snapshots are 404 page (which crawler doesn't find obviously) and old code-snippets page, which in Jekyll site was an empty HTML with a redirect.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ before_install:
   # specifically install bundler 1.16.6 (as per Gemfile)
   - gem install bundler:1.16.6
 install:
-  - gem install percy-cli
   - gem install sass
   - bundle install
   - yarn install
@@ -15,7 +14,9 @@ before_script:
 script:
   - yarn run test
   - yarn run build-docs
-  - percy snapshot --widths "375,1280" docs/_site/
+  - yarn run serve &
+  - sleep 5
+  - yarn run percy
 notifications:
   irc:
     channels:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_script:
   - yarn run build-docs-css
 script:
   - yarn run test
-  - yarn run serve &
+  - ./entrypoint 0.0.0.0:${PORT} &
   - sleep 5
   - yarn run percy
 notifications:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,21 +1,17 @@
...
   - yarn install
+  - pip3 install -r requirements.txt
 before_script:
   - yarn run build
+  - yarn run build-docs-css
 script:
   - yarn run test
-  - yarn run build-docs
-  - percy snapshot --widths "375,1280" docs/_site/
+  - ./entrypoint 0.0.0.0:${PORT} &
```
bartaz: @nottrobin Not really sure what to do with output of this command. 
When the site is crawled for links and then snapshot by percy there is a lot of logs noise from the server running in the background.

Should we do something about it? (`> /dev/null`?) or better to leave it as it is in case we want to look at the logs?

nottrobin: I don't like `> /dev/null` because often when there's an error, it's because the server actually failed to start for some reason. I think it's important to be able to see that output.

In that case we should probably go with a solution more similar to [ubuntu.com](https://github.com/canonical-web-and-design/ubuntu.com/blob/master/.circleci/config.yml#L44) where we start the server as a background task in Travis. Although I'm struggling with finding how to do that in Travis, or in GitHub Actions..

nottrobin: Wait... maybe if we redirect STDOUT to `/dev/null` but not STDERR then we would still see all the important errors? So maybe your suggestion is a good one?

bartaz: @nottrobin As much as current server output is noisy it's not like we are looking at these logs all the time. We are only going to check them when build fails, and I guess thats when they may be helpful.

On the other hand, seems like we should move to circle CI or GH Actions anyway soon (as all our other project do that), so I guess we should just leave it as it is for now (as it is completely functional) and consider any changes when we move to another CI or find current setup really problematic.

PR #2785, opened 2020-01-27
Ensure nested fixed-witdh containers clear their side padding

lyubomir-popov:
## Done

- Fix https://github.com/canonical-web-and-design/vanilla-framework/issues/2784
- Minor refactor

## QA

- Pull code
- Run `./run serve --watch`
- Open any example page 
- add `.u-fixed-width` inside a '.row'
- observe padding is cleared

## Screenshots

image

### DIFF
--- a/docs/examples/utilities/fixed-width-container.html
+++ b/docs/examples/utilities/fixed-width-container.html
@@ -7,29 +7,3 @@ category: _utilities
 <div class="u-fixed-width">
   Content inside a container with the class `u-fixed-width` matches the width and position of content inside a grid.
 </div>
-
-<script>
-/**
-  Attaches event listener to switch baseline grid visibility on click.
-  @param {HTMLElement} toggle Switch element to toggle baseline grid.
-*/
-function setupBaselineGridSwitch(toggle) {
-  var target = toggle.getAttribute('aria-controls');
-
-  if (target) {
-    target = document.getElementById(target);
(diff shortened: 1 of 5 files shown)


scss/_base_grid-definitions.scss
```diff
@@ -51,4 +51,31 @@
       }
     }
   }
+
+  // Grid
+  %vf-grid-container-padding {
```
bartaz: Could these be inside vf-b-grid-definitions mixin? So they are not included multiple times by mistake?

lyubomir-popov: well spotted, thanks. fixed along with a typo in the md file.

docs/examples/utilities/fixed-width-container.html
```diff
@@ -0,0 +1,35 @@
...
+layout: examples
+title: Fixed width container
+category: _utilities
+---
+
+<div class="u-fixed-width">
+  Content inside a container with the class `u-fixed-width` matches the width and position of content inside a grid.
+</div>
+
+<script>
```
bartaz: This shouldn't be needed. Every example page has a grid switch built into layout.

lyubomir-popov: Good catch, I duplicated another example and forgot to delete this. Deleted now.

PR #2786, opened 2020-01-27
Improve import sections of components' docs

### LINKED ISSUE #2777: Update docs of components that have some dependencies

bartaz:
## Done

Improves "Import" section of each component docs to make sure they show both `@import` and `@include` needed for given component, also should mention any related components that are required.

This also adds `vf-p-icons-common` mixin that can be used to include individual icons instead of whole set.

Fixes #2777 

## QA

- Pull code
- Run `./run serve --watch`
- Go to pattern docs pages
- Scroll down to "Import" section of the docs
- Make sure docs include both `@import` and `@include` in SCSS snippet, some patterns (search box, table of contents, pagination, etc) should also mention other related patterns that need to be included

## Screenshots

Screenshot 2020-01-27 at 12 12 56

### DIFF
--- a/docs/patterns/buttons.md
+++ b/docs/patterns/buttons.md
@@ -84,6 +84,11 @@ To import just this component into your project, copy the snippet below and incl
 
 ```scss
 @import 'patterns_buttons';
+@include vf-p-buttons;
+
+// when using icons within buttons you need to include icons as well
+@import '../patterns_icons';
+@include vf-p-icons;
 ```
 
 For more information see [Customising Vanilla](/customising-vanilla/) in your projects, which includes overrides and importing instructions.
--- a/docs/patterns/buttons.md
+++ b/docs/patterns/buttons.md
@@ -91,6 +91,8 @@ To import just this component into your project, copy the snippet below and incl
 @include vf-p-icons;
 ```
 
+If you only need a small subset of the icons consider [including them individually](/patterns/icons/#import) to reduce the size of your CSS file.
+
 For more information see [Customising Vanilla](/customising-vanilla/) in your projects, which includes overrides and importing instructions.
 
 ### Design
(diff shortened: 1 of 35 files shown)


docs/patterns/buttons.md
```diff
@@ -84,6 +84,11 @@ To import just this component into your project, copy the snippet below and incl
 
 ```scss
 @import 'patterns_buttons';
+@include vf-p-buttons;
+
+// when using icons within buttons you need to include icons as well
+@import '../patterns_icons';
+@include vf-p-icons;
```
lyubomir-popov: this is a big mixin, shouldn't we advise including the individual icon needed?

bartaz: Fair point, but for that we would probably need to improve icons docs themselves, to make sure we give detailed instructions how to include each of the icons.

I'll try to add something quickly

docs/patterns/buttons.md
```diff
@@ -91,6 +91,8 @@ To import just this component into your project, copy the snippet below and incl
 @include vf-p-icons;
 ```
 
+If you only need small subset of the icons consider [including them individually](/patterns/icons/#import) to reduce the size of your CSS file.
```
lyubomir-popov: I think it should be  'If you only need a small subset of icons'

docs/patterns/contextual-menu.md
```diff
@@ -57,9 +57,9 @@ To import just this component into your project, copy the snippet below and incl
 @import 'patterns_contextual-menu';
 @include vf-p-contextual-menu;
 
-// when using indicator variant you need to include icons as well
+// when using indicator variant you need to include contextual menu icon as well
```
lyubomir-popov: 'Indicator variant' sounds a bit confusing - we then refer to it as "contextual menu icon" - shouldn't these use the same term in both places?

Something like "If using the variant with an icon, you need to include that separately" - but maybe a native speaker like @deadlight could help?

bartaz: Good point. And I just noticed you actually need `p-button` as well, so I updated that.

PR #2819, opened 2020-02-10
Search results page

### LINKED ISSUE #2805: Add search results page

bartaz:
Updates layout of search results page.
Removes filter functionality from existing search box and replaces it with full docs search.

Fixes #2805 

## QA

See https://github.com/canonical-web-and-design/vanilla-framework/pull/2812 for instructions on how to get search API token for local use.

Note: search box will be moved to main nav separately in future PR.

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/
- Type anything to search box
- Search page with results should appear

## Screenshots

Screenshot 2020-02-10 at 13 20 20


docs/404.html
```diff
@@ -1,329 +1,10 @@
...
-<!-- end usabilla live embed code -->
-  </body>
-</html>
+{% extends "_layouts/default.html" %}
+
+{% block content %}
+<div class="p-strip is-deep">
+  <div class="row">
+    <h1>404: Page not found</h1>
+    <p class="p-heading--four" style="max-width: 35em">Sorry, we can't find the page you're looking for.<br>If you think this is an error on our site, please <a class="p-link--external" href="https://github.com/canonical-websites/vanillaframework.io/issues/new">file a bug.</a></p>
```
anthonydillon: Why not use the grid, col-8 instead of an inline style?

anthonydillon: Please update the link to the correct org.

docs/500.html
```diff
@@ -1,329 +1,10 @@
...
-<!-- end usabilla live embed code -->
-  </body>
-</html>
+{% extends "_layouts/default.html" %}
+
+{% block content %}
+<div class="p-strip is-deep">
+  <div class="row">
+    <h1>500: Server error</h1>
+    <p class="p-heading--four" style="max-width: 35em">Something went wrong.<br>Please <a class="p-link--external" href="https://github.com/canonical-websites/vanillaframework.io/issues/new">file a bug.</a></p>
```
anthonydillon: Same comments here.

docs/static/js/scripts.js
```diff
@@ -18,35 +18,20 @@ var links = document.querySelectorAll('a');
...
-        item.classList.remove('u-hide');
-      } else {
-        item.classList.add('u-hide');
-      }
-    });
-  } else {
-    resetFilter();
-  }
-}
+searchDocsReset.addEventListener('click', function(e) {
```
anthonydillon: Didn't a type reset in a form do this without js?

bartaz: It doesn't on input type="search" 🤷‍♂ https://jsfiddle.net/bartaz/L1a9npse/
I guess because browsers have custom 'clear' functionality for search that we hide.

docs/404.html
bartaz: Ah, good catch!

bartaz: I was just moving existing code, but will have a look into cleaning it up.

docs/static/js/scripts.js
bartaz: Oh, it's actually something different: https://jsfiddle.net/bartaz/L1a9npse/

"Reset" button does work (both on text and search inputs) but it doesn't clear them, it sets them to whatever was set as value in HTML. So if we set value in HTML (what we do with search input on search page) it will not clear it, just reset to original value.

anthonydillon: Oh yea, I remember now. Thanks

PR #2821, opened 2020-02-11
Number classes

### LINKED ISSUE #1406: Inconsistent numbers in class names
Numbers in class names should either be numeric - such as `col-4` - or spelled out as words - such as `p-heading--four` - but we shouldn't mix them in the codebase.

I have no real preference which system is chosen but we should pick and stick to one.

bartaz:
## Done

Adds new heading classes with numbers instead of words (`p-heading-1`, `p-heading-2`, etc).
Existing class names still work but are marked deprecated until 3.0 (`p-heading--one`, etc).

Updates all examples to use new class names.

Fixes #1406 

## QA

- Pull code
- Run `./run serve --watch`
- Open http://0.0.0.0:8101/
- Make sure new headings classes work as expected.
- Check heading examples (use direct example pages, CodePen uses live Vanilla version which doesn't have these changes yet)


docs/component-status.md
```diff
@@ -21,7 +21,31 @@ When we add, make significant updates, or deprecate a component we update their
     </tr>
   </thead>
   <tbody>
-    <!-- 2.6.0 -->
+    <!-- 2.7.0 -->
+    <tr>
+      <th><a href="/base/typography#heading-classes">Heading classes</a></th>
+      <td><div class="p-label--updated">Updated</div></td>
+      <td>New heading classes with numbers (<code>p-heading-1</code>, ...) were added for consistency with other class names we use.</td>
```
anthonydillon: `p-heading--1` double dash

bartaz: @anthonydillon 

I decided not to do this as it seems confusing when compared with other classes with numbers like `col-1`.
Also, we don't really have `p-heading` pattern, so the numbers are not really modifiers, they are integral part of the pattern name. It doesn't exist without it.

But I'm open to arguments.

anthonydillon: I agree the grid is a special case but we shouldn't continue to spread special cases. The reason the grid uses none BEM names (`row`, `*-col`) is due to the fact it used to come from a separate package called [Shelves](http://shelvesgrid.org/). I think the heading pattern denoted by the prefixed `p-` should continue to be a pattern or become a utility, in any case, should continue to follow the BEM naming convention.

bartaz: OK. You are right that it seems like our use of `p-heading--` classes is more like an utility than pattern. But let's not do that now.

I'll update the names to use double dashes.

scss/_base_typography.scss
```diff
@@ -89,30 +89,36 @@
   }
 
   //@section Adjusted spacing for headings (or a p) following a paragraph
-  p:not([class*='p-heading--']):not([class*='p-muted-heading']) {
+  p:not([class*='p-heading-']):not([class*='p-muted-heading']) {
```
anthonydillon: Not going for the double dash in this selector?

scss/_patterns_card.scss
```diff
@@ -79,7 +79,7 @@
   }
 
   [class*='p-card'] {
-    > p:not([class^='p-heading--']),
+    > p:not([class^='p-heading-']),
```
anthonydillon: Same here

anthonydillon:
## QA
If you are adding updated to the heading classes section of the docs should you add it to the navigation item?

scss/_base_typography.scss
bartaz: Technically either would work, but good catch, I searched/replaced only the one with numbers, but forgot about those. Will fix.

bartaz:
> If you are adding updated to the heading classes section of the docs should you add it to the navigation item?

I haven't figure out good process around it yet. Should I update them all immediately (meaning - remove old ones and add new ones? or do it just before the release)?
For now I'm adding changes to status page and docs pages on the way and update navigation right before the release.

So, I'm aware of this. Will update the navigation before release.

PR #2827, opened 2020-02-12
Position contextual menu using JS

### LINKED ISSUE #2816: Contextual menu dropdown overlaps links
**Describe the bug**
`.p-contextual-menu__dropdown` has 1.2rem negative margin-top, which seams to be intended to negate the margin-bottom on a preceding button with the class `.p-contextual-menu__toggle`.

Problem is, this class can be applied on buttons and links, which have different margin bottom, so the negative margin backfires on text links.

**To Reproduce**
1. Go to 'examples/patterns/contextual-menu/default'
2. Click on a text link
3. See the overlap

**Expected behavior**
The contextual menu should be aligned correctly with both buttons and text links. 

**Screenshots**
image

bartaz:
## Done

Removes any margin compensation code from contextual menu styles.
Updates contextual menu script to position dropdown correctly regardless of the margin.
Removes unnecessary light example.

Fixes #2816

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/
- Check contextual menu examples to see if position of contextual menu is correct

## Screenshots

Screenshot 2020-02-12 at 16 53 50

Screenshot 2020-02-12 at 16 53 42


scss/_patterns_contextual-menu.scss
```diff
@@ -63,11 +63,9 @@
     }
   }
 
-  .p-contextual-menu__toggle {
-    margin-bottom: $spv-inner--small + ($spv-nudge-compensation * 2);
```
bartaz: @lyubomir-popov I removed that, because I wasn't sure if it's necessary. This was changing the bottom margin on toggle buttons, so the value of `$input-margin-bottom` was not even correct.

So now the buttons have their default margin bottom, and we compensate it by moving contextual menu up (only if toggle is button).

lyubomir-popov: Well spotted, I can confirm the amount of spacing inserted after matches  that of buttons and inputs.

scss/_patterns_contextual-menu.scss
```diff
@@ -63,11 +63,9 @@
...
-  }
-
-  .p-contextual-menu__dropdown {
+  // compensate for bottom margin when toggle is a button
+  button.p-contextual-menu__toggle + .p-contextual-menu__dropdown, // sass-lint:disable-line no-qualifying-elements
```
bartaz: @anthonydillon How do you feel about this?

It seems it's the only way, because we need to address all the buttons (`<button>` without class and with `p-button`). 

Unless we would just introduce yet another class, for buttons with contextual menu specifically (`p-contextual-menu__toggle--button`). Which we would do just to avoid using `button` in selector.

bartaz:

On a second thought I don't think this solution is good. All it does it adds an exception to cover one special case - a default button that is contextual menu toggle. It will not work anyway if contextual menu is used on anything else that has a margin (like input for example). More so - it will not work if button is using `.is-dense` (because it has much smaller margin) or if someone used `u-no-margin` on it (and believe me, in application UIs such as snapcraft we use `u-no-margin` on buttons a lot).

There is no way in which we can predict and cover all the different margins contextual toggles may have.
I currently don't think there is a CSS-only solution to that that will not involve changing HTML structure (for example to similar one that we use in tooltips, that don't suffer from this issue). On the other hand tooltips have other issues because of their HTML structure (they suffer from overflow issues, etc) - so it's not perfect solution either.

Unless we have any other idea I think the best we can do is remove any kind of margin 'compensation' from contextual menu, position it by default under the element (including it's margin) - so it will work well out of the box on elements with no margin, and assume JavaScript that opens the contextual menu makes sure it's positioned correctly.

anthonydillon:
@bartaz I agree this is not the way to solve this issue. I would prefer to us cut it back to basics. As I see it there are three options:
1. Remove the `margin-top` from the menu and it will just be the margin distance away from it (the small arrow here my help this gap visually and if it overlays the text slightly I think that is ok)
2. Remove the `margin-top` from the menu and remove `margin-bottom` from the contextual-menu__toggle. This could be dangerous but in all the examples we have documented it seems to work ok.
3. As Bartek mentioned, add a mark-up container to aid the alignment of the menu. If we introduce the mark up I would like to try and keep the current structure of the mark up functional and so document the container method and allow people to switch slowly. _The more I think about this I am unsure how more mark-up could help here_

lyubomir-popov:
@anthonydillon @bartaz I made an example to show what option 2 will cause:

Currently: (margin bottom matches inputs, buttons etc.):
image

With margin-bottom removed:
image

docs/examples/patterns/contextual-menu/dark.html
```diff
@@ -65,71 +65,6 @@
...
-    if (e.keyCode === 27) {
-      for (var i = 0, l = toggles.length; i < l; i++) {
-        toggleMenu(toggles[i], false);
-      }
-    }
-  });
-}
-
-setupAllContextualMenus('.p-contextual-menu__toggle');
+  {% include 'examples/patterns/contextual-menu/_script.js' %}
```
anthonydillon: Will this work for the JS tab in codepen include in docs?

docs/examples/patterns/contextual-menu/_script.js
```diff
@@ -0,0 +1,78 @@
...
+*/
+function toggleMenu(element, show, top) {
+  var target = document.getElementById(element.getAttribute('aria-controls'));
+
+  if (target) {
+    element.setAttribute('aria-expanded', show);
+    target.setAttribute('aria-hidden', !show);
+
+    if (typeof top !== 'undefined') {
+      target.style.top = top + 'px';
```
anthonydillon: My worry with this is that this will be the first time we apply styling via JS not just functionality. I think if people use the mark up in the example they should get what they see in the example. I know that the JS needs to be used but simply toggling aria is easier to get right and document.

docs/examples/patterns/contextual-menu/dark.html
bartaz: Yes, it does:

Screenshot 2020-02-13 at 12 04 04

CodePen (similar to example-js before) uses rendered example HTML (fetched via ajax), doesn't matter how it's done in the example template.

docs/examples/patterns/contextual-menu/_script.js
bartaz: @anthonydillon 

Ok. Stepping back a bit.

We currently have an issue that contextual menu is misaligned (too much to the top):

Screenshot 2020-02-13 at 12 43 05

It's because we have some styling that adds negative margin (possibly to compensate for negative margin on button).
If we remove any compensation and changes to margins we get:

Screenshot 2020-02-13 at 12 55 45

Which fixes issue for inline links, but makes it wrong the opposite way for buttons.
So it's hard to call it 'fix'. The margin bottom on button is large enough that contextual menu looks visually unattached. 

If we can leave with that, we can leave it like this. And document that you could add `u-no-margin` to this button or fix it with JS.

Or, we could fix it with JS ourselves as well - as an example.
Yes, it's more complicated that toggling visibility. But there is no good way around it. It either looks bad and broken or we have more code somewhere (weird custom complex rules in SCSS or a bit of positioning in JS). If we want Vanilla CSS to be cleaner (because we can't simply cover all possible cases for stuff with different margins) we need to have a bit more complex JS. And for contextual menu you need some custom JS anyway. It's likely that its main use is in apps rather than websites, so they already have custom JS to handle showing this dropdown and all the actions it triggers.

On the opposite side - for Vanilla tooltips, which are currently CSS only - there were number of complaints form apps devs that they are impossible to control and position with JS (which is how apps handle them). So I think it's OK to leave some of responsibility of positioning to JS. 

Both for tooltips and dropdowns you may easily get to the point where you want to change their position (top/bottom/left/right) depending on where they are on screen, so they are not off screen when opened. We can't do it from Vanilla CSS. It would only be possible with JS.

Bootstrap doesn't even bother - they use external library for dropdowns and position them with position absolute. They control styling, but position is controlled by JS. 

It's not perfect, but it's hard to avoid if we want to avoid having exceptions in SCSS.

anthonydillon: I agree, fair point :+1:

PR #2837, opened 2020-02-19
Update release drafter template, use it as GH Workflow

bartaz:
## Done

Updates release drafter template.
Uses Release Drafter as GH Actions Workflow.

Fixes https://github.com/canonical-web-and-design/vanilla-squad/issues/771

## QA

There is no way to QA it until it's merged and runs on GH, so trust in code review...

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,16 +2,17 @@
 
 [List of work items including drive-bys]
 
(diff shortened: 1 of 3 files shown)


bartaz:
> All looks good. You are missing the `Enhancement ✨` label on the project though.

Ah, good catch @anthonydillon . This probably means we don't need it. I was copying the template from maas-ui and thought these were default labels every project has. Will remove that.

PR #2840, opened 2020-02-20
Add workflow to publish Vanilla to npm and assets on release

### LINKED ISSUE #2327: Set up GH actions to publish to NPM

### LINKED ISSUE #2328: Set up GH actions (or jenkins) to publish to assets server

bartaz:
## Done

Add GitHub workflow to build Vanilla and publish it to npm and assets server when release in GitHub is created.

Fixes #2327
Fixes #2328 

## QA

There is no good way to QA it in place until it's merged and release is triggered.

You can review [test run of the workflow](https://github.com/bartaz/vanilla-framework/actions/runs/42250611) in the fork.
Note - test run fails to publish, because version in package.json didn't change, but you can review if all steps are called correctly.

### DIFF
--- a/.github/workflows/publish-on-release.yml
+++ b/.github/workflows/publish-on-release.yml
@@ -1,0 +1,56 @@
+name: Publish Vanilla release
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
+    name: Build Vanilla
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - name: Read Vanilla version from package.json
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-on-release.yml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          node-version: 12
+          registry-url: https://registry.npmjs.org/
+      - run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+
+  publish-assets:
+    name: Publish to assets server
+    needs: build
```
tbille: Do you think we should wait for `publish-npm` before uploading it to our asset server?

bartaz: Good point. Updated, so we don't upload to assets if npm publish fails.

PR #2850, opened 2020-02-26
Better dotrun support

bartaz:
## Done

- adds dotrun generated files to gitignore
- adds `start` script to package.json

## QA

- Pull code
- If you have dotrun snap installed, run `dotrun`
- Otherwise `./run exec yarn start`
- Vanilla should be built, server should start


anthonydillon:
## QA:
`dotrun` works well. Ran some benchmarks on my system and they show the snap is much faster.

Command | dotrun | ./run
------------ | ------------- | -------------
Clean build | 35.948 | 1:14.46
Rebuild build | 8.834 | 14.442
Clean | 1.273 | 2.877
Clean test | error | error
Test after build | 3.814 | 47.085

PR #2852, opened 2020-02-26
Improve class selectors for navigation

### LINKED ISSUE #1746: Deeply nested selectors of navigation pattern
Navigation pattern seems to have very deeply nested selectors, which makes it very difficult to extend/override when some custom styling is needed.

Found in Vanilla 1.6.7.

For example this (https://github.com/vanilla-framework/vanilla-framework/blob/v1.6.7/scss/_patterns_navigation.scss#L269-L271):
```
       &:hover {
          background-color: $color-light;
        }
```

ends up as:
```
.p-navigation .p-navigation__links .p-navigation__link:hover {
    background-color: #f7f7f7
}
```

## Expected result

Ideally Vanilla classes should not be nested at all. Simply this should be enough:
```
.p-navigation__link:hover {
    background-color: #f7f7f7
}
```

If it's not possible or hard to achieve due to how Vanilla is structured (patterns generated from mixins) amount of nesting should be at least limited to 2 levels:

```
.p-navigation .p-navigation__link:hover {
    background-color: #f7f7f7
}
```

### NOTE

I had a look at current master code of navigation and it looks to be a bit better in terms of nesting in SCSS, but still there are selectors that can be quite deep: https://github.com/vanilla-framework/vanilla-framework/blob/master/scss/_patterns_navigation.scss#L243-L245

### LINKED ISSUE #2168: p-navigation relies on element selectors
**Describe the bug**
p-navigation uses the `p-navigation__link > a`, which is excessively greedy and makes overriding without also using a `> a` rule in other patterns impossible.

We should use classes for all of these selectors.

bartaz:
## Done

- introduces new navigation class names (`p-navigation__items`, `p-navigation__item`, `p-navigation__link`) to be used in place of existing `p-navigation__links`, `p-navigation__link` and classless `<a>`
- deprecates the use of old classes and selectors
- updates the documentation and examples

Fixes #2168 
Fixes #1746

## QA

- Pull code and `./run`
- Open http://0.0.0.0:8101/ or [demo](https://vanilla-framework-canonical-web-and-design-pr-2852.run.demo.haus/examples)
- Check all navigation examples
- They should work as expected with new classes
- New "deprecated" examples should also work with old classes
- Check [component status](https://vanilla-framework-canonical-web-and-design-pr-2852.run.demo.haus/component-status) - make sure changes to navigation are listed
- Check [navigation documentation](https://vanilla-framework-canonical-web-and-design-pr-2852.run.demo.haus/patterns/navigation) - make sure deprecated classes are documented

Please note - embedded CodePen examples will not work because they use published version of Vanilla, not latest one in development.


anthonydillon:
@bartaz out of curiosity. What is the benefit of moving the styles into one use placeholders? I tend to try and put the code where a developer would look first for something.

bartaz:
@anthonydillon These parts will be duplicated for new class names and current deprecated ones.

Not to duplicate the code, I moved it to placeholders so both new and old class names can extent from it. Old ones will be wrapped in `deprecate`.

anthonydillon:
Makes sense :+1: forgot about the wrapping deprecate mixin. Thanks for explaining.

lyubomir-popov:
@bartaz would it be scope creep to fix the p-subnav situation? Shouldn't `subnav` just be a modifier instead of a separate pattern that is however tied theme-wise to the nav, which causes some unnecessary complications.

bartaz:
@lyubomir-popov I agree that we should fix subnav, but this PR got quite complex even without it.

I created #2857 so we have this in backlog.

docs/examples/patterns/navigation/_script.js
```diff
@@ -0,0 +1,81 @@
+/**
+  Opens a given subnav by applying is-active class to it
+  and setting aria-hidden attribute on dropdown contents.
+  @param {HTMLElement} subnav Root element of subnavigation to open.
+*/
+function openSubnav(subnav) {
+  subnav.classList.add('is-active');
+  var toggle = subnav.querySelector('.p-subnav__toggle');
+  var dropdown = document.getElementById(toggle.getAttribute('aria-controls'));
+  dropdown.setAttribute('aria-hidden', 'true');
```
anthonydillon: Might want to null check this for better reuse.

docs/examples/patterns/navigation/_script.js
```diff
@@ -0,0 +1,81 @@
...
+/**
+  Closes a given subnav by removing is-active class to it
+  and setting aria-hidden attribute on dropdown contents.
+  @param {HTMLElement} subnav Root element of subnavigation to open.
+*/
+function closeSubnav(subnav) {
+  subnav.classList.remove('is-active');
+  var toggle = subnav.querySelector('.p-subnav__toggle');
+  var dropdown = document.getElementById(toggle.getAttribute('aria-controls'));
+  dropdown.setAttribute('aria-hidden', 'false');
```
anthonydillon: Same here

docs/examples/patterns/navigation/_script.js
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/**
+  Closes all subnavs on the page.
+*/
+function closeAllSubnavs() {
+  var subnavs = document.querySelectorAll('.p-subnav');
```
anthonydillon: I wonder if its ever likely to have more then one on a page at a time? In that case we could limit the search to the used subnav.

docs/examples/patterns/navigation/_script.js
```diff
@@ -0,0 +1,81 @@
...
+/**
+  Attaches click event listener to subnav toggle.
+  @param {HTMLElement} subnavToggle Toggle element of subnavigation.
+*/
+function setupSubnavToggle(subnavToggle) {
+  subnavToggle.addEventListener('click', function(event) {
+    event.preventDefault();
+    event.stopPropagation();
+
+    var subnav = subnavToggle.parentElement;
```
anthonydillon: Maybe use `closest` so the JS doesnt need to change if the markup does

docs/patterns/navigation.md
```diff
@@ -34,6 +36,18 @@ by adjusting the `$breakpoint-navigation-threshold` in `_settings_breakpoints.sc
 View example of the navigation pattern
 </a>
 
+<span class="p-label--deprecated">Deprecated</span>
+
+In Vanilla 2.8 we deprecated the use of `p-navigation__links`, `p-navigation__link` and classless `<a>` in the navigation. Support for these classes will be removed in future version 3.0.
+
+You should use `p-navigation__items`, `p-navigation__item` and `<a class="p-navigation__link">` instead.
```
anthonydillon: Maybe we should start a 3.0 upgrade document now so its easier when it comes to it?

scss/_patterns_navigation.scss
```diff
@@ -296,14 +362,41 @@ $navigation-hover-opacity: 0.58;
...
     background: $colors--dark-theme--border-default;
   }
+
+  .p-navigation__item > .p-navigation__link, // remove nesting in 3.0 when deprecated __link styles are removed
+  .p-navigation__toggle--close,
+  .p-navigation__toggle--open {
+    @extend %navigation-link-dark;
+  }
+
+  .p-navigation__item.is-selected > .p-navigation__link {
```
anthonydillon: Do we need the sibling selector now that we are directly targeting classes?

scss/_patterns_subnav.scss
```diff
@@ -33,13 +33,21 @@
...
 
-  // XXX: Fix required because of greedy rule in p-navigation
-  // Can be updated with resolution of:
-  // https://github.com/canonical-web-and-design/vanilla-framework/issues/2168
-  .p-subnav > a {
+  %subnav-link {
     padding-right: 2 * $sph-inner--small + map-get($icon-sizes, default); // icon padded with the default padding-right of selects, inputs etc.
   }
 
+  .p-subnav > .p-navigation__link {
```
anthonydillon: Seems strange to cross use two patterns here?

bartaz: It's been like this before. I don't want to change it in this PR. I opened a separate issue to move p-subnav into navigation #2857.

scss/_patterns_navigation.scss
bartaz: Yes. In this case it depends on `is-selected` state of the parent. I guess we could simplify it to `.p-navigation__item.is-selected .p-navigation__link` if that's what you mean.

docs/examples/patterns/navigation/_script.js
bartaz: `closest` is not supported by IE. We would need to polyfill that.

docs/examples/patterns/navigation/_script.js
bartaz: Yes. Why wouldn't you have multiple dropdowns in the navigation? We have this in our example where each item in menu is a dropdown.

docs/examples/patterns/navigation/_script.js
bartaz: Good catch. Just copied this existing code, but will fix it while we are at it.

docs/patterns/navigation.md
bartaz: Created issue about it #2858

anthonydillon:
Not really a breaking change 🏅

bartaz:
> Not really a breaking change 🏅

@anthonydillon Yes, when we deprecate things there are never really breaking changes (at least not intentionally). I use this label to remind myself to mention the deprecated classes in release notes.

PR #2875, opened 2020-03-03
Move vanillaframework.io pages to the docs

jkfran:
## Done

Move site pages

## QA

- Pull code
- Run `./run`


site/_layouts/site.html
```diff
@@ -0,0 +1,126 @@
+<!DOCTYPE html>
```
nottrobin: What was wrong with `default.html`?

nottrobin: Oh I see, you're using two different templates... Are we not planning to merge the layouts at this stage?

site/accessibility.html
```diff
@@ -0,0 +1,105 @@
+{# ---
```
nottrobin: Why keep this?

site/404.html
```diff
@@ -1,11 +1,18 @@
-{% extends "_layouts/default.html" %}
```
nottrobin: In most of our projects, these templates live in a `templates` folder. I get that vanilla-framework is different in that it's not in and of itself a website project, but I'm not sure `templates` is a much worse name than `site`. Don't you think it would probably be okay to simply call it `templates` for consistency?

site/accessibility.html
jkfran: We are leaving this as a reminder, we will add a way to have this functionality later.

site/404.html
jkfran: I don't mind, if we are calling it `templates` in others projects we should do the same. WDYT @bartaz ?

site/_layouts/site.html
jkfran: Probably some parts of the layouts will be unified later, but they are quite different, right?

nottrobin: Yes, I assume they will be unified, and I was hoping that would be happening this week. But if you want to phase the work such that at the moment you preserve 2 different layouts and then a future PR will come and unify them, that's fine.

bartaz: @nottrobin Yes. We want them unified. This is just a first step to get it running with 2 separate layouts, so there are no changes to the look of site or docs. Next step would be to unify them both on design and template levels.

site/404.html
bartaz: Yes, I added it to todos, let's do it separately to avoid too moving files again that will cause conflicts with Lyubos design changes.

PR #2915, opened 2020-03-11
Add side navigation component with default styling

### LINKED ISSUE #2860: Build basic sidenav styles

### LINKED ISSUE #2861: Build nested sidenav levels

bartaz:
## Done

- adds basic version of side navigation
- adds examples and simple docs

Fixes #2860 
Fixes #2861

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/ or [demo](https://vanilla-framework-canonical-web-and-design-pr-2915.run.demo.haus/docs/examples/patterns/side-navigation/default)
- Check new side navigation pattern examples, make sure they work correctly
  - [default](https://vanilla-framework-canonical-web-and-design-pr-2915.run.demo.haus/docs/examples/patterns/side-navigation/default)
  - [docs](https://vanilla-framework-canonical-web-and-design-pr-2915.run.demo.haus/docs/examples/patterns/side-navigation/docs)
- make sure side nav documentation is added: https://vanilla-framework-canonical-web-and-design-pr-2915.run.demo.haus/docs/patterns/navigation/#side-navigation
  - NOTE: codepen will not render the styles for sidenav, because it uses published version of Vanilla

## Screenshots

Screenshot 2020-03-12 at 10 13 11

### DIFF
--- a/scss/_patterns_side-navigation.scss
+++ b/scss/_patterns_side-navigation.scss
@@ -58,15 +58,15 @@
 // theme
 @mixin vf-side-navigation-theme(
   // default text color in sidebar
-    $color-sidenav-text-default: $color-mid-dark,
+    $color-sidenav-text-default: $colors--light-theme--text-disabled,
   // text color of highlighted items
-    $color-sidenav-text-active: $color-dark,
+    $color-sidenav-text-active: $colors--light-theme--text-default,
   // background color of active/hovered items
-    $color-sidenav-item-background-highlight: $color-light,
+    $color-sidenav-item-background-highlight: $colors--light-theme--background-highlighted,
   // border color of active item
     $color-sidenav-item-border-highlight: $color-brand,
   // border color of items list
-    $color-sidenav-list-border: $color-mid-x-light
+    $color-sidenav-list-border: $colors--light-theme--border-default
 ) {
   .p-side-navigation {
     color: $color-sidenav-text-default;
--- a/scss/_patterns_side-navigation.scss
+++ b/scss/_patterns_side-navigation.scss
@@ -9,12 +9,12 @@
   .p-side-navigation__list {
     @extend %vf-list;
 
-    &:not(:last-child) {
-      @extend %vf-pseudo-border--bottom;
+    & + & {
+      @extend %vf-pseudo-border--top;
 
       &::after {
-        bottom: -0.5 * $spv-outer--scaleable; // place border in the middle of the margin
         left: $sph-inner;
+        top: -0.5 * $spv-outer--scaleable; // place border in the middle of the margin
       }
     }
   }
--- a/scss/_patterns_side-navigation.scss
+++ b/scss/_patterns_side-navigation.scss
@@ -19,7 +19,7 @@
 
   .p-side-navigation {
     color: $color-sidenav-text-default;
-    padding-top: 2px;
+    padding-top: 0.5 * $spv-inner--x-small;
   }
 
   .p-side-navigation__list {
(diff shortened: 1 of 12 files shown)


scss/_patterns_side-navigation.scss
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  // default light theme
+  @include vf-side-navigation-theme;
+}
+
+// theme
+@mixin vf-side-navigation-theme(
+  // default text color in sidebar
+    $color-sidenav-text-default: $color-mid-dark,
```
lyubomir-popov: When it comes to colours, in my opinion these local vars should point to the new [theme colour vars](e.g. `$colors--light-theme--border-default`). They are named after what they're supposed to be used for, and allow us to audit colour use easier. With the generic ones, it's difficult to know what they were meant for, or what is the other theme's counterpart for a specific colour.

lyubomir-popov: so this line would be: 
`$color-sidenav-text-default: $colors--light-theme--text-default;`

bartaz: Sure, thanks for pointers. Will update that.

bartaz: @lyubomir-popov So, for side nav we actually use lighter text, so default text will use `--text-disabled` and highlighted version will use `--text-default`.

spencerbygraves: Is this related to https://github.com/canonical-web-and-design/vanilla-framework/issues/2851?

bartaz: @spencerbygraves A little bit I guess. I think the only place where we used 'faded out' text colour was disabled text, so the name of the variable reflects that.

spencerbygraves:
@bartaz we've just had our weekly product design review which produced some interesting feedback (thanks @cristinadresch), which I've just discussed with @lyubomir-popov.

Can we try updating the selected states please, to simplify things a bit:

- Remove the bold and darker style on items you have previously selected (this does feel a bit confusing when looking at the current version)
- Remove the bold style on the currently selected items (this would help to differentiate selected items from the section titles, and bring it in line with tabs)

Now that we can see it in action, we may need to tweak things like this to get it right, hope that's ok.

This shows the difference, on the left is what we have now:

image

Thanks.

bartaz:
@spencerbygraves Sure, this will simplify the code as well and remove a need for confusing additional 'selected' state on non active items.

spencerbygraves:
@bartaz @anthonydillon @lyubomir-popov  if you look at the charmhub side nav, you'll see functionality to toggle on/off filters for the main content. Should we add that to the Vanilla side nav at some point?

https://charmhub-io-canonical-web-and-design-pr-5.run.demo.haus/

The side nav here will be updated with the Vanilla one.

bartaz:
@spencerbygraves Couple of notes on that:

- charmhub seems to be using custom colors of the side nav (and they have it on gray background) - we currently assume white background and use light theme colors, if we want project to be able to change colors only in sidenav component without affecting rest of light theme we need separate variables for that
- side nav is going to have support for right hand icons, this should cover the X case for filters - the easiest way would be to have X always visible (when item is highlighted), having it appear on hover would require a special case

So, I think what we have in plan to implement should be allow for such functionality (with some small compromises). We could later iterate on improvements if needed.

lyubomir-popov:
Would it be possible to keep the focus outline above the black bar?
image

bartaz:
> Would it be possible to keep the focus outline above the black bar?

@lyubomir-popov I'll have a look - I just used existing mixins. I'll check if we can control it.

bartaz:
@lyubomir-popov So it seems currently the only way to display outline above our absolutely positioned border is to hide the border on focus.

I'm personally not sure if it's better - looks a bit like border is "disappearing" rather than getting covered by outline:

side-nav-focus

But will leave that design decision to you - do we keep current default (dark border over outline) or should we implement workaround and hide border on focus?

Issue #2862, opened 2020-02-28
Build optional sidenav icons


bartaz:

bartaz:
@spencerbygraves @lyubomir-popov 

I think our current designs don't specify that.
Do we allow nested levels to have icons as well?

Should they be all aligned with 1st level?

![Screenshot 2020-03-12 at 16.13.50.png](https://images.zenhubusercontent.com/593a660851aa2c30578102bb/a0e0cd62-c35e-46fe-91ec-e69036721769)

Or should they be nested as well?

![Screenshot 2020-03-12 at 16.15.10.png](https://images.zenhubusercontent.com/593a660851aa2c30578102bb/c5a92242-6c37-443a-8215-281b8936f9a2)

spencerbygraves:
@bartaz we think it's unlikely that icons are needed for second and third level links, perhaps we can leave that for now.

lyubomir-popov:
if it's not too much hassle would be nice to align icons  with the previous level's text:
```
[•]  Level n <li> 
     [•]  Level n+1 <li>
```
Just  so it doesn't look broken if someone does add icons.

bartaz:
@lyubomir-popov 

To do that we would need to add quite big paddings to nested elements (even those without icons) to make them all align:

![Screenshot 2020-03-13 at 08.00.41.png](https://images.zenhubusercontent.com/593a660851aa2c30578102bb/edfcfb62-1690-4872-8cae-9106fc747111)

To avoid that (have default smaller padding if icons are not used in nested levels) we would need to add "has icons" modifier on each nested level, not just on root of component - which adds complexity to code and usage of the pattern, that is probably unnecessary - assuming that we don't really expect icons on nested levels.

I would suggest going with a simplest solution that covers our case (icons on first level) and document that icons are not supported on nested levels.

Both proposals from my previous comment are fairly easy to implement. Currently the first one (all icons aligned on the left, not nested) requires the least code (as it doesn't add rules to nest the icons).

Issue #2865, opened 2020-02-28
Build example for sidenav use cases


bartaz:
Make sure we have examples for:

- wrapping
- truncating
- icons on the right (on different nesting levels)
- labels on the right (on different nesting levels)
- icons/labels on the right with wrapping and truncating
- items on different levels and titles that are not links
- side nav in grid
- icons on the left
- icons on the left + icons on the right

PR #2928, opened 2020-03-16
Side navigation status labels and icons

bartaz:
## Done

Add right side status icons/labels to side navigation.

Part of #2862 

## QA

- `./run` or [demo](https://vanilla-framework-canonical-web-and-design-pr-2928.run.demo.haus/docs/examples/patterns/side-navigation/default)
- Check if labels and icons are displayed correctly on different levels of navigation

## Screenshots

Screenshot 2020-03-16 at 16 22 57

### DIFF
--- a/scss/_patterns_side-navigation.scss
+++ b/scss/_patterns_side-navigation.scss
@@ -25,7 +25,7 @@
 
   .p-side-navigation__text,
   .p-side-navigation__link {
-    display: block;
+    display: flex;
     padding: $spv-inner--x-small $sph-inner--small $spv-inner--x-small $sph-inner;
 
     // nested 2 or 3 levels
@@ -50,6 +50,11 @@
     &:hover {
       text-decoration: none;
     }
+  }
+
+  .p-side-navigation__status {
+    margin-left: auto;
+    padding-left: $spv-outer--small;
   }
 
   // default light theme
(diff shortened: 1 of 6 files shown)


lyubomir-popov:
@bartaz I think the margin-right on the icon should be 2x bigger. This is so make it consistent with [rules](https://discourse.ubuntu.com/t/icon-positioning-within-tables/14435) we're creating for icons elsewhere: .5rem offset between icon and the text it belongs to, 1rem between icon and unrelated text.

lyubomir-popov:
@bartaz We already have an example for truncation in the utilities section, but wondering if it would be helpful to have one here as well. The gotcha there is, the truncated text needs to be wrapped in a span with class"u-truncate", rather than adding that class on the `<li>`. Only if you think it is a good idea, not insisting.
image

bartaz:
@lyubomir-popov 

I updated the spacing on the right of icons.

As for truncating - I have a separate task #2865 to make sure we have all cases in examples, so will make sure I add it there.

scss/_patterns_side-navigation.scss
```diff
@@ -52,6 +52,11 @@
     }
   }
 
+  .p-side-navigation__status {
+    margin-left: auto;
+    padding-left: $spv-outer--small;
```
lyubomir-popov: This should be $spv-inner--small (because it is inside, not outside of the element)

PR #111, opened 2015-03-17
consolidate global settings and make neccessary edits in dependent files

deadlight:
Done:
updated global settings, made some of the variables more semantic, removed silly ones

QA:
gulp build

PR #2932, opened 2020-03-17
Add side navigation variant with icons on the left

### LINKED ISSUE #2862: Build optional sidenav icons

bartaz:
## Done

Adds side navigation variant with icons on the left.

Fixes #2862 

## QA

- Run `./run` or [demo](https://vanilla-framework-canonical-web-and-design-pr-2932.run.demo.haus/docs/examples/patterns/side-navigation/icons)
- Make sure side nav works correctly with right side icons
- Make sure side nav default variant is not affected
- Make sure there are [docs](https://vanilla-framework-canonical-web-and-design-pr-2932.run.demo.haus/docs/patterns/navigation#side-navigation) describing the use of variant with icons
  - please note CodePen example will show unstyled side nav, because it was not published yet

Screenshot 2020-03-17 at 14 38 34

### DIFF
--- a/scss/_patterns_side-navigation.scss
+++ b/scss/_patterns_side-navigation.scss
@@ -1,13 +1,16 @@
 @import 'settings';
 
 @mixin vf-p-side-navigation {
-  .p-side-navigation__list {
-    @extend %vf-list;
-
-    .p-side-navigation > &:first-child {
+  .p-side-navigation,
+  .p-side-navigation--icons {
+    & > .p-side-navigation__list:first-child {
       // nudge from top to put text on baseline grid
       padding-top: 0.5 * $spv-inner--x-small;
     }
+  }
+
+  .p-side-navigation__list {
+    @extend %vf-list;
 
     & + & {
       @extend %vf-pseudo-border--top;
@@ -35,6 +38,27 @@
 
     .p-side-navigation__item .p-side-navigation__item .p-side-navigation__item & {
       padding-left: 3 * $sph-inner;
+    }
+
+    .p-side-navigation--icons & {
+      padding-left: 2 * $sph-inner + map-get($icon-sizes, default);
+      position: relative;
+    }
+
+    .p-side-navigation--icons .p-side-navigation__item .p-side-navigation__item & {
+      padding-left: 2 * $sph-inner + map-get($icon-sizes, default) + $sph-inner;
+    }
+
+    .p-side-navigation__item .p-side-navigation__item .p-side-navigation__item & {
+      padding-left: 2 * $sph-inner + map-get($icon-sizes, default) + 2 * $sph-inner;
+    }
+  }
+
+  .p-side-navigation--icons {
+    .p-side-navigation__icon {
+      left: $sph-inner;
+      position: absolute;
+      top: $sph-inner--small;
     }
   }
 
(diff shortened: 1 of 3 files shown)


lyubomir-popov:
Nice work, one suggestion: 

can we clip the border? @spencerbygraves what do you think?
image

scss/_patterns_side-navigation.scss
```diff
@@ -28,13 +31,40 @@
...
+    $icon-spacing-width: map-get($icon-sizes, default) + $sph-inner;
+
+    .p-side-navigation--icons & {
+      padding-left: $sph-inner + $icon-spacing-width;
+      position: relative;
+    }
+
+    // nested 2nd level of navigation
     .p-side-navigation__item .p-side-navigation__item & {
       padding-left: 2 * $sph-inner;
```
lyubomir-popov: I think the levels should be configurable when you invoke the mixin; many sites won't need 3 levels, and some might need 5; my proposal: @mixin p-side-navigation($depth: 3)

spencerbygraves:
>can we clip the border?

@lyubomir-popov to draw attention to the text more than the icon?

I've just noticed that in the design I made the hr #e5e5e5 rather than #cdcdcd, but I guess that would be a change to the hr everywhere?

lyubomir-popov:
@spencerbygraves yes, to reenforce the text keyline. 
I also prefer #e5e5e5, but it feels as something we need to define at a more global level. SOme borders need to be more prominent than others:

- first line in a table (currently #111, but could be #cdcdcd)
- tabstrip line (I think that should remain #cdcdcd)
- hr in general (As this is a section separator, I think it should remain #cdcdcd

then there are many lines that could be lighter:
- accordion separators 
- table rows after the first one
- separators inside list-like patterns in general

bartaz:
@spencerbygraves For the border I used `#cdcdcd` as this is our light theme border color. If possible we should keep this consistent, so either keep that or update to #e5e5e5 everywhere where light theme borders are used.

And what do you say about border width? Do we want to shorten it to align with text as per @lyubomir-popov suggestion?

scss/_patterns_side-navigation.scss
bartaz: Feels a bit overkill.

The design was discussed many times and it seems for now we barely have examples that would use all 3 level of navigation. If anything really needs more nesting we could add it later.

Also nesting is just 2 not very long rules (4 precisely if you count `--icons` variant), so there is no much benefit in file size to reduce it to just one.

The complexity of adding configurable mixin that probably no one will use doesn't seem worth it. It doesn't seem like we do anything like this anywhere else. 

While flexibility and ability to configure things is great it comes with cost of maintenance and ease of use. 

I guess the biggest problem we are trying to solve with Vanilla is to keep our products consistent, so lets start with good default variants of components. If anything different is needed we can consider it on case by case basis.

spencerbygraves:
@bartaz @lyubomir-popov ok, let's keep the border as it is, it would be good to look at it's use as you mention above.

Happy to try it aligned with the text. Would we need to bear this in mind for tables? 

Thanks

lyubomir-popov:
no implications for tables, but could be a nice example to follow in the contextual menu pattern (if / when we add an example with icons)

PR #2937, opened 2020-03-18
Side navigation examples 

bartaz:
## Done

Adds more examples for side navigation

Fixes #2865

## QA

- Run `./run` or [demo]
- Clicking on side nav in these examples is possible (JS is now added to make them interactive without navigating to other pages)

Check updated side nav examples (with wrapping, truncating, icons, etc):
https://vanilla-framework-canonical-web-and-design-pr-2937.run.demo.haus/docs/examples/patterns/side-navigation/default

Screenshot 2020-03-18 at 15 50 26

MAAS current custom docs layout: https://vanilla-framework-canonical-web-and-design-pr-2937.run.demo.haus/docs/examples/templates/maas-docs

Screenshot 2020-03-18 at 13 49 03

MAAS docs in grid: https://vanilla-framework-canonical-web-and-design-pr-2937.run.demo.haus/docs/examples/templates/maas-docs-grid

Screenshot 2020-03-18 at 13 49 17

Snapcraft publicise: https://vanilla-framework-canonical-web-and-design-pr-2937.run.demo.haus/docs/examples/templates/snapcraft-publicise

Screenshot 2020-03-18 at 13 49 35

PR #2944, opened 2020-03-25
Side nav mobile toggle

bartaz:
## Done

Adds menu toggle component to be used with side navigation.

Fixes #2937

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/ or [demo](https://vanilla-framework-canonical-web-and-design-pr-2944.run.demo.haus/docs/examples/patterns/side-navigation/toggle)
- Open [side navigation menu toggle example](https://vanilla-framework-canonical-web-and-design-pr-2944.run.demo.haus/docs/examples/patterns/side-navigation/toggle)
  - menu toggle should only be visible on small screens
  - make sure on small screens navigation can be expanded/collapsed

## Screenshots

Screenshot 2020-03-25 at 08 37 13

Screenshot 2020-03-25 at 08 23 29

### DIFF
--- a/scss/_patterns_menu-toggle.scss
+++ b/scss/_patterns_menu-toggle.scss
@@ -1,0 +1,33 @@
+@mixin vf-p-menu-toggle {
+  .p-menu-toggle {
+    display: flex;
+  }
+
+  .p-menu-toggle__button {
+    // reset default button styling
+    background: none;
+    border: 0;
+    display: block;
+    line-height: inherit;
+    margin: 0;
+    margin-left: auto;
+    padding: 1rem;
+    width: auto;
+  }
+
+  .p-menu-toggle__button--hide {
+    display: none;
+
+    .p-menu-toggle__button[aria-expanded='true'] & {
+      display: block;
+    }
+  }
+
+  .p-menu-toggle__button--show {
+    display: block;
+
+    .p-menu-toggle__button[aria-expanded='true'] & {
+      display: none;
+    }
+  }
+}
--- a/scss/standalone/patterns_side-navigation.scss
+++ b/scss/standalone/patterns_side-navigation.scss
@@ -12,5 +12,18 @@
 @import '../patterns_grid';
 @include vf-p-grid;
 
+// notification used in menu toggle example
+@import '../patterns_notifications';
+@include vf-p-notification;
+
+// show/hide utilities used in mobile toggle example
+@import '../utilities_hide';
+@include vf-u-hide;
+@import '../utilities_show';
+@include vf-u-show;
+
+@import '../patterns_menu-toggle';
+@include vf-p-menu-toggle;
+
 @import '../patterns_side-navigation';
 @include vf-p-side-navigation;
--- a/templates/docs/examples/patterns/side-navigation/toggle.html
+++ b/templates/docs/examples/patterns/side-navigation/toggle.html
@@ -1,0 +1,63 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Side navigation / Mobile toggle{% endblock %}
+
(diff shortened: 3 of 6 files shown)


bartaz:
@lyubomir-popov For a bit of context.

This is meant to be based on what our docs (https://maas.io/docs) do on small screens with side navigation.

If we feel that it's not good enough to go like this into Vanilla and we want better design for mobile look and behaviour of side nav I'm OK with putting it on hold.

We can still release side navigation as it is now and let products (docs, apps, etc) do their own custom mobile collapsing or whatever.

templates/docs/examples/patterns/side-navigation/toggle.html
```diff
@@ -0,0 +1,67 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Side navigation / Mobile toggle{% endblock %}
+
+{% block standalone_css %}patterns_side-navigation{% endblock %}
+
+{% block content %}
+<div class="p-notification--caution u-hide--small">
```
lyubomir-popov: @bartaz why the notification in the example btw? why not have the copy in the md file

scss/_patterns_menu-toggle.scss
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  .p-menu-toggle__button {
+    // reset default button styling
+    background: none;
+    border: 0;
+    display: block;
+    line-height: inherit;
+    margin: 0;
+    margin-left: auto;
```
lyubomir-popov: why not use the align utilities instead of hardcoding algnment?

cassiocassio:
It feels like there's two/three difference versions of collapsed

in cases where it's a question of **responsiveness**

- small screen = show it's a menu (humburger, "menu" or other icon
- no left to right  sense of motion - it's more of a big overlay dropdown

in cases where it's a **user choice**, _like Zenhub_

- big screen, start with toggled open, see choices
- choose to collapse to get more space for content
- icon:words collapse to icons
- animation and << double-chevron emmphasise collase and expand, left to righ, right to left

command line up from the bottom
- built as toggle
- probably really want a resize-able thing. Challenge for another day

templates/docs/examples/patterns/side-navigation/toggle.html
bartaz: I thought it could be useful to have it in the example to make sure the message is shown not only in documentation text, but also if example is opened directly from `Examples` page.

I guess before we usually add notes in docs instead, so I can move it out if we feel it obstructs the example itself.

scss/standalone/patterns_side-navigation.scss
```diff
@@ -12,5 +12,18 @@
 @import '../patterns_grid';
 @include vf-p-grid;
 
+// notification used in menu toggle example
+@import '../patterns_notifications';
+@include vf-p-notification;
```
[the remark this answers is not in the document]
bartaz: These `standalone` scss files are solely for our own use to make sure the `standalone` examples have all (and only) what they need (patterns that we use in the example HTML).

I agree that the notification here is a stretch, so maybe we should leave it out of the example to the docs text.

scss/_patterns_menu-toggle.scss
bartaz: If we want this to be a pattern it should come with some default styling.
Utilities should be used to override some styling if necessary. I don't think by default we should make use of utility classes as part of patterns.

I guess the only exception would be show/hide utilities that may be used in JavaScript to toggle visibility of component, like in this case.

bartaz:
As discussed, we are parking it for now.

We don't want to release temporary solution. We will explore and design mobile responsive behaviour soon.

bartaz:
Superseded by #2977

PR #2950, opened 2020-03-25
Dark theme for side navigation dark

### LINKED ISSUE #2863: Build sidenav dark theme

bartaz:
## Done

Adds dark theme for side navigation component.
Updates themes' color variable names to be more consistent.
Updates dark theme colors to use opacity.
Consistently use text color for side navigation active item border (rather than brand color).

Fixes #2863 

## QA

- Run `./run` or [demo](https://vanilla-framework-canonical-web-and-design-pr-2950.run.demo.haus/docs/examples/patterns/side-navigation/dark)
- Review [dark side navigation example](https://vanilla-framework-canonical-web-and-design-pr-2950.run.demo.haus/docs/examples/patterns/side-navigation/dark)
- Review [light side navigation example](https://vanilla-framework-canonical-web-and-design-pr-2950.run.demo.haus/docs/examples/patterns/side-navigation/default) - because active element border color changed
- Review other dark patterns (colors were updated):
  - [main navigation](https://vanilla-framework-canonical-web-and-design-pr-2950.run.demo.haus/docs/examples/patterns/navigation/subnav-dark)
  - [contextual menu](https://vanilla-framework-canonical-web-and-design-pr-2950.run.demo.haus/docs/examples/patterns/contextual-menu/dark)
  - [hr](https://vanilla-framework-canonical-web-and-design-pr-2950.run.demo.haus/docs/examples/base/hr-dark)

Screenshot 2020-03-25 at 15 17 46

### DIFF
--- a/scss/_patterns_side-navigation.scss
+++ b/scss/_patterns_side-navigation.scss
@@ -187,7 +187,7 @@
     $color-sidenav-text-active: $colors--light-theme--text-default,
     $color-sidenav-item-background-highlight: $colors--light-theme--background-highlight,
     $color-sidenav-item-border-highlight: $colors--light-theme--text-default,
-    $color-sidenav-list-border: $colors--light-theme--border-default
+    $color-sidenav-list-border: $colors--light-theme--border-low-contrast
   );
 }
 
@@ -197,6 +197,6 @@
     $color-sidenav-text-active: $colors--dark-theme--text-default,
     $color-sidenav-item-background-highlight: $colors--dark-theme--background-highlight,
     $color-sidenav-item-border-highlight: $colors--dark-theme--text-default,
-    $color-sidenav-list-border: $colors--dark-theme--border-default
+    $color-sidenav-list-border: $colors--dark-theme--border-low-contrast
   );
 }
--- a/scss/_patterns_side-navigation.scss
+++ b/scss/_patterns_side-navigation.scss
@@ -185,18 +185,18 @@
   @include vf-side-navigation-theme(
     $color-sidenav-text-default: $colors--light-theme--text-muted,
     $color-sidenav-text-active: $colors--light-theme--text-default,
-    $color-sidenav-item-background-highlight: $colors--light-theme--background-alt,
-    $color-sidenav-item-border-highlight: $color-brand,
+    $color-sidenav-item-background-highlight: $colors--light-theme--background-highlight,
+    $color-sidenav-item-border-highlight: $colors--light-theme--text-default,
     $color-sidenav-list-border: $colors--light-theme--border-default
   );
 }
 
 @mixin vf-side-navigation-theme-dark {
   @include vf-side-navigation-theme(
-    $color-sidenav-text-default: rgba($colors--dark-theme--text-default, 0.55),
+    $color-sidenav-text-default: $colors--dark-theme--text-muted,
     $color-sidenav-text-active: $colors--dark-theme--text-default,
-    $color-sidenav-item-background-highlight: rgba($colors--dark-theme--text-default, 0.05),
+    $color-sidenav-item-background-highlight: $colors--dark-theme--background-highlight,
     $color-sidenav-item-border-highlight: $colors--dark-theme--text-default,
-    $color-sidenav-list-border: rgba($colors--dark-theme--text-default, 0.05)
+    $color-sidenav-list-border: $colors--dark-theme--border-default
   );
 }
--- a/scss/_patterns_side-navigation.scss
+++ b/scss/_patterns_side-navigation.scss
@@ -185,7 +185,7 @@
   @include vf-side-navigation-theme(
     $color-sidenav-text-default: $colors--light-theme--text-muted,
     $color-sidenav-text-active: $colors--light-theme--text-default,
-    $color-sidenav-item-background-highlight: $colors--light-theme--background-highlighted,
+    $color-sidenav-item-background-highlight: $colors--light-theme--background-alt,
     $color-sidenav-item-border-highlight: $color-brand,
     $color-sidenav-list-border: $colors--light-theme--border-default
   );
--- a/scss/_patterns_side-navigation.scss
+++ b/scss/_patterns_side-navigation.scss
@@ -96,7 +96,14 @@
   }
 
   // default light theme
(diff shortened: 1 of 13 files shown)


scss/_settings_colors.scss
```diff
@@ -45,11 +45,11 @@ $colors--light-theme--text-muted: #666 !default;
...
 
 //text-hover: minimum contrast on primary that satisfies contrast checker AA
+$colors--dark-theme--text-default: hsl(0, 0%, 100%) !default;
+$colors--dark-theme--text-disabled: rgba($colors--dark-theme--text-default, 0.55) !default;
+$colors--dark-theme--text-muted: rgba($colors--dark-theme--text-default, 0.55) !default;
+$colors--dark-theme--text-hover: hsl(0, 0%, 56%) !default;
 $colors--dark-theme--background: hsl(0, 0%, 15%) !default;
-$colors--dark-theme--background-highlighted: hsl(0, 0%, 20%) !default;
-$colors--dark-theme--border-default: hsl(0, 0%, 27%) !default;
+$colors--dark-theme--background-highlighted: rgba($colors--dark-theme--text-default, 0.05) !default;
```
bartaz: @lyubomir-popov  It feels like 5% is a bit too subtle, the background/borders are barely visible.

scss/_settings_colors.scss
```diff
@@ -45,11 +45,11 @@ $colors--light-theme--text-muted: #666 !default;
...
+$colors--dark-theme--border-default: rgba($colors--dark-theme--text-default, 0.05) !default;
```
bartaz: @lyubomir-popov 

Now that we have opacity based values here, I guess we should be consistent and define `$colors--dark-theme--text-hover` and `$colors--dark-theme--border-high-contrast` in the same way - could you suggest the values?

scss/_patterns_side-navigation.scss
```diff
@@ -163,3 +169,23 @@
...
+    $color-sidenav-list-border: $colors--light-theme--border-default
+  );
+}
+
+@mixin vf-side-navigation-theme-dark {
+  @include vf-side-navigation-theme(
+    $color-sidenav-text-default: $colors--dark-theme--text-muted,
+    $color-sidenav-text-active: $colors--dark-theme--text-default,
+    $color-sidenav-item-background-highlight: $colors--dark-theme--background-highlighted,
+    $color-sidenav-item-border-highlight: $colors--dark-theme--text-default,
```
bartaz: @lyubomir-popov 
We wanted to use brand color for this and we do in light theme.

Shouldn't we do it as well in dark theme? Although in Vanilla brand color is dark grey, so it doesn't really work good in this case...

lyubomir-popov: I think the vanilla default should focus on being accessible. Any of these variables can be assigned the value of $color-brand.

scss/_settings_colors.scss
lyubomir-popov: Yes, definitely. I'll come up with values.

scss/_settings_colors.scss
lyubomir-popov: it is what is used in jaas.ai, and they are not essential as the white-space and bar reinforce the meaning of the line / active element. For example, vscode only has a bar and not a background:

image

So I thikn it is acceptableto keep them subtle.

scss/_patterns_side-navigation.scss
bartaz: I don't think it can be easily overridden. If dark theme of side navigation uses `$colors--dark-theme--text-default` they would need to override text default to brand color to make border use that - but it will change all text in dark theme to brand color.

Maybe we need a separate color for that in every theme - like accent color? So it can be adjusted without affecting other color in a theme?

lyubomir-popov: @bartaz did you mean $color-sidenav-item-border-highlight? sorry got confused which colour we're talking about.

lyubomir-popov: just checked the tabstrip, it uses $color-mid-dark. Arguably, these should be tied together. Since they are not, I'd leave it as is at the moment and then we can do a pass and fixe everywhere if we decide we want the brand color.

bartaz: @lyubomir-popov `$color-sidenav-item-border-highlight` is a variable used inside of the `vf-side-navigation-theme` mixin. It's not a global setting, you can't change its value directly.

Currently in 'light' theme the value of this variable is `$color-brand`, so to change it, devs can change `$color-brand` value.
In dark theme we implemented it as `$colors--dark-theme--text-default`, so the only way to change it (when theme is dark) is to change `$colors--dark-theme--text-default` - but this will also affect all the text color in all the dark components.

If we want things like `$color-sidenav-item-background-highlight`, etc (every single theme color PER COMPONENT) exposed to be potentially changed individually (and not by changing shared 'theme' colors) it's a completely new feature that we currently don't have.

Not a hard thing to implement - we just need to expose these variables (with default values from theme colors), but the question is if we really need this level of granularity.
Because once we allow it and give it to users we will have to maintain it.

bartaz:
@lyubomir-popov Just had a look at percy, and it seems we are not really ready to replace dark theme colors with opacity based ones:
https://percy.io/vanilla-framework/vanilla-framework/builds/4652940

Screenshot 2020-03-26 at 07 39 49

For example, main navigation uses `$colors--dark-theme--background-highlighted` as background, rather than `$colors--dark-theme--background`.
So we can't change `$colors--dark-theme--background-highlighted` to opacity based color, as it needs to be solid dark.

To achieve what we want we would need to be much more consistent in using theme colors across components. And possibly we would need more theme color variables - for solid backgrounds, for transparent backgrounds, for borders with different steps of intensity, etc.

I'm reverting it now, so the opacity based colors will only be used in dark side navigation. But this means dark side navigation will not directly use dark theme colors, but their transparent variants. Which leads to yet another inconsistency in how we deal with themes between components.

bartaz:
@lyubomir-popov 

HR dark theme now with opacity based border color feels VERY subtle (barely visible) compared to light theme equivalent:

Screenshot 2020-03-26 at 15 35 04
Screenshot 2020-03-26 at 15 34 42

PR #2968, opened 2020-04-02
Add documentation layout example

bartaz:
## Done

Adds documentation layout build with existing Vanilla grid and classes.

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/ or [demo](https://vanilla-framework-canonical-web-and-design-pr-2968.run.demo.haus/docs/examples/)
- Go to [documentation layout example](https://vanilla-framework-canonical-web-and-design-pr-2968.run.demo.haus/docs/examples/layouts/documentation-layout)
  - check if it renders correctly

## Screenshots

Screenshot 2020-04-06 at 11 57 55

### DIFF
--- a/scss/_layouts_docs.scss
+++ b/scss/_layouts_docs.scss
@@ -1,25 +1,35 @@
 @mixin vf-l-docs {
+  .l-docs .l-content,
   .l-fixed-width--with-sidebar {
     @extend %vf-row;
   }
 
-  .l-sidebar,
-  .l-content {
+  .l-aside,
+  .l-main {
     @extend %span-full-grid-on-mobile;
     @extend %span-full-grid-on-tablet;
     @extend %display-block;
   }
 
   @media (min-width: $threshold-6-12-col) {
+    .l-docs .l-content,
     .l-fixed-width--with-sidebar {
+      // we use 12 column grid from %vf-row
       grid-template-areas: 'side side side main main main main main main main main main';
     }
 
-    .l-sidebar {
+    .l-docs .l-aside, // FIXME: fighting specificity
+    .l-aside {
       grid-area: side;
+
+      // content area takes 3 columns, so we reset nested rows to use 9 columns as well
+      & .row {
+        grid-template-columns: repeat(3, minmax(0, 1fr));
+      }
     }
 
-    .l-content {
+    .l-docs .l-main, // FIXME: fighting specificity
+    .l-main {
       grid-area: main;
 
       // content area takes 9 columns, so we reset nested rows to use 9 columns as well
--- a/scss/_layouts_docs.scss
+++ b/scss/_layouts_docs.scss
@@ -1,0 +1,34 @@
+/// Animations
+@mixin vf-l-docs {
+  //.l-docs {}
+
+  .l-docs__wrapper {
+    @extend %fixed-width-container;
+
+    @media (min-width: $breakpoint-navigation-threshold) {
+      display: flex;
+      flex-wrap: wrap;
+    }
+  }
+
+  .l-docs__sidebar {
+    background: transparentize($color-positive, 0.9);
+
+    flex-basis: 16rem;
+    flex-grow: 0;
+    flex-shrink: 0;
+    margin-right: map-get($grid-gutter-widths, large);
+  }
+
+  .l-docs__content {
+    background: transparentize($color-positive, 0.9);
+
+    flex-basis: 0;
+    flex-grow: 3;
+    flex-shrink: 1;
+
+    .row {
+      padding: 0;
+    }
+  }
+}
--- a/scss/_layouts_docs.scss
+++ b/scss/_layouts_docs.scss
@@ -1,34 +1,31 @@
-/// Animations
 @mixin vf-l-docs {
-  //.l-docs {}
+  .l-fixed-width--with-sidebar {
+    @extend %vf-row;
+  }
 
-  .l-docs__wrapper {
-    @extend %fixed-width-container;
+  .l-sidebar,
+  .l-content {
+    @extend %span-full-grid-on-mobile;
+    @extend %span-full-grid-on-tablet;
+    @extend %display-block;
+  }
 
-    @media (min-width: $breakpoint-navigation-threshold) {
-      display: flex;
-      flex-wrap: wrap;
+  @media (min-width: $threshold-6-12-col) {
+    .l-fixed-width--with-sidebar {
+      grid-template-areas: 'side side side main main main main main main main main main';
     }
-  }
 
-  .l-docs__sidebar {
-    background: transparentize($color-positive, 0.9);
+    .l-sidebar {
+      grid-area: side;
+    }
 
-    flex-basis: 16rem;
-    flex-grow: 0;
-    flex-shrink: 0;
-    margin-right: map-get($grid-gutter-widths, large);
-  }
+    .l-content {
+      grid-area: main;
 
-  .l-docs__content {
-    background: transparentize($color-positive, 0.9);
-
-    flex-basis: 0;
-    flex-grow: 3;
-    flex-shrink: 1;
-
-    .row {
-      padding: 0;
+      // content area takes 9 columns, so we reset nested rows to use 9 columns as well
+      & .row {
+        grid-template-columns: repeat(9, minmax(0, 1fr));
+      }
     }
   }
 }
--- a/scss/_vanilla.scss
+++ b/scss/_vanilla.scss
@@ -66,6 +66,8 @@
 @import 'utilities_vertically-center';
 @import 'utilities_no-print';
 
+@import 'layouts_docs';
+
 // Include all the CSS
 @mixin vanilla {
   @include vf-base;
@@ -136,4 +138,6 @@
   @include vf-u-vertically-center;
   @include vf-u-visualise-baseline;
   @include vf-u-no-print;
+
+  @include vf-l-docs;
 }
--- a/templates/_layouts/docs.html
+++ b/templates/_layouts/docs.html
@@ -1,12 +1,14 @@
 {% extends "_layouts/_root.html" %}
 
+{# TODO: move to layout #}
 {% block custom_head %}
-    <link rel="stylesheet" href="/static/build/css/docs/docs.css" />
+    <link rel="stylesheet" href="/static/build/css/docs/site.css" />
 {% endblock %}
 
 {% block body %}
+    {# TODO: move to layout #}
     <header id="navigation" class="p-navigation">
-      <div class="p-navigation__row--full-width">
+      <div class="p-navigation__row">
         <div class="p-navigation__banner">
           <div class="p-navigation__logo">
             <a class="p-navigation__item" href="/">
@@ -40,8 +42,10 @@
     </header>
 
     <hr class="u-no-margin u-hide--small">
-    <div class="docs-container">
-      <aside class="p-sidebar">
+
+    <div class="p-strip is-shallow is-bordered">
+    <div class="l-fixed-width--with-sidebar">
+      <aside class="l-aside">
         <nav class="p-side-navigation">
 
             <ul class="p-side-navigation__list">
@@ -149,19 +153,15 @@
 
       </aside>
 
-      <main class="p-content" id="main-content">
+      <main class="l-main" id="main-content">
         {% block banner %}{% endblock %}
-        <div class="p-strip is-shallow">
-          <div class="p-content__row">
-            {% block content %}{{ content | safe }}{% endblock content %}
-          </div>
-        </div>
 
-        <hr>
+        {% block content %}{{ content | safe }}{% endblock content %}
 
-        {% include "_layouts/_footer.html" %}
       </main>
 
+      {# TODO: remove or reimplement page TOC #}
+      <!--
       <aside id="side-content" class="u-hide--small p-aside js-contents" style="display: none">
         <div class="p-aside__section">
           <h4 class="p-aside__header">Contents</h4>
@@ -170,6 +170,7 @@
           </nav>
         </div>
       </aside>
+      -->
 
       <script>
         let aside = document.querySelector('.js-contents');
@@ -198,6 +199,9 @@
         }
       </script>
     </div>
+    </div>
+
+    {% include "_layouts/_footer.html" %}
     <script src="/static/js/scripts.js"></script>
     <script defer src="https://static.codepen.io/assets/embed/ei.js"></script>
     <script defer src="/static/js/example.js"></script>
--- a/templates/_layouts/examples.html
+++ b/templates/_layouts/examples.html
@@ -68,7 +68,7 @@
     </script>
   </head>
 
-  <body>
+  <body class="{% block body_class %}{% endblock %}">
     {% block content %}{% endblock %}
   </body>
 </html>
--- a/templates/docs/examples/index.html
+++ b/templates/docs/examples/index.html
@@ -45,6 +45,16 @@
     </nav>
   </div>
   <div class="col-3">
+    <h3>Layouts</h3>
+    <nav>
+      <ul class="p-list">
+        {% for example in examples.layouts %}
+          <li class="p-list__item">
+            <a href="/docs/examples/{{ example.path }}" class="p-link">{{ example.title }}</a>
+          </li>
+        {% endfor %}
+      </ul>
+    </nav>
     <h3>Templates</h3>
     <nav>
       <ul class="p-list">
--- a/templates/docs/examples/layouts/documentation-layout-2.html
+++ b/templates/docs/examples/layouts/documentation-layout-2.html
@@ -1,7 +1,9 @@
 {% extends "_layouts/examples.html" %}
-{% block title %}Documentation - Layout{% endblock %}
+{% block title %}Documentation - Layout with regions{% endblock %}
 
 {% block content %}
+<div class="l-header">
+
 <header id="navigation" class="p-navigation is-dark">
   <div class="p-navigation__row">
     <div class="p-navigation__banner">
@@ -29,6 +31,10 @@
   </div>
 </header>
 
+</div>
+
+<div class="l-hero">
+
 <section id="search-docs" class="p-strip--light is-shallow">
   <div class="row">
     <form class="p-search-box u-no-margin--bottom" action="#">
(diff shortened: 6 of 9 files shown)


bartaz:
@anthonydillon Do you think it would be enough to use existing `row` `col-*` to build the layout?

Or should we start adding `l-content-wrapper` (= `row`)  `l-sidebar` (= `col-3`) `l-content` (= `col-9`), etc?

anthonydillon:
@bartaz nice work, but I can think of a few benefits from adding additional layout classes to the mark up:
- It gives the framework the ability to adjust the layout from a central place without requiring site updates. For example, if we decide to increase the sidebar width we can do it once and it will apply in the next update.
- It allows us to decouple the grid from the layout. The grid is to keep regions of the same document visually aligned. But, when it comes to sidebar and main content they do not need to align to a grid as they will remain aligned all the way down the page by default.
- Ideally, we should initialise the grid within the content area so we get the full 12-cols within the main content area. I understand this may cause issues with media queries and the like but I would like us to give it a go.
- We may want to provide features such as collapsing sidebar which would act as a modifier on the sidebar layout class. This becomes difficult to do using the Vanilla grid.
- Providing bespoke layouts using CSS gird is very easy to read, understand and maintain.

Let me know if you want to chat these through.

lyubomir-popov:
@anthonydillon said it all, here's a sidebar that appears/disappears/changes width depending on the case: https://zpl.io/a7879ZM

Also, having a wrapper class allows introduing responsive changes without having to add tons of classes.

lyubomir-popov:
@bartaz @anthonydillon The one point I'm not convinced about is having 12 columns in the main area. What is this useful for when the width there is ~800pixels at best?

anthonydillon:
> What is this useful for when the width there is ~800px at best?

Alignment always matters and it shouldn't matter what the container width would be? Are you suggesting we don't use the grid at all the in main content? I agree that the main use case will be block content from a data source but we should allow developer to rely on a grid when creating content.

> If you had a documentation with a brochure-style facade as in vanillaframework.io, content on the brochure part would be impossible to align with content inside the docs, as the sidebar would be off the grid.

What do you mean by brochure-style facade? Do we need things outside of the main content to align with the layout? No. For example, the navigation could align with the sidebar on the left or not either is fine as they are totally different regions.

> In my opinion, having a sidebar that is off the grid is only useful if you're in a full-width layout.

Why is that? Surely it's useful to allow flexibility of global layout widths that are not limited to the grid.

> ... we discussed at length in the global navigation project, this is only applicable to certain types of sites like apps, and causes problems on brochure sites.

Sorry, I must have missed this or forgotten the discussion. Is there a link to the discussion to catch me up?

lyubomir-popov:
> Alignment always matters and it shouldn't matter what the container width would be? Are you suggesting we don't use the grid at all the in main content? I agree that the main use case will be block content from a data source but we should allow developer to rely on a grid when creating content.

I just think 8 or 9 columns is enough in the main content area. The number of columns is generally the least common denominator of  the different layouts that might be used, so 8 or 9 col gives you plenty of options for a main content docs column. 

> What do you mean by brochure-style facade? Do we need things outside of the main content to align with the layout? No. For example, the navigation could align with the sidebar on the left or not either is fine as they are totally different regions.

Well, we may need things outside the main content area to align. A bit contrived, but not unthinkable example:

image

If the sidebar is using 3 grid columns, then it's easy to align something in the footer or header for example to the main column. If ti is defined as width: 18rem, then we'd need to be able to offset by the same value in other places, or we would get slight misalignments as in the image above. 

> > In my opinion, having a sidebar that is off the grid is only useful if you're in a full-width layout.
> 
> Why is that? Surely it's useful to allow flexibility of global layout widths that are not limited to the grid.

This just made me realise we need both options. For jaas.ai/models, as it is full width, the sidebar  should be a fixed value like width: 18rem. For a fixed width documentation site, it would be far easier to align things if the sidenav is say 3 columns, main content 9 columns. 


> Sorry, I must have missed this or forgotten the discussion. Is there a link to the discussion to catch me up?

[This](https://docs.google.com/presentation/d/19-1TsmFiFgWZBec7IcEFSZKwkPKahazxlcRcw96jPxE/edit#slide=id.g493de585b3_0_93) is the guideline we wrote at the end of the discussion, can't find links to the multiple threads where the actual discussion happened, but the gist of it was:

- site width determines nav width
- docs etc follow main (brochure) site width, which for cases like snapcraft and maas.io etc. means fixed width

anthonydillon:
> I just think 8 or 9 columns is enough in the main content area. The number of columns is generally the least common denominator of the different layouts that might be used, so 8 or 9 col gives you plenty of options for a main content docs column.

Sure it might be enough but why change what the developer/author is used to having and have to provide an entirely new set of grid documentation for?

> Well, we may need things outside the main content area to align. A bit contrived, but not unthinkable example:

This alignment should derive from the layout not the grid. We will end up in a bad place if people start mixing grid usage and layout usage. We will end up with no confidence in the layout we are effecting with changes.

> This just made me realise we need both options. For jaas.ai/models, as it is full width, the sidebar should be a fixed value like width: 18rem. For a fixed width documentation site, it would be far easier to align things if the sidenav is say 3 columns, main content 9 columns.

The JAAS dashboard will have its own layout which may in time become a centralised application layout provided by Vanilla. This layout is likely to involve a double sidebar with a main content area. How do we use the grid in that situation? The grid should reliably provide 12 columns.

When it comes to the global navigation, that is entirely different as it is a module that is applied independently from the site/application or styling.

lyubomir-popov:
> This alignment should derive from the layout not the grid. We will end up in a bad place if people start mixing grid usage and layout usage. We will end up with no confidence in the layout we are effecting with changes.

Does that mean a footer layout would be needed where you get sidenav-width offset using a modifier? That can work, but you wouldn't be able to have e.g. the 3-column divider pattern align with the main column. 

> The JAAS dashboard will have its own layout which may in time become a centralised application layout provided by Vanilla. This layout is likely to involve a double sidebar with a main content area. How do we use the grid in that situation? The grid should reliably provide 12 columns.

The jaas layout is already full width, so it doesn't need to put the sidenav on the grid and is free to use 12 columns in the main area. My comments are limited to the fixed-width docs as part of a fixed-width brochure site like maas.io or snapcraft.

lyubomir-popov:
@anthonydillon thinking about this some more, I think you're right. If existing layouts make use of 12 columns in the main area, reducing the number of columns would break them. Haven't seen any and thought it would be safe to do, but if you have, then I guess what I'm proposing is out of the question.

anthonydillon:
> Does that mean a footer layout would be needed where you get sidenav-width offset using a modifier? That can work.

I am suggesting the documentation layout (_there is probably a better name for it_) would position the footer inline with the main-content if it was required for the design. There would be no need to modifiers as layouts would be set.

> I guess the only issue then is that you wouldn't be able to have e.g. the 3-column divider pattern, where the second column aligns with the main column. Granted, not a major concern:

If the layout supported a second content region is could offset the navigation width and initialise a grid aligning with the main-content.

lyubomir-popov:
> would position the footer inline with the main-content

Question is how does this work when the sidenav appears on only some of the pages, while the footer appears on all, as in maas.io for example. Not a difficult question to answer, just a small limitation for how content is laid out within. 

But given the prospect of breaking existing pages, I guess my proposal is a non-starter. We'll find a way to work around the various small issues as needed.

anthonydillon:
> Question is how does this work when the sidenav appears on only some of the pages, while the footer appears on all, as in maas.io for example. Not a difficult question to answer, just a small limitation for how content is laid out within the footer.

They would use different layouts as they do now. The docs would use the docs layout and the rest of the site would either use no layout or a brochure one if needed.

lyubomir-popov:
Yes, this is how it would be implemented, which is fine, I'm talking more about the visual side of things. You don't expect the footer to change between pages of the same site, and indeed it doesn't on maas.io. The only difference between the two pages below is that on the homepage, main content and footer grid align, whereas on the docs page they don't. Again - not much of an issue and I guess this is how it will have to be given the alternative risks breaking stuff. 

maas.io:

image

maas.io/docs:

image

bartaz:
For the sake of experimentation and demo for discussion I made a quick'n'dirty version of documentation layout with full 12 col grin within main content:
https://vanilla-framework-canonical-web-and-design-pr-2968.run.demo.haus/docs/examples/layouts/documentation-layout

Screenshot 2020-04-03 at 08 02 51

Below docs content (sidebar + main) I added standard grid as well for comparison, as I guess there may be cases where 12 col grid is used also in footer, additional content below sidebar or something like that.

I suppose that's the direction we want to go?

lyubomir-popov:
@bartaz I see you've made the sidenav nearly match 3 col on desktop once the window width is bigger than the max-width. 

What if we have 12 columns in the main area, but nested within a 9-col parent? this way it would be best of both worlds unless I'm missing something:
sidenav 3 col, main area 9 col wide, but inside, we still hae 12 columns to avoid breaking layouts placed inside that rely on having 12 columns.

bartaz:

I updated the branch with layout implementation that hopefully serves as a good compromise between keeping in line with brochure site grid and having new layout class names for simplicity of implementation in projects.

The current implementation introduces `l-fixed-width--with-sidebar` layout.
This may feel a bit overly expressive class name, but is based on the idea that we will likely have our brochure site (centered) layout as `l-fixed-width` (to replace existing `u-fixed-width` used for similar purpose). Potentially (apart from `with-sidebar` used for docs) this layout may in future have additional modifiers (2 sidebars for example).
The 'app' layout in this convention may be called `l-full-width`.

Children of `l-fixed-width` are `l-sidebar` and `l-content`. Vanilla controls sizing and positioning of those and makes sure they are aligned to the grid.
As per current designs sidebar uses 3 columns, content 9 columns. This number of columns is available for any nested rows.

Let me know if this shape of things is OK. I'll work on improving the docs and cleaning the implementation a bit.

lyubomir-popov:
- l-sidebar or l-aside?
- l-content or l-main? - I have no preference, just thought it's worth discussing
- a utility to obtain 12 columns inside the main column for compatibility with layouts that assume 12 col are available in the main column?

bartaz:
> * l-sidebar or l-aside?
> * l-content or l-main? - I have no preference, just thought it's worth discussing

Both work for me. But if we call is `l-aside` should it be `l-fixed-width--with-aside` then?
`l-main` and `l-aside` are shorter than `l-sidebar` `l-content`, so we may go with that.

> * a utility to obtain 12 columns inside the main column for compatibility with layouts that assume 12 col are available in the main column?

Personally I wouldn't want to add it if we really have an example where we need it that cannot be easily adapted to correct use of nested grid.
We are introducing new class names and a whole new concept to Vanilla. To use it projects need to adapt their HTML anyway, so I think we can expect them to nest the grid if they need one.

If we will have a project that will be blocked on introducing this layout because of nested grid we can think about adding workaround at that point.

lyubomir-popov:
> Personally I wouldn't want to add it if we really have an example where we need it that cannot be easily adapted to correct use of nested grid.

Maybe it's just me being overly cautious as this is not a major release and I don't wan't breaking changes, but you're right - this is a new thing so maybe the rules can be relaxed.

lyubomir-popov:
How do we feel about responsiveness? Between 620 and 1040, do we want the aside column 2-col wide, or stacking on top of the main one?

anthonydillon:
I like the use of `row` and `col-*` classes providing the 12 columns in the main content. 

A `Layout` should contain and define all regions of the page. In the demo, that would include `header`, `search, `footer`.

scss/_layouts_docs.scss
```diff
@@ -0,0 +1,31 @@
+@mixin vf-l-docs {
+  .l-fixed-width--with-sidebar {
+    @extend %vf-row;
+  }
+
+  .l-sidebar,
+  .l-content {
```
anthonydillon, line 7: These should be nested in the main layout class to avoid cross layout overspill

scss/_layouts_docs.scss
```diff
@@ -0,0 +1,31 @@
...
+    @extend %span-full-grid-on-tablet;
+    @extend %display-block;
+  }
+
+  @media (min-width: $threshold-6-12-col) {
+    .l-fixed-width--with-sidebar {
+      grid-template-areas: 'side side side main main main main main main main main main';
+    }
+
+    .l-sidebar {
```
anthonydillon, line 18: Same here

scss/_layouts_docs.scss
```diff
@@ -0,0 +1,31 @@
...
+  .l-sidebar,
+  .l-content {
+    @extend %span-full-grid-on-mobile;
+    @extend %span-full-grid-on-tablet;
+    @extend %display-block;
+  }
+
+  @media (min-width: $threshold-6-12-col) {
+    .l-fixed-width--with-sidebar {
+      grid-template-areas: 'side side side main main main main main main main main main';
```
anthonydillon, line 15: Can you use `repeat` here? For example:
```css 
.l-fixed-width--with-sidebar {
  grid-template-areas: "repeat(3, 'side') . repeat(8. 'main')";
}
```

bartaz:
> I like the use of row and col-* classes providing the 12 columns in the main content.

We experimented with that before, but current implementation doesn't do that.
The whole page layout is 12 cols which is split between sidebar (currently 3 cols) and main (currently 9 cols). Any nested rows only have as many cols as their parent layout.

Because for fixed-width brochure site 12 col grid defines the layout and overall alignment of things on whole page we want to make sure that everything (even nested) is placed on the same grid.
We don't want to reset the grid to 12 columns  if the main content is supposed to take 9 columns.

It would be fine if there is only 1 region in layout that uses the grid - for example main content region in 'app' layout.
But on brochure/docs site - whole page uses grid. If main content uses only part of the grid, nested rows need to do so as well. Because main content is not a separate thing. It needs to be aligned with rest of the content on the page (other regions/strips below and above it).

> A `Layout` should contain and define all regions of the page. In the demo, that would include `header`, `search, `footer`.

@anthonydillon I think we still have a bit different understanding of how to implement this.

Your idea of layout defining all the regions on the page works perfectly for full width application layouts.
It doesn't seem to work well with our current approach for 'brochure' site layouts (which also include documentation).
These pages are build of full-width strips with fixed-width grid inside them.

How do we define regions for that? Each strip is a separate region? There is one 'main content' region (what is just a full width block) and we put all content strips inside of it?
How does docs layout (that has sidebar and content within the fixed-width grid) work with that if it needs to be implemented INSIDE the strip.
Do we allow layout regions to be put inside other components (like strips)? or should they be only top level elements?

Maybe `p-strips` should be a layout components - but they are not really regions - they provide spacing and background, etc.

scss/_layouts_docs.scss
bartaz: Not really, I tried that. It's only for `grid-template-columns` or `grid-template-rows` where these functions are available.
With `grid-template-areas` we use the row/col pattern defined (12 columns in our case) and we need to fill it with areas. To span things across cells you need to repeat in the template.

scss/_layouts_docs.scss
bartaz: Yes, I'm thinking how to do that.
Technically should be `l-fixed-width__sidebar`. But because sidebar works only for `--with-sidebar`, should it be `l-fixed-width--with-sidebar__sidebar` - this feels overkill and wrong.

So maybe using `--with-sidebar` is wrong. Maybe it should just be a top level layout, not a modifier?
`l-fixed-width-with-sidebar` `l-fixed-width-with-sidebar__sidebar` `l-fixed-width-with-sidebar__content`?

Avoiding calling it `l-docs` to have more generic name that is not tied to one use case makes these class names very verbose and long. I'm struggling with giving them short and descriptive names.

bartaz:
@anthonydillon Also, for the `header` `footer` regions.

Why do we need them? Our patterns (main navigation, strip that is used for footer) perfectly serve their purpose. 
Should we add `l-something__header` `l-something__footer` just for the sake of having these class names, if they don't add any styling?

lyubomir-popov:
@bartaz having footer as part of the layout nicely solves the sticky footer problem

bartaz:
> @bartaz having footer as part of the layout nicely solves the sticky footer problem

Yes, that's why I added sticky footer to layouts work.
But, technically speaking, currently footer doesn't need any additional class or wrapper UNTIL it needs to be sticky.

And having sticky footer on brochure sites, would also mean brochure site starts to use layouts.

But, lets have this discussion once we get to sticky footer case.

lyubomir-popov:
how about:

full width apps:
`l-header + l-aside + l-main + l-footer`
then inside the main area in the app:
`p-strip > row > col-*` classes


fixed-width sites with sidebar:
`l-header + l-main + l-footer` - the outer, vertical partitioning into 3 areas, which gives you sticky footer, then:

`l-main > p-strip > row > l-fixed-width__aside + l-fixed-width__main` - the use of custom classes allowing cross website, responsive, easily adjustable behaviour via css

bartaz:
> `l-main` > `p-strip` > `row` > `l-fixed-width__aside` + `l-fixed-width__main` - the use of custom classes allowing cross website, responsive, easily adjustable behaviour via css

@lyubomir-popov Yes, that's more or less what I have in mind to try. I'll push another example that will have more top level layout regions.

bartaz:

I added 2 more examples (hard refresh to make sure styles are reloaded) with 'regions'.

Let's keep the naming of classes aside for now, just getting overall feel how things may be structured.

### 2

https://vanilla-framework-canonical-web-and-design-pr-2968.run.demo.haus/docs/examples/layouts/documentation-layout-2

This one simply adds wrappers around parts of the page to create regions.
There is no 'central' layout. The root 'regions' are simply wrappers with no styling (100% width).
The layout itself (sidebar / main content) is still done with 'child' layout elements.
There is hardly a way around that - we need both full width wrapper and another one to keep 'fixed-width' of the site.


### 3

https://vanilla-framework-canonical-web-and-design-pr-2968.run.demo.haus/docs/examples/layouts/documentation-layout-3

This one ties to explore adding 'layout' class name on body, this class name may affect placement of regions. In this case it only affects content (gives it styles to handle sidebar and main content).
Because `l-content` region in this case is fixed width, if we want to give it more space by using p-strip), we need to wrap the region in the strip.

---

Not sure if any of these is better or worse than initial (`#1`) example (https://vanilla-framework-canonical-web-and-design-pr-2968.run.demo.haus/docs/examples/layouts/documentation-layout), but just leaving it here for further discussion.

@anthonydillon Does any of these seem closer to what you think the layout structure should be?

anthonydillon:
If it helps here is a very quick example I was thinking: https://codepen.io/anthonydillon/pen/oNXKjed

bartaz:
Blocked - Waiting for side navigation responsive styles to be implemented.

bartaz:
Superseded by #2996

PR #2977, opened 2020-04-09
Build responsive side navigation drawer

### LINKED ISSUE #2934: Build side nav mobile toggle
Build basic side nav mobile toggle based on current sidenav toggle in docs (maas.io/docs).

bartaz:
## Done

Adds responsive styles for side navigation.
See **Tasks** below for details on specific areas touched.

Fixes #2934

## QA

Run `./run` or [demo](https://vanilla-framework-canonical-web-and-design-pr-2977.run.demo.haus/docs/examples/templates/maas-docs-grid)

Check steps below on the example template page:

- load docs on large screen, sidenav should be in the grid
- resize to small screen, sidenav should hide
- toggle sidenav open - it should open
- toggle it back
- resize screen back to large - snap back to grid
- resize to small, open side nav, resize to large with open side nav
- make screen small (height and width) so sidenav needs to be scrollable, check if that works
- QA small screen on real phone

## Screenshots

Screenshot 2020-04-15 at 15 02 39

### Tasks

- [x] move colors to themes
- [x] dark theme
  - what to do with buttons? (buttons don't have themes)
- [x] buttons (use p-button vs own pattern classes)
- [x] icons for < >
- [ ] expand/collapse aria
  - [x] added `aria-controls`
  - [ ] `aria-hidden`/`aria-expanded` turns out problematic because technically we would need to change it on resize
- [x] don't animate when resizing the screen
  - possible but requires a bit of JS to toggle additional class and listen to resize event (see commit https://github.com/canonical-web-and-design/vanilla-framework/pull/2977/commits/54a488f042fa6a24875f56ed7f9cf0f45ba0f43f)
- [x] when sidenav is expanded on small screen, when you resize to bigger screen sidenav goes to content, but overlay stays open
- [ ] overflow hidden on body when drawer is open (?)
   - this will need to react to resize as well to prevent situation when JS adds this to body but doesn't take it away on resize
- [x] close on overlay click


anthonydillon:
@bartaz This is looking awesome! great work. I have a few things if they help:
- The Toggle doesn't look like it aligns with the content because the base button is adding some padding. @lyubomir-popov I am not sure minus margin is good here as the button has a hover color so maybe `p-button--neutral` would be better?
image

- When crossing the breakpoint from large to small you see a quick flash of animation.
- Clicking on the overlay should close the menu too.

All small things but thought I should mention them.

bartaz:
@anthonydillon Thanks for comments. I'm aware of these issues.

> * The Toggle doesn't look like it aligns with the content because the base button is adding some padding. @lyubomir-popov I am not sure minus margin is good here as the button has a hover color so maybe `p-button--neutral` would be better?

I just used buttons cause they were straightforward here. I'm not sure if the design assumed these will be buttons or bigger clickable area.

It would make sense to make whole this area clickable on mobile:
Screenshot 2020-04-10 at 08 17 07

Neutral button works in content (as border aligns):
Screenshot 2020-04-10 at 08 16 27

but on the sidenav it feels busy with all the borders:

Screenshot 2020-04-10 at 08 16 33

I guess the content button could be neutral (to keep border) and sidnav close button can be base (to hide borders).

We can adjust these on design review level.

But also, because they are simply a buttons, we can just put them in the example code, but they are external patterns. So project will be free to use any other button anyway.
Unless we want to codify it in the pattern, using custom class names (not `p-button--`) to make sure these styles are always controlled by side nav.

I'm not sure if that's something we've been doing in Vanilla before. I guess other patterns with buttons just use `p-button`.

> * When crossing the breakpoint from large to small you see a quick flash of animation.

Yes, I noticed that, but there is no easy way around that.
We want sidenav hiding / showing animated. So it animates when hiding 'automatically' on resizing as well.
To prevent it we would need to have additional class name for hiding with animation. And JS for toggling would need to be aware of these 3 states of just two.

I'll experiment, but I wouldn't like to overcomplicate JS to avoid that flicker.

> * Clicking on the overlay should close the menu too.

Of course.

anthonydillon:
I would like to see the buttons codified into the pattern to give us consistent styling across projects and a clean upgrade path for the component.

Yes, agree we should not over complicated the js here just worth an issue to think of elegant solutions.

bartaz:
> I would like to see the buttons codified into the pattern to give us consistent styling across projects and a clean upgrade path for the component.

@anthonydillon Agreed. Having a quick look it seems like that would require a dependancy on `_patterns_button.scss` - because that's where different flavors of buttons are defined.
But they are all defined as mixins, so maybe it's a good time to extract those button mixins to some shared global/base space? So any component can consistently reuse button mixins without relying on `p-button` class names?

anthonydillon:
Do you mean to achieve something like this?
p-side-navigation
```scss
@mixin side-navigation {
  @import "p-buttons";

  .p-side-navigation__toggle {
    @extend vf-base-button;
    ....
  }
}
```

bartaz:
Yes, but to avoid pattern dependency do it as:

```scss
@mixin side-navigation {
  @import "base-button-definitions";

  .p-side-navigation__toggle {
    @extend vf-base-button;
    ....
  }
}
```

anthonydillon:
Let's catch up on Tuesday. I would like to keep the styles for buttons in the buttons file. Otherwise it becomes difficult to maintain and for other developers to know where the styles are defined. Let's see if we can define the styles within the buttons file but make them available globally. This could be a change to the method of inclusion.

lyubomir-popov:
@bartaz can we use the base button (no border)? Also can we apply a compensation margin so the icon, not the button edge, aligns with left edge of things above/under?

bartaz:
@lyubomir-popov We can, but should we?

I think @anthonydillon specifically mentioned using base button (with border) to avoid compensating with negative margin.

Negative margin makes the button stick out on hover:


Screenshot 2020-04-14 at 16 06 51

We shouldn't override styling with negative margins and other overrides. If given style of the button doesn't work in given context we should use different button style.
Or - if any button style doesn't work for us - maybe it shouldn't be a button?

lyubomir-popov:
@bartaz you're right. I would vote for note using a button. On mobile, the latest requested change floats the icon to the right, and the whole area should be clickable.

To indicate a hover, the icon colour could change from #666 to #111: 
image

What do you think @spencerbygraves ?

bartaz:
> @bartaz you're right. I would vote for note using a button. On mobile, the latest requested change floats the icon to the right, and the whole area should be clickable.

@lyubomir-popov @spencerbygraves 

Technically it should still be a button (from HTML point of view). Of course, we can remove all default styling from it and pretend it's not a button.
But, why can't we use one of a handful of existing button styles? Do we need to invent new styles for a 'clickable thing' just for this toggle? It is that different from other buttons we have on the page?
On hand we want to keep things consistent across patterns and project, do we need to invent new styles here?

If this toggle shouldn't look like a button - what about the other one (on the top of open side navigation menu) - I suppose it should have the same style?

lyubomir-popov:
> But, why can't we use one of a handful of existing button styles?

If we decide the alternatives are too costly, I can live with what is currently implemented.
I think a base button (no outline, icon visibly misaligned) is a lesser compromise than a bordered button.

I also think this is a problem we've come across before, in different contexts so maybe we can use this case as an opportunity to discuss it in a little more detail.

1. For example, if you look at the snapcraft Products menu and User name dropdown, we have a similar misalignment. Both buttons include padding, which is proportional to the button size, and both are offset from the edge of the fixed width container by different amounts:
image

2. Jaas.ai/models - similar problem, button and label not aligned:
image

One way to fix these would be to have a modifier button/menu/select etc that doesn't include a hover background, and thus doesn't require left/right padding.

templates/docs/examples/templates/maas-docs-grid.html
```diff
@@ -52,113 +31,119 @@
...
-                <a class="p-side-navigation__link" href="#">Upgrade from 2.3 to 2.4</a>
-              </li>
-              <li class="p-side-navigation__item ">
-                <a class="p-side-navigation__link" href="#">Upgrade from 1.9 to 2.x</a>
-              </li>
-            </ul>
-          </li>
-        </ul>
+        <button class="p-side-navigation__toggle js-drawer-toggle" aria-controls="drawer">
+          <span>Toggle side navigation</span>
```
anthonydillon, line 55: Is this span required?

scss/_patterns_side-navigation.scss
```diff
@@ -1,15 +1,109 @@
...
+
+  .p-side-navigation__toggle {
+    &::before {
+      transform: rotate(90deg);
+    }
+  }
+
+  .p-side-navigation__toggle--in-drawer {
+    &::before {
+      transform: rotate(-90deg);
```
anthonydillon: I think these transforms are the wrong way around.

scss/_settings_colors.scss
```diff
@@ -77,6 +79,7 @@ $colors--dark-theme--text-muted: rgba($colors--dark-theme--text-default, 0.55) !
 $colors--dark-theme--background-default: hsl(0, 0%, 15%) !default;
 $colors--dark-theme--background-alt: hsl(0, 0%, 20%) !default;
 $colors--dark-theme--background-highlight: rgba($colors--dark-theme--text-default, 0.05) !default;
+$colors--dark-theme--background-overlay: transparentize($color-dark, 0.15) !default;
```
anthonydillon, line 82: Should this switched to `$color-light` to add contrast to the dark theme version?

templates/docs/examples/templates/maas-docs-grid.html
```diff
@@ -52,113 +31,119 @@
...
+        </button>
+
+        <div class="p-side-navigation__overlay js-drawer-toggle" aria-controls="drawer"></div>
+
+        <div class="p-side-navigation__drawer" id="drawer">
+          <div class="p-side-navigation__drawer-header">
+            <button class="p-side-navigation__toggle--in-drawer js-drawer-toggle" aria-controls="drawer">
+              <span>Toggle side navigation</span>
```
anthonydillon, line 63: Is this span required?

scss/_patterns_side-navigation.scss
```diff
@@ -1,15 +1,109 @@
...
+    }
+  }
+
+  .p-side-navigation__drawer-header {
+    border-bottom-style: solid;
+    border-bottom-width: 1px;
+    padding-bottom: $spv-inner--small;
+    padding-left: $sph-inner--small;
+    padding-top: $spv-inner--small;
+    position: sticky;
```
anthonydillon, line 49: Nice touch :ok_hand:

scss/_settings_colors.scss
bartaz: @lyubomir-popov ? What should be the color of overlay in dark theme?

templates/docs/examples/templates/maas-docs-grid.html
bartaz: `p-button has-icon` requires span, and this is based on it.
But because we always have icon on the left I think `<span>` may not be needed. I'll have a look. 

Good catch, thanks.

scss/_settings_colors.scss
lyubomir-popov: In the design, the overlay styling was taken from the RBAC overlay, which is 40% #111 for the light theme.  

Inverting to light overlay seems strange, the background has a high contrast:
image

Dark p strip underneath not ideal too at this opacity:
image

I think dark overlay is better for both, and I much prefer the .85 opacity used by @bartaz. In rbac, the user is supposed to be able make out what is underneath while the overlay is on, but here that's not necessary. So I think what is currently happening on this pr is best:
image

Separately, can we have a box-shadow too, like cards? Then once we have  a proposal for container shadows/elevaiton we can go back and update the shadow depth to something more suitable.

lyubomir-popov:
@bartaz still concerned about the button behaviours, it is even worse on mobile:

image

That full width centered button is something optimised for purchase flows, not for things like this.

anthonydillon:
The width behaviour on small breakpoints can be modified here as this is a pattern.

scss/_settings_colors.scss
bartaz: @lyubomir-popov What would be the purpose of the box shadow if we have an overlay?

Dark box shadow is completely invisible on such overlay.

It is visible when it shouldn't (when drawer is offscreen it sticks from the left):
Screenshot 2020-04-16 at 07 38 18

We can prevent this from happening of course, but why applying shadow if it will be not visible when drawer is open anyway?

bartaz:

I updated the PR based on your comments:

- removed unnecessary spans from buttons
- prevented buttons from growing on mobile

I have not added box shadow because it's invisible on dark overlay anyway. More details in comment thread above.

scss/_settings_colors.scss
anthonydillon: I did think drop shadow would help but may need to be increased from the normal card shadow. Also, I was thinking it might look nice if we transition the shadow spread. So off canvas has 0 spread were as when the menu is revealed the shadowed spread is set and transitioned to at the same rate as the slide out. This would solve the shadow visible off screen and might make the shadow more visible to the eye.

bartaz: Box shadow will be useful, if it's visible on top of overlay. We would need to adjust the opacity of overlay and shadow to achieve that. With overlay appearing immediately when sidenav starts animating into the screen there is no way to make shadow visible (with our current colors).

Animating shadow is an interesting idea. But I wonder if this will not make it look like drawer is 'elevating' during animation, which could be a bit unexpected.

templates/docs/examples/patterns/side-navigation/_toggle_script.js
```diff
@@ -0,0 +1,70 @@
...
+*/
+function setupSideNavigations(sideNavigationSelector) {
+  // Setup all side navigations on the page.
+  var sideNavigations = [].slice.call(document.querySelectorAll(sideNavigationSelector));
+
+  sideNavigations.forEach(setupSideNavigation);
+
+  // on resize event remove `is-collapsed` class from drawers to prevent them from animating back in
+  // TODO: this should probably be throttled
+  window.addEventListener('resize', function() {
```
bartaz, line 60: @anthonydillon Are you ok with component script listening on resize?

We need that to prevent sidenav animating on screen resize, but it feels like it's getting into area or 'too much JS' for CSS framework.

To use this pattern this (or equivalent) JS is required. And in most cases it will be copied from here blindly.
And it's not bulletproof as it is now. Resize event should be probably throttled (so it's not called million times). This adds more and more to JS that we don't officially publish or support that is required by pattern to work as expected.

scss/_settings_colors.scss
lyubomir-popov: We recently created a set of rules for different things, one of them was anything that overlaps something else should have a box shadow, so it's a consistency thing. I was wary of suggesting a bespoke drop shadow, but ideally it should be something like the google maps one: 
image we just have to define our own in the shadow hierarchy piece of work.

I know it won't be very visible here, but in the future we may change the opacity of the overlay. It doesn't hurt to apply the rules consistently.

lyubomir-popov:
@bartaz nice work, just a couple of niggles:
- is it possible to make the flashing black overlay appear with a fade?
- the button still worries me, but I'm prepard to accept it asa temporary compromise, then work separately on a button that fits the design  (and maye solves similar issues highlighted above with menus with chevrons etc.) I'll make a separate issue for this.

lyubomir-popov:
@bartaz I've created a working group [issue ](https://github.com/canonical-web-and-design/vanilla-framework/issues/2982) to discuss how we can achieve the properly aligned version of the toggle.

templates/docs/examples/patterns/side-navigation/_toggle_script.js
bartaz: @anthonydillon Thanks. It was good call.

I changed transition to animation and it allowed to get rid of resize event.
Moving collapsed/expanded to root element also helped with simplifying JS part.

templates/docs/examples/patterns/side-navigation/_icons.html
```diff
@@ -1,95 +1,112 @@
...
+<div class="p-side-navigation--icons {% if is_dark %}is-dark{% endif %}" id="drawer-icons">
+  <button class="p-side-navigation__toggle js-drawer-toggle" aria-controls="drawer-icons">
+    Toggle side navigation
+  </button>
+
+  <div class="p-side-navigation__overlay js-drawer-toggle" aria-controls="drawer-icons"></div>
+
+  {% if is_dark %}
+  <!-- dark background color should be changed via color theme variables, inline style is used here for example purposes -->
+  {% endif %}
```
anthonydillon, line 10: Do we need this?

anthonydillon:
The demo's toggle doesn't seem to work on Google Chrome on Android?

scss/_patterns_side-navigation.scss
```diff
@@ -1,15 +1,145 @@
...
+    bottom: 0;
+    left: 0;
+    overflow: auto;
+    position: fixed;
+    top: 0;
+    transform: translateX(-100%);
+    width: 100%;
+    z-index: 12;
+
+    [class*='p-side-navigation'].is-expanded & {
```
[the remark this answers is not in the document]
bartaz: We already have `.p-side-navigation` and `.p-side-navigation--icons` to cover.
It's not much, we can explicitly list them both here. But this just adds when another variants are added.

It really is limiting that we can't rely on main class `p-side-navigation` being ALWAYS on the parent.

templates/docs/examples/patterns/side-navigation/_icons.html
bartaz: I wanted to make it clear that setting background inline is not recommended way. Just in case someone uses example as a reference.

I can remove it. Especially that this example is not exposed in CodePen embeds, it's just listed on Examples pages and screenshoted for Percy.

scss/_patterns_side-navigation.scss
```diff
@@ -1,15 +1,145 @@
...
+      @extend %vf-has-box-shadow;
+
+      animation: vf-p-side-navigation-expand map-get($animation-duration, brisk);
+      transform: translateX(0);
+    }
+
+    [class*='p-side-navigation'].is-collapsed & {
```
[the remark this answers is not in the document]
bartaz: We can't if we want to prevent unexpected animations.

Having 'default' (collapsed) state as one without class means that close animation will trigger on page load for example.

To have animation only triggered by user intent, we need to have both of these classes.

bartaz:
> The demo's toggle doesn't seem to work on Google Chrome on Android?

@anthonydillon Any specific URL. I tested the MAAS docs example template and seems to be working fine on Android Chrome.

Screenshot_20200416-163103

anthonydillon:
Yeah, this URL: https://vanilla-framework-canonical-web-and-design-pr-2977.run.demo.haus/docs/examples/templates/maas-docs-grid

bartaz:
@anthonydillon Hmmm... Works for me. And I think we both have OnePlus with recent Chrome?

scss/_patterns_side-navigation.scss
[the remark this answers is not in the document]
anthonydillon: Oh I see, they are both under the same mixin. Normally the variants are there own mixin with shared placeholders.

anthonydillon:
Checking browserstack, working on FF on my phone.

anthonydillon:
Works on BrowserStack so must be cache or something, nevermind.

scss/_patterns_side-navigation.scss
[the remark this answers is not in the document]
bartaz: Oh, I haven't notice that. I guess not all patterns do it consistently.

anthonydillon:
The double navigations in the docs are causing issues. We should only have a single example per example (if that makes sense):
image

templates/docs/patterns/navigation.md
```diff
@@ -89,6 +89,9 @@ To add icons on the left side of the items in side navigation use the `.p-side-n
 View example of the side navigation pattern with icons
 </a>
 
+On small screens side navigation is rendered off-screen as a drawer. To open the side navigation drawer, add `is-expanded` class to
```
anthonydillon, line 92: Can we title this section JS functionality or something?

PR #2993, opened 2020-04-21
Sticky side navigation

bartaz:
## Done

Adds sticky positioning to side navigation when page is scrolled.
When side navigation is longer than screen size it's separately scrolled.

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/ or [demo](https://vanilla-framework-canonical-web-and-design-pr-2993.run.demo.haus/docs/examples/templates/maas-docs-grid)
- Check [docs example with sticky navigation](https://vanilla-framework-canonical-web-and-design-pr-2993.run.demo.haus/docs/examples/templates/maas-docs-grid)
- Make sure side nav works as expected
- Check if mobile view is not affected

## Screenshots

sticky

### DIFF
--- a/templates/docs/examples/templates/maas-docs-grid.html
+++ b/templates/docs/examples/templates/maas-docs-grid.html
@@ -39,7 +39,7 @@
 <section id="search-docs" class="p-strip--image is-shallow" style="background-image: url('https://assets.ubuntu.com/v1/e54487e2-maas-docs-suru.png')">
   <div class="row">
     <form class="p-search-box u-no-margin--bottom" action="#">
-      <input type="search" class="p-search-box__input" name="q" placeholder="Search documentation" required="">
+      <input type="search" class="p-search-box__input" name="q" placeholder="Search documentation" required="" />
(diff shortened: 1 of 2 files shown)


templates/docs/examples/templates/maas-docs-grid.html
```diff
@@ -123,6 +122,78 @@
                 </li>
               </ul>
             </li>
+            <li class="p-side-navigation__item--title">
```
bartaz, line 125: Just added more content to side nav and doc example to make both scroll.

lyubomir-popov:
thanks @bartaz, scrolling works as expected, I like it. One possible enhancement on mobile - disable <main> scroll while the drawer is open?

bartaz:
> disable scroll while the drawer is open?

We discussed it before. That requires JS and adding/removing style on `<body>`. And (quite likely) making sure on resize event to remove it as well - so felt to complicated to add to example JS.

lyubomir-popov:
@bartaz ok, maybe we can do this in the actual vanilla docs then. +1'ed it.

PR #3000, opened 2020-04-23
Bump Vanilla version to 2.10.0

bartaz:
## Done

Just a version bump for new Vanilla release.

## QA

Check if version in package.json is correct


anthonydillon:
@bartaz Ready when you are

PR #2996, opened 2020-04-22
Documentation layout documentation

bartaz:
## Done

Needs #2988 

Adds documentation layout example and documentation page about documentation layout structure.

3xnur2

Fixes #2953

## QA

DEMO: https://vanilla-framework-canonical-web-and-design-pr-2996.run.demo.haus/docs/layouts/documentation

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/ or [demo](https://vanilla-framework-canonical-web-and-design-pr-2996.run.demo.haus/docs/layouts/documentation)
- Go to [documentation layout page](https://vanilla-framework-canonical-web-and-design-pr-2996.run.demo.haus/docs/layouts/documentation), check if it has enough information
- Check [documentation layout example](https://vanilla-framework-canonical-web-and-design-pr-2996.run.demo.haus/docs/examples/layouts/documentation) if it has good structure to be used as a base for documentation sites to copy from

Screenshot 2020-04-24 at 15 31 21

PR #3003, opened 2020-04-23
Visualise layouts in a screen-width independent manner

lyubomir-popov:
## Done

NOTE: this is prototyping code, not meant for merging. Just to agree on the look.

- Based on https://github.com/canonical-web-and-design/vanilla-framework/pull/2996.
- Add a visualisation of the areas of the documentation layout, in a way that isn't affected by screen width. 

The idea was to generate with code, so it is easily update-able, then replace with an image, but I wonder - do we even need an image if we can get a screen-independent behaviour with 5 lines of css instead?

## QA

- Pull code
- Run `./run`
- Open docs/layouts/documentation or [demo](https://vanilla-framework-canonical-web-and-design-pr-3003.run.demo.haus/docs/layouts/documentation)
- Review the reworded text and thumbnail implementation. 
Should we have these as inline styles? Should we have a dedicated place in the scss for demo styles? 

## Screenshots

image

### DIFF
--- a/templates/docs/layouts/documentation.md
+++ b/templates/docs/layouts/documentation.md
@@ -1,0 +1,82 @@
+---
+wrapper_template: '_layouts/docs.html'
+context:
+  title: Documentation | Layouts
+---
+
+## Documentation layout
+
+<hr>
+
+Documentation page layout can be build using Vanilla grid and common components. It consists of header with main navigation, optional hero strip (that may contain search field), grid based content area and a footer.
+
+### Structure
+
+#### Heading
+
+Heading with the main navigation is build with [navigation component](/docs/patterns/navigation#global-navigation).
+
+Style and contents of documentation main navigation should be consistent with rest of the site.
+
+Documentation pages may have an optional search box in the main navigation.
+
+#### Hero
+
+Documentation pages can have a hero area above main content area. This part of the page will usually contain a search field.
+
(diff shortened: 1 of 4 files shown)


templates/docs/layouts/documentation.md
```diff
@@ -0,0 +1,168 @@
...
+context:
+  title: Documentation | Layouts
+---
+
+<style>
+
+.u-thumb-row {
+  --grid-gap: 0.5rem;
+  display: grid;
+  grid-gap: var(--grid-gap);
```
bartaz, line 12: I feel like we are overcomplicating this. 
We don't have to make it this specific (to literally repeat the grid implementation). Also, we probably don't want to use CSS grid because it will not work in older browsers making this page just look broken and even more confusing.

If we want this to be HTML, the CSS should be really minimal.

Or it should be an image. Maybe SVG if we want better scaling and font rendering.

All we need it couple boxes with labels

lyubomir-popov:
@bartaz png preview of the svg:
Documentation-thumb

lyubomir-popov:
@bartaz Please close once you've reviewed the text changes. Svg [here](https://assets.ubuntu.com/v1/cdf09557-Documentation+layout%401x.svg)

bartaz:
Suggestions applied in #2996

PR #3010, opened 2020-04-27
Use Stylelint

### LINKED ISSUE #2945: Use stylelint instead of sass-lint
We discussed on dev catch-up that we should consistently use stylelint instead of sass-lint in our projects.

bartaz:
## Done

Fixes  #2945

- Replaces sass-lint with stylelint
- Makes sure stylelint is using configuration compatible with SCSS and Prettier
- Adds additional config rules based on restrictive sass-lint we used before:
  - disallowing classes in @extend
  - disallowing vendor prefixes
  - disallowing selectors with qualifying elements
  - disallowing selectors with #ids 
  - disallowing deep nesting (note - stylelint counts nested levels differently, every mixin, media query, etc is counted as nesting level)
  - disallowing http urls of assets
  - disallowing underscores and extensions in @import

To do later:

Some things are not possible or not easily achievable with stylelint:
- allowing color values to be used only in $variables is not possible currently
- enforcing BEM class names requires custom regexp or additional external tooling - to be explored
- enforcing order of mixins/extends/props requires additional config - to be explored

## QA

- PR checks should pass
- Percy should not show any diffs to approve
- Review the code


.stylelintrc.json
```diff
@@ -0,0 +1,23 @@
+{
+  "extends": ["stylelint-config-standard", "stylelint-prettier/recommended"],
+  "plugins": ["stylelint-order", "stylelint-scss"],
+  "rules": {
+    "order/properties-alphabetical-order": true,
+    "at-rule-no-unknown": null,
+    "scss/at-rule-no-unknown": true,
+    "no-descending-specificity": [true, {"severity": "warning "}],
```
bartaz, line 8: This is turned on in  "stylelint-config-standard" by default.

I updated the files in Vanilla that were easy to fix, but we have quite some that are not that easy.
Nested selectors and many parent-child dependencies make it hard to reorder files in a way that specificity will be in order.

I turned it from error to warning, so we see it as a problem, but it doesn't fail the build.

BTW, Love how GH shows errors and warnings from linting on GitHub actions inline in the diff:
Screenshot 2020-04-27 at 16 31 13

PR #3018, opened 2020-04-28
Fix scrolling with modal

### LINKED ISSUE #2801: Modal title doesn't look great on large modals
## Pattern to amend

Modal

## Visual

Screen Shot 2020-01-30 at 3 05 33 PM

## Context

I don't have any suggestions for an improvement here but maybe providing a couple header style options would help.

### LINKED ISSUE #2970: Modal is scrollable on long pages
**Describe the bug**
When on a long page a modal stays at the top of the page when the user scrolls. This is never the expected behaviour.

**To Reproduce**
Steps to reproduce the behaviour:

1. Go to https://vanillaframework.io/docs/patterns/modal
2. Click on Edit in codepen
3. Add lots of content behind the model
4. See when you scroll the model is not fixed.

**Expected behavior**
The model should stay over the top of the content. I used `position: fixed` on the `p-model` block to achieve this.

It would be better if we could block the page scrolling at all but I think the only way we can achieve this is by adding `overflow: hidden` to the body element which would allow a lot of leakage but might be ok using `:root body` from within the pattern when open. But I think we would need a add a modal class to the body via the js. 

**Screenshots**
image

bartaz:
## Done

- Makes modal fixed to the window, so it's not affected by body scroll.
- Updates the styling to make modal dialog itself scrollable if it has more content than fits on the screen.
- Adds the border under modal header as per new design #2801

Fixes #2970
Fixes #2801

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/ or [demo](https://vanilla-framework-canonical-web-and-design-pr-3018.run.demo.haus/docs/examples/patterns/modal/modal-scroll)
- Open [scrollable modal example](https://vanilla-framework-canonical-web-and-design-pr-3018.run.demo.haus/docs/examples/patterns/modal/modal-scroll)
  - make sure large modal works on different screen sizes
- Open existing [modal example](https://vanilla-framework-canonical-web-and-design-pr-3018.run.demo.haus/docs/examples/patterns/modal/default)
  - make sure it works as expected

## Screenshots

Screenshot 2020-04-28 at 12 18 41

### DIFF
--- a/scss/_patterns_modal.scss
+++ b/scss/_patterns_modal.scss
@@ -24,6 +24,7 @@
 
     bottom: $spv-inner--large;
     left: $sph-inner--large;
+    margin-bottom: 0;
     max-height: calc(100% - #{2 * $spv-inner--large});
     max-width: $grid-max-width;
     overflow: scroll;
@@ -68,9 +69,12 @@
     border: 0;
     box-sizing: content-box;
     height: $icon-size;
-    margin: -#{$icon-size} -#{$icon-size} 0 0;
-    padding: $icon-size;
+    margin: 0;
+    padding: $spv-inner--small $sph-inner--small;
+    position: absolute;
+    right: 0;
     text-indent: -999em;
+    top: 0;
     width: $icon-size;
 
     &:focus {
--- a/scss/_patterns_modal.scss
+++ b/scss/_patterns_modal.scss
@@ -4,15 +4,15 @@
   .p-modal {
     align-items: center;
     background: transparentize($color-dark, 0.15);
-    content: '';
+    bottom: 0;
     display: flex;
     height: 100vh;
     justify-content: center;
     left: 0;
     margin: 0;
-    overflow: scroll;
-    padding: $sp-large;
-    position: absolute;
+    padding: $spv-inner--large;
+    position: fixed;
+    right: 0;
     top: 0;
     width: 100%;
   }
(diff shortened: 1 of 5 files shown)


scss/_patterns_modal.scss
```diff
@@ -4,15 +4,15 @@
   .p-modal {
     align-items: center;
```
lyubomir-popov, line 5: margin-bottom on the child throws this vertical centering off. I would set it to 0.

scss/_patterns_modal.scss
```diff
@@ -53,20 +56,22 @@
...
+      size: $icon-size;
     }
 
     border: 0;
     box-sizing: content-box;
-    height: $sp-medium;
-    margin: -#{$sp-medium} -#{$sp-medium} 0 0;
-    padding: $sp-medium;
+    height: $icon-size;
+    margin: -#{$icon-size} -#{$icon-size} 0 0;
```
lyubomir-popov, line 71: Since there's a border under the title, the button is identified as part of a title bar. As such, it would be better to position it so it aligns with the title. I'll add a screenshot as a regular comment, the code for I used (hardcoded rem values just for testing, but it should use the var for modal padding):
```{
    position: absolute;
    right: 1rem;
    top: .5rem;
}
```

lyubomir-popov:
@bartaz (Optional improvements)

Made a suggestion for the button positioning, it takes the button's invisible padding into account horizontally, and vertically aligns it with the heading:

image

Separately, if you take the bottom marign off, it becomes centered:

image

lyubomir-popov:
Is it meant to cover the whole screen on medium/small screens?
image

bartaz:
> Is it meant to cover the whole screen on medium/small screens?

Yes, it already does that currently:
https://vanillaframework.io/docs/examples/patterns/modal

PR #3015, opened 2020-04-28
Remove duplicated side-navigation id from the template

bartaz:
## Done

Remove duplicated side-navigation id from the template to make sure automatic scrolling to #side-navigation works properly

## QA

- Pull code
- Run `./run` or [demo](https://vanilla-framework-canonical-web-and-design-pr-3015.run.demo.haus/docs/patterns/navigation)
- Go to [Navigation page](https://vanilla-framework-canonical-web-and-design-pr-3015.run.demo.haus/docs/patterns/navigation)
- Click on [Side navigation](https://vanilla-framework-canonical-web-and-design-pr-3015.run.demo.haus/docs/patterns/navigation#side-navigation) sub item in side navigation
- Make sure the page scroll to side navigation part of documentation
- Go directly to [Side navigation](https://vanilla-framework-canonical-web-and-design-pr-3015.run.demo.haus/docs/patterns/navigation#side-navigation)
- Make sure page loads with Sub navigation part visible on top of screen

## Screenshots

Screenshot 2020-04-28 at 08 01 08

PR #3023, opened 2020-04-29
Fix the small screen side navigation

anthonydillon:
## Done
Added the missing `id` to the navigation element.

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/3022

## QA
- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/docs
- Shrink your screen to mobile size
- Click the navigation toggle and check it works.

### DIFF
--- a/templates/_layouts/docs.html
+++ b/templates/_layouts/docs.html
@@ -12,7 +12,7 @@
     <div class="p-strip">
     <div class="row">
       <aside class="col-3">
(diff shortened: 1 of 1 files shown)


bartaz:
This makes #3015 not work again.
We need a different id for side-navigation, because "side-navigation" is autogenerated from markdown on documentation page.

PR #3059, opened 2020-05-11
Update blue link colours

lyubomir-popov:
## Done

- Update link blues for all states
- Make checkboxes, switch, radios etc use the link blue

Drive-by's:
- remove border radius from inputs, something we had agreed a long time ago but never implemented
- Update focus state bar thickness for some components which had it hardcoded to 2px. The correct variable is $bar-thickness, which is 3px.

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/3042

## QA

- Pull code
- Run `./run`
- Open as many examples as possible, QA link colour and focus states. E.g.:
docs/examples/patterns/buttons/icon
docs/examples/patterns/forms/form-inline
docs/examples/patterns/forms/form-stacked
docs/examples/patterns/forms/form-validation

Open the pagination specifically, verify visited (or other link colours) do not show when the buttons are in various states:
docs/examples/patterns/pagination/pagination-truncated

## Screenshots

various form elements:

image

green button with outline/background contrast ratio above 3:1:
image

Negative button with outline/background contrast ratio above 3:1:
image

### DIFF
--- a/scss/_base_forms.scss
+++ b/scss/_base_forms.scss
@@ -26,7 +26,6 @@
 
     background-color: $color-x-light;
     border: 1px solid $color-mid;
-    border-radius: 0;
     box-shadow: inset 0 1px 1px $color-input-shadow;
     color: $color-dark;
     font-family: unquote($font-base-family);
--- a/scss/_base_forms.scss
+++ b/scss/_base_forms.scss
@@ -26,6 +26,7 @@
 
     background-color: $color-x-light;
     border: 1px solid $color-mid;
+    border-radius: 0;
     box-shadow: inset 0 1px 1px $color-input-shadow;
     color: $color-dark;
     font-family: unquote($font-base-family);
--- a/scss/_base_forms.scss
+++ b/scss/_base_forms.scss
@@ -26,7 +26,7 @@
 
     background-color: $color-x-light;
     border: 1px solid $color-mid;
-    border-radius: $border-radius;
+    border-radius: 0;
     box-shadow: inset 0 1px 1px $color-input-shadow;
     color: $color-dark;
     font-family: unquote($font-base-family);
--- a/scss/_base_links.scss
+++ b/scss/_base_links.scss
@@ -7,7 +7,7 @@
     text-decoration: none;
 
     &:focus {
-      outline: thin dotted $color-mid-light;
+      outline: thin dotted $color-focus;
     }
 
     &:hover {
@@ -16,7 +16,7 @@
     }
 
     &:visited {
-      color: darken($color-link, 10%);
+      color: $color-link-visited;
     }
   }
 }
--- a/scss/_base_links.scss
+++ b/scss/_base_links.scss
@@ -3,11 +3,13 @@
 // Base styling for links
 @mixin vf-b-links {
   a {
+    @include vf-focus;
+
     color: $color-link;
     text-decoration: none;
 
     &:focus {
-      outline: thin dotted $color-focus;
+      outline-offset: 0;
     }
 
     &:hover {
--- a/scss/_settings_colors.scss
+++ b/scss/_settings_colors.scss
@@ -19,9 +19,9 @@ $color-label-validated: #006b75;
 
 $color-negative: #c7162b !default;
 $color-caution: #f99b11 !default;
-$color-positive: hsl(129, 81%, 29%) !default;
+$color-positive: #0e8620 !default;
 
-$color-link: #06c !default;
+$color-link: #0066cc !default;
 $color-link-visited: #7d42b8 !default;
 $color-focus: #2e96ff !default;
 $color-focus-positive: #003008 !default;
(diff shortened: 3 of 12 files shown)


scss/_base_forms.scss
```diff
@@ -26,7 +26,7 @@
 
     background-color: $color-x-light;
     border: 1px solid $color-mid;
-    border-radius: $border-radius;
+    border-radius: 0;
```
anthonydillon, line 29: Can't this just be removed?

scss/_settings_colors.scss
```diff
@@ -12,21 +13,20 @@ $color-mid: #999 !default;
...
 
 $color-label-validated: #006b75;
 
-$color-input-shadow: rgba($color-x-dark, 0.12) !default;
-
-$color-accent: $color-brand !default;
-$color-accent-background: $color-accent !default;
+$color-negative: #c7162b !default;
+$color-caution: #f99b11 !default;
+$color-positive: hsl(129, 81%, 29%) !default;
```
anthonydillon, line 22: Why switch to use hsl here?

scss/_settings_colors.scss
```diff
@@ -12,21 +13,20 @@ $color-mid: #999 !default;
...
 
-// for visual outline on :active / :focus elements
-$color-focus: #19b6ee !default;
+$color-link: #06c !default;
```
anthonydillon, line 24: Please use fully qualified hex values please `#0066cc`

scss/_settings_colors.scss
```diff
@@ -12,21 +13,20 @@ $color-mid: #999 !default;
...
+$color-link-visited: #7d42b8 !default;
```
anthonydillon, line 25: Is this the visitied color we want to use now? `#7d42b8`

anthonydillon: This new visited links look very different?
image

scss/_base_forms.scss
lyubomir-popov: removed

scss/_settings_colors.scss
lyubomir-popov: shouldn't be, I was just trying to check the hue in vscode - reverted.

scss/_settings_colors.scss
lyubomir-popov: sure, changed. Can't help but wonder though, vscode underlined it in red, and suggested the shorthand. We're also using shorthand for fff, 111, etc - so why should this be the full version?

scss/_settings_colors.scss
lyubomir-popov: yes, turns out there are requirements for contrast ratio between link states. There's a good article here: https://webaim.org/blog/wcag-2-0-and-link-colors/

lyubomir-popov:
@anthonydillon the linter doesn't like #0066cc:
image

have to change it back to #06c...

scss/_settings_colors.scss
bartaz: I believe our linter enforces short hex values if they are possible. This was the case both with previous sass-lint and current stylelint.
As Lyubo pointed out, we have shorthand for some colors (`#111`, `#666`, `#fff`, `#999`).

scss/_settings_colors.scss
```diff
@@ -12,21 +13,20 @@ $color-mid: #999 !default;
...
+$color-positive: #0e8620 !default;
 
-// for visual outline on :active / :focus elements
-$color-focus: #19b6ee !default;
+$color-link: #06c !default;
+$color-link-visited: #7d42b8 !default;
+$color-focus: #2e96ff !default;
+$color-focus-positive: #003008 !default;
+$color-focus-negative: #3b0006 !default;
+$color-information: #24598f !default;
```
bartaz, line 29: Could `$color-information` be moved back up with its "siblings" (positive, negative, caution)? I thought it was removed by mistake when it wasn't there.
We seem to have colors in logical groups in this file, so let's keep them this way for easier reference.

bartaz:
Because range input in Chrome doesn't support changing progress in CSS we update the color in JS where it is hardcoded. We should update the value to new `$color-information` value there: https://github.com/canonical-web-and-design/vanilla-framework/blob/master/templates/docs/examples/base/forms/range.html#L13

bartaz:
Selects have even more prominent border radius than before, is that expected?

Screenshot 2020-05-12 at 07 40 28

bartaz:
Changes in this 'update blue link color' seem to go quite far outside updating colors. This makes it difficult to review and QA.

Now, focused links are barely readable because of the outline:
Screenshot 2020-05-12 at 12 40 11

lyubomir-popov:
@bartaz sorry I thought I had pushed that.. pushed now. this is the last change, anything else will go onto separate prs.

scss/_base_links.scss
```diff
@@ -7,7 +7,9 @@
     text-decoration: none;
 
     &:focus {
-      outline: thin dotted $color-mid-light;
+      @include vf-focus;
```
bartaz, line 10: `vf-focus` adds `&:focus` by itself, so it shouldn't be nested in `:focus` selector.

PR #3072, opened 2020-05-18
scope margin-left on p-stepped-list__content  to stepped list only

lyubomir-popov:
## Done

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/3066

## QA

- Pull code
- Run `./run`
- Open docs/examples/patterns/lists/lists-stepped-detailed
- Verify .p-stepped-list__content has margin-left:0 beyond $breakpoint-medium 

## Screenshots
Large
image

Small:
image

### DIFF
--- a/scss/_patterns_lists.scss
+++ b/scss/_patterns_lists.scss
@@ -187,7 +187,7 @@ $spv-list-item--inner: null;
         }
       }
 
-      & + .p-stepped-list__content {
+      .p-stepped-list & + .p-stepped-list__content {
         @media (max-width: $breakpoint-heading-threshold) {
           margin-left: $bullet-width-mobile + $sph-inner;
         }
--- a/scss/_patterns_lists.scss
+++ b/scss/_patterns_lists.scss
@@ -199,11 +199,9 @@ $spv-list-item--inner: null;
       }
     }
 
-    .p-stepped-list__title {
-      .p-stepped-list--detailed & + .p-stepped-list__content {
-        @media (min-width: $threshold-6-12-col) {
-          margin-left: 0;
-        }
+    .p-stepped-list--detailed .p-stepped-list__title + .p-stepped-list__content {
+      @media (min-width: $threshold-6-12-col) {
+        margin-left: 0;
       }
     }
   }
--- a/scss/_patterns_lists.scss
+++ b/scss/_patterns_lists.scss
@@ -198,11 +198,11 @@ $spv-list-item--inner: null;
         margin-left: $bullet-width + $sph-inner;
       }
     }
+  }
 
-    .p-stepped-list--detailed .p-stepped-list__title + .p-stepped-list__content {
-      @media (min-width: $threshold-6-12-col) {
-        margin-left: 0;
-      }
+  .p-stepped-list--detailed .p-stepped-list__title + .p-stepped-list__content {
+    @media (min-width: $threshold-6-12-col) {
+      margin-left: 0;
     }
   }
 }
--- a/scss/_patterns_lists.scss
+++ b/scss/_patterns_lists.scss
@@ -163,6 +163,7 @@ $spv-list-item--inner: null;
     @extend %numbered-step-title;
   }
 
+  // Indent titles
   @for $i from 6 through 1 {
     // Bullet sizes for each heading level
     $bullet-width: map-get($line-heights, default-text);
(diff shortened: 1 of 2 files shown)


scss/_patterns_lists.scss
```diff
@@ -186,14 +187,22 @@ $spv-list-item--inner: null;
...
 
-        @media (min-width: $breakpoint-heading-threshold) {
-          margin-left: $bullet-width + $sph-inner;
+      @media (min-width: $breakpoint-heading-threshold) {
+        margin-left: $bullet-width + $sph-inner;
+      }
+    }
+
+    .p-stepped-list__title {
+      .p-stepped-list--detailed & + .p-stepped-list__content {
```
bartaz, line 203: This feels unnecessarily nested.

The only thing within `.p-stepped-list__title` is `.p-stepped-list--detailed & + .p-stepped-list__content` - so, couldn't it be just `.p-stepped-list--detailed .p-stepped-list__title + .p-stepped-list__content` without nesting?

lyubomir-popov:
@bartaz updated to reduce the nesting

scss/_patterns_lists.scss
```diff
@@ -186,15 +187,21 @@ $spv-list-item--inner: null;
...
 
-        @media (min-width: $breakpoint-heading-threshold) {
-          margin-left: $bullet-width + $sph-inner;
-        }
+      @media (min-width: $breakpoint-heading-threshold) {
+        margin-left: $bullet-width + $sph-inner;
+      }
+    }
+
+    .p-stepped-list--detailed .p-stepped-list__title + .p-stepped-list__content {
```
bartaz, line 202: I think this part doesn't need to be inside `for` loop, it's unnecessarily repeated in output CSS:

Screenshot 2020-05-18 at 13 34 13

scss/_patterns_lists.scss
```diff
@@ -186,18 +187,24 @@ $spv-list-item--inner: null;
           width: $bullet-width;
         }
       }
+    }
 
-      & + .p-stepped-list__content {
-        @media (max-width: $breakpoint-heading-threshold) {
-          margin-left: $bullet-width-mobile + $sph-inner;
-        }
+    h#{$i}.p-stepped-list__title + .p-stepped-list__content {
```
anthonydillon, line 192: Why does this require adjacent sibling selector? This could complicate the specificity.

anthonydillon: It looks like the use of `$i` is outside of the for loop

bartaz: > It looks like the use of `$i` is outside of the for loop

No. Loop ends in line 201 if I look at nesting correctly.

lyubomir-popov: @anthonydillon github has omitted some lines, which makes it look outside the loop. But the brace on line 190 closes an opening brace on line 177:
image

As to why this is needed - it's a very manual way of creating a hanging indent that works for any heading size (that's what is currently implemented, I'm not adding it with this pr). Feels like there should be a simpler way to do this so the pseudo element pushes anything underneath in line with the heading it sits next to. But that's a separate exploration.

scss/_patterns_links.scss
```diff
@@ -70,6 +70,7 @@ $external-icon-size--headings: #{map-get($icon-sizes, heading-icon--x-small)};
       background-color: currentColor;
       content: '\00A0';
       display: inline-block;
+      line-height: 1em;
```
bartaz, line 73: What does that have to do with this PR?

bartaz: Oh, I thought it was a fix for external icon size. Because I think these icons used to be a bit bigger, and there is some weird bug with size of the CSS map.

Seems to be a different issue.

Anyway, it's OK to fix some drive bys when they are related to issue in place, unrelated drive bys make it confusing - during review, what should be tested; they may accidently cause regressions (because were not QAed properly becaused seemed small), later they are harder to track. In 6 months we may wonder why line-height is 1em, if we track history, we see this in completely unrelated PR and we don't know why it happened.

I know it's a small thing. But if it is a small thing, the easier it should be to create a small, dedicated PR to properly review and QA it.

lyubomir-popov: you're right, I'll move it to the correct pr

bartaz: Thanks, could you reset this branch to previous commit and force push again?

`git reset --hard 3131fe7` (while in the `list-fix` branch)

PR #3080, opened 2020-05-19
fix external link icon lineheight so it doesn't distort focus outline

lyubomir-popov:
## QA

- Pull code
- Run `./run`
- Open /docs/examples/patterns/links/links-external#www.youtube.com/watch?v=Sv3Z_gbRVgQ, verify the appearance of the focused link around the external icon matches the "After" screenshot shown below:

Before: 
image

After:
image

### DIFF
--- a/scss/_patterns_links.scss
+++ b/scss/_patterns_links.scss
@@ -34,12 +34,9 @@ $external-icon-size--headings: 1rem; // slghtly increased
     }
   }
 
+  // include external link styles only if CSS mask is supported
   @supports (mask-size: 1em) or (-webkit-mask-size: 1em) {
     @include vf-mask-supported;
-  }
-
-  @supports not ((mask-size: 1em) or (-webkit-mask-size: 1em)) {
-    @include vf-mask-unsupported;
   }
 
   .p-top {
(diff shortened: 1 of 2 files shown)


scss/_patterns_links.scss
```diff
@@ -70,6 +70,7 @@ $external-icon-size--headings: #{map-get($icon-sizes, heading-icon--x-small)};
       background-color: currentColor;
       content: '\00A0';
       display: inline-block;
+      line-height: 1em;
       // stylelint-disable property-no-vendor-prefix
       -webkit-mask: url("data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24' %3E%3Cg fill='none' fill-rule='evenodd'%3E%3Cpath d='M-1-1h16v16H-1'/%3E%3Cpath fill='%23111' d='M4.867 1.313C.6 1.32.067 1.443.067 4.51v6.4c0 3.2.533 3.2 5.333 3.2h2.133c4.8 0 5.334 0 5.334-3.2v-1.6h-1.6v1.068c0 2.133 0 2.133-4.267 2.133H5.933c-4.266 0-4.266 0-4.266-2.132V5.044c0-1.93.034-2.112 3.2-2.13v-1.6z'/%3E%3Cpath d='M-1-1h16v16H-1'/%3E%3Cpath fill='%23111' d='M6.435 2.16c.11-.446 7.113-2.196 7.448-1.86.335.334-1.416 7.335-1.863 7.447-.447.112-5.697-5.14-5.586-5.586z'/%3E%3Cpath fill='%23111' d='M9.032 3.38L4.705 7.708l1.767 1.767L10.8 5.148'/%3E%3C/g%3E%3C/svg%3E")
         no-repeat 0.25em 0 / #{$mask-size};
```
bartaz, line 76: Not entirely related, but as we update the link we can verify that:

Size of the mask is `$mask-size` here, for `-webkit-mask`. Couple lines below, in unprefixed `mask` it uses `$size` instead. This makes the external icon use different size in Firefox and webkit based browsers.

Is this expected?

bartaz:
Thanks for extracting it to separate PR.

Looks good, works fine.
I added "Design needed" and leave Percy for approve, because this change moves the external icon vertically a bit, so want to make sure we are OK with this change.

Also, in inline comment I mentioned difference in size between Firefox and Chrome.

scss/_patterns_links.scss
bartaz: Firefox (1rem):

Screenshot 2020-05-20 at 07 50 57

Chrome (0.6667rem):

Screenshot 2020-05-20 at 07 51 02

scss/_patterns_links.scss
```diff
@@ -61,144 +56,24 @@ $external-icon-size--headings: #{map-get($icon-sizes, heading-icon--x-small)};
 
 // For browsers that support CSS masks
 @mixin vf-mask-supported {
+  $external-icon-size--default-text: 1rem;
```
bartaz, line 59: If there is only one var, maybe we can drop `--default-text` suffix.

PR #3089, opened 2020-05-22
Remove vertical offsets from divider lines

### LINKED ISSUE #3086: .p-divider has no min-height
**Describe the bug**

For p-divider blocks with single lines of text, the blocks are too small to make the divider visible.

**To Reproduce**

1. Go to https://canonical.com/#partners
2. In the inspector, remove has-min-height from the row of links
3. See error that the divider lines do not have enough height to be seen

**Expected behavior**

Even with a single line of text, you should be able to see the divider line

**Screenshots**

original
image

fix
image

Here is the code used to fix the issue:

```scss
  .p-divider.has-min-height {
    @media screen and(min-width: $breakpoint-medium) {
      .p-divider__block:not(:nth-child(1))::before {
        min-height: 1.5rem;
        top: 0;
      }
    }
  }
```

lyubomir-popov:
## Done

Remove vertical offsets from divider lines

Fixes #3086

## QA

- Pull code
- Run `./run`
- Open docs/examples/patterns/lists/divider
- Verify divider lines span the full height of the component.

image

### DIFF
--- a/scss/_patterns_divider.scss
+++ b/scss/_patterns_divider.scss
@@ -26,7 +26,7 @@
     }
 
     @media (min-width: $threshold-6-12-col) {
-      &:not(:nth-child(1))::before {
+      &:not(:first-child)::before {
         background-color: $color-mid-light;
         bottom: 0;
         content: '';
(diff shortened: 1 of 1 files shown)


scss/_patterns_divider.scss
```diff
@@ -28,11 +28,11 @@
     @media (min-width: $threshold-6-12-col) {
       &:not(:nth-child(1))::before {
```
bartaz, line 29: While we are at it, can it be simplified to:

```suggestion
      &:not(:first-child)::before {
```

PR #3114, opened 2020-06-05
Consistent hover background color

lyubomir-popov:
## Done

Make buttons, sidenav, contextual menu etc use the same hover color.

## QA

- Pull code
- Run `./run`
Paste the following in any example, inspect the hover states and make sure background color matches:

```
<div class="p-strip--light">
  <div class="row">
    <div class="col-6">
      <input type="search" name="" id="" placeholder="search" class="u-sv-2 u-no-margin--bottom">
      <div class="p-chips__chip--with-category"><span class="p-chips__category"></div>
        <button class="p-button--positive">Positive</button>
        <button class="p-button--negative">p-button--negative</button>
        <button>Hover state</button>
        <button>Active state</button><br>
      {% include "docs/examples/patterns/contextual-menu/default.html" %}
      {% include "docs/examples/patterns/accordion.html" %}
    </div>
    <div class="col-6">
      {% include "docs/examples/patterns/side-navigation/_default.html" %}
    </div>
  </div>
</div>
```

## Screenshots
image


scss/_settings_colors.scss
```diff
@@ -78,7 +78,7 @@ $colors--light-theme--text-muted: #666 !default;
 
 $colors--light-theme--background-default: #fff !default;
 $colors--light-theme--background-alt: #f7f7f7 !default;
-$colors--light-theme--background-highlight: rgba($color-x-dark, 0.03) !default;
+$colors--light-theme--background-highlight: rgba($color-x-dark, $hover-background-darken-amount) !default;
```
bartaz, line 81: Should dark theme be updated as well?

scss/_patterns_buttons.scss
```diff
@@ -95,12 +87,11 @@
...
       $button-text-color: $color-x-light,
-      $button-hover-background-color: darken($color-negative, $hover-background-darken-percentage),
-      $button-hover-border-color: darken($color-negative, $hover-background-darken-percentage),
+      $button-hover-background-color: scale-color($color-negative, $lightness: -$hover-background-darken-amount * 100),
       $button-disabled-background-color: $color-negative,
       $button-disabled-border-color: $color-negative,
-      $button-active-background-color: darken($color-negative, $active-background-darken-percentage),
-      $button-active-border-color: darken($color-negative, $active-background-darken-percentage)
+      $button-active-background-color: scale-color($color-negative, $lightness: -$hover-background-darken-amount * 100),
+      $button-active-border-color: scale-color($color-negative, $lightness: -$hover-background-darken-amount * 100)
```
bartaz, line 94: These are active colors, but they use hover darken amount - is that intended?

scss/_patterns_buttons.scss
```diff
@@ -119,9 +110,9 @@
       $button-background-color: $color-transparent,
       $button-border-color: $color-transparent,
       $button-text-color: $color-dark,
-      $button-hover-background-color: darken($color-x-light, $hover-background-darken-percentage),
+      $button-hover-background-color: scale-color($color-x-light, $lightness: -$hover-background-darken-amount * 100),
       $button-hover-border-color: $color-transparent,
-      $button-active-background-color: darken($color-light, $active-background-darken-percentage),
+      $button-active-background-color: scale-color($color-x-light, $lightness: -$active-background-darken-amount * 100),
```
bartaz, line 115: These are repeated a lot, maybe it would be good to create our own functions for those. So we can have `scale-active-color($color)` and `scale-hover-color($color)` and not repeat lightness value every time.

scss/_patterns_side-navigation.scss
```diff
@@ -518,17 +530,22 @@
     }
 
     &:hover,
-    &.is-active,
     &[aria-current='page'],
     &[aria-current='true'] {
-      background: $color-sidenav-item-background-highlight;
+      background: $color-sidenav-item-background-hover;
```
bartaz, line 535: `[aria-current=...]` mean the same as `.is-active`, so if `.is-active` was removed from this selector, so should be these two.

bartaz: But I think we should actually keep `.is-active` here, because it needs to have the same background as hover.

scss/_patterns_side-navigation.scss
```diff
@@ -518,17 +530,22 @@
...
+    &:active {
+      background: $color-sidenav-item-background-active;
+    }
+
     &.is-active,
     &[aria-current='page'],
     &[aria-current='true'] {
       @include vf-highlight-bar($color-sidenav-item-border-highlight, left);
+
+      background: $color-sidenav-item-background-hover;
```
bartaz, line 548: This doesn't seem needed, it's already added in line 535.

lyubomir-popov: addressed

scss/_patterns_side-navigation.scss
lyubomir-popov: addressed

scss/_settings_colors.scss
```diff
@@ -61,7 +61,7 @@ $states: (
 //
 // Background colors:
 // --background           - default background color
-// --background-alt       - alternative version of background color, to be used to diffrenciate given surface
+// --background-alt       - alternative version of background color, to be used to differentiate given surface
 // --background-highlight - background used on highlighted elements (for example selected elements in the navigation)
```
bartaz, line 65: Comment needs to be updated, we don't have `--background-highlight`, instead both `--background-hover` and `--background-active` should be documented in the comment

lyubomir-popov: updated

PR #3117, opened 2020-06-09
Ensure <pre> elements use ubuntu mono instead of browser default

### LINKED ISSUE #2037: Code snippet: font-family not set for pre element
**Describe the bug**
Some blogs create a code snippet only using pre, not ```<code>```; this results in the browser agent font being used for the snippet.

I believe pre needs to be added to the list of selectors in https://github.com/vanilla-framework/vanilla-framework/blob/bf66cea8ba6d7525c3f5cd080a3785421fa160d6/scss/_base_code.scss#L7

**To Reproduce**
1. Go to https://blog.ubuntu.com/2017/04/13/ros-production-our-prototype
2. Inspect the font used for the code snippet 

**Expected behavior**
pre should use ubuntu monospace, not default to the user agent font as it does now.

**Screenshots**
image

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

lyubomir-popov:
## Done

Fix #2037

## QA

- Pull code
- Run `./run`
- paste this anywhere, ensure the font-family is ubuntu mono: `<pre>Test</pre>`

### DIFF
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -19,8 +19,8 @@
     font-weight: bold;
   }
 
-  pre,
-  code {
+  code,
+  pre {
     direction: ltr;
     // stylelint-disable property-no-vendor-prefix
     -webkit-hyphens: none;
@@ -50,6 +50,7 @@
     border-radius: $border-radius;
     color: $color-dark;
     display: block;
+    font-family: unquote($font-monospace);
     margin-bottom: $spv-outer--scaleable;
     margin-top: 0;
     overflow: auto;
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -3,20 +3,26 @@
 // Base code styles
 @mixin vf-b-code {
   code,
-  samp,
-  kbd {
-    background-color: $color-mid-x-light;
-    box-shadow: 0 0 0 0.25rem $color-mid-x-light;
+  kbd,
+  pre,
+  samp {
     font-family: unquote($font-monospace);
     font-weight: $font-weight-regular-text;
-    margin-left: 0.25rem;
-    margin-right: 0.25rem;
     text-align: left;
+
+    b,
+    strong {
+      font-weight: bold;
+    }
   }
 
-  code b,
-  code strong {
-    font-weight: bold;
+  code,
+  kbd,
+  samp {
+    background-color: $color-mid-x-light;
+    box-shadow: 0 0 0 0.25rem $color-mid-x-light;
+    margin-left: 0.25rem;
+    margin-right: 0.25rem;
   }
 
   code,
@@ -50,11 +56,10 @@
     border-radius: $border-radius;
     color: $color-dark;
     display: block;
-    font-family: unquote($font-monospace);
     margin-bottom: $spv-outer--scaleable;
     margin-top: 0;
     overflow: auto;
-    padding: calc(#{map-get($nudges, nudge--p-ubuntumono)} - 1px) $sph-inner calc(#{map-get($nudges, nudge--p-ubuntumono)} - 1px);
+    padding: map-get($nudges, nudge--p-ubuntumono) $sph-inner;
     text-align: left;
     text-shadow: none;
     white-space: pre;
--- a/templates/docs/examples/base/code-block.html
+++ b/templates/docs/examples/base/code-block.html
@@ -1,14 +1,12 @@
 {% extends "_layouts/examples.html" %}
-{% block title %}Code - inline{% endblock %}
+{% block title %}Code block{% endblock %}
 
 {% block standalone_css %}base{% endblock %}
 
 {% block content %}
-  <pre>
-  <code>bolla.internal
-  - lldp:descr:
-    label: SysDescr
-    Ubuntu 18.04.4 LTS Linux 4.15.0-106-generic #107-Ubuntu SMP Thu Jun 4 11:27:52 UTC 2020 x86_64
-  </code></pre>
+<pre>
+<code>function() {
+  alert("Hello world");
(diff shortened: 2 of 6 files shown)


bartaz:
Should we add or update example for that? 

Now, that `<pre>` works without the code, why do we have `<pre><code>` in the example? They seem to be doing the same. Maybe we should document only using one of the variants?

But if we aim to support both, probably both should have an example to make sure we screenshot it and not have regressions.

lyubomir-popov:
According to [this](https://stackoverflow.com/questions/4611591/code-vs-pre-vs-samp-for-inline-and-block-code-snippets) stackoverflow thread, pre doesn't wrap, pre>code should.

lyubomir-popov:
@bartaz I updated this old pr and the accompanying documentation. To obtain the expected behaviour as specified by the spec, we need to change the current default. 

Let me know what you think.

scss/_base_code.scss
```diff
@@ -45,6 +45,7 @@
     border-radius: $border-radius;
     color: $color-dark;
     display: block;
+    font-family: unquote($font-monospace);
```
anthonydillon, line 48: We should check if Ubuntu is in the sites font stack before setting this.

bartaz:
> According to [this](https://stackoverflow.com/questions/4611591/code-vs-pre-vs-samp-for-inline-and-block-code-snippets) stackoverflow thread, pre doesn't wrap, pre>code should.

@lyubomir-popov Where does it say that pre>code should wrap? For what I see it that it says something completely opposite:

> Use `<code>` for inline code that can wrap and `<pre><code>` for block code that must not wrap.

HTML spec doesn't seem to say anything about wrapping in pre and it seems that by default nothing wraps in `<pre>`, regardless if `<code>` is in it or not. There is specifically `white-space: pre-wrap` that needs to be set to make code inside `<pre>` wrap.

Because it doesn't seem to be a common convention, I'm not sure if we should introduce any arbitrary difference between `<pre>` and `<pre><code>`. Especially that very often there is no control over how code is generated (for example markdown always produces `<pre><code>` for blocks of code). If we want to introduce wrapping and non-wrapping variants they should probably be added as modifier classnames.

But this seems to be something that was not even intended as in scope of this fix. This was supposed to be just about font used, not wrapping.

scss/_base_code.scss
lyubomir-popov: How can we check?

lyubomir-popov: I just copied this from the code-copyable component, this is not a new addition.

lyubomir-popov (commented):
@bartaz ugh I had missed the "inline" qualifier. You're right.

templates/docs/examples/base/code-block.html
```diff
@@ -0,0 +1,14 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Code - inline{% endblock %}
```
bartaz, line 2:
```suggestion
{% block title %}Code - block{% endblock %}
```

templates/docs/examples/base/code-block.html
```diff
@@ -0,0 +1,14 @@
...
+
+{% block standalone_css %}base{% endblock %}
+
+{% block content %}
+  <pre>
+  <code>bolla.internal
```
bartaz, line 8: As we correctly suggest `<pre><code>` should be used for code blocks, so would be good to give good example, so instead of command line output we should put some basic code here.

I guess something simple would be enough:

```
function() {
  alert("Hello world");
}
```

lyubomir-popov: updated

templates/docs/examples/base/code-block.html
lyubomir-popov: updated

scss/_base_code.scss
anthonydillon: Sorry I was wrong yesterday, this is correct.

anthonydillon: Can you add `pre` to the font-family set block above? 

If we are setting the font-family to mono then shouldn't it inherit the other aspects the font such as:
```css
font-family: unquote($font-monospace);
font-weight: 300;
text-align: left;
```

lyubomir-popov:
Added ability to have bold in all kbd, samp, code, pre:

image
 

(the default b, strong do not work in these elements as the mono weight needed is 600, and strong only applies 400). 

Drive by: add b to strong selector so both render as 400 outside of the kbd,samp, code, pre elements.

scss/_base_code.scss
```diff
@@ -3,20 +3,26 @@
...
+  samp {
     font-family: unquote($font-monospace);
     font-weight: $font-weight-regular-text;
-    margin-left: 0.25rem;
-    margin-right: 0.25rem;
     text-align: left;
+
+    b,
+    strong {
+      font-weight: bold;
```
anthonydillon, line 15: :+1: thanks

scss/_base_code.scss
```diff
@@ -50,11 +56,10 @@
     border-radius: $border-radius;
     color: $color-dark;
     display: block;
-    font-family: unquote($font-monospace);
     margin-bottom: $spv-outer--scaleable;
     margin-top: 0;
     overflow: auto;
-    padding: calc(#{map-get($nudges, nudge--p-ubuntumono)} - 1px) $sph-inner calc(#{map-get($nudges, nudge--p-ubuntumono)} - 1px);
+    padding: map-get($nudges, nudge--p-ubuntumono) $sph-inner;
```
anthonydillon, line 62: Much cleaner

anthonydillon (approved):
LGTM :+1:

PR #3140, opened 2020-07-02
Make list placeholders includable separately from patterns

### LINKED ISSUE #2808: List pattern placeholders cannot be included without all list patterns
**Describe the bug**
The placeholders here are inside the @mixin vf-p-lists:
https://github.com/canonical-web-and-design/vanilla-framework/blob/64fe763f601f2f3702b84749d84804a15024e629/scss/_patterns_lists.scss#L267

In order to extend say  %vf-list from maas for example, you need to include the entire mixin, which includes all [list patterns](https://github.com/canonical-web-and-design/vanilla-framework/blob/64fe763f601f2f3702b84749d84804a15024e629/scss/_patterns_lists.scss#L314)

**Expected behavior**
The placeholders should be in a separate mixin that doesn't output anything, so they can be included cleanly.

bartaz:
## Done

Separates list placeholders from list pattern styles to allow including them independently.

Fixes #2808 

## QA

- CI checks should pass, there should be no regressions

### DIFF
--- a/scss/_patterns_lists.scss
+++ b/scss/_patterns_lists.scss
@@ -7,6 +7,49 @@ $spv-list-item--inner: null;
   $spv-list-item--inner: 0;
 }
 
+@mixin vf-p-list-placeholders {
+  %numbered-step-container {
+    counter-reset: li;
+    display: flex;
+    flex-direction: column;
+    list-style: none;
+    padding-left: 0;
+  }
+
+  %numbered-step-title {
+    display: flex;
+    list-style: none;
+    margin-left: 0;
+    padding-left: 0;
+
+    &::before {
+      align-self: start;
+      background-color: $color-mid-x-light;
+      border-radius: 100%;
+      content: counter(li);
+      counter-increment: li;
+      direction: rtl;
+      display: block;
+      text-align: center;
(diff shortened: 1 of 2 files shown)


scss/_patterns_lists.scss
```diff
@@ -7,6 +7,49 @@ $spv-list-item--inner: null;
   $spv-list-item--inner: 0;
 }
 
+@mixin vf-p-list-placeholders {
```
anthonydillon, line 10: Why have it as a separate mixin and not global?

bartaz: I see it as a convention in all Vanilla scss file, that nothing is added without a mixin.

While placeholders don't add to the size of CSS if they are not used (so potentially they can be global and included automatically with import) they also can duplicate - when same file would be imported multiple times, any global placeholders will also multiply, which can lead to unexpected increase of file size. Keeping them in placeholder and explicitly including them makes it more controlled, so developer can make sure they are included only once.

lyubomir-popov:
@anthonydillon @bartaz Rules placed outside of mixins end up at the top of the css file. The reset may then override them.

PR #3164, opened 2020-07-10
Change default font-display value to fallback

### LINKED ISSUE #3157: Update the default font-display
## Pattern to amend
Default font display 

## Context
The new default font-display is set to optional. It turns out that means never switch even if the font loads very quickly. When using it on ubuntu.com I believe `fallback` would be better.

**swap**: Gives the font face an extremely small block period and an infinite swap period.
**fallback**: Gives the font face an extremely small block period and a short swap period.
**optional**: Gives the font face an extremely small block period and no swap period.

bartaz:
## Done

Because of the cases of Ubuntu font not rendering at all in some cases, even on fast connections, we change default value to 'fallback' - this makes sure the text is rendered immediately, but web font still has a small time window to load to be rendered.

Fixes #3157 

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/ or [demo](https://vanilla-framework-3164.demos.haus)
- Make sure text is rendered immediately with fallback font, but Ubuntu font is loaded on fast connection

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "clean": "rm -rf build docs/static/css node_modules/ yarn-error.log",
     "percy": "percy exec -- node snapshots.js"
   },
-  "version": "2.14.0",
+  "version": "2.14.1",
   "files": [
     "/scss",
     "!/scss/docs"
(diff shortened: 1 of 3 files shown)


bartaz:
> Sorry do you need to update the docs?

Ah, good catch, of course. Updated.

PR #3175, opened 2020-07-20
Add application layout with examples and docs

bartaz:
## Done

Adds application layout and panel component with documentation and examples.

## QA

- `./run` or [demo](https://vanilla-framework-3175.demos.haus/docs/examples/layout/application)
- Check application layout documentation: https://vanilla-framework-3175.demos.haus/docs/layouts/application
  - please note that CodePen will show unstyled examples (as this version is not published yet)
- Check [Application layout structure example](https://vanilla-framework-3175.demos.haus/docs/examples/layout/application-structure)
     - verify it behaves as expected on all screen sizes
- Check full [Application layout example](https://vanilla-framework-3175.demos.haus/docs/examples/layout/application)
     - verify it behaves as expected on all screen sizes

### Structure

Screenshot 2020-08-04 at 12 07 32

### Small screen

Screenshot 2020-08-04 at 12 08 42

### Medium screen

Screenshot 2020-08-04 at 12 08 28

### Large screen

Screenshot 2020-08-04 at 12 07 53

### Tasks:

- [x] move is-pinned class to sidenav instead of top level layout
- [x] ~~make side nav 'mobile 'toggle' bar' part of side navigation panel~~ (not possible without additional wrapper markup)
- [x] overlay and aside should be one panel, just different variant (pinned/unpinned)
- [x] separate styling classes into consistent child panel element
  - make sure it works with small amount of content (nothing to scroll) and with lot of content (scroll)
  - move border and shadow to layout styling from panel
- [x] remove any unnecessary demo styling
- [x] create panel component based on existing demo styling
- [x] create minimal layout demo (just layout, no content)
- [x] fix demo side navigation include (to avoid docs side navigation drawer)
- [x] fix sidebar panel scrolling issue (side nav scrolls and whole panel scrolls)
- [x] fix stacked form demo HTML (includes)
- [x] add application layout documentation
- [x] remove styling from layout
- [x] separate file for panel styling

PR #3179, opened 2020-07-22
Chips improvements

### LINKED ISSUE #3192: Enhancments to chip overflow
## Pattern to amend

The current chip proposal may result in the "dismiss" button overflowing and not being visible. Creating this issue so we can  merge the proposal, then consider the best approach to solve this. 

For some context - all elements inside the chip are display:inline, with the chip itself inline-block. This ensures wrapping happens as expected. However, It means a simple overflow on the parent hides the dismiss button. In itself, this can easily be acheived with a flex layout, but that would cause vertical align issues and wrapping issues.

image

The simplest solution I can think of is to wrap the text in a div, whose width is equal to the parent minus the width of the buttton.

lyubomir-popov:
## Done

Based on #3175; improve styling of chips. Includes code for the popover menu. This is just to ensure the chips fit as expected in the bigger picture. This will be deleted before merging.

## QA

- Pull code
- Run `./run`
- Open docs/examples/patterns/chip/multiple
To evaluate in situ, add the following [scss](https://pastebin.canonical.com/p/Y7HqmMBFMD/) and replace the markup of any chips example with [this](https://pastebin.canonical.com/p/KZtFNSd8zW/)

- Verify it looks like this:

image

Fixes #3179 
Fixes #3192


lyubomir-popov:
@anthonydillon I made some changes, mainly opted for display:inline on the chip and everything in it. 
The height of the chips wrapper  uses the height of individual chips and the margin between them to ensure precise fit.
I've also adjusted margins and paddings here and there to get it closer to the mockup.

lyubomir-popov:
@anthonydillon this is ready for another look

anthonydillon:
Unfortunately, display inline causes some wrapping issues on small screens:
image

anthonydillon:
The dismiss button has margin-bottom:
Screenshot from 2020-07-24 22-34-55

lyubomir-popov:
@anthonydillon updated to fix the inline issues, ready for another look.

scss/_patterns_chip.scss
```diff
@@ -1,18 +1,19 @@
...
 
     background-color: rgba($color-x-dark, 0.05);
     border-radius: $chip-height * 0.5;
-    display: inline;
-    margin-bottom: $chip-margin-bottom;
-    margin-right: $sph-inner--small;
+    display: inline-block;
+    line-height: $chip-line-height;
+    margin: $chip-margin-top $sph-inner--small 0 0;
     padding: $chip-vertical-padding $sph-inner;
```
anthonydillon, line 17: I would suggest changing this from a result of:
```css
padding: 0 1rem;
```
to
```scss
padding-left: $sph-inner;
padding-right: $sph-inner;
```
Long-form is generally preferred over shorthand as its more specific and results it less complicated overrides.

scss/_patterns_chip.scss
```diff
@@ -1,18 +1,19 @@
 @import 'settings';
 
 @mixin vf-p-chip {
-  $chip-vertical-padding: map-get($nudges, nudge--small);
-  $chip-height: map-get($line-heights, small) + 2 * $chip-vertical-padding;
-  $chip-margin-bottom: $spv-outer--small-scaleable - 2 * $chip-vertical-padding;
+  $chip-line-height: map-get($line-heights, default-text);
+  $chip-vertical-padding: 0;
```
anthonydillon, line 5: All uses of this variable are obsolete so you can remove it once addressed below.

scss/_patterns_chip.scss
```diff
@@ -0,0 +1,75 @@
...
+    @extend %small-text;
+
+    background-color: rgba($color-x-dark, 0.05);
+    border-radius: $chip-height * 0.5;
+    display: inline-block;
+    line-height: $chip-line-height;
+    margin: $chip-margin-top $sph-inner--small 0 0;
+    padding: $chip-vertical-padding $sph-inner;
+    user-select: none;
+    white-space: nowrap;
```
anthonydillon, line 19: It might also be nice to add overflow control with the following:
```css
max-width: 100%;
overflow: hidden;
text-overflow: ellipsis;
```

scss/_patterns_chip.scss
```diff
@@ -0,0 +1,75 @@
...
+      line-height: $chip-line-height;
+      text-transform: uppercase;
+    }
+
+    .p-chip__value {
+      @extend %small-text;
+
+      display: inline;
+      line-height: $chip-line-height;
+      position: relative;
```
anthonydillon, line 50: I don't think you need display: relative here as your not changing the position.

lyubomir-popov:
@anthonydillon I've addressed the feedback above; with regards to overflow, I'm not sure we can properly solve this with the current markup:

image

I think the two pieces of text (__lead and __value) need to be in a wrapper so I can fix the width of the wrapper to that of the chip - the width of the button. Otherwise, the dismiss button is pushed out of view and clipped by the overflow: hidden.

lyubomir-popov:
@anthonydillon Ok, created an issue to track the [overflow problem](https://github.com/canonical-web-and-design/vanilla-framework/issues/3192)

lyubomir-popov:
Hmm, overflow still needs some work in firefox:

lyubomir-popov:
@anthonydillon I think I finally got it to work:

image

templates/docs/examples/patterns/chip/multiple.html
```diff
@@ -0,0 +1,12 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Chip / Multiple{% endblock %}
+
+{% block standalone_css %}patterns_chip{% endblock %}
+
+{% block content %}
+
+<div class="p-chip">AWS</div>
```
anthonydillon, line 8: Shouldn't these have `p-chip__value` wrapping them otherwise users wont use them.

lyubomir-popov:
@anthonydillon updated.

anthonydillon:
Thanks 👍

PR #3193, opened 2020-07-29
Add suru strips to Vanilla

### LINKED ISSUE #2539: Build new Suru template set
## Suru templates
Build new templates based on [defined set](https://github.com/canonical-web-and-design/design-vanilla-framework/issues/426) and [colorways](https://github.com/canonical-web-and-design/design-vanilla-framework/issues/427).

_Note:_ Designs will be added once complete.

MeltySnow:
## Done

- Added new strip classes: p-strip--suru and p-strip--suru-topped
- Added new example templates for the new classes
- Updated the documentation to include the new classes

Fixes #2539 

## QA

- Pull code
- Run `./run`
- Navigate to http://0.0.0.0:8101/docs/examples/patterns/strips/suru and ensure that the pattern is congruent with the [specification](https://github.com/canonical-web-and-design/vanilla-framework/issues/2539) and [meeting notes](https://github.com/canonical-web-and-design/meeting-notes/issues/596).
  - Do the same for http://0.0.0.0:8101/docs/examples/patterns/strips/suru-topped
- Navigate to the strip section in the [documentation](http://0.0.0.0:8101/docs/patterns/strip) and ensure that the copy is sensible and consistent in style to the rest of the documentation.
- Ensure that the code is sensible and consistent in style to the rest of the code.

## Screenshots

image

image

### DIFF
--- a/scss/_patterns_strip.scss
+++ b/scss/_patterns_strip.scss
@@ -119,6 +119,11 @@
 @mixin vf-p-strip-suru {
   .p-strip--suru {
     @extend %vf-strip;
+    $brand-color: #333333;
+    $color-suru-start: lighten($brand-color, 10%);
+    $color-suru-middle: $brand-color;
+    $color-suru-end: darken($brand-color, 10%);
+
     background-blend-mode: multiply, multiply, normal, normal, normal;
     background-image:
       linear-gradient(
@@ -148,9 +153,9 @@
       ),
       linear-gradient(
         111deg,
-        #4e4e4e 10%,
-        #333333 37%,
-        #111111 100%
+        $color-suru-start 10%,
+        $color-suru-middle 37%,
+        $color-suru-end 100%
       );
     background-position: 0% 0%, top right, right 0 bottom 4rem, right bottom,
     0% 0%;
--- a/scss/_patterns_strip.scss
+++ b/scss/_patterns_strip.scss
@@ -233,7 +233,7 @@ $color-suru-end: darken($color-brand, 10%) !default;
     }
 
     &.is-shallow {
-      padding: 2rem 0rem 6rem 0rem;     
+      padding: 1.5rem 0rem 6rem 0rem;     
     }
   }
 }
--- a/scss/_patterns_strip.scss
+++ b/scss/_patterns_strip.scss
@@ -106,7 +106,8 @@
 }
 
 @mixin vf-p-strip-shallow {
-  [class*='p-strip']:not([class*='suru']).is-shallow {
+  [class*='p-strip']:not([class*='-suru']).is-shallow {
+    // Suru strips were not designed to be shallow, so they are excluded
     @extend %section-padding--shallow;
   }
 }
--- a/scss/_patterns_strip.scss
+++ b/scss/_patterns_strip.scss
@@ -49,6 +49,7 @@
   @include vf-p-strip-shallow;
   @include vf-p-strip-deep;
   @include vf-p-strip-suru;
+  @include vf-p-strip-suru-topped;
 }
 
 @mixin vf-p-strip-default {
@@ -116,14 +117,14 @@
   }
 }
 
+$brand-color: #333333;
+$color-suru-start: lighten($brand-color, 10%);
+$color-suru-middle: $brand-color;
+$color-suru-end: darken($brand-color, 10%);
+
 @mixin vf-p-strip-suru {
   .p-strip--suru {
     @extend %vf-strip;
-    $brand-color: #333333;
-    $color-suru-start: lighten($brand-color, 10%);
-    $color-suru-middle: $brand-color;
-    $color-suru-end: darken($brand-color, 10%);
-
     background-blend-mode: multiply, multiply, normal, normal, normal;
     background-image:
       linear-gradient(
@@ -230,5 +231,45 @@
         padding-bottom: ($padding * 3) !important;
       }
     }
+  }
+}
+
+@mixin vf-p-strip-suru-topped {
+  .p-strip--suru-topped {
+    @extend %vf-strip;
+    background-blend-mode: multiply, multiply, normal, normal;
+    background-image:
+      linear-gradient(
+        to bottom left,
+        rgba(229, 229, 229, 0.5) 0%,
+        rgba(229, 229, 229, 0.5) 49%,
+        transparent 50%,
+        transparent 100%
+      ),
+      linear-gradient(
+        to bottom left,
+        rgba(205, 205, 205, 0.16) 0%,
+        rgba(205, 205, 205, 0.16) 49%,
+        transparent 50%,
+        transparent 100%
+      ),
+      linear-gradient(
+        to bottom right,
+        transparent 0%,
+        transparent 49%,
+        rgba(255, 255, 255, 1) 50%,
+        rgba(255, 255, 255, 1) 100%
+      ),
+      linear-gradient(
+        -75deg,
+        $color-suru-start 2%,
+        $color-suru-middle 28%,
+        $color-suru-end 89%
+    );
+    background-position: top right, top right, top left, top left;
+    background-repeat: no-repeat;
+    background-size: 39.4% 6rem, 54% 4rem, 63% 4rem, 62.6% 4rem;
+    padding-bottom: 4rem;
+    padding-top: 6rem;
   }
 }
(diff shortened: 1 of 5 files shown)


bartaz:
Thanks for the PR @MeltySnow.

Couple quick comments.

First, prettier check seems to be failing on your changes. Please make sure you format the code with Prettier in the editor.

Second, not very important, but maybe something you may look into, seems like the git user details you are using to commit, are not the same as your GitHub user name, so GitHub doesn't know it's @MeltySnow who made the commit:

Screenshot 2020-07-29 at 13 25 08

bartaz:
The standalone example for suru strip uses `p-button--positive` which is not included in standalone strip stylesheet (standalone stylesheets contain only the styles of given component, so we can test if components don't depend on each other).

This causes button not to render on standalone example (https://vanilla-framework-3193.demos.haus/docs/examples/standalone/patterns/strips/suru):
Screenshot 2020-07-30 at 07 36 09

I think the easiest way to workaround this would be to change `<a class="p-button--positive">Get started</a>` (which requires button pattern) to `<button>Get started</button>` that only requires base styles that are always available.

bartaz:
Topped Suru doesn't seem to be supporting `is-shallow`

Screenshot 2020-07-30 at 09 35 27

@anthonydillon @lyubomir-popov I don't know if it needs to support it, but for the consistency I guess it would be good. The problem with that is that we would need to make the topped suru itself smaller to make sure text doesn't overlap it (on any screen size).

lyubomir-popov:
@waynecrosby do you think the shallow strip should be allowed to use this background, albeit modified to not overlap the text in the available padding of 24px?

image

bartaz:
The shallow version of regular Suru also has bottom line overlapping with the slanted background, which is not expected I guess?

Screenshot 2020-07-30 at 09 46 45

I guess the issue here is that surus (as on Ubuntu) were designed for a specific spacing in mind and don't really support flexibility of Vanilla shallow/deep variants out of the box.

@anthonydillon @lyubomir-popov  Do we want to update them and make sure they work with shallow/deep variants, or should we just assume they come as they are and should not be altered with other variants?

bartaz:
@lyubomir-popov I'm also not sure about light/dark variants for surus.

Currently `p-strip--suru` seems to be implementing `is-light` and `is-dark` variants, but it seems unnecessary. To my understanding suru backround is always going to be dark gradient, so text color should always be light. There shouldn't be a need for `is-dark` class to change it, as it's dark by default. Also, `is-light` is probably not needed, because it doesn't affect the background at all.

Similarly, `p-strip--suru-topped` is has always text on light background, so it also doesn't need light/dark variants (and these are not implemented currently).

scss/_patterns_strip.scss
```diff
@@ -114,3 +116,88 @@
...
+@mixin vf-p-strip-suru {
+  .p-strip--suru {
+    @extend %vf-strip;
+
+    background-blend-mode: multiply, multiply, normal, normal, normal;
+    background-image: linear-gradient(to bottom right, rgba(205, 205, 205, 0.55) 0%, rgba(205, 205, 205, 0.55) 49.8%, transparent 50%, transparent 100%),
+      linear-gradient(to bottom left, rgba(205, 205, 205, 0.55) 0%, rgba(205, 205, 205, 0.55) 49.8%, transparent 50%, transparent 100%),
+      linear-gradient(to top right, #fff 0%, #fff 49%, transparent 50%, transparent 100%), linear-gradient(#fff 0%, #fff 100%),
+      linear-gradient(111deg, $color-suru-start 10%, $color-suru-middle 37%, $color-suru-end 100%);
+    @supports not (background-blend-mode: multiply) {
```
bartaz, line 133: I think we should do it another way around, by default provide the styling for browsers not supporting blend mode multiply (because they may not support `@support` as well...) and use `@support` to detect browsers that do.

So something like:

```scss

background-image: // gradients without multiply

@supports (background-blend-mode: multiply) {
   background-blend-mode: multiply, multiply, normal, normal, normal;
   background-image: // gradients with multiply
}
```

lyubomir-popov:
I think this highlights a few open questions that need to be discussed in design / brand before making a final call on the implementation: 
- What is the minimum height at which we consider the diagonals effective?
- Does that height differ per background? Should it be specified per background?
- How should overlap with text be resolved? For example, the options could be reducing the anles, or prohibiting use under a certain height.
- What colours can be used in place of the current black one? Substituting with orange for example will create an accessibility issue against both white and link blue

waynecrosby:
@lyubomir-popov If we are to use the shallow strip the cross over point would need to be moved across to the left along way so the dark suru doesn't impact the type. There would be a lot more light grey which the type will overlap so we would have to make a call on how that look.

@bartaz The shallow version of regular Suru also has bottom line overlapping with the slanted background, which is not expected I guess? - No there should definitely not be any overlap like that.

MeltySnow:
Alright, I've pushed up changes that should address the suggestions to use html button elements instead of the p-button--positive class, and to assume multiply isn't supported by default.

I'll wait until a consensus is reached on the design issues before changing anything there, but I'd just like to raise that the non-multiply versions of the suru strips currently use the same rgb values as the regular versions but at a lower opacity (the implementation on ubuntu.com does this as well). This makes the gradients look rather light and I'm not sure whether this behaviour is intended.
Screenshot 2020-07-30 at 10 35 57

bartaz:
@MeltySnow Hey, we talked with @anthonydillon and @lyubomir-popov and not to block it for too long on design decisions we'd like to give suru a go as it is, but remove problematic variants.

It seems that `is-shallow` variant is causing issues currently (both in suru and suru-topped), so please remove these variants from the code and documentation.

Also, as I mentioned before, suru shouldn't need `is-dark` to change color. It should use light text color by default (just as on ubuntu). I also think that we probably don't need `is-light` variant at all for now. Suru backgrounds are meant to be dark (maybe we need some documentation guidance on that). I would not add the light variant until we have a case that really needs it.

MeltySnow:
Alright, I've pushed up changes that should address these points.

bartaz:
@MeltySnow Thanks. Could you also rebase with master? There has been a release in the meantime, would be good to get all the changes to this branch as well before merging (and to fix unnecessary Percy check fails).

scss/_patterns_strip.scss
```diff
@@ -104,7 +106,7 @@
 }
 
 @mixin vf-p-strip-shallow {
-  [class*='p-strip'].is-shallow {
+  [class*='p-strip']:not([class*='suru']).is-shallow {
```
bartaz, line 109: To avoid situation that `suru` is part of some longer word I'd add a dash before it `-suru`.

```suggestion
  [class*='p-strip']:not([class*='-suru']).is-shallow {
```

Also it would be nice to add a code comment saying why we are adding this exception. Something like: `// Suru strips were not designed to be shallow, so they are excluded`

MeltySnow:
Done!

bartaz:
Thanks to Percy screenshots we noticed an unexpected regression:

Screenshot 2020-08-03 at 08 10 12

If you compare [current "MAAS docs" example](https://vanillaframework.io/docs/examples/templates/maas-docs) to the one [after adding suru patterns](https://vanilla-framework-3193.demos.haus/docs/examples/templates/maas-docs) you can see that there is additional padding on the bottom of the title.

"MAAS docs" here are a copy of some of our existing product styles, while there are not pure Vanilla they reflect what may be on one on our websites. It happens to be affected by Suru changes, even if they are not using suru at all. This is unexpected and we should avoid this. 

It seems that the issue is caused by updated `is-shallow` selector.

Previously (as in current Vanilla) selector was simply `.is-shallow[class*='p-strip']` which MAAS docs override to remove bottom padding in the title.
New selector `.is-shallow[class*='p-strip']:not([class*='-suru'])` has a higher specificity and is not overridden anymore.

I think the biggest issue in this case is that it may (as in this example) affect strips that are not suru related at all.

To avoid that we would need to leave the `.is-shallow[class*='p-strip']` selector intact, making sure all existing styling works as expected. This will mean `-suru` strips will be affected by this as well, so we should separately "disable" padding change on suru strips.

This may look something like this (I haven't tested it, just a rough shape of SCSS) - for both `-suru` and `-suru-topped`:

```scss
.p-strip--suru {
   // Suru strips were not designed to be shallow, so they are excluded
   &.is-shallow {
      @extend %section-padding--default;
   }
}
```

MeltySnow:
Alright, I've reverted the is-shallow selector. `@extend %section-padding--default` didn't work as both variants of Suru strip have more padding than the default for p-strips, so I made custom overrides for both.

lyubomir-popov:
@MeltySnow @bartaz design looks good, well done!

PR #3203, opened 2020-08-12
Update navigation highlight: option 1

lyubomir-popov:
## Done

Increase the prominance of hover and selected navigation items. 
This is option one of two, option 2 (as in charmhub) will be on a separate branch.

Do not merge: the two options need to be presented and whichever is chosen can then be merged.

## QA

- Pull code
- Run `./run`
- Open /docs/examples/patterns/navigation/default-dark#navigation or [demo]
- Verify it looks like this:

## Screenshots
image

### DIFF
--- a/scss/_patterns_navigation.scss
+++ b/scss/_patterns_navigation.scss
@@ -320,6 +320,8 @@ $lightness-threshold: 70;
   %navigation-link-selected {
     @extend %navigation-link-hover;
 
+    font-weight: $font-weight-bold;
+
     &::before {
       background-color: $color-navigation-text;
       content: '';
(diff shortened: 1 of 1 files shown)


cassiocassio:
Can we make a comparison between the menu like this, and with the Canonical/Product strip above - the more common version?

Perhaps bar at the top -right at the edge of the screen- will feel very different to a bar floating above one menu and below the other?

lyubomir-popov:
@cassiocassio It is most sensitive to light OS chrome, no global nav. I've added a one pixel ofset so it doesn't blend with the browser's chrome, most visible in the third screenshot:

image
image
image
image

PR #3204, opened 2020-08-12
Navigation highlight update option 2

lyubomir-popov:
## Done

Increase the prominance of hover and selected navigation items.
This is option 2, (as in charmhub).

Do not merge: the two options need to be presented and whichever is chosen can then be merged.

## QA

- Pull code
- Run `./run`
- Open /docs/examples/patterns/navigation/default-dark#navigation or [demo]
- Verify it looks like this:

## Screenshots

image

### DIFF
--- a/scss/_patterns_navigation.scss
+++ b/scss/_patterns_navigation.scss
@@ -306,26 +306,8 @@ $lightness-threshold: 70;
   }
 
   %navigation-link-selected {
-    &::after {
-      background-color: $color-background-hover;
-      content: '';
-      height: 100%;
-      left: 0;
-      position: absolute;
-      right: 0;
-      top: 0;
-    }
-
-    &::before {
-      background-color: $color-navigation-text;
-      content: '';
-      height: $bar-thickness;
-      left: 0;
-      position: absolute;
-      right: 0;
-      top: 0;
-      z-index: 1;
-    }
+    background-color: $color-navigation-text;
+    color: $color-navigation-background;
   }
 
   %navigation-link-before {
@@ -369,7 +351,7 @@ $lightness-threshold: 70;
       $color-navigation-background: $colors--light-theme--background-default,
       $color-navigation-text: $colors--light-theme--text-default,
       $color-navigation-separator: $colors--light-theme--border-default,
-      $color-background-hover: $colors--light-theme--background-hover
+      $color-background-hover: $colors--dark-theme--text-default
     );
   } @else {
     // take $color-navigation-background (that can be overridden) into account
@@ -380,7 +362,7 @@ $lightness-threshold: 70;
       $color-navigation-background: $color-navigation-background,
       $color-navigation-text: $color-navigation-text,
       $color-navigation-separator: $colors--light-theme--border-default,
-      $color-background-hover: $colors--light-theme--background-hover
+      $color-background-hover: $colors--dark-theme--text-default
     );
   }
 }
@@ -391,7 +373,7 @@ $lightness-threshold: 70;
       $color-navigation-background: $colors--dark-theme--background-alt,
       $color-navigation-text: $colors--dark-theme--text-default,
       $color-navigation-separator: $colors--dark-theme--border-default,
-      $color-background-hover: $colors--dark-theme--background-hover
+      $color-background-hover: $colors--dark-theme--text-default
     );
   } @else {
     // take $color-navigation-background (that can be overridden) into account
@@ -402,7 +384,7 @@ $lightness-threshold: 70;
       $color-navigation-background: $color-navigation-background,
       $color-navigation-text: $color-navigation-text,
       $color-navigation-separator: $colors--dark-theme--border-default,
-      $color-background-hover: $colors--dark-theme--background-hover
+      $color-background-hover: $colors--dark-theme--text-default
     );
   }
 }
--- a/scss/_patterns_navigation.scss
+++ b/scss/_patterns_navigation.scss
@@ -115,6 +115,10 @@ $lightness-threshold: 70;
     display: flex;
   }
 
+  %vf-reset-horizontal-padding {
+    padding-left: 0;
+    padding-right: 0;
+  }
   // no colours
   .p-navigation {
     display: flex;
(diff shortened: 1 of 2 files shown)


lyubomir-popov:
@bartaz rebased and ready for review.

bartaz:
While on dark navigation it kind of looks OK in such high contrast, on light variant it looks like a bug:

Screenshot 2020-09-10 at 14 07 54

With light navigation the current item in the nav may be the only extremely dark thing in the navigation.

scss/_patterns_navigation.scss
```diff
@@ -283,28 +282,32 @@ $lightness-threshold: 70;
 
 @mixin vf-navigation-theme(
   // color for navigation separator on small screens
-  $color-navigation-separator,
+    $color-navigation-separator,
```
bartaz, line 285: Prettier is complaining on something, possibly these spaces.

It was changed in recent version of prettier, so make sure you have up to date dependencies.

bartaz:
Could hover be less prominent maybe?

It looks and behaves quite jarred with this white block jumping on mouse move (the framerate of the gif make it look better than it is in reality)...

nav-hover

scss/_patterns_navigation.scss
lyubomir-popov: these are added automatically on save. I did dotrun clean and dotrun, is that enough to refresh dependancies?

lyubomir-popov:
I think we should remove the animation. very annoying and I think material also only animate on mobile.

spencerbygraves:
Let's keep the current hover state, which is also used on https://charmhub.io/

scss/_patterns_navigation.scss
```diff
@@ -287,24 +286,28 @@ $lightness-threshold: 70;
   // color for navigation background
   $color-navigation-background,
   // color for navigation text
-  $color-navigation-text
+  $color-navigation-text,
+  // transparent color of selected and hovered menu item
+  $color-background-hover
```
bartaz, line 291: Is this used anywhere?

lyubomir-popov: I've picked the wrong choice during conflict resolution, removed now.

scss/_patterns_navigation.scss
```diff
@@ -292,19 +291,21 @@ $lightness-threshold: 70;
   background-color: $color-navigation-background;
 
   %navigation-link-theme {
+    position: relative;
```
bartaz, line 294: If this is needed (doesn't seem to) it shouldn't be in the theme, but in the component styling.

bartaz:
Hover in the subnav is now inconsistent:

nav-hover-subnav

There are even more issues on small screens (especially with sub navigations being different)

nav-hover-mobile

scss/_patterns_navigation.scss
lyubomir-popov: not needed for this version of the design

bartaz:
@lyubomir-popov @spencerbygraves 

I just want to be sure there is a Design +1 for that:

Screenshot 2020-09-10 at 16 18 04

While it is not ideal in our dark navigations it at least doesn't look terrible.
Black square in the middle of white navigation does look terrible.

It's (probably) not being used on any of our internal sites, but that's the default that we ship with Vanilla as open-source project. We are going to break navigation for other people.

lyubomir-popov:
@bartaz Just discussed this with @spencerbygraves, we decided to go with a f7f7f7 light grey highlight for the light theme. 
I will fix the other subnav issues, apologies for missing them the first time. Being a separate pattern I forgot it will also be afected by this.

lyubomir-popov:
@bartaz had to signifnicantly rewrite nav and subnav padding to get this to work, but I think it's ok now.

Screenshot 2020-09-14 at 13 23 19
Screenshot 2020-09-14 at 13 23 06

image

Screenshot 2020-09-14 at 13 23 54

bartaz:
There is some issue with disappearing border on hover on small screen navigation.
It's most apparent on light theme. It's not really visible on dark theme, but this may be because of the colors.

nav-hover-mobile-border

lyubomir-popov:
The separators issue is fixed now:
image

bartaz:
Noticed this in percy - there seems to be some additional padding around search field now. Is this expected?
It makes it not aligned with right side of the page anymore.

Before:
Screenshot 2020-09-15 at 13 20 18


After:

Screenshot 2020-09-15 at 13 20 05

scss/_patterns_navigation.scss
```diff
@@ -14,24 +13,61 @@ $lightness-threshold: 70;
...
   }
 
+  %navigation-link-responsive-padding-vertical {
+    padding-bottom: $spv-inner--medium;
+    padding-top: $spv-inner--medium;
+  }
+
+  %navigation-link-responsive-padding-left {
+    // follows grid padding to ensure nav items on small breakpoints align with grid padding
+    @media (max-width: $threshold-4-6-col) {
```
bartaz, line 23: It's probably question for later, but do we need that many different paddings on different screen sizes?
It creates a lot of CSS in media queries in places where we need to align with the grid.

Especially that medium and large are actually the same:
https://github.com/canonical-web-and-design/vanilla-framework/blob/6e1d68232ce8611eb2aa887972fd8afb4af13ef2/scss/_settings_grid.scss#L24-L25

lyubomir-popov:
@bartaz no, the margin on the search box was a duplicate, thanks. Removed.

scss/_patterns_navigation.scss
lyubomir-popov: No, I guess small/default would suffice. Will create a new issue and tackle separately.

bartaz (approved):
LGTM.

Great work, thanks for the updates and little tweaks, despite some "controversy" around design it feels like improvement in the end.

bartaz:
@lyubomir-popov Sorry, another thing noticed in Percy. There seems to be double margin in full width example:
Screenshot 2020-09-15 at 14 18 52

lyubomir-popov:
Tested dqlite running locally on this bransh, nothing looks broken. (The bold body text is just to verify it runs off of my local vanilla)

image
image

lyubomir-popov:
@bartaz ah the full width row, I'd missed that. Fixed now, thanks for spotting!

scss/_patterns_navigation.scss
```diff
@@ -95,10 +135,12 @@ $lightness-threshold: 70;
     &__row {
       @extend %fixed-width-container;
       @extend %navigation-row;
+      @extend %vf-reset-horizontal-padding;
 
       &--full-width {
         @extend %vf-grid-container-padding;
         @extend %navigation-row;
+        @extend %vf-reset-horizontal-padding;
```
bartaz, line 143: If we need to reset paddings to 0, why do we set them 2 lines above with `%vf-grid-container-padding`?

PR #3215, opened 2020-08-25
Base form tick elements refactoring

bartaz:
## Done

Extracts styled tick boxes styling into placeholders, so they can be reused when building new checkbox pattern.
This is just a refactor branch, it only restructures the code, no functionality or styling changes in the branch.

## QA

- `./run` or [demo](https://vanilla-framework-3215.demos.haus)
- Make sure [tick elements](https://vanilla-framework-3215.demos.haus/docs/examples/templates/tick-element-comparison) work as before, there should be no regressions or changes to styling

### DIFF
--- a/scss/_base_forms-tick-elements.scss
+++ b/scss/_base_forms-tick-elements.scss
@@ -1,5 +1,10 @@
 // tick element variables
 $form-tick-box-size: 1rem;
+$form-tick-height: 0.375rem;
+$form-tick-offset-top: 0.1875rem;
+
+$form-radio-inner-circle-diameter: 0.375rem;
+$form-radio-circle-offset: 0.5 * ($form-tick-box-size - $form-radio-inner-circle-diameter);
 
 // Offsets to align tick elements to the baseline of headings.
 $box-offsets-top: (
@@ -30,6 +35,9 @@ $box-offsets-top: (
   }
 
   %vf-pseudo-tick-box {
+    padding-left: $sph-inner + $form-tick-box-size;
+    position: relative;
+
     &::before,
     &::after {
       $properties: #{background-color, border-color};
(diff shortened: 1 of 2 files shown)


scss/_base_forms-tick-elements.scss
```diff
@@ -18,40 +23,114 @@ $box-offsets-top: (
...
-      &::before {
-        content: '';
-        height: $box-size;
-        left: 0;
-        outline-offset: 1px;
-        width: $box-size;
-      }
+  %vf-pseudo-radio {
+    &::before,
+    &::after {
```
lyubomir-popov, line 101: border-radius is defined on 106 too - should one be dropped? I think separate makes it easier to work with

scss/_base_forms-tick-elements.scss
```diff
@@ -18,40 +23,114 @@ $box-offsets-top: (
...
+      $properties: #{background-color, border-color};
+      @include vf-animation($properties);
+
+      position: absolute;
+    }
+
+    // container
+    &::before {
```
lyubomir-popov, line 50: An idea - feel free to ignore as this may complicate things. But should we hardcode the styling of container / tick onto before/after? Or is it likely that we may want to change that because of things like the *required functionality? 

One thing I was thinking of was creating a mixin that allows you to attach to before/after using arguments with default values like` before: container, after: tick ` maybe?

lyubomir-popov (commented):
Looks good, added a couple of comments.

scss/_base_forms-tick-elements.scss
bartaz: I would prefer not to complicate things until it's needed. I don't expect `required` to be a problem, because ::before and ::after of the tick will be on different element (`<span class="p-checkbox__label">`) than ::after from the required (on `<label class="p-checkbox">`.

We can always restructure the placeholders/mixins if needed, but hopefully what is here will allow us to build components, and then in 3.0, if we drop base styling, we won't need that much of reuse of checkboxes.

scss/_base_forms-tick-elements.scss
bartaz: Oh, didn't notice that. I just copied what we had before.

lyubomir-popov (approved):

PR #3260, opened 2020-09-04
Add alt attribute to images that do not have one

lyubomir-popov:
## Done

Fixes https://github.com/canonical-web-and-design/vanilla-squad/issues/877

## QA

- Search for images without alt or with empty alt. I used the following: (]*)(>)
- Verify there are none
- Go through the changes, verify the alt text is plausible

### DIFF
--- a/templates/_layouts/examples.html
+++ b/templates/_layouts/examples.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 
-<html>
+<html lang="en">
   <head>
     <meta charset="utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1">
--- a/templates/contribute.html
+++ b/templates/contribute.html
@@ -108,7 +108,7 @@
     <div class="col-6">
       <div class="p-heading-icon">
         <div class="p-heading-icon__header is-stacked">
-          <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/60d9b81e-picto-canonical.svg" alt="Canonical pictogram"/>
+          <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/60d9b81e-picto-canonical.svg" alt=""/>
           <h3 class="p-heading-icon__title">Licences</h3>
         </div>
       </div>
--- a/templates/contribute.html
+++ b/templates/contribute.html
@@ -108,7 +108,7 @@
     <div class="col-6">
       <div class="p-heading-icon">
         <div class="p-heading-icon__header is-stacked">
-          <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/60d9b81e-picto-canonical.svg" alt="">
+          <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/60d9b81e-picto-canonical.svg" alt="Canonical pictogram"/>
           <h3 class="p-heading-icon__title">Licences</h3>
         </div>
       </div>
--- a/templates/docs/examples/layouts/application.html
+++ b/templates/docs/examples/layouts/application.html
@@ -19,7 +19,7 @@
             <div class="p-inline-list--stretch">
               <div class="p-inline-list__item">
                 <a class="p-panel__logo" href="#">
-                  <img src="https://assets.ubuntu.com/v1/0b2b7bce-Vanilla-logo_88x24px_white.svg" alt="Vanilla framework logo" height="24">
+                  <img src="https://assets.ubuntu.com/v1/0b2b7bce-Vanilla-logo_88x24px_white.svg" alt="" height="24">
                 </a>
               </div>
               <div class="p-inline-list__item u-align--right u-hide--large">
--- a/templates/docs/examples/layouts/application.html
+++ b/templates/docs/examples/layouts/application.html
@@ -19,7 +19,7 @@
             <div class="p-inline-list--stretch">
               <div class="p-inline-list__item">
                 <a class="p-panel__logo" href="#">
-                  <img src="https://assets.ubuntu.com/v1/0b2b7bce-Vanilla-logo_88x24px_white.svg" alt="" height="24">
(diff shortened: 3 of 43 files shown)


templates/contribute.html
```diff
@@ -108,7 +108,7 @@ <h3 class="p-heading-icon__title">Running tests locally</h3>
     <div class="col-6">
       <div class="p-heading-icon">
         <div class="p-heading-icon__header is-stacked">
-          <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/60d9b81e-picto-canonical.svg" alt="">
```
bartaz, line 111: Again, this added `alt` has no content value.

templates/_layouts/examples.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 
-<html>
```
bartaz, line 3: Changes from another PR

templates/showcase.html
```diff
@@ -23,12 +23,12 @@ <h2>Projects using Vanilla</h2>
   </div>
   <div class="row u-equal-height">
     <div class="col-4 p-card">
-      <img src="https://assets.ubuntu.com/v1/25df3b87-ubuntu-vf.io-showcase.png" alt="">
```
bartaz, line 26: Repeating "Showcase screenshot" over and over again is not really helpful.
As each example has its own text it should be sufficient as description of given page. Images don't provide any additional value (other than visual), so are completely fine with empty alt text.

templates/docs/examples/templates/typographic-spacing.html
```diff
@@ -1639,42 +1639,42 @@ <h3>This is a heading following an image</h3>
     <h1>Heading</h1>
     <div class="p-strip--dark">
       <h1>This is an h1 heading inside a dark strip</h1>
-      <img src="https://assets.ubuntu.com/v1/06153e05-screenshot-tv-homepage-wide-image-900x334.jpg" alt="">
```
bartaz, line 1642: Empty alt is good enough.

templates/docs/examples/templates/z-index.html
```diff
@@ -4,7 +4,7 @@
     <div class="p-navigation__banner">
       <div class="p-navigation__logo">
         <a class="p-navigation__item" href="#">
-          <img class="p-navigation__image" src="https://assets.ubuntu.com/v1/20673d9f-lxd_primary--for-dark-bg.svg" alt="" width="95" />
```
bartaz, line 7: This is a place where adding alt makes sense, because there is no other text representation of each link.

But I'd say the alt should say "LXD", "Kubernetes", etc (without the 'icon') - same as a label on the link would.

bartaz (changes requested):
There seem to be a lot of unnecessarily added alt text in this PR.

Every image should have alt attribute, but empty `alt=""` are completely fine and recommended for images that are visual and don't provide any content value. And, arguably, many of our images are like that.

When we add the `alt` is should provide the meaningful description of the contents of the image (if it adds any value to the content), alt text such as "Screenshot" "icon" are meaningless and users don't get anything from screen reader reading those.

If the content is understandable and accessible with given image removed, it should have empty alt.

Most of images in vanilla and our examples are there for visual reasons only and they don't need any text in alt attribute.

bartaz:
Please make sure PR has a meaningful title, they end up in our release changelog.

templates/docs/examples/patterns/heading-icon/heading-icon-small.html
```diff
@@ -6,7 +6,7 @@
 {% block content %}
 <div class="p-heading-icon--small">
   <div class="p-heading-icon__header">
-    <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/425efe3a-lxd.svg" />
+    <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/425efe3a-lxd.svg" alt="LXD icon"/>
```
matthewpaulthomas, line 9: Example of both problems at once: it’s not representing the LXD icon, it’s representing LXD — and therefore `alt` text is redundant, because it’s immediately followed by the heading “LXD — system containers”.
```suggestion
    <img class="p-heading-icon__img" src="https://assets.ubuntu.com/v1/425efe3a-lxd.svg" alt=""/>
```

matthewpaulthomas (dismissed):
All the empty values you’ve added look correct.

Unfortunately, apart from the cases @bartaz flagged, all the non-empty values make one or both of these mistakes [described in the HTML spec](https://html.spec.whatwg.org/multipage/images.html#a-short-phrase-or-label-with-an-alternative-graphical-representation:-icons,-logos):
> If the logo is being used to represent the entity, e.g. as a page heading, the `alt` attribute must contain the name of the entity being represented by the logo. The `alt` attribute must _not_ contain text like the word "logo", as it is not the fact that it is a logo that is being conveyed, it's the entity itself.
>
> If the logo is being used next to the name of the entity that it represents, then the logo is supplemental, and its `alt` attribute must instead be empty.

lyubomir-popov:
@matthewpaulthomas good points, thanks, I'll update. I'm no accessibility expert so UX feedback on the accessibility prs is very welcome!

lyubomir-popov:
@matthewpaulthomas what about something like this:

image

The text explains do's and dont's, so the alts seem unnecessary. Is this correct?

templates/docs/examples/layouts/application.html
```diff
@@ -19,7 +19,7 @@
             <div class="p-inline-list--stretch">
               <div class="p-inline-list__item">
                 <a class="p-panel__logo" href="#">
-                  <img src="https://assets.ubuntu.com/v1/0b2b7bce-Vanilla-logo_88x24px_white.svg" alt="Vanilla framework logo" height="24">
```
bartaz, line 22: Shouldn't this one be "Vanilla framework"? It seems to be the same as 'home page logo' link in main nav.

bartaz (approved):

bartaz:
One comment to potentially address, but otherwise LGTM, thanks!

PR #3271, opened 2020-09-10
Add roles and correct tags to side navigation

lyubomir-popov:
## Done

Add landmark region roles to side navigation. 
Fixes https://github.com/canonical-web-and-design/vanilla-squad/issues/865

## QA

- Pull code
- Run `./run`
- Review use of roles in all side navigation examples
- [Add additional steps]

### DIFF
--- a/templates/docs/examples/layouts/documentation.html
+++ b/templates/docs/examples/layouts/documentation.html
@@ -4,7 +4,7 @@
 {% block content %}
 {% include "docs/examples/layouts/_global-nav.html" %}
 
-<header id="navigation" class="p-navigation is-dark">
+<header id="navigation" class="p-navigation is-dark" role="banner">
   <div class="p-navigation__row">
     <div class="p-navigation__banner">
       <div class="p-navigation__logo">
--- a/templates/docs/examples/layouts/documentation.html
+++ b/templates/docs/examples/layouts/documentation.html
@@ -4,7 +4,7 @@
 {% block content %}
 {% include "docs/examples/layouts/_global-nav.html" %}
 
-<header id="navigation" class="p-navigation is-dark" role="banner">
+<header id="navigation" class="p-navigation is-dark">
   <div class="p-navigation__row">
     <div class="p-navigation__banner">
       <div class="p-navigation__logo">
--- a/templates/docs/examples/patterns/side-navigation/_default.html
+++ b/templates/docs/examples/patterns/side-navigation/_default.html
@@ -1,4 +1,4 @@
-<div class="p-side-navigation {% if is_dark %}is-dark{% endif %}" id="drawer" role="banner">
+<div class="p-side-navigation {% if is_dark %}is-dark{% endif %}" id="drawer">
 
   <a href="#drawer" class="p-side-navigation__toggle js-drawer-toggle" aria-controls="drawer">
     Toggle side navigation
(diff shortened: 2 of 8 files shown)


templates/docs/examples/layouts/documentation.html
```diff
@@ -4,7 +4,7 @@
 {% block content %}
 {% include "docs/examples/layouts/_global-nav.html" %}
 
-<header id="navigation" class="p-navigation is-dark">
+<header id="navigation" class="p-navigation is-dark" role="banner">
```
bartaz, line 7: Docs say that `header` is equivalent of `banner`: "By default, the HTML5 <header> element has an identical meaning to the banner landmark, unless it is a descendant of <aside>, <article>, <main>, <nav>, or <section>.".

So maybe we can use that instead?

bartaz: Oh, it's a main navigation not side navigation... I got confused.

So it safe to drop banner role. It is a header already, and it will be top level element (not in any section, etc).

templates/docs/examples/patterns/side-navigation/_default.html
```diff
@@ -1,12 +1,12 @@
-<div class="p-side-navigation {% if is_dark %}is-dark{% endif %}" id="drawer">
+<div class="p-side-navigation {% if is_dark %}is-dark{% endif %}" id="drawer" role="banner">
```
bartaz, line 1: Why is side navigation pattern a banner?

"A banner role represents general and informative content frequently placed at the beginning of the page. This usually includes a logo, company name, search icon, photo related to the page, or slogan."

Side navigation doesn't include logo, company name, search, slogan. It's navigation, nothing more nothing less. I don't feel like "banner" applies here at all.

bartaz (changes requested):
I don't think that banner role applies to the side navigation component.

templates/docs/examples/patterns/side-navigation/_default.html
lyubomir-popov: I think it applie in the case of a sidenav that serves as the main nav, as in the application layout.
I find this [MDN definition](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Banner_role#:~:text=A%20banner%20landmark%20role%20converts,is%20applied%20into%20a%20header.&text=Assistive%20technologies%20can%20identify%20the,main%20%2C%20nav%20or%20section%20subsection.) more instructive. 

In the case of the docs sidenav it doesn't apply, indeed. I wil lremove it from there. Do you think we can keep it in the application sidenav example?

templates/docs/examples/patterns/side-navigation/_layout-application.html
```diff
@@ -1,6 +1,6 @@
 {# used by the application layout example #}
 
-<div class="p-side-navigation--icons {% if is_dark %}is-dark{% endif %}" id="drawer-icons">
+<nav class="p-side-navigation--icons {% if is_dark %}is-dark{% endif %}" id="drawer-icons">
```
bartaz, line 3: It feels a bit inconsistent that we use `<nav>` here on root element of the pattern, while on other cases on `.p-side-navigation__drawer`. Maybe we should do it always on the root element?

templates/docs/examples/patterns/side-navigation/_default.html
bartaz: In the application layout (full layout example) we used banner landmark (via `<header>`) on `l-navigation` - whole navigation region.
Side navigation component is just one part of this region (which also contains logo, etc).
So yes, banner landmark should be in application layout navigation region, but not on side navigation component (which by itself is not a banner).

bartaz: I think Ant was right here, that components should not have landmarks themselves. Landmarks should come from layout containers.

Possibly the only exception is the main navigation. Which by itself is kind of a layout type component and can have banner role.
Other patterns should not define landmarks.

bartaz (approved):
LGTM, thanks

PR #3295, opened 2020-09-22
Add aria-labels to tick elements, make ids in examples unique

lyubomir-popov:
## Done

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/3222

## QA

Inspect changes, verify the ids do not repeat and that all checkboxes/radios either:
- have an aria-labelled-by attr
have accompanying labels with `for`

### DIFF
--- a/templates/_layouts/examples.html
+++ b/templates/_layouts/examples.html
@@ -57,7 +57,7 @@
         if (!inIframe()) {
           document.addEventListener('DOMContentLoaded', function() {
             var body = document.body;
-            var controls = fragmentFromString('<div class="u-baseline-grid__toggle"><label>Toggle baseline grid<input type="checkbox" class="p-switch js-baseline-toggle" /><span class="p-switch__slider"></span></label></div>');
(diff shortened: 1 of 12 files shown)


bartaz (changes requested):
You have duplicated ids in same files (in many cases checkbox/radio "2" is duplicated for some reason).

More so, the files _checkobox-heading, _checkbox, _checkbox-inline etc are partials that are later merged together in single big comparison example, so their ids should be unique between them as well. Something like "checkboxHeading1" "checkboxInline1" etc would be needed.

Also, in some cases you are using `for` and `id` on an input, and in others `id` on span and `aria-labeledby` - why the inconsistency?

lyubomir-popov:
> Also, in some cases you are using `for` and `id` on an input, and in others `id` on span and `aria-labeledby` - why the inconsistency?

`for` not allowed on spans. So wherever the label value is in a span, we have to use `aria-labelledby`

bartaz (approved):
LGTM, thanks!

PR #3306, opened 2020-09-24
Add class names for states controlled by aria attributes

### LINKED ISSUE #1507: Style based on class and not only aria attr
Its great to style based on aria labels as this encourages accessibility but we should also use classes as some platforms (Angular 1) does not support conditional aria attributes.

bartaz:
## Done

Adds class names for states controlled by `aria-` attributes that can be used instead of `aria` for styling.

Fixes #1507

## QA

- Run `./run` or [demo](https://vanilla-framework-3306.demos.haus)
- Make sure affected patterns work as before:

### DIFF
--- a/scss/_patterns_accordion.scss
+++ b/scss/_patterns_accordion.scss
@@ -34,6 +34,8 @@
   }
 
   .p-accordion__tab {
+    @include vf-icon-plus($color-mid-dark);
+
     // calculate icon top position based on top padding (from %single-border-text-vpadding--scaling)
     // and offset to align it with baseline of accordion tab text element
     $icon-top-position: calc(#{$table-cell-vertical-padding} - 1px + 0.25rem);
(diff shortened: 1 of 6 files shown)


bartaz:

I am bit torn about this one. On one hand it would be nice to consistently provide class names for styling everything, but on the other having styles based on `aria` promotes using aria attributes and improves accessibility.

I guess the initial source of this issue (aria not being supported by Angular) is not really such pressing for us anymore.
Also we seem to use aria inconsistently (for example sometimes `aria-hidden=false` is default, another time `aria-hidden=true`) making it harder to convert those into consistent class names without (possibly breaking) changes to styling structure.

@anthonydillon Do you think we should still do it and deprecate aria in styles? If so, I assume we should provide consistent class names like `is-expanded`, `is-closed`, etc across the board, and not just patch whatever aria styling we have?

anthonydillon:
I think your right about the angular.js not being a real concern anymore. I would prefer to style based on aria personally. It enforces good practises.

lyubomir-popov:
I also think it would be nice to predominantly style based on aria where possible. Articles like [this one ](https://tink.uk/using-the-aria-current-attribute/) suggest the intention is to replace classes with aria-attribs, rather than append them to the existing class selectors.

bartaz:
OK, as we agree on this I'll close it for now.

PR #3316, opened 2020-09-28
Remove duplicate ids

lyubomir-popov:
## Done

Fixes https://github.com/canonical-web-and-design/vanilla-squad/issues/871

## QA

- Pull code
- Inspect changes, verify they don't break anything
- Bonus: use a better tool than regex to locate duplicate ids. [Apparently ](https://stackoverflow.com/questions/29681442/find-duplicated-id-keys-in-html-using-a-regex) regex is not ideal for parsing html, so another tool may reveal more duplicates.

### DIFF
--- a/templates/docs/examples/layouts/application.html
+++ b/templates/docs/examples/layouts/application.html
@@ -167,13 +167,13 @@
                 <td class="u-overflow--visible" data-test-column="summary">
                   <div class="u-flex">
                     <div class="u-flex--block p-tooltip--top-center" aria-describedby="tp-cntr">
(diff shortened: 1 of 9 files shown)


bartaz (approved):

bartaz:
LGTM, thanks.

This PR makes me wonder if we need all those ids. In some cases (labels for form fields) we do for sure, but many others are just copy pasted data from some our products and the ids there only clutter the HTML.

lyubomir-popov:
@bartaz probably not in many cases, but I'd rather limit the refactoring to only pass the validator. A lot of these pages are templates for design decisions, not really linked to from the main docs. We could maybe optimise unnecessary markup at a later stage.

PR #3317, opened 2020-09-28
Accordion: fix accessibility validator errors

lyubomir-popov:
## Done

Fixes the validator complaints on https://github.com/canonical-web-and-design/vanilla-framework/issues/3217

## QA

- Pull code
- Verify the changes look correct and do not break functionality.

### DIFF
--- a/templates/docs/examples/patterns/accordion/default.html
+++ b/templates/docs/examples/patterns/accordion/default.html
@@ -4,7 +4,7 @@
 {% block standalone_css %}patterns_accordion{% endblock %}
 
 {% block content %}
-<aside class="p-accordion" role="tablist" aria-multiselect="true">
+<aside class="p-accordion" role="tablist" >
   <ul class="p-accordion__list">
     <li class="p-accordion__group">
       <button type="button" class="p-accordion__tab" id="tab1" aria-controls="tab1-section" aria-expanded="true">Advanced topics</button>
--- a/templates/docs/examples/patterns/accordion/default.html
+++ b/templates/docs/examples/patterns/accordion/default.html
@@ -7,8 +7,8 @@
 <aside class="p-accordion" role="tablist" aria-multiselect="true">
   <ul class="p-accordion__list">
     <li class="p-accordion__group">
-      <button type="button" class="p-accordion__tab" id="tab1" role="tab" aria-controls="tab1-section" aria-expanded="true">Advanced topics</button>
-      <section class="p-accordion__panel" id="tab1-section" role="tabpanel" aria-hidden="false" aria-labelledby="tab1-section">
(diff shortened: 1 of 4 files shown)


templates/docs/examples/templates/maas-layout.html
```diff
@@ -95,23 +95,23 @@ <h1 class="p-heading--3">Machines</h1>
     <div class="col-3">
       <div class="">
         <h4 class="u-sv-1">Filter by</h4>
-        <aside class="p-accordion" role="tablist" aria-multiselect="true">
+        <aside class="p-accordion" role="tablist" >
           <ul class="p-accordion__list">
             <li class="p-accordion__group">
-              <button class="p-accordion__tab" id="owner-tab" role="tab" aria-controls="#owner" aria-expanded="false">Owner</button>
-              <section class="p-accordion__panel" id="owner" role="tabpanel" aria-hidden="true" aria-labelledby="owner-tab">
+              <button class="p-accordion__tab" id="owner" role="region" aria-controls="#owner" aria-expanded="false">Owner</button>
```
bartaz, line 101: Why does it have `role=region` here? It's the only accordion where this change was made.

bartaz, line 101: You changed the id (removed the `-tab`, and now the button has the same id as the section it triggers).
We should not change the ids, or if we want to remove the `tab` from id, we should make it `owner-trigger` or something like that.

templates/docs/examples/patterns/accordion/default.html
```diff
@@ -4,11 +4,11 @@
...
 
 {% block content %}
-<aside class="p-accordion" role="tablist" aria-multiselect="true">
+<aside class="p-accordion" role="tablist" >
   <ul class="p-accordion__list">
     <li class="p-accordion__group">
-      <button type="button" class="p-accordion__tab" id="tab1" role="tab" aria-controls="tab1-section" aria-expanded="true">Advanced topics</button>
-      <section class="p-accordion__panel" id="tab1-section" role="tabpanel" aria-hidden="false" aria-labelledby="tab1-section">
+      <button type="button" class="p-accordion__tab" id="tab1" aria-controls="tab1-section" aria-expanded="true">Advanced topics</button>
+      <section class="p-accordion__panel" id="tab1-section" aria-hidden="false" aria-labelledby="tab1-section">
```
bartaz, line 11: I just noticed that is seems that our `aria-labelledby` may be wrong on all accordion examples.
It seems to be pointing to itself, which doesn't make much sense. I think it was supposed to be pointing to the accordion header button (that is a 'label' of the section).

So it would be `aria-labelled-by="tab1"` in this case.

It's the same error in all accordion examples as far as I can see.

bartaz (commented):
LGTM, thanks!

bartaz (approved):

PR #3320, opened 2020-09-28
Fix tree component roles as per spec

lyubomir-popov:
## Done

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/3228

## QA

- Pull code
- Run `./run`
- Ensure tree example isn't broken
- Verify roles down the hierarchy are exactly as described in the [spec](https://www.w3.org/TR/wai-aria-practices/#tree_roles_states_props). There's an [example](https://www.w3.org/TR/wai-aria-practices/examples/treeview/treeview-1/treeview-1a.html) to compare to.


anthonydillon (approved):
LGTM :+1:

PR #3325, opened 2020-09-30
Applicaiton layout: Point aria-describedby to correct ids

lyubomir-popov:
## Done

aria-describedby on all tooltips in the application layout were pointing to the same id. This pr corrects that.

##QA

- Inspect changes, verify the aria-describedby points to the correct id for each tooltip.


bartaz (approved):
LGTM.

The tooltips don't seem to show (I think they are cut by overflow), but it's not a related issue, so we can have a look at it separately.

PR #3329, opened 2020-10-01
Add cookie policy to Vanilla site

pmahnke:
## Done

Add cookie policy
*Note: running locally, the site only loads one page and then the connection starts dropping?*

Fixes https://github.com/canonical-web-and-design/web-squad/issues/3265

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/
- See the new cookie popup and try the settings
- Test the "Manage your tracker settings" link from the footer


bartaz:
@pmahnke Why adding global nav?

pmahnke:
@bartaz re: global nav, I wasn't sure why it wasn't there, asked Ant, he said 'go for it'.  Personally I don't mind either way.

anthonydillon:
It's part of the canonical group and would bring it inline with our other sites. 

@bartaz to you have concerns with it being added?

bartaz:
> @bartaz to you have concerns with it being added?

@anthonydillon

Vanilla site never had it and I thought it was deliberate decision, as it's an open-source framework and not really a Canonical product.

So I don't mind - it indeed would make it consistent, but I wonder if it brings any value. Vanilla doesn't seem to be itself in a suite of Canonical products (it's not itself listed in global nav) and the audience of Vanilla framework and our products seems different. So I don't feel like anyone on Vanilla site would be looking for anything that is in global nav.
But surely it is a nice way to signify it under Canonical brand and promote other products.

My other concern is that it's simply another dependency to maintain, update, make sure it works, another piece of JS to download by the browser. And the way it injects itself with JS pushing the page down, another separate CSS (with separate Vanilla internal dependency)...
But it's also a benefit - because other projects are using it and it lets us make sure Vanilla works with global nav itself (we had issues before that we only discovered in products because we didn't have it).

I'm not completely against it, but because it's not a small thing I'd rather consider it separately from obligatory (and much smaller) change that is cookie policy.

anthonydillon:
Sure ok, let's leave it of for now. Sorry Peter. We can discuss it as a separate PR.

pmahnke:
@bartaz @anthonydillon global nav has been removed

templates/_layouts/_footer.html
```diff
@@ -13,6 +13,9 @@
           <li class="p-inline-list__item">
             <a class="p-link--external" href="https://github.com/canonical-web-and-design/vanilla-framework/issues/new/choose">Report a bug</a>
           </li>
+          <li class="p-inline-list__item">
+            <a href="" class="js-revoke-cookie-manager">Manage your tracker settings</a>
```
bartaz, line 17: Would be good not to leave href empty.

```suggestion
            <a href="#tracker-settings" class="js-revoke-cookie-manager">Manage your tracker settings</a>
```

anthonydillon: There is no where for this to default. If JS is switched off tracking won't work as they are injected via scripts.

anthonydillon (approved):
LGTM does the script need to be added to gitignore?

bartaz:
> does the script need to be added to gitignore?

@anthonydillon The script is copied into `build` folder which is ignored as a whole (both by git and by prettier)

PR #3330, opened 2020-10-02
Fix show utility to correctly set initial value

### LINKED ISSUE #3321: `u-show` utilities set display to `inherit` which causes rendering issues
`u-show` utilities override the display property to `inherit`, which is not always a correct value.
For example for spans nested in block elements it makes them block, etc.

The correct value should be `initial` (to set display what it was initially for the element).

bartaz:
## Done

Fix `u-show` utility to correctly set display to `initial` (but keep `inherit` for IE).

Fixes #3321

## QA

- Run `./run` or [demo](https://vanilla-framework-3330.demos.haus/docs/examples/utilities/show)
- Check the [show utility example](https://vanilla-framework-3330.demos.haus/docs/examples/utilities/show), make sure it correctly displays elements on relevant screen sizes, resetting inline elements back to inline state.

### DIFF
--- a/scss/_utilities_show.scss
+++ b/scss/_utilities_show.scss
@@ -3,23 +3,30 @@
 // Show elements with explicit breakpoints
 @mixin vf-u-show {
   .u-show {
+    // We keep inherit for IE that doesn't support `initial`.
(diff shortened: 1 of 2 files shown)


sowasred2012 (approved):
LGTM

PR #3332, opened 2020-10-02
Make sure cookie policy is correctly styled in docs

bartaz:
## Done

Vanilla docs pages were not using the same stylesheet as 'brochure' site, so cookie policy dialog was not styled properly on docs pages.
This PR fixes it by including site CSS on all pages.

## QA

- Run `./run` or [demo](https://vanilla-framework-3332.demos.haus/docs)
- Go to [docs](https://vanilla-framework-3332.demos.haus/docs),  cookie dialog should be correctly styled

### Before

Screenshot 2020-10-02 at 15 00 03

### After

Screenshot 2020-10-02 at 15 00 12

### DIFF
--- a/templates/_layouts/docs.html
+++ b/templates/_layouts/docs.html
@@ -1,7 +1,7 @@
 {% extends "_layouts/_root.html" %}
 
 {% block custom_head %}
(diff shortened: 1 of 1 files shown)


anthonydillon (approved):
LGTM :+1:

PR #3335, opened 2020-10-05
Change h3 font-weight to 100

lyubomir-popov:
## Done

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/3334

## QA

- Inspect files changed. Verify value 300 changed to 100.

Screenshot: 

image

### DIFF
--- a/scss/_patterns_pull-quotes.scss
+++ b/scss/_patterns_pull-quotes.scss
@@ -27,7 +27,9 @@
 
     // this needs to be nested to increase the specificity
     .p-pull-quote__citation {
-      font-style: italic;
+      @extend %vf-heading-6;
+
+      // font-style: italic;
       margin-top: $spv-inner--x-small--scaleable;
     }
   }
--- a/scss/_patterns_pull-quotes.scss
+++ b/scss/_patterns_pull-quotes.scss
@@ -29,7 +29,6 @@
     .p-pull-quote__citation {
       @extend %vf-heading-6;
 
-      // font-style: italic;
       margin-top: $spv-inner--x-small--scaleable;
     }
   }
@@ -76,10 +75,6 @@
         font-size: 2em;
       }
     }
-
-    .p-pull-quote__citation {
-      @extend %vf-heading-4;
-    }
   }
 
   .p-pull-quote--small {
@@ -92,10 +87,6 @@
       &::after {
         font-size: 1.5em;
       }
-    }
-
-    .p-pull-quote__citation {
-      @extend %default-text;
     }
   }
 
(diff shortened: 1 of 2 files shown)


spencerbygraves:
@lyubomir-popov the difference isn't as pronounced, but this seems to shift the problem down a size.

When h3 is placed above h4 after this change, I wonder if we would we get the same comment?

It doesn't block this PR but something to think about [next cycle](https://github.com/canonical-web-and-design/vanilla-framework/issues/3334).

image

lyubomir-popov:
@spencerbygraves I know, the only way to avoid this is to stop using the thin wait unfortunately. 

In order to ensure lower headings have thinner or same thickness contours, the ratio by which we scale them needs to be at least equal to the thickness of weight 300 divided by the thickness of weight 100. That is around 1.9, which would be huge to use as a type scale ratio. I've added more details in the attached [issue](https://github.com/canonical-web-and-design/vanilla-framework/issues/3334)

bartaz:
Not sure if we want to address this now, but the change affects large pull quotes:

Screenshot 2020-10-07 at 10 43 43

lyubomir-popov:
@bartaz changed to this:
image 

@spencerbygraves are you ok with this fix until we have the font weights conversation? 

I'm reducing the font-size of the bylines in quotes to be effectively `h6`'s.

scss/_patterns_pull-quotes.scss
```diff
@@ -27,7 +27,9 @@
 
     // this needs to be nested to increase the specificity
     .p-pull-quote__citation {
-      font-style: italic;
+      @extend %vf-heading-6;
+
+      // font-style: italic;
```
bartaz, line 32: If we don't need this, it should be removed.

scss/_patterns_pull-quotes.scss
```diff
@@ -27,7 +27,9 @@
 
     // this needs to be nested to increase the specificity
     .p-pull-quote__citation {
-      font-style: italic;
+      @extend %vf-heading-6;
```
bartaz, line 30: Do we wan't the same h6 size for all the variants?

This part of SCSS file is defining common shared styles of all quotes. Specific sizes for all variants are defined below:
line 79 defines size of default variant (as h4)
line 96 defines small variant as default text size
line 114 defines large variant as h3

I don't know why this shared definition of h6 wins in specificity and overrides all of the specific styles defined later.

If we really want the same size for all variants, we should remove individual size definitions below.

bartaz (changes requested):
Some changes needed around pull-quote styles

bartaz (approved):
LGTM, thanks

spencerbygraves:
>are you ok with this fix until we have the font weights conversation?

@lyubomir-popov I think that would mean the large pull quote has a smaller citation than the default one. It should probably be the same size, what do you think?

lyubomir-popov:
@spencerbygraves no they all have the same h6, I tagged you in percy on the image that shows them all.

PR #3346, opened 2020-10-08
Set muted heading font weight to 300

### LINKED ISSUE #3308: p-muted-heading needs to set font-weight
When applied to h1 or h2, which use `font-weight:100;` you get this:

image

bartaz:
## Done

Fixes #3308

## QA

- Pull code
- Run `./run`
- Make sure muted heading renders as expected

### DIFF
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -17,7 +17,7 @@
     cursor: pointer;
     display: inline-block;
     font-size: 1rem;
-    font-weight: 300;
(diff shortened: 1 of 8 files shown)


lyubomir-popov (approved):

anthonydillon (changes requested):
The reason to make this weight a variable and default is to have it editable by consumers. If that is what we want then we should document this variable.

bartaz:
Fair point @anthonydillon, thanks. Created as #3348

PR #3363, opened 2020-10-19
Update u-hide util for table columns and fix table expanding issues

### LINKED ISSUE #2726: Table expanding pattern is broken in IE
Table expanding doesn't render in full width in IE11 making it completely broken and unusable.

Screenshot 2020-01-10 at 09 12 13

### LINKED ISSUE #3337: Show/Hide utilities don't work well with table columns
**Describe the bug**

`u-hide` and `u-show` utilities that Vanilla provides work on `display` property.
While it successfully hides/shows table cells if applied it messes up the `colspan` functionality, because cells with `display: none` are not counted.

To hide columns (but still count them in `colspan`) we should hide them using `width: 0` instead.

**To Reproduce**

See codepen:
https://codepen.io/bartoszopka/pen/qBNWvqX

bartaz:
## Done

- Updates `u-hide` utility to work properly with table columns (we collapse columns to 0 width instead of hiding with display), with an exception of expanding table (which uses flex layout and needs to use display none for hiding).
- Updates expanding table to remove use of `u-hide` util and provide hiding using `aria-hidden`
- Fix IE rendering issues with expanding table

Fixes #3337
Fixes #2726

## QA

- `./run` or [demo](https://vanilla-framework-3363.demos.haus/docs)
- Check [new example for hiding table columns](https://vanilla-framework-3363.demos.haus/docs/examples/utilities/hide-table-columns)
  - check different screen widths, different browsers
- Check updated [example of expanding table](https://vanilla-framework-3363.demos.haus/docs/examples/patterns/tables/table-expanding)
  - Make sure to check it in IE
- Check updated docs for [expanding table](https://vanilla-framework-3363.demos.haus/docs/base/tables#expanding)

### DIFF
--- a/scss/_patterns_table-expanding.scss
+++ b/scss/_patterns_table-expanding.scss
@@ -6,7 +6,9 @@
     flex-flow: column nowrap;
     justify-content: space-between;
 
+    thead,
     tbody {
+      display: block; // for IE we need to force display block
       margin: 0;
     }
 
@@ -50,6 +52,10 @@
           padding: 0;
           width: 100%;
         }
+      }
+
+      &[aria-hidden='true'] {
+        display: none;
       }
     }
     // stylelint-enable selector-no-qualifying-type
--- a/scss/_utilities_hide.scss
+++ b/scss/_utilities_hide.scss
@@ -1,22 +1,81 @@
 @import 'settings';
 
+// mixin for hiding table columns
+@mixin vf-hide-table-column {
+  display: table-cell !important; // override display: none from standard u-hide
+  opacity: 0 !important;
+  overflow: hidden !important; // make sure content doesn't overflow
+  padding: 0 !important;
+  white-space: nowrap !important; // prevent content from wrapping in 0 width column
+  width: 0 !important;
+}
+
 // Hide elements with explicit breakpoints
 @mixin vf-u-hide {
   .u-hide-text {
     @extend %vf-hide-text;
   }
 
+  // mixin for hiding using display none (for all breakpoints)
   .u-hide {
     display: none !important;
 
     &--small {
-      @media screen and (max-width: $breakpoint-medium) {
+      @media screen and (max-width: $breakpoint-medium - 1) {
         display: none !important;
       }
     }
 
     &--medium {
-      @media (min-width: $breakpoint-medium) and (max-width: $breakpoint-large) {
+      @media (min-width: $breakpoint-medium) and (max-width: $breakpoint-large - 1) {
+        display: none !important;
+      }
+    }
+
+    &--large {
+      @media screen and (min-width: $breakpoint-large) {
+        display: none !important;
+      }
+    }
+  }
+
+  // stylelint-disable selector-no-qualifying-type
+  td.u-hide,
+  th.u-hide {
+    @include vf-hide-table-column;
+
+    &--small {
+      @media screen and (max-width: $breakpoint-medium - 1) {
+        @include vf-hide-table-column;
+      }
+    }
+
+    &--medium {
+      @media (min-width: $breakpoint-medium) and (max-width: $breakpoint-large - 1) {
+        @include vf-hide-table-column;
+      }
+    }
+
+    &--large {
+      @media screen and (min-width: $breakpoint-large) {
+        @include vf-hide-table-column;
+      }
+    }
+  }
+  // stylelint-enable selector-no-qualifying-type
+
+  // expanding table uses flex layout, so use standard u-hide for it
+  .p-table-expanding .u-hide {
+    display: none !important;
+
+    &--small {
+      @media screen and (max-width: $breakpoint-medium - 1) {
+        display: none !important;
+      }
+    }
+
+    &--medium {
+      @media (min-width: $breakpoint-medium) and (max-width: $breakpoint-large - 1) {
         display: none !important;
       }
     }
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "clean": "rm -rf build docs/static/css node_modules/ yarn-error.log",
     "percy": "percy exec -- node snapshots.js"
   },
-  "version": "2.19.1",
+  "version": "2.19.2",
   "files": [
     "/scss",
     "!/scss/docs"
--- a/templates/docs/base/tables.md
+++ b/templates/docs/base/tables.md
@@ -39,11 +39,13 @@ View example of the table sortable pattern
 
 ### Expanding
 
+<span class="p-label--updated">Updated</span>
+
 Using `.p-table-expanding` in conjunction with the `<table>` element will allow expanding and hidden table cells which take up the full width of the table row element.
 
 This pattern should be used when a table requires configuration fields (add, edit). Expandable rows can also be used to supply additional information not visible on the table row.
 
-Using `p-table-expanding__panel` it can be hidden using the `aria-hidden` attribute. The table must contain all table cells required.
+Using `p-table-expanding__panel` it can be hidden using the `aria-hidden` attribute. The table must contain all table cells required. The expanding panel is implemented as additional cell in each row, so to keep the markup of the table valid, an additional cell is also needed in the table `<thead>`. This placeholder heading cell should be hidden using `aria-hidden="true"` attribute.
 
 <div class="embedded-example"><a href="/docs/examples/patterns/tables/table-expanding/" class="js-example">
 View example of the expanding table pattern
--- a/templates/docs/component-status.md
+++ b/templates/docs/component-status.md
@@ -12,7 +12,7 @@ When we add, make significant updates, or deprecate a component we update their 
 
 ### What's new in Vanilla 2.19
 
-<table aria-label="What's new in Vanilla 2.18">
+<table aria-label="What's new in Vanilla 2.19">
   <thead>
     <tr>
       <th style="width: 20%">Component</th>
@@ -23,6 +23,12 @@ When we add, make significant updates, or deprecate a component we update their 
   </thead>
   <tbody>
     <!-- 2.19 -->
+    <tr>
+      <th><a href="/docs/base/tables#expanding">Expanding table column placeholder</a></th>
+      <td><div class="p-label--updated">Updated</div></td>
+      <td>2.20.0</td>
+      <td>We started using <code>aria-hidden="true"</code> attribute to hide the dummy table header in place of previously used <code>u-hide</code> utility.</td>
+    </tr>
     <tr>
       <th><a href="/docs/base/tables#icons">Table icons</a></th>
       <td><div class="p-label--new">New</div></td>
--- a/templates/docs/examples/patterns/tables/table-expanding.html
+++ b/templates/docs/examples/patterns/tables/table-expanding.html
@@ -13,8 +13,8 @@
             <th id="t-units1" aria-sort="none">Rack</th>
             <th id="t-units2" aria-sort="none">Last seen</th>
             <th id="t-revenue" aria-sort="none" class="u-align--right">Actions</th>
-            <th class="u-hide">
-                <!-- empty cell required for validation -->
+            <th aria-hidden="true">
+                <!-- hidden empty cell required for validation -->
             </th>
         </tr>
     </thead>
--- a/templates/docs/examples/utilities/hide-table-columns.html
+++ b/templates/docs/examples/utilities/hide-table-columns.html
@@ -1,0 +1,28 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Hide / Table columns{% endblock %}
+
+{% block content %}
+<table class="p-table">
+    <thead>
+        <tr>
+            <th>Always visible</th>
+            <th class="u-hide--small">Hidden on small screens</th>
+            <th class="u-hide--medium">Hidden on medium screens</th>
+            <th class="u-hide--large">Hidden on large screens</th>
+            <th class="u-hide">Always hidden</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td>Always visible</td>
+            <td class="u-hide--small">Hidden on small screens</td>
+            <td class="u-hide--medium">Hidden on medium screens</td>
+            <td class="u-hide--large">Hidden on large screens</td>
+            <td class="u-hide">Always hidden</td>
+        </tr>
+        <tr>
+            <td colspan="5">Cells with <code>colspan</code> should work correctly regardless of hidden columns</th>
+        </tr>
+    </tbody>
+</table>
+{% endblock %}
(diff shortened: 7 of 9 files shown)


anthonydillon:
I am trying and failing to think of why this would be a bad idea:
> scoping existing u-hide utils to do something different in tables is probably a bad idea

The con is that user have to learn a new class which does what you would expect the visible utility classes should do. Not to mention, the changes to existing codebases and the impact on the version.

We adapted the alignment utilities to work with both block and flex elements. I know this is an abstraction of the utility in the table context. I can not foresee any issues with this approach.

bartaz:
@anthonydillon The issue is that utilities ideally (even more than any other class) should do as little as possible and not depend on the context. If utility such as `u-hide` does `display: none` in 99% but something completely different inside the table may simply cause unexpected behaviour.

And it already did in case of expanding table (that for some reason uses `u-hide` as required part of the pattern).

At the very beginning I did just that - make sure `u-hide` when applied to th/td uses `width: 0` instead of display. It seemed to work fine, was released with modular table in react-components. But apparently it broke expaning table (that expects u-hide to do display: none).

So it simply feels to be safer to introduce new class that will do a specific thing rather that add contextual difference inside `u-hide` (that apparently in some contexts fails). And avoid a breaking change in expanding table.
Because otherwise (to avoid breaking change) we would need to 1) introduce an exception to make `u-hide` work differently in tables and 2) add exception to the exception to make sure expanding table does use `display: none` version of `u-hide`.

anthonydillon:
I agree its clearer and I am normally a fan of that but its a shame to have to provide new classes but I understand. Thanks for stepping through it @bartaz

bartaz:
@anthonydillon Out of curiosity I tied how it would work to add an exception for expanding table and it doesn't turn out terrible (thx to using `:not`):

```scss
  // stylelint-disable selector-no-qualifying-type
  table:not(.p-table-expanding) td.u-hide,
  table:not(.p-table-expanding) th.u-hide {
    display: table-cell !important; // override u-hide

    // hide table cell using width
    opacity: 0 !important;
    overflow: hidden !important; // make sure content doesn't overflow
    padding: 0 !important;
    white-space: nowrap; // prevent content from wrapping in 0 width column
    width: 0 !important;
  }
  // stylelint-enable selector-no-qualifying-type
```

This ^ times 4 (to cover all breakpoint variants).

Obviously this kind of selector (`table:not(.p-table-expanding) td.u-hide`) is much more fragile that `u-hide-table-column`, but it does keep the complexity in Vanilla rather that on developer. And is not as complicated as I initially thought it would be.

So I'm fine to do that to avoid adding new class if you also agree. We can always introduce new one if that fails to work in some use case again.

scss/_utilities_hide.scss
```diff
@@ -1,5 +1,14 @@
 @import 'settings';
 
+// mixin for hiding table columns
+@mixin vf-u-hide-table-column {
```
anthonydillon, line 4: Is this better as part of the hide utilities or a table pattern utility? This is scoped to table and so would expect to be able to use it if I import the table pattern.

scss/_patterns_table-expanding.scss
```diff
@@ -53,5 +55,9 @@
       }
     }
     // stylelint-enable selector-no-qualifying-type
+
+    .p-table-expanding__column-placeholder {
```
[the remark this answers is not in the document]
anthonydillon: I thought the reason for this change was to remove the dependency of hide utilities on the table pattern. As this is not the case due to the `u-hide-table-column` class being part of the hide utilities. Why not continue to use `u-hide` here for `display: none;`?

anthonydillon (changes requested):
Few questions in the codebase.

scss/_patterns_table-expanding.scss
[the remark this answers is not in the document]
bartaz: This is a weird one. The expanding table is built in a way that requires one additional cell in every row (the hidden 'expandable' row). But, because it's additional `<td>` in a row, it needs to be accompanied by corresponding cell in header row for validity. So in the markup we have a `<td>` that is empty, not needed for the pattern itself (just for validity of markup) and has to be hidden.
Currently expanding table uses `u-hide` utility for it. But I don't think a pattern should depend on utility to provide a styling for its integral part. That's why I thought it would be better to have a pattern specific class name for this column placeholder (hence the name).

Personally I don't think the way expanding table was built is a good one from semantic and styling point of view, but it's not a place to change it completely. We can at least improve it to remove dependency on utility (that may or may not be available).

scss/_utilities_hide.scss
bartaz: True. I thought of it as modification of `u-hide` pattern (especially that it shares the `--small`, etc variants), so I kept it here.
It would especially make sense if we decide to do it as `u-hide` scoped to the table in the end.

If we keep it as new utility it may make sense to separate it, although, I don't think we have any other example of utility bound with a pattern, so I'd probably just separate it to different file.

scss/_patterns_table-expanding.scss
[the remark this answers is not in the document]
bartaz: As for the name, I don't know how to best name it. Is it an 'expanding column'? It doesn't seem to be a column at all - it may be in markup, but in the table we don't want it as column (we don't want it in header, and in rows we expand it below the row). So it's this weird implementation artifact that we want hidden from the view. I thought about just hiding the last cell in header row `.p-table-expanding thead td:last-child` but this feels quite fragile and much less explicit.

The benefit of `u-hide` here is that it's explicit what it does. But makes pattern dependent on a utility.
Maybe `p-table-expanding__hidden-cell` would be best of both words?

scss/_utilities_hide.scss
anthonydillon: If a utility is bound so tightly to a pattern then it might not be a utility at all but a pattern element. This might be much nicer as a state class on table cells. For example: `is-hidden--small`?

anthonydillon: This would be much easier if we used classes on each element but tables are difficult in this case.

anthonydillon: Or even doubling down on a11y and targeting just
```css
&[aria-hidden='true'] {
  ...
}
```

scss/_patterns_table-expanding.scss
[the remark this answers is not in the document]
anthonydillon: We could target `&[aria-hidden='true'] {` and document that is how to hide the final header cell.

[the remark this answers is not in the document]
bartaz: Interesting idea. I like that better than inventing new class names, so will do that.

scss/_utilities_hide.scss
bartaz: We can make `aria-hidden` replace `u-hide-table-column` (or even `u-hide`), but this will not work with breakpoints. We still need a classname to put `--small`, etc variants on.

Using table specific styling is indeed tempting as these are table specific styles. The semantic problem is that we don't really have "table pattern" (`p-table`), so we don't have a class names to modify.
This leads us to modifier classes like `is-hidden--small` you suggested. But this class name by itself doesn't tell it's table specific. It could be easily mistaken for `u-hide--small`. So technically we should do it as `p-table__cell--hidden--small` or something along that lines - which get's weirdly long and specific, but on the other hand does exactly what it says, where it says.

bartaz:
Thanks for suggestions @anthonydillon.

PR updated to use `aria-hidden` in the expanding table, and table specific utility is now a table specific class `p-table__cell--hidden`.

anthonydillon:
Just saw this? Would this work? https://css-tricks.com/almanac/properties/e/empty-cells/

scss/_patterns_table-hidden.scss
```diff
@@ -0,0 +1,34 @@
...
+  padding: 0;
+  white-space: nowrap; // prevent content from wrapping in 0 width column
+  width: 0;
+}
+
+@mixin vf-p-table-hidden {
+  .p-table__cell--hidden {
+    @include vf-u-hide-table-column;
+
+    &--small {
```
anthonydillon, line 16: I don't think we want to start double modifying classes could this be `p-table__cell-hidden--small`?

anthonydillon (changes requested):
Small comment in the code.

scss/_patterns_table-hidden.scss
bartaz: @anthonydillon True, I don't like double modifiers either, but `hidden` is modifier on `cell` element, so `p-table__cell--hidden-small`?

bartaz:
> Just saw this? Would this work? https://css-tricks.com/almanac/properties/e/empty-cells/

This seems to be about rendering borders/backgrounds of empty cells, not about hiding cells (possibly not empty) from view.

scss/_patterns_table-hidden.scss
anthonydillon: Doesn't feel consistent with the rest of the framework. Considering the options:

| Interface | Usage |
| --- | --- |
| Table pattern element | `p-table-cell--hidden-small` |
| Table pattern utility | `u-hide-cell--small` |
| Extend the existing utility | `u-hide--small` |
| State class | `is-hidden--small` |

Please add our suggest and options that are not here. Looking at the usage table I believe extending the existing utility pattern feels like the most elegant approach.
 
**Foot note**
A part for the fact a dislike this approach of responsive table management. I believe the same content and functionally should be available on all screen sizes and devices. I have lost this battle a number of times.

bartaz: Sure, I agree with consistency, and consistency is why I started with `--hidden--small` (to keep `--small` as a modifier itself).

While we currently don't have `p-table` I believe in long term we will start to need that for more complex tables (modular table component). So I think we can start assuming that base table styling (HTML only) is for basic static tables, but any table functionality (icons, hiding, etc) would require `p-table` family of classes.

With that assumption to add icons to cells we added `p-table__cell--icon-placeholder`. `p-table` pattern, `__cell` child element `--icon-placeholder` modifier. 
Similarly, for consistency we would hide the cells using `p-table__cell--hidden`.  But the problem starts with modifiers of this hidden modifier, should they be `p-table__cell--hidden--small` (double modifier)? or `p-table__cell--hidden-small` (part of the same modifier, but inconsistent with rest of the framework).

In other places where we have multiple modifiers on same element we tend to use `is-something` kinds of class names. It's not BEM, and I'm not a huge fan of those, but they serve their purpose.
I don't like the idea of having just `is-hidden--small` (and scoping it to table HTML elements like td), but if we treat table as `p-table` pattern, and cell as `p-table__cell` we could have `p-table__cell.is-hidden--small`. So `is-hidden--small` state class, but only on a pattern element `p-table__cell`.

| Interface | Usage |
| --- | --- |
| State class on table pattern element | `p-table__cell.is-hidden--small` |

bartaz: @anthonydillon thoughts? ^

anthonydillon: Yeah I agree feels strange to create a new global sounding state class but it is consistent with the rest of the framework. What about using a different adverb. Maybe something like: `p-table__cell.is-removed--small` which would make it clearer and remove the confusion between this class and the `u-hide--small` class.

I keep coming back to the fact people familiar with Vanilla are likely to use `u-hide--small` before having to check why that doesn't work as expected. I wonder if we shouldn't just set `.p-table .u-hide` in the table pattern scss file. Then we don't need to document anything particularly differently.

bartaz: Thanks @anthonydillon, there is a lot of back and forth here, but valuable feedback to get us to good compromise.

> Yeah I agree feels strange to create a new global sounding state class but it is consistent with the rest of the framework. What about using a different adverb. Maybe something like: `p-table__cell.is-removed--small` which would make it clearer and remove the confusion between this class and the `u-hide--small` class.

We could try with different verb, but `removed` sounds more permanent than `hidden` which seems to be more accurate in case if responsively hiding columns. I thought about `collapsed` (which would be more in line with what the class actually does - collapses the column to 0 width without removing it) but this on the other hand sounds like something that can also be 'expanded'.

> I keep coming back to the fact people familiar with Vanilla are likely to use `u-hide--small` before having to check why that doesn't work as expected. I wonder if we shouldn't just set `.p-table .u-hide` in the table pattern scss file. Then we don't need to document anything particularly differently.

I agree that reusing `u-hide` and redefining it for tables is the least invasive approach. It doesn't introduce new class names, removes from developer the need to understand different concepts and keeps the complexity within the framework. But, this does mean there is a bit of complexity added.

First of all, it can't be just `.p-table .u-hide` - these new styles are specifically for table columns, not anything in the table, so we need `.p-table td.u-hide, .p-table th.u-hide`, and I guess people would expect it working consistently across the tables, not just with this `p-table` class, so why not `table td.u-hide, .table th.u-hide`, but this will again break in the expanding table which we want excluded from this...

The other thing is that it creates a dependency between a pattern and utility. And the question is where do we maintain this dependency. Should pattern define its exceptions to how utilities work, or should utility define how it works in different contexts? Either side is not perfect.
The other aspect of this is overriding and specificity. Utilities are built to override any other styling, they use !important all over the place. So it's even harder to override them - and that's what we would need to do to have `.p-table u-hide`. Sure, higher specificity and another bunch of !importants should do the trick, but I'm not sure if that should be our goto approach.

At this point, I am leaning towards implementing some variant of `.p-table th.u-hide, .p-table td.u-hide`, because, as you mentioned, it doesn't introduce new artifacts and we maybe don't need to document anything new (although `.p-table` would be kinda new, as it's not mentioned anywhere yet, because was never directly used). And this fix is also needed to unblock react-components for MAAS so would be good to release something soon (or we need to put a workaround in react-components).

anthonydillon: Yes, I like that solution with a slight twist. `th.u-hide, td.u-hide` which divorces the cross pattern/utility dependency. 

Regarding where it resides I believe we should add this variant to `_utilities_hide.scss`. The utility code has to live up to the promise that it will "hide" things as the user expects in all cases. Therefore, it feels right from a maintenance and responsibility point of view for this code to live there. 
 - **Maintenance**: because if there was an issue with `u-hide` not doing what I expected in any situation the first place I would go would be `_utilities_hide.scss`. It can also help us to rationale overrides but using `:not()` on the main utility and providing an extension very near each other in the code.
- **Responsibility**: I think of a utility as a promise to the user and therefore should handle any complexities that, that promise entails.

bartaz: @anthonydillon `th.u-hide, td.u-hide` is what I did in react components first, and this resulted in a broken expanding table. The problem with utility that addresses ALL tables is that we need to EXCLUDE p-table-expanding, because it uses flex layout instead of table layout and needs to use standard (display: none) u-hide.

If we say that `u-hide` should be responsible for how it works it means we need to do `td.u-hide, th.u-hide` override for tables but create another override for .p-expanding table. So (in simplified form):

```scss
.u-hide {
   display: none !important;
}

td.u-hide, th.u-hide {
   display: table-cell !important;
   width: 0 !important;
   // etc...
}

.p-table-expanding td.u-hide, .p-table-expanding th.u-hide {
   display: none !important;
}
```

anthonydillon: That seems reasonable as a temporary fix until 3.0 when we require the user to switch from using `u-hide` to using `aria-hidden='true'` in the expanding pattern.

anthonydillon (approved):
This is awesome, thanks @bartaz

Issue #3366, opened 2020-10-19
Icon color usage improvements


hatched:
The documentation for the white icons indicates that you must place them in a `.p-strip--dark` which adds a lot of unnecessary other styles.

The actual style in question only requires that its parent have the suffix `--dark` for it to switch to the white icon.

While this is a simple documentation update, I feel like the long term fix to this would be to have icons be the masters of their own domain and have each icon switch to a light or dark based on it's own class `p-icon--share__dark` for example. It's also a bit odd that the colour is respective to the background it's placed on, not the icon so `p-icon--share__light` might be more appropriate for the white icon. and `p-icon--share__dark` for the black one.

PR #3367, opened 2020-10-19
Clarify the usage of the dark-themed icons.

hatched:
## Done

Update the documentation around using the dark-themed icons.

Related #3366

### DIFF
--- a/templates/docs/patterns/icons.md
+++ b/templates/docs/patterns/icons.md
@@ -109,9 +109,7 @@ Our icons have two predefined color styles: light and dark. The light variant is
 
 ### Dark theme
 
-Our dark-themed icons are available when placed within `.p-strip--dark`, icon colors are inverted to ensure legibility as shown in our example.
-
-Alternatively to use light version of icons outside of the dark strip, add `is-light` class to the icon element (e.g. `p-icon--plus is-light`).
+Our dark-themed icons are available when placed within a container that has the suffix `--dark` ex) `help-container--dark`. Icon colors will then be inverted to ensure legibility as shown in our example.
 
 <section>
 <div class="p-strip--dark is-shallow u-no-padding--top" style="background-color: transparent;">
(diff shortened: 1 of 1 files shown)


anthonydillon (commented):
What was the reason we wildcarded the `--dark` parent selector? I thought `p-strip--dark` and `is-dark` are the only parent selectors that should auto trigger the light icons?

lyubomir-popov (approved):
LGTM. Technically the selector will be matched regardless of whether `--dark` is a suffix. Wonder if we should explain that precisely - to reassure people that if they have to chain modifiers for whatever reason (e.g. `--dark-small`) it would still work.

bartaz:
> What was the reason we wildcarded the `--dark` parent selector? I thought `p-strip--dark` and `is-dark` are the only parent selectors that should auto trigger the light icons?

I looked at history, and I don't know. It's 3 years old and basically first time we started having 'dark' version of icons it was using `--dark` already instead of `.p-strip--dark`. Having a quick search strips are the only ones that use `--dark` naming convention (because it predates our themed `is-dark` versions of other components).

While we technically use `--dark` I'm not sure if we should document and encourage usage of it. It's an exception that should not be used elsewhere and promoted. I would even say that it's arguable if the icons should even take external class names into account. While it makes things 'simpler' if icons are thrown directly into darks strips, it can easily lead to issues when dark parent container will be further up the document tree and actually a light version of icon will be expected.
In a long term (3.0, later?) I think we should remove icons being styled by strips and allow icon classnames to affect icon look.

anthonydillon:
I agree, it would be good to remove this complexity and simply provide `p-icon--anchor is-inverted` which with switch to dark if using the light theme and vise versa.

bartaz:
> I agree, it would be good to remove this complexity and simply provide `p-icon--anchor is-inverted` which with switch to dark if using the light theme and vise versa.

`is-inverted` would still make icons depend on theme of external component. For clarity and simplicity components should define their own theme, including icons.
But, that's not something we can change now without breaking changes, so we can think about it and decide closer to 3.0 release.

PR #3370, opened 2020-10-20
Unquote fontfamily so the $font-monospace can be used directly

### LINKED ISSUE #3369: $font-monospace returns a non-functional quoted value
When using `font-family: $font-monospace;` as outlined here https://vanillaframework.io/docs/settings/font-settings the resulting CSS is `font-family: '"Ubuntu Mono", Consolas, Monaco, Courier, monospace'` which is incorrect formatting. To fix this I had to use `font-family: unquote($font-monospace);` which isn't documented.

I'd have expected `font-family: $font-monospace` to work out of the box.

hatched:
## Done

Unquoted the values of `$font-base-family` and `$font-monospace` so that their values can be used directly without `unquote()`.

Fixes #3369 

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/ or [demo]()
- Make sure all the fonts are rendered appropriately.

### DIFF
--- a/scss/_base_fontfaces.scss
+++ b/scss/_base_fontfaces.scss
@@ -1,5 +1,5 @@
 @mixin vf-b-typography-fontfaces {
-  @if str-index(quote($font-base-family), 'Ubuntu') {
+  @if str-index($font-base-family + '', 'Ubuntu') {
     @if $font-use-subset-latin {
       @font-face {
         font-display: $font-display-option;
--- a/scss/_base_fontfaces.scss
+++ b/scss/_base_fontfaces.scss
@@ -1,5 +1,5 @@
 @mixin vf-b-typography-fontfaces {
-  @if str-index($font-base-family, 'Ubuntu') {
+  @if str-index(quote($font-base-family), 'Ubuntu') {
     @if $font-use-subset-latin {
       @font-face {
         font-display: $font-display-option;
--- a/scss/_settings_font.scss
+++ b/scss/_settings_font.scss
@@ -1,6 +1,6 @@
 // Global font settings
-$font-base-family: '"Ubuntu", -apple-system, "Segoe UI", "Roboto", "Oxygen", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif' !default;
-$font-monospace: '"Ubuntu Mono", Consolas, Monaco, Courier, monospace' !default;
+$font-base-family: 'Ubuntu', -apple-system, 'Segoe UI', 'Roboto', 'Oxygen', 'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif !default;
+$font-monospace: 'Ubuntu Mono', Consolas, Monaco, Courier, monospace !default;
 $font-heading-family: $font-base-family !default;
 $font-use-subset-latin: false !default;
 $font-display-option: fallback !default;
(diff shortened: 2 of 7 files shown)


scss/_base_fontfaces.scss
```diff
@@ -1,5 +1,5 @@
 @mixin vf-b-typography-fontfaces {
-  @if str-index($font-base-family, 'Ubuntu') {
+  @if str-index(quote($font-base-family), 'Ubuntu') {
```
barrymcgee, line 2: Shouldn't this be unquote?

scss/_settings_font.scss
```diff
@@ -1,6 +1,6 @@
 // Global font settings
-$font-base-family: '"Ubuntu", -apple-system, "Segoe UI", "Roboto", "Oxygen", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif' !default;
-$font-monospace: '"Ubuntu Mono", Consolas, Monaco, Courier, monospace' !default;
+$font-base-family: 'Ubuntu', -apple-system, 'Segoe UI', 'Roboto', 'Oxygen', 'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif !default;
+$font-monospace: 'Ubuntu Mono', consolas, monaco, courier, monospace !default;
```
lyubomir-popov, line 3: property values are case-insensitive, but if we're changing to lowercase maybe we should change all? Or leave all uppercase as this change isn't really needed.

bartaz: Linter was complaining about upper case values, so that's why the change was made.
I guess uppercase is fine in case of names in quotes ('Ubuntu Mono', 'Roboto'), so we can put them all in quotes (uppercase) or remove quotes from single word ones and lowercase them too, still, we will need quotes around 'Ubuntu Mono' etc - not sure if they also are case insensitive? I guess they should...

hatched:
Does anyone have any context as to why the Dart build is failing?

lyubomir-popov:
> Does anyone have any context as to why the Dart build is failing?

Seems it doesn't recognize this as a string for some reason?
image

hatched:
Sorry that's what I meant, why the dart build thinks that it's not a string.

bartaz:
@hatched Apparently the fact that it "works" in libsass (that we use) is a bug, and dart implementation is correct:
https://github.com/sass/sass/issues/2597

In one of the comments a workaround is suggested to use `$variable + ''` to force value into a string (which is a nice nod to all of us JS devs 😉 ), but I haven't checked if it would work in our case, and would feel like a bit of codesmell.

bartaz:
There are couple places (jp.ubuntu.com, ubuntu.com, maas ui) that use font settings directly and will be affected by the change.
If we want to implement this breaking change we need to make sure with the release we properly update the affected projects that define their own (currently quoted) font stacks:
https://github.com/search?q=org%3Acanonical-web-and-design+%24font-base-family&type=code
https://github.com/search?q=org%3Acanonical-web-and-design+%24font-monospace&type=code

scss/_base_fontfaces.scss
```diff
@@ -1,5 +1,5 @@
 @mixin vf-b-typography-fontfaces {
-  @if str-index($font-base-family, 'Ubuntu') {
+  @if str-index($font-base-family + '', 'Ubuntu') {
```
bartaz, line 2: This seems to do the trick (casting font stack list to string), but I'm not sure if I like that. Seems like a hack.

bartaz:
We discussed it on Vanilla sync meeting (https://github.com/canonical-web-and-design/meeting-notes/issues/677). Changing the way the font is defined seems bit risky at this point (it's been like this since beginning of Vanilla and may affect sites that redefine fonts).

More importantly using the font stack variable directly is not really recommended (as it only uses font definitions, but not anything else (weights, line-height, etc). It would be best to have a 'monospace text' SASS placeholder that can be extended for consistent use of monospace text.

We don't have it currently unfortunately, but created an issue for that:
https://github.com/canonical-web-and-design/vanilla-framework/issues/3417

PR #3379, opened 2020-10-30
WIP: POC side navigation for apps improvements

bartaz:
For demo purposes.

- improved icon spacing (to center icons when side nav is collapsed)
  - currently spacing is the same as grid margin (changes with breakpoints), this doesn't work with collapsed view
  - we need to hardcode specific values to make icons centered - should this be just for apps, could it be for all side navigations?
- expanding on focus
- fading secondary levels when collapsed
  - fading makes sure expanded levels are not visible in collapsed view, but this requires a lot of assumptions and hardcoding - may be hard (or impossible) to build a flexible solution that fits all possible uses

### DIFF
--- a/scss/_layouts_application.scss
+++ b/scss/_layouts_application.scss
@@ -42,7 +42,7 @@
     }
   }
 
-  $layout-navigation-collapsed-width: 3rem;
+  $layout-navigation-collapsed-width: 4rem;
   $layout-navigation-expanded-width: 15rem;
 
   .l-navigation__drawer {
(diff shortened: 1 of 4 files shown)


bartaz:

I'm not really sold on 'make the collapsed sidebar wider so it fits bigger padding'. It really seems like there is too much dead space here in a case, where we collapse sidebar to save space on the screen.

Screenshot 2020-11-04 at 15 05 13

How about shifting the padding a bit when collapsing? So expanded sidebar has standard wide padding of side navigation pattern, but when it collapses it's moved a bit to the side, so the icons are centered. And it animates back when expanding?

expand-padding

PR #3283, opened 2020-09-18
Accessibility fix: Remove alt attribs from all buttons

lyubomir-popov:
## Done

Fixes https://github.com/canonical-web-and-design/vanilla-squad/issues/872

## QA

- Pull code
- Inspect list of changes, verify all alt attributes on buttons have been removed

PR #3382, opened 2020-11-02
WIP: Updates to application layout panels responsiveness

bartaz:
## Done

- make sure aside panel gets unpinned on small screens
- make sure aside panel doesn't exceed small screen width
- provide 3 width options for panels (is-narrow, is-medium, is-wide)

Fixes #3283

## QA

- Run `./run` or [demo](https://vanilla-framework-3382.demos.haus/docs/examples/layouts/application)
- Check application layout panels example https://vanilla-framework-3382.demos.haus/docs/examples/layouts/application
  - make sure aside panel works well responsively (unpins on small screens, is never wider than the screen)
  - check different built-in widths of the panel

### DIFF
--- a/scss/_layouts_application-panels.scss
+++ b/scss/_layouts_application-panels.scss
@@ -1,15 +1,27 @@
 @mixin vf-l-application-panels {
   $header-padding-bottom: $sp-unit * 6;
 
-  .p-panel,
-  .p-panel--aside {
+  .p-panel {
     background: $colors--light-theme--background-default;
     color: $colors--light-theme--text-default;
+    max-width: 100%;
     min-height: 100%; // make sure it fills whole container even when has little content
 
     &.is-dark {
       background: $colors--dark-theme--background-default;
       color: $colors--dark-theme--text-default;
+    }
+
+    &.is-wide {
+      width: 720px; // ~ col-8
+    }
+
+    &.is-medium {
+      width: 530px; // ~ col-6
+    }
+
+    &.is-narrow {
+      width: 340px; // ~ col-4
     }
   }
 
--- a/scss/_layouts_application-panels.scss
+++ b/scss/_layouts_application-panels.scss
@@ -11,18 +11,6 @@
       background: $colors--dark-theme--background-default;
       color: $colors--dark-theme--text-default;
     }
-
-    &.is-wide {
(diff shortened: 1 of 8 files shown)


anthonydillon:
Do you need `is-medium`? That seems like the sensible default.

bartaz:
> Do you need `is-medium`? That seems like the sensible default.

By default I didn't want to set any width to let content define the width or app developers to decide what width suits given content the best.

scss/_layouts_application-panels.scss
```diff
@@ -1,16 +1,28 @@
...
+    &.is-wide {
+      width: 720px; // ~ col-8
+    }
+
+    &.is-medium {
+      width: 530px; // ~ col-6
+    }
+
+    &.is-narrow {
+      width: 340px; // ~ col-4
```
anthonydillon, line 24: What about using `vw` unit instead to be responsive on all screen sizes. Such as `70vw`, `50vw` and `30vw`?

scss/_layouts_application-panels.scss
```diff
@@ -1,16 +1,28 @@
...
     &.is-dark {
       background: $colors--dark-theme--background-default;
       color: $colors--dark-theme--text-default;
     }
+
+    &.is-wide {
+      width: 720px; // ~ col-8
+    }
+
+    &.is-medium {
```
anthonydillon, line 19: I think we want this as the default. There is no case where we just want the content to dictate the width.

anthonydillon:
@bartaz This is much better, thanks!

scss/_layouts_application-panels.scss
bartaz: These are just placeholders for now, we are discussing different variants with @lyubomir-popov.
We discussed flexible vw / % values as well, but it's not something we currently use often in Vanilla.
Currently layout is usually proportional to font size (rems) and not to screen size. Using 'responsive' values such as 30vw for narrow panel would mean we will need to redefine it on different breakpoints, because 30vw on large screen will fit much more content than on small screen. And we should make sure they are more consistent.

bartaz:
This PR seems stuck. It doesn't show new commits on the branch.
Will close and try to recreate new one.

PR #3384, opened 2020-11-03
Fix spacing between different `p-button` element types

### LINKED ISSUE #1897: Margin of buttons missing when .p-button-* is applied to different tags
1. Bug template

## Process

1. go to https://codepen.io/lyubomir-popov/pen/RJYyPV
2. change the <a> tag of the last button to <button>
3. observe how the buttons get closer 

## Expected result

changing markup shouldn't affect the maegins

## Current result

due to the :not(only-of-type), the margin is not applied because the two buttons are no longer of the same type

## Screenshot
screen shot 2018-06-26 at 14 50 41

sowasred2012:
## Done

Multiple sibling elements with the `p-button` class should have extra margin applied to ensure nice spacing between them, but this wouldn't happen if the siblings were different types of element.

~This PR updates the selectors used to apply the margin so that the type of element isn't a factor.~
See comments below - using various combinations of selectors introduced other issues, so this PR now applies a right margin to all `p-button` elements except a `:last-child`.

Fixes #1897 

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/docs/examples/patterns/buttons/neutral or [demo]
- See that the three elements have consistent margin between them, and compare it to what happens in Vanilla currently https://codepen.io/sowasred2012/pen/GRqGEjv

Make sure PR has one of the following labels to automatically categorise it in release notes:
`Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.

## Screenshots

Before:
Screenshot from 2020-11-03 14-36-13

After:
Screenshot from 2020-11-04 11-05-45

### DIFF
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -42,10 +42,11 @@
     }
 
     @media only screen and (min-width: $breakpoint-x-small + 1) {
+      margin-right: $sph-outer;
       width: auto;
 
-      &:not(:last-of-type):not(:only-of-type) {
-        margin-right: $sph-outer;
+      &:last-child {
+        margin-right: 0;
       }
     }
 
--- a/templates/docs/examples/patterns/buttons/alignment.html
+++ b/templates/docs/examples/patterns/buttons/alignment.html
@@ -1,0 +1,30 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Buttons / Alignment{% endblock %}
+
+{% block standalone_css %}patterns_buttons{% endblock %}
+
+{% block content %}
+<p>
+  <button>Classless button</button>
+  <button>Classless button</button>
+</p>
+
+<p>
+  <button class="p-button--neutral">Neutral button</button>
+  <button class="p-button--positive">Positive button</button>
+  <button class="p-button--negative">Negative button</button>
+</p>
+
+<p>
+  <a href="#" class="p-button--neutral">Neutral link</a>
+  <a href="#" class="p-button--positive">Positive link</a>
+  <a href="#" class="p-button--negative">Negative link</a>
+</p>
+
+<p>
+  <button>Classless button</button>
+  <button class="p-button--positive">Positive button</button>
+  <a href="#" class="p-button--neutral">Neutral link</a>
+  <button class="p-button--negative">Negative button</button>
+</p>
+{% endblock %}
--- a/templates/docs/examples/patterns/buttons/neutral.html
+++ b/templates/docs/examples/patterns/buttons/neutral.html
@@ -5,5 +5,8 @@
 
 {% block content %}
 <button class="p-button--neutral">Neutral button</button>
+
+<a href="#" class="p-button--neutral">Neutral button (link)</a>
+
 <button class="p-button--neutral" disabled>Neutral button disabled</button>
 {% endblock %}


sowasred2012:
@bartaz I've updated the example markup to demonstrate, wasn't sure the best way to demo the fix, I'll remove the markup if that's not how we do things!

scss/_base_button.scss
```diff
@@ -44,8 +44,9 @@
     @media only screen and (min-width: $breakpoint-x-small + 1) {
       width: auto;
 
-      &:not(:last-of-type):not(:only-of-type) {
-        margin-right: $sph-outer;
+      + [class^='p-button'],
+      + [class*='p-button'] {
```
bartaz, line 48: Why both `^=` and `*=` are needed?
And I guess this will not cover `<button /> <button />` case.

bartaz:
> @bartaz I've updated the example markup to demonstrate, wasn't sure the best way to demo the fix, I'll remove the markup if that's not how we do things!

As for that, I would keep this example as it is.

But create a separate one, maybe "Buttons / Alignment" or something specifically for that. And make sure to cover all cases in it:

- classless buttons next to each other
- buttons with `p-button` classes next to each other
- links with `p-button` next to each other
- mixed buttons and links next to each other

sowasred2012:
Adding -1 labels - Percy pointed out a few things...

scss/_base_button.scss
sowasred2012: True - fixing

sowasred2012:
@lyubomir-popov I thought I'd figured this out, but adding margin-left causes problems when multiple `p-button` elements wrap in their container:

Screenshot from 2020-11-03 15-54-22

This is using sibling selectors, and almost gets us there, but I'm forced to use left margin, rather than right, leading to the problem above.

To style an element directly (rather than its sibling), there is no `last-of-class` selector, and `last-of-type`/`only-of-type` aren't suitable, as you pointed out in the issue. `last-child` isn't suitable either, we can't be certain a button is the last child in a parent element.

I wondered if the fix was simpler: apply the right margin to all buttons, and let users remove it where they need to with `u-no-margin--right`? Obviously that would potentially have an impact on sites using Vanilla.

lyubomir-popov:
@sowasred2012 using the `no-margin-*` sounds reasonable to me. Since there's no last-of-class selector, wer'e doomed to fail regardless of how clever we try to be with the selectors.

bartaz:
@sowasred2012 @lyubomir-popov 

Currently buttons have margins on the right and last/only button doesn't.
We should make sure to keep this working and not rely too much on utilities.

Adding right margin on all buttons by default may affect a lot of places that have right aligned buttons.

I guess `:last-of-type` will not work in case of buttons mixed with links.
So maybe we could at least do `:last-child` and remove right margin on that? This should hopefully cover most cases of right aligned buttons.

templates/docs/examples/patterns/buttons/alignment.html
```diff
@@ -0,0 +1,22 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Buttons / Alignment{% endblock %}
+
+{% block standalone_css %}patterns_buttons{% endblock %}
+
+{% block content %}
+<button>Classless button</button>
+<button>Classless button</button>
+
+<button class="p-button--neutral">Neutral button</button>
```
bartaz, line 10: Could you group these into separate paragraphs?

sowasred2012: Cool - I've updated

sowasred2012:
@bartaz I agree `last-child` will cover a decent amount of cases, but even in Vanilla examples, there are instances where we would still need to use the utility class - the MAAS layout has a hidden element after the button that prevents the `last-child` rule from being applied:

Screenshot from 2020-11-04 11-35-47

bartaz:
@sowasred2012

I had a look at the code that we currently have: `&:not(:last-of-type):not(:only-of-type)`. So we used to say that buttons that are `:last-of-type` or `:only-of-type` should not have a margin.

Now, we know that `:last-of-type` will not work and gives us 'false positives' (removes margin from buttons that are followed by buttons of different type).
But `:only-of-type` should still be fine.

`&:only-of-type, &:last-child { margin-right: 0 }`

This will hopefully cover some more cases (when there is one button followed by something else). Of course, there still may be cases that will break (multiple buttons followed by something else), but there should be less of those...

The other way around that may be (but I haven't tried that) to stop trying to make ALL types of buttons aligned correctly, and focus on fixing this for `p-button` only (not base `<button>`). But, as there is no `:last-of-class` in CSS this may be hard or impossible to achieve anyway. The fact that we need this to be right margin is limiting here as well (as we don't want left margin when line wraps for inline buttons).

lyubomir-popov:
Should we earmark this for a proper solution when we're closer to 3.0? When we can do something like margin-right: on all, with manual reset to zero with a class?

sowasred2012:
> Now, we know that `:last-of-type` will not work and gives us 'false positives' (removes margin from buttons that are followed by buttons of different type).
> But `:only-of-type` should still be fine.
> 
> `&:only-of-type, &:last-child { margin-right: 0 }`

This would still cause the problem mentioned in the original issue, which had a `button`, an`a`, and another `button` - the two `button`s would have margins but the `a`, being the only one of its type, wouldn't. In the example template I've added, you can see the problem occurring when `only-of-type` is applied:

Screenshot from 2020-11-04 12-10-21

bartaz:
> This would still cause the problem mentioned in the original issue, which had a `button`, an`a`, and another `button` 

Right. So, both `last-of-type` and `only-of-type` are of no use when we mix types.

Seems like there is no way of fixing this without breaking changes.


I quickly had a look at bootstrap and material UI to see what they are doing about it and... they don't have margins on buttons at all. They seem to handle spacing separately from components.


I start to wonder if fixing this (and introducing other issues) is worth it.
The original bug doesn't point to any site having this issue. So I'm not sure if we actually have a problem of buttons mixed with links in real use.

Sure, technically it is a bug, and it arguably is better to have a margin that you can remove with utility (which is proposed in this PR) then having no margin and no way of adding it (which is current case if you mix button with link).
But the way to 'fix' this may affect many places in our sites and there is no way to easily verify who may be affected. It may take a long time until someone notices misaligned button on some page.

@anthonydillon @lyubomir-popov @sowasred2012 Any thoughts on that?

lyubomir-popov:
I filed this in response to an issue somewhere, but can't remember where - so it is not sa purely theoretical problem. I would vote to hold off until we can fix it properly and don't have to worry about a breaking change, e.g. 3.0 release.

sowasred2012:
@bartaz I lean towards agreeing with you, on this not being worth the time spent trying to fix - each potential solution seems to trade one problem for another, as seen in the Percy checks, and if I've ever encountered the issue (though I can't think of a specific example), I imagine the simpler solution was to add an inline margin style, shameful as that is.

sowasred2012:
Closing - as discussed, the solutions we've attempted each introduce new problems that would have a relatively large impact on sites using Vanilla.

PR #3385, opened 2020-11-03
remove top border from first item in matrix on small screens

### LINKED ISSUE #2517: Remove top border from Matrix on small screens
## Pattern to amend
Remove the top border on the first element on the Matrix pattern on small screens

## Visual
Screenshot_2019-08-28 Vanilla documentation

Request:
Screenshot_2019-08-28 Vanilla documentation

sowasred2012:
## Done

Removed top border from first matrix item on small screens

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/docs/examples/patterns/matrix or [demo]
- At the small screen breakpoint, see that the first list item doesn't have a top border
- Resize your browser through the wider breakpoints and see that the top border shows/hides appropriately
- Compare to https://vanillaframework.io/docs/examples/patterns/matrix

Fixes #2517 

## Screenshots

Before:
Screenshot from 2020-11-03 16-52-40

After:
Screenshot from 2020-11-03 16-50-24

### DIFF
--- a/scss/_patterns_matrix.scss
+++ b/scss/_patterns_matrix.scss
@@ -25,6 +25,10 @@
     padding-bottom: $spv-inner--scaleable;
     padding-top: calc(#{$spv-inner--scaleable} - 1px);
 
+    &:first-child {
+      border-top: none;
+    }
+
     @media (min-width: $breakpoint-small) {
       display: flex;
       flex-wrap: wrap;


anthonydillon (approved):
Works, do you think `first-child` would be clearer for other developers to understand what's going on?

bartaz:
> Works, do you think `first-child` would be clearer for other developers to understand what's going on?

I agree with Ant on that. We tend to do a lot of magic in Vanilla with `& + &` and similar selectors, but on a glance it's hard to understand what it's actually meant to be doing.

Could we try if `&:first-child { border-top: none }` does the trick?

sowasred2012:
@anthonydillon @bartaz ahhh good point - I had fallen out with psuedo selectors on a different issue, hence my original, slightly roundabout solution. Updated.

anthonydillon:
@sowasred2012 thanks LGTM :+1:

Issue #3269, opened 2020-09-10
Side navigation: make titles #111


lyubomir-popov:
## Pattern to amend

Side navigation

## Visual

We received a request to make sidenav titles stand out more against links. After a [discussion](https://docs.google.com/document/d/1YuEQLWLKaf-Rxbn0g1St-FdsrCnTAKihVoBWZX_60M4/edit#bookmark=id.ouhsa0yjifir) in the design review meeting it was decided to darken titles like this:

![image](https://user-images.githubusercontent.com/2741678/92724995-5dd87480-f363-11ea-930a-ea95c63ed994.png)

bartaz:
Just to be clear, technically titles can be links as well. How should they be styled in such case?

lyubomir-popov:
My proposal would be to apply the standard hover to it in case it is a link.

PR #3394, opened 2020-11-09
Add psuedo element on tooltip hover to allow interaction with tooltip message

### LINKED ISSUE #2325: Tooltips should stay active on hover

sowasred2012:
#3269  Done

Added a small pseudo element to tooltip elements on hover, so that it's possible for a user to mouseover the tooltip message.

Fixes #2325 

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/docs/examples/patterns/tooltips or [demo]
- Hover over a tooltip button, see that you are able to move your cursor over the tooltip message without it disappearing.

## Screenshots

Before:
tooltip1

After:
tooltip2


bartaz:
@sowasred2012 Works nice on the buttons, thanks. But tooltips may be also on smaller elements, like icons, and in such case having the 'hover' area only extending from element that triggers the tooltip doesn't seem enough:

icon-tooltip

Would it be possible to extend area below the tooltip itself?

Screenshot 2020-11-10 at 09 37 15

Or, even better try to play around with border radius to make it more elliptic and cut the corners?

Screenshot 2020-11-10 at 09 26 27

sowasred2012:
@bartaz thanks for feedback, have updated with the Chrome fix :+1:

scss/_patterns_tooltips.scss
```diff
@@ -53,12 +53,26 @@ $triangle-height: $sp-unit;
...
+      right: 0;
+      top: -15px;
+    }
   }
 
   // Bottom center tooltip
   .p-tooltip--btm-center {
     @extend %tooltip;
 
+    &:hover::before {
```
bartaz, line 72: Is this needed? It doesn't seem to do anything.
It's targeting `.p-tooltip--btm-center:hover::before`, but we don't have `::before` defined for it, so these styles don't target any element.
It's redefined for all position variants of the tooltip.

bartaz: Or is it a leftover of previous implementation?

sowasred2012: Sorry, that's left over from the previous solution - updated and removed!

scss/_patterns_tooltips.scss
```diff
@@ -53,6 +53,16 @@ $triangle-height: $sp-unit;
       position: absolute;
       width: 0;
     }
+
+    &::after {
+      border-radius: 10% 90% 0% 100% / 100% 100% 0% 0%;
+      content: '';
+      height: 15px;
```
bartaz, line 60: I guess the 15px here is related to triangle height, so maybe we could reuse `$triangle-height` variable for that (and other positioning values)?

scss/_patterns_tooltips.scss
```diff
@@ -71,6 +81,11 @@ $triangle-height: $sp-unit;
         -webkit-transform: translateX(-50%); // stylelint-disable-line property-no-vendor-prefix
         transform: translateX(-50%);
       }
+
+      &::after {
+        border-radius: 50% 50% 0% 100% / 100% 100% 0% 0%;
```
bartaz, line 86: Would using direct props such as `border-top-right-radius` etc help and reduce the number of overrides we do here in all the variants?
Would be good to only modify borders that need to be modified, not redefine everything.

scss/_patterns_tooltips.scss
```diff
@@ -53,6 +53,16 @@ $triangle-height: $sp-unit;
       position: absolute;
       width: 0;
     }
+
+    &::after {
```
bartaz, line 57: As these pseudo-elements (both `::before` and `::after`) don't have any classnames that could describe their meaning it would be good to put a comment above each of those to describe their use.

So `::before` would be the tooltip triangle 'tip', and `::after` a hover overlay?

scss/_patterns_tooltips.scss
```diff
@@ -53,6 +56,17 @@ $triangle-height: $sp-unit;
       position: absolute;
       width: 0;
     }
+
+    // hover overlay
+    &::after {
+      border-radius: 10% 90% 0% 100% / $bottom-border-radius;
+      content: '';
+      height: #{2 * $triangle-height};
```
bartaz, line 64: Isn't `height: 2 * $triange-height;` working here? `#{}` should not be needed on such simple calculations.
Should also be true with `-2 * $triangle-height`.

scss/_patterns_tooltips.scss
```diff
@@ -1,5 +1,7 @@
 @import 'settings';
 $triangle-height: $sp-unit;
+$bottom-border-radius: 100% 100% 0% 0%;
```
bartaz, line 3: I know I suggested such names, but the may be confusing without context.
It sounds like it's related to bottom border radius of some element, while it is a border radius of bottom positioned tooltip.

I would at least prefix the variables with tooltip `$tooltip-bottom-border-radius` (to make sure it doesn't collide with some other variable in sass).
And maybe add a comment above those to give a bit more context. Something like:
`// these are the values of second border radius used for hover overlays in bottom and top aligned tooltips` ?

scss/_patterns_tooltips.scss
sowasred2012: You're right - I had used the same format I saw elsewhere in this file, but on testing it makes no difference, so have removed it from the example I used too.

bartaz (approved):
LGTM, thanks!

PR #3401, opened 2020-11-11
Allow .p-form--inline formgroups to wrap when necessary

### LINKED ISSUE #1930: p-form--inline elements do not wrap
The following bug was reported against MAAS recently: https://bugs.launchpad.net/maas/+bug/1782230

It appears that divs styled with _p-form--inline_ on the repository settings page render incorrectly, such that they don't appear to be visible. The divs are present in the markup, but as you can see from the attached screenshots, there appears to be a bug whereby the elements rendering outside the viewport fail to wrap appropriately.

# divs with p-form--inline
image
The "disabled pockets" div is selected here in the inspector.

# divs without p-form--inline
image

# Markup
The markup in question is in the form:

```html
<div class="row">
  <div class="col-6">
      <div class="p-form--inline">
          ...
      </div>
      <div class="p-form--inline">
          ...
      </div>
  </div>
</div>
```

sowasred2012:
## Done

Fixed a bug where formgroups and buttons inside a `.p-form--inline` would not wrap and could break out of their parent element.

Fixes #1930 

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/docs/examples/patterns/forms/form-inline or [demo]
- Reduce the width of the viewport, see that the form elements wrap when necessary

### [canonical.com](https://canonical.com/careers/engineering#available-roles) 
Before:

Screenshot from 2020-11-11 14-32-55

After:

Screenshot from 2020-11-11 14-28-59

### Vanilla docs example
Before:
form-inline-2

After:
form-inline-1

### DIFF
--- a/scss/_patterns_forms.scss
+++ b/scss/_patterns_forms.scss
@@ -31,20 +31,21 @@
       align-items: baseline;
       display: inline-flex;
       flex-direction: row;
+      flex-wrap: wrap;
 
       > * {
-        margin: 0;
+        margin-right: $sph-inner--large;
       }
     }
 
     .p-form__group {
       @media screen and (min-width: $breakpoint-medium) {
         display: flex;
+        flex-shrink: 1;
         width: auto;
 
-        + .p-form__group,
         + [class*='p-button'] {
-          margin-left: $sph-inner--large;
+          flex-shrink: 1;
         }
 
         .p-form__label,
--- a/templates/docs/examples/patterns/forms/form-inline.html
+++ b/templates/docs/examples/patterns/forms/form-inline.html
@@ -7,7 +7,7 @@
 <form class="p-form p-form--inline">
   <div class="p-form__group">
     <label for="username-inline" class="p-form__label">Username</label>
-    <div class="p-form__control u-clearfix">
+    <div class="p-form__control">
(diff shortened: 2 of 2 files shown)


lyubomir-popov:
Just a clearfix that doesn't seem needed, other thanthat looks good!

lyubomir-popov (approved):
LGTM! Please just  remove the clearfix.

scss/_patterns_forms.scss
```diff
@@ -31,20 +31,22 @@
       align-items: baseline;
       display: inline-flex;
       flex-direction: row;
+      flex-wrap: wrap;
+      padding-left: $sph-inner--large;
 
       > * {
-        margin: 0;
+        margin: 0 #{2 * $sph-inner--large} 0 #{-$sph-inner--large};
```
bartaz, line 38: I may not fully understand the context here. Why are we adding a padding-left to the form, just to remove it with negative margin on the child elements?
It seems to work the same with them being set to 0.

bartaz (changes requested):
I'd like to discuss the purpose of negative margin first.

scss/_patterns_forms.scss
sowasred2012: Good point!

bartaz (approved):
LGTM

PR #3404, opened 2020-11-12
Add .is-small button class

### LINKED ISSUE #3341: Add option for small button
## Pattern to amend

Buttons

## Visual

Screenshot 2020-10-07 at 14 56 33
Screenshot 2020-10-07 at 14 56 19

## Context

As part of modular table (and possibly other use cases) we need buttons to work along small text and in dense areas.

This was initially done as part of SummaryButton component https://github.com/canonical-web-and-design/react-components/pull/300
Prototyped in https://github.com/lyubomir-popov/vanilla-framework/commit/5cb82b2dee79bbb0ed26cdc76118513776335484

sowasred2012:
## Done

Added `.is-small` modifier class for buttons, and `.is-dense` combo modifier.

Fixes #3341 

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/docs/examples/patterns/buttons/small
- See that the buttons match the visual given in the issue https://github.com/canonical-web-and-design/vanilla-framework/issues/3341

## Screenshots

Screenshot from 2020-11-12 12-19-31

### DIFF
--- a/templates/docs/component-status.md
+++ b/templates/docs/component-status.md
@@ -22,18 +22,12 @@ When we add, make significant updates, or deprecate a component we update their 
     </tr>
   </thead>
   <tbody>
-    <!-- 2.19 -->
+    <!-- 2.20 -->
     <tr>
-      <th><a href="/docs/base/tables#expanding">Expanding table column placeholder</a></th>
-      <td><div class="p-label--updated">Updated</div></td>
-      <td>2.20.0</td>
-      <td>We started using <code>aria-hidden="true"</code> attribute to hide the dummy table header in place of previously used <code>u-hide</code> utility.</td>
-    </tr>
-    <tr>
-      <th><a href="/docs/base/tables#icons">Table icons</a></th>
+      <th><a href="/docs/patterns/buttons#small">Small buttons</a></th>
       <td><div class="p-label--new">New</div></td>
-      <td>2.19.0</td>
-      <td>We added a <code>p-table__cell--icon-placeholder</code> class to properly align icons in table cells.</td>
+      <td>2.20.0</td>
+      <td>We added an <code>is-small</code> modifier class for buttons, which can be combined with <code>is-dense</code>.</td>
     </tr>
   </tbody>
 </table>
@@ -50,6 +44,19 @@ When we add, make significant updates, or deprecate a component we update their 
     </tr>
   </thead>
   <tbody>
+    <!-- 2.19 -->
+    <tr>
+      <th><a href="/docs/base/tables#expanding">Expanding table column placeholder</a></th>
+      <td><div class="p-label--updated">Updated</div></td>
+      <td>2.19.2</td>
+      <td>We started using <code>aria-hidden="true"</code> attribute to hide the dummy table header in place of previously used <code>u-hide</code> utility.</td>
+    </tr>
+    <tr>
+      <th><a href="/docs/base/tables#icons">Table icons</a></th>
+      <td><div class="p-label--new">New</div></td>
+      <td>2.19.0</td>
+      <td>We added a <code>p-table__cell--icon-placeholder</code> class to properly align icons in table cells.</td>
+    </tr>
     <!-- 2.18 -->
     <tr>
       <th><a href="/docs/patterns/breadcrumbs">Breadcrumbs</a></th>
--- a/templates/docs/examples/patterns/buttons/small.html
+++ b/templates/docs/examples/patterns/buttons/small.html
@@ -1,0 +1,8 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Buttons / Small{% endblock %}
+
+{% block standalone_css %}patterns_buttons{% endblock %}
+
+{% block content %}
+<small>This is small text <button class="p-button--neutral is-small">This is a small button</button><button class="p-button--neutral is-small is-dense">This is a small, dense button</button></small>
+{% endblock %}
--- a/templates/docs/examples/patterns/buttons/small.html
+++ b/templates/docs/examples/patterns/buttons/small.html
@@ -4,5 +4,5 @@
 {% block standalone_css %}patterns_buttons{% endblock %}
 
 {% block content %}
-<small>This is small text <button class="p-button--neutral is-small">This is a small button</button><button class="p-button--neutral is-small is-dense">This is a small, dense button</button></small>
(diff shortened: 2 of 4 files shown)


lyubomir-popov:
@sowasred2012 they are off the baseline grid though

Screenshot 2020-11-12 at 12 37 17

bartaz:
@lyubomir-popov @sowasred2012 

I don't think it's an issue with the buttons themselves.
Small text in Vanilla doesn't stay on baseline grid. So the buttons just follow that.

Small buttons are in line with small text, and that's expexted:

Screenshot 2020-11-12 at 14 09 20

sowasred2012:
@lyubomir-popov I think @bartaz is correct: https://vanillaframework.io/docs/examples/base/small

Screenshot from 2020-11-12 13-37-06

lyubomir-popov:
@sowasred2012 @bartaz there's a slight distinction. <small> is inline, so impossible to apply padding or margin to. As a result it doesn't align to the baseline grid.

On the other hand, if you apply the .p-text--small class to a block level element, the text inside aligns nicely to the grid:
Screenshot 2020-11-12 at 13 53 38

Given that the buttons are inline-block, so essentially block level, I think it would be reasonable to have them on the baseline grid. This way, placed next to text or other elements in an adjacent column, the baselines will line up and produce a nicer looking combination.

bartaz:
`inline-block` doesn't make it block element. It has block properties (width/height/margins, etc), but is still inline in the flow of text around it.

Once you put small button in paragraph (not directly in the document) it is on baseline - because paragraph defines baseline:

Screenshot 2020-11-12 at 15 38 27

If we start adjusting position of the button we will need to position it differently when it's in paragraph, when it's not, when it is in `<small>` or not.

I guess should make a better example that shows small buttons in context we expect them to be (and they should be on baseline there). Not just thrown into body.

Currently it's being used inside the table in `<small>`:

Screenshot 2020-11-12 at 15 43 12

So I guess the idea is that when it's places alongside small text it's in line with it?

lyubomir-popov:
You're right @bartaz, and yes - we need to be able to align it to the baseline of small text. An example that shows how to do that would be great.

bartaz:
The easiest way I found (which seems reasonable) is simply putting it in paragraph:

`<p><small>This is small text <button class="is-small">This is small button</button></small></p>`

anthonydillon:
Why not just make the font-size and padding of the button relative to the font size of its container with `em`'s?

bartaz:
> Why not just make the font-size and padding of the button relative to the font size of its container with `em`'s?

Would we want external context to define styles of the button? Do we want buttons in headings have same font as heading?

In many cases in Vanilla I see that we are not very consistent in what controls the styling of given element. On one hand we use BEM class names and want the styles only to be defined by given component and classes on it.
In others we let context of other parent elements define some aspects of the style. Both have their benefits and issues, but we don't seem to have any good rules when to do what.

Buttons are very common component and I think they are used in many different contexts. We also seem to want to have a lot of control of how they are styled (the size, alignment, margins, etc). If we keep button styling in the button component only and provide any needed (and supported by us) variants we can assure that given button (with given set of modifier classes) will behave consistently regardless of where it's put in the document.

templates/docs/examples/patterns/buttons/small.html
```diff
@@ -0,0 +1,8 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Buttons / Small{% endblock %}
+
+{% block standalone_css %}patterns_buttons{% endblock %}
+
+{% block content %}
+<small>This is small text <button class="p-button--neutral is-small">This is a small button</button><button class="p-button--neutral is-small is-dense">This is a small, dense button</button></small>
```
bartaz, line 7: Could we wrap it in paragraph as well to make sure it's properly placed on baseline grid?

```suggestion
<p><small>This is small text <button class="p-button--neutral is-small">This is a small button</button><button class="p-button--neutral is-small is-dense">This is a small, dense button</button></small></p>
```

bartaz (approved):
LGTM, with little suggestion

bartaz:
@sowasred2012 Ah, one more thing worth adding. Could you add new button to [component status page](https://vanilla-framework-3404.demos.haus/docs/component-status)?

Update the version to 2.20 (we are adding new classname so it a new feature), move the rows from 2.19 to table below, and add a new one shortly describing addition of the new button.

package.json
```diff
@@ -42,7 +42,7 @@
     "clean": "rm -rf build docs/static/css node_modules/ yarn-error.log",
     "percy": "percy exec -- node snapshots.js"
   },
-  "version": "2.19.3",
+  "version": "2.20.0",
```
bartaz, line 45: Oh, sorry, I meant updating the version in component status page "What's new in Vanilla 2.20" heading, not in package.json yet - lets leave it for the release.

templates/docs/component-status.md
```diff
@@ -23,18 +23,18 @@ When we add, make significant updates, or deprecate a component we update their
...
+    <tr>
+      <th><a href="/docs/patterns/buttons#small">Small buttons</a></th>
+      <td><div class="p-label--new">New</div></td>
+      <td>2.20.0</td>
+      <td>We added an <code>is-small</code> modifier class for buttons, which can be combined with <code>is-dense</code>.</td>
+    </tr>
     <tr>
       <th><a href="/docs/base/tables#expanding">Expanding table column placeholder</a></th>
       <td><div class="p-label--updated">Updated</div></td>
       <td>2.20.0</td>
```
bartaz, line 35: This is a typo actually. It was released in 2.19.2 (.3?) I think. So this as well should drop down to table below.
(and update the version to the correct one).

```suggestion
      <td>2.19.2</td>
```

templates/docs/component-status.md
```diff
@@ -23,18 +23,18 @@ When we add, make significant updates, or deprecate a component we update their
   </thead>
   <tbody>
     <!-- 2.19 -->
```
bartaz, line 25: Move this comment along with the rows. Or just remove it.

bartaz (changes requested):
Some clarifications.

templates/docs/component-status.md
sowasred2012: Ah OK - I thought it had crept in there, updated!

bartaz (approved):

PR #3410, opened 2020-11-16
Add ability to set foreground (tick) color on success icon

sowasred2012:
## Done

Adds a `fg-color` parameter to the success icon mixin and, when the icon is a child of a button, switches the foreground/tick color to transparent, and sets the background/circle color to white.

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/3142
## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/docs/examples/patterns/buttons/icon or [demo](https://vanilla-framework-3410.demos.haus/docs/examples/patterns/buttons/icon)
- See the success icon on each button looks good.

### DIFF
--- a/scss/_base_icon-definitions.scss
+++ b/scss/_base_icon-definitions.scss
@@ -83,8 +83,8 @@ $social-icon-size: map-get($icon-sizes, heading-icon--small);
   background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M6.964 1a5.964 5.964 0 014.709 9.623l4.303 4.305-1.06 1.06-4.306-4.305A5.964 5.964 0 116.963 1zm0 1.5a4.464 4.464 0 100 8.927 4.464 4.464 0 000-8.927z' fill='#{vf-url-friendly-color($color)}' fill-rule='nonzero'/%3E%3C/svg%3E");
 }
 
-@mixin vf-icon-success($color: $color-positive) {
-  background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' xmlns='http://www.w3.org/2000/svg'%3E%3Cg fill='none' fill-rule='evenodd'%3E%3Ccircle fill='#{vf-url-friendly-color($color)}' cx='8' cy='8' r='7'/%3E%3Cpath fill='%23FFF' fill-rule='nonzero' d='M10.83 4.501l1.208.89-5.033 6.83-2.922-3.096 1.091-1.029 1.689 1.789z'/%3E%3C/g%3E%3C/svg%3E");
+@mixin vf-icon-success($bg-color: $color-positive, $fg-color: $color-x-light) {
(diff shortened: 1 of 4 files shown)


scss/_base_icon-definitions.scss
```diff
@@ -83,8 +83,8 @@ $social-icon-size: map-get($icon-sizes, heading-icon--small);
   background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M6.964 1a5.964 5.964 0 014.709 9.623l4.303 4.305-1.06 1.06-4.306-4.305A5.964 5.964 0 116.963 1zm0 1.5a4.464 4.464 0 100 8.927 4.464 4.464 0 000-8.927z' fill='#{vf-url-friendly-color($color)}' fill-rule='nonzero'/%3E%3C/svg%3E");
 }
 
-@mixin vf-icon-success($color: $color-positive) {
-  background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' xmlns='http://www.w3.org/2000/svg'%3E%3Cg fill='none' fill-rule='evenodd'%3E%3Ccircle fill='#{vf-url-friendly-color($color)}' cx='8' cy='8' r='7'/%3E%3Cpath fill='%23FFF' fill-rule='nonzero' d='M10.83 4.501l1.208.89-5.033 6.83-2.922-3.096 1.091-1.029 1.689 1.789z'/%3E%3C/g%3E%3C/svg%3E");
+@mixin vf-icon-success($bg-color: $color-positive, $fg-color: $color-x-light) {
```
bartaz, line 86: Not sure if it was, but could we pass it through SVGO first? https://jakearchibald.github.io/svgomg/

templates/docs/examples/patterns/buttons/icon.html
```diff
@@ -4,7 +4,9 @@
 {% block standalone_css %}patterns_buttons{% endblock %}
 
 {% block content %}
-<button class="p-button--positive has-icon"><i class="p-icon--minus is-light"></i></button>
+<button class="p-button--positive has-icon"><i class="p-icon--success"></i></button>
+<button class="p-button--negative has-icon"><i class="p-icon--success"></i></button>
```
bartaz, line 8: This example is supposed to demonstrate different uses of icons in buttons, I wouldn't turn it into success icon exploration.
So leave just the first positive button with the success icon, and drop the rest.

We have separate example https://vanilla-framework-3410.demos.haus/docs/examples/templates/tick-element-comparison where we explore different form elements, so I guess we can put buttons with success icons alongside the buttons there?

Screenshot 2020-11-17 at 14 28 47

scss/_patterns_buttons.scss
```diff
@@ -149,5 +149,9 @@ $hover-background-opacity-percentage: $hover-background-opacity-amount * 100%;
         margin-right: -#{$sph-inner--small};
       }
     }
+
+    & .p-icon--success {
```
bartaz, line 153: Adding this here makes it work on all the buttons, not just positive/negative/brand.
Now if you put success icon in base button it's invisible (white on white background).

bartaz (changes requested):
We should change the success icon only in positive/negative/brand buttons, not all the buttons.

scss/_patterns_buttons.scss
bartaz: Maybe we can create custom placeholder :

```scss
%vf-button-white-success-icon {
   & .p-icon--success {
      @include vf-icon-success($color-x-light, $color-transparent);
   }
}
```
And then `@extend` it where `--positive`, `--negative`, `--brand` buttons are defined?

sowasred2012: good idea - I'd added that to the classes directly, I've just changed it to a placeholder.

bartaz (approved):
LGTM, thanks

PR #3413, opened 2020-11-18
Add "is-active" button state

sowasred2012:
## Done

Adds an `is-active` state class to buttons

Fixes another aspect of https://github.com/canonical-web-and-design/vanilla-framework/issues/3142

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/docs/examples/patterns/buttons/active or [demo]
- See that the button is not opaque, and also not interactive.

### DIFF
--- a/templates/docs/patterns/buttons.md
+++ b/templates/docs/patterns/buttons.md
@@ -88,6 +88,14 @@ Should you wish to place an icon in a button. You will not want to button to bec
 View example of the icon button pattern
 </a></div>
 
+### Active
+
+In cases where a button needs to indicate that an action is occurring (e.g. saving data, processing a payment) while also preventing user interaction, the state class `is-active` can be added to a disabled button to maintain full opacity.
+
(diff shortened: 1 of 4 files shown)


templates/docs/patterns/buttons.md
```diff
@@ -64,6 +64,14 @@ Should you wish to place a button after a line of inline text, as a CTA for exam
 View example of the inline button pattern
 </a></div>
 
+### Active
```
bartaz, line 67: Could we move it to be the last variant on the docs page, I think the following (dense, small, icon) are more common and could be earlier on the page.

`is-active` is quite an edge case, so it could be mentioned last.

bartaz (approved):
LGTM, just a little comment about order on the docs page.

templates/docs/patterns/buttons.md
sowasred2012: :+1: updated

PR #3414, opened 2020-11-18
Set position: relative on .p-checkbox

### LINKED ISSUE #3411: Jumping checkboxes
**Describe the bug**
When you use a list with the new `p-checkbox` component in it. The input is absolutely positioned at the bottom of the nearest relative element (which is most likely the `strip`). This causes the screen to jump down if the content is long enough.

**To Reproduce**
Steps to reproduce the behaviour:

1. Go to https://codepen.io/anthonydillon/pen/vYKMELv
2. Click on "Jumping one" check box
3. See that it scrolls down the page
4. Scroll back up and click the "Fixed one" (see that the `p-checkbox` has `position: relative` applied)

**Expected behaviour**
Act as the Fixed one does.

**Additional context**
This is for https://charmhub.io

sowasred2012:
## Done

Set `position: relative;` on `.p-checkbox` to contain the absolutely positioned inputs.

Fixes #3411 

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/docs/examples/patterns/forms/checkbox
- Inspect the `p-checkbox` elements, see that the hidden checkbox element is now contained within its bounds

## Screenshots
 Before:
Screenshot from 2020-11-18 15-19-33

After: 
Screenshot from 2020-11-18 15-22-54

### DIFF
--- a/scss/_patterns_form-tick-elements.scss
+++ b/scss/_patterns_form-tick-elements.scss
@@ -17,6 +17,13 @@
     @extend %vf-pseudo-radio;
   }
 
+  // contain the absolutely positioned input
+  .p-checkbox,
+  .p-checkbox--heading,
+  .p-checkbox--inline {
+    position: relative;
+  }
+
   // inline variants
   .p-checkbox--heading,
   .p-checkbox--inline,


anthonydillon (approved):
LGTM :+1:

bartaz:
@sowasred2012 Thanks, looks good. Only one thing that bugs me a bit is that this makes input aligned to 'bottom: 0' now so it doesn't fully line up with the fake checkbox.

It's visible on your 'After' screenshot, but even more so with bigger fonts like with headings:

Screenshot 2020-11-18 at 17 13 35

It's not a hugely bug deal, but may for example make browser validation tooltips a bit misaligned.

I'll have a look tomorrow morning if there is a reasonably simple solution to that that wouldn't require us adjusting it for every possible font size.

sowasred2012:
@bartaz yeah, I spotted the same thing, but couldn't see a clean way to align it perfectly - I did try setting `top: 0; margin: auto;` on the input element to center it vertically in the parent element, but the top padding on the parent foiled that plan.

anthonydillon:
Could you do something like `top: calc(50% - $checkbox-height)`?

bartaz:
They are not centered in the container. Input lays on the baseline of the text.

I think that's what we use to position the 'fake' checkbox on baseline. As they are same size it should work on hidden input as well hopefully?

https://github.com/canonical-web-and-design/vanilla-framework/blob/d4b1ac7477023a18f1fe21d7e06385c396b6eda1/scss/_base_forms-tick-elements.scss#L56

PR #3429, opened 2020-11-25
Add focus-visible support to vf-focus mixin

### LINKED ISSUE #3352: Consider using focus-visible
Chrome adds support for focus-visible in recent 86 version, should be followed by Firefox soon.

This allows to show focus styles only on keyboard focus, not when clicking.

It seems to be possible to keep fallback :focus support for older browsers and override it with :focus-visible for those that support it.

https://matthiasott.com/notes/focus-visible-is-here
https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible

sowasred2012:
## Done

Added `:focus-visible` to the vf-focus mixin.

Fixes #3352 

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/ in Chrome/Chromium
- Tab through the page, see that links still receive a focus state. Click any of the links and see that the focus state is _not_ applied.

This does not yet work on browsers other than Chrome ([MDN Compatibility list](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#Browser_compatibility)). Firefox supposedly supports it using the alternative name `:moz-focusring`, but it seemed to have no effect. Browsers that don't support `:focus-visible` will fall back to the current `:focus` styling.

### DIFF
--- a/scss/_global_functions.scss
+++ b/scss/_global_functions.scss
@@ -58,12 +58,24 @@
   }
 }
 
-// Adds visiual focus to elements on :focus
+// Adds visual focus to elements on :focus-visible,
+// or :focus if the browser doesn't support the former
 @mixin vf-focus($color: $color-focus, $width: $bar-thickness, $has-validation: false) {
   &:focus {
     outline: $width solid $color;
     outline-offset: -#{$width};
   }
+
+  &:focus-visible {
+    outline: $width solid $color;
+    outline-offset: -#{$width};
+  }
+
+  &:focus:not(:focus-visible) {
+    outline: 0;
+    outline-offset: 0;
+  }
+
   @if ($has-validation) {
     .is-error &:focus {
       outline-color: $color-negative;


bartaz:
Firefox seems to be working on it, and it seems it was recently implemented (https://bugzilla.mozilla.org/show_bug.cgi?id=1445482) for version 85.

Not sure if it will be simply renaming `-mox-focusring` that we have some trouble getting to work or will it actually be fixed.
But I don't think we should try to address browser quirks. Let's wait for it to show up in Firefox in couple weeks.

scss/_global_functions.scss
```diff
@@ -58,12 +58,24 @@
...
-// Adds visiual focus to elements on :focus
+// Adds visual focus to elements on :focus-visible,
+// or :focus if the browser doesn't support the former
 @mixin vf-focus($color: $color-focus, $width: $bar-thickness, $has-validation: false) {
   &:focus {
     outline: $width solid $color;
     outline-offset: -#{$width};
   }
+
+  &:focus-visible {
```
bartaz, line 69: Could this be merged with focus above?

```
&:focus,
&:focus-visible {
  outline: ...
  outline-offset: ...
}
```

sowasred2012: I tried that too, but it didn't work - I _think_ because when compiled the selector is `foo:focus, foo:focus-visible: { ... }`, browsers that don't recognise `:focus-visible` ignore the entire thing. In practice, this meant Firefox didn't even apply `:focus` styling.

bartaz: Ah, makes sense. Thanks!

bartaz (approved):
LGTM, thanks!

PR #3430, opened 2020-11-25
Update command prompt font family

### LINKED ISSUE #3428: Code snippet uses wrong shell prompt icon
## Pattern to amend

Current dollar sign:
image

Correct icon:
https://assets.ubuntu.com/v1/27e7d804-code-snippet.svg
image

sowasred2012:
## Done

Set the `$font-monospace` font family on the command prompt character in the code copyable component.

Fixes #3428 

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/docs/examples/patterns/code-copyable
- See the updated command prompt icon and check that the design is OK

## Screenshots

Before:
Screenshot from 2020-11-25 11-56-12

After:
Screenshot from 2020-11-26 09-25-47

### DIFF
--- a/scss/_patterns_code-copyable.scss
+++ b/scss/_patterns_code-copyable.scss
@@ -19,15 +19,13 @@ $cc-button-size: 36px;
     }
 
     &::before {
-      color: $color-mid;
+      color: $colors--light-theme--text-muted;
       content: '$';
       display: block;
-      left: 0;
+      font-family: unquote($font-monospace);
       line-height: map-get($line-heights, default-text);
-      padding: 0 calc(#{$cc-button-size} + #{$sph-inner--small}) 0 $sph-inner--small;
-      position: absolute;
-      top: $spv-inner--x-small;
-      width: 1rem;
+      padding: 0 $sph-inner--small;
+      position: relative;
     }
   }
 
@@ -38,7 +36,7 @@ $cc-button-size: 36px;
     line-height: map-get($line-heights, default-text);
     margin-bottom: 0;
     margin-top: 0;
-    padding: 0 calc(#{$cc-button-size} + #{$sph-inner--small}) 0 $sph-inner * 1.5;
+    padding: 0 calc(#{$cc-button-size} + #{$sph-inner--small}) 0 0;
     width: 100%;
   }
 


scss/_patterns_code-copyable.scss
```diff
@@ -20,13 +20,13 @@ $cc-button-size: 36px;
 
     &::before {
       color: $color-mid;
-      content: '$';
+      content: url('#{$assets-path}27e7d804-code-snippet.svg');
```
bartaz, line 23: We shouldn't make Vanilla patterns depend on external resources. If we need this icon it should be added to Vanilla icon mixins as data URL.

bartaz (changes requested):
Please add the icon to _base_icon-definitions.scss as data URL in mixin, not external resource.

scss/_patterns_code-copyable.scss
```diff
@@ -20,13 +20,13 @@ $cc-button-size: 36px;
 
     &::before {
       color: $color-mid;
```
bartaz, line 22: It seems that currently many of these styles on `::before` were needed for sake of rendering $ as text (for example `color`, `line-height`, but possibly more). We should keep only things that are necessary to style the icon.

This will likely result in icon not being added as `content`, but as background-image, so the `::before` needs to be correctly sized and positioned as well.

sowasred2012: good point!

scss/_patterns_code-copyable.scss
sowasred2012: Ah ok - updated!

bartaz:
@sowasred2012 Please rebase as percy contains changes agains master

lyubomir-popov:
@sowasred2012 sorry this icon doesn't look right. Can we just update the current character to match the font-family and use $colors--light-theme--text-muted for color?

lyubomir-popov:
@sowasred2012 now that it is a character, it doesn't need abs position and can be simplified to this:

image

And the input padding no longer needs to create space for the pseudo element, so could look like so:
image

lyubomir-popov (approved):
Looks great, thanks @sowasred2012

bartaz (approved):

PR #3441, opened 2020-12-02
Show team members and contributors on contribute page

sowasred2012:
## Done

Get list of contributors from github and show them on /contribute, along with current team members.

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/contribute
- See the list of team members and contributors

Screenshot from 2020-12-02 17-09-31

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -2,8 +2,10 @@
 import glob
 import json
 import os
+import random
 
 # Packages
+import requests
 import flask
 from canonicalwebteam.flask_base.app import FlaskBase
 from canonicalwebteam.templatefinder import TemplateFinder
@@ -24,6 +26,17 @@ app = FlaskBase(
     template_500="500.html",
 )
 
+TEAM_MEMBERS = [
+    {"login": "anthonydillon", "role": "Engineering Director"},
+    {"login": "bartaz", "role": "Web Developer"},
+    {"login": "lyubomir-popov", "role": "Senior Visual Designer"},
+    {
+        "login": "wgx",
+        "role": "Lead UX Designer",
+    },
+    {"login": "sowasred2012", "role": "Web Developer"},
+]
+
 
 # Helpers
 # ===
@@ -33,7 +46,9 @@ def _get_title(title):
 
 def _get_examples():
     # get all example files (but ignore partials that start with _)
-    example_files = glob.glob("templates/docs/examples/*/**/[!_]*.html", recursive=True)
+    example_files = glob.glob(
+        "templates/docs/examples/*/**/[!_]*.html", recursive=True
+    )
     examples = {}
 
     for filepath in sorted(example_files):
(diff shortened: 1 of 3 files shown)


sowasred2012:
OK, this is causing the page to timeout on the demo - will investigate: https://sentry.is.canonical.com/canonical/vanillaframework-io/issues/13434/?referrer=alert_email

bartaz:
> OK, this is causing the page to timeout on the demo

I don't know much about our API backends, but could it be the case that demo may not have access to external APIs? Maybe it needs to be allowed for it to call github?

On another note - it would be worth having some fallback mechanism if API fails or timeouts. Like rendering the page without contributors or something? I don't know what are the practices around that in our backends?

sowasred2012:
> On another note - it would be worth having some fallback mechanism if API fails or timeouts. Like rendering the page without contributors or something? I don't know what are the practices around that in our backends?

It has a fallback if the API fails, it just returns an empty array of contributors, and if that's empty we don't show that section - maybe it could be updated to handle a timeout too. The only janky thing is that, if the API fails, the team member avatar images don't show.

webapp/app.py
```diff
@@ -63,6 +73,80 @@ def _get_examples():
     return examples
 
 
+def _make_github_request(endpoint):
+    # Passing a gh username as a header increases
+    # the API rate limit from 60 to 5000 per hour
+    headers = { "username": random.choice(team_members)["login"] }
```
bartaz, line 79: What's the practice in the team with such things? It feels bit weird to use team members quota for that.
Don't we have some API keys that we could use to call GH API?

sowasred2012: I agree it's weird - I was working from [the API docs](https://docs.github.com/en/free-pro-team@latest/rest/overview/resources-in-the-rest-api#authentication), which talks about basic auth and oauth tokens, both of which are tied to users, rather than organisations. I couldn't see an organisation-based solution. 

The reason I went with basic auth is we'd have to store the OAuth token as a secret somewhere, which felt like it would make this nice-to-have feature more complex - but maybe it's easier than I thought?

jpmartinspt: You should be [generating an access-token](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token) within our service account and authenticate using that(read it from an env var). You can test by creating one under your account first(read-only permissions should be enough for this), and then we can create one for staging and production. To authenticate you pass the token in the headers of the request `Authorization: token <generate-token>`

webapp/app.py
```diff
@@ -63,6 +73,80 @@ def _get_examples():
...
+
+    try: 
+        response = requests.get(
+            f'https://api.github.com/{endpoint}',
+            headers,
+            timeout=1
```
jpmartinspt, line 85: this timeout is really low, I don't think we have gone bellow `3` seconds anywhere

webapp/app.py
```diff
@@ -63,6 +73,80 @@ def _get_examples():
...
+        )
+
+        response.raise_for_status()
+    except:
+        return {}
+        
+    return response.json()
+
+
+def _get_team_members(contributors):
```
jpmartinspt, line 95: I attempted to simplify this a bit, what o you think?
```python3
def _get_team_members(contributors):
    contributors = {contributor["login"]: contributor for contributor in contributors}

    for team_member in team_members:
        member = contributors.get(team_member["login"]) or _make_github_request(f'users/{team_member["login"]}')
        member["role"] = team_member["role"]
        yield member
```

webapp/app.py
```diff
@@ -63,6 +73,80 @@ def _get_examples():
...
+
+def _get_contributors():
+    contributors = _make_github_request(
+        "repos/canonical-web-and-design/vanilla-framework/contributors"
+    )
+
+    return contributors
+
+
+def _filter_contributors(contributors):
```
jpmartinspt, line 132: I also had a go at this one, what do you think about:
```python3
def _filter_contributors(contributors):
    # Filter team members from contributors list
    member_usernames = [member["login"] for member in team_members]
    return [contributor for contributor in contributors if contributor["login"] not in member_usernames]
```

webapp/app.py
sowasred2012: Nice!

lyubomir-popov:
@sowasred2012 Could we cache and periodically update the cache to avoid an empty page?

jpmartinspt:
Since the page is going to be behind content-cache you could probably relax the timeout of the API call, and increase the time the page is cached by, assuming we are not that worried about showing slightly outdated contribution counters. It will also help with us eventually hitting the API rate limit.

bartaz:
> assuming we are not that worried about showing slightly outdated contribution counters.

Haha, this page was outdated for months, we can probably live with it sitting in a cache for hours or days even ;)

webapp/app.py
```diff
@@ -93,6 +168,24 @@ def standalone_example(example_path):
...
+    contributors = _filter_contributors(all_contributors)
+
+    response = flask.make_response(
+        flask.render_template(
+            "contribute.html",
+            team_members=team_members,
+            contributors=contributors,
+        )
+    )
+
```
jpmartinspt, line 184: I forgot `response.cache_control.public = True`

bartaz:
@sowasred2012 @jpmartinspt Does this PR have everything needed for it to run on staging/prod? Do the GitHub keys need to be added somewhere (like `/konf/site.yaml`)?

sowasred2012:
> @sowasred2012 @jpmartinspt Does this PR have everything needed for it to run on staging/prod? Do the GitHub keys need to be added somewhere (like `/konf/site.yaml`)?

Yes they do, once they've been setup on the demo/staging/prod clusters, I'll update the site.yaml and merge.

jpmartinspt (approved):

PR #3442, opened 2020-12-03
Include %vf-button-white-success-icon in base button mixin

### LINKED ISSUE #3438: The `%vf-button-white-success-icon` placeholder is not available when including individual buttons
As noted by @barrymcgee in https://github.com/canonical-web-and-design/cookie-policy/pull/105/files the `%vf-button-white-success-icon` placeholder breaks the SCSS build if project includes individual buttons (like `vf-button-positive`) without including whole `vf-p-buttons`.

This dependency is not very clear and not documented.

We should probably move the definition of `%vf-button-white-success-icon` to the `_base_buttons.scss`, so it's always available when including individual buttons.

sowasred2012:
## Done

Moved `%vf-button-white-success-icon` to the base button styles, so as not to break builds when including just Vanilla base.

Fixes #3438

### DIFF
--- a/scss/_patterns_buttons.scss
+++ b/scss/_patterns_buttons.scss
@@ -4,7 +4,6 @@ $hover-background-opacity-percentage: $hover-background-opacity-amount * 100%;
 
 // Default button styling
 @mixin vf-p-buttons {
-  @include vf-button-white-success-icon;
   @include vf-button-plain;
   @include vf-button-neutral;
   @include vf-button-brand;
@@ -17,10 +16,8 @@ $hover-background-opacity-percentage: $hover-background-opacity-amount * 100%;
 }
 
 @mixin vf-button-white-success-icon {
-  %vf-button-white-success-icon {
-    & .p-icon--success {
-      @include vf-icon-success($color-x-light, $color-transparent);
-    }
+  @include deprecate('3.0.0', 'vf-button-white-success-icon mixin will be removed, use @extend %vf-button-white-success-icon instead') {
+    //
   }
 }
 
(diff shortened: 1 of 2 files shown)


scss/_patterns_buttons.scss
```diff
@@ -16,14 +15,6 @@ $hover-background-opacity-percentage: $hover-background-opacity-amount * 100%;
   @include vf-button-icon;
 }
 
-@mixin vf-button-white-success-icon {
```
bartaz, line 19: I would leave the mixin empty, deprecate it, but keep it for the sake of not breaking the builds that may include it (like cookie policy now).

Something similar to this (but empty):
https://github.com/canonical-web-and-design/vanilla-framework/blob/387102aef5aea9b1aa5dd17f79b4853b4eb650af/scss/_patterns_grid.scss#L3-L9

sowasred2012: Ah yes - updating

scss/_patterns_buttons.scss
```diff
@@ -17,10 +16,8 @@ $hover-background-opacity-percentage: $hover-background-opacity-amount * 100%;
 }
 
 @mixin vf-button-white-success-icon {
-  %vf-button-white-success-icon {
-    & .p-icon--success {
-      @include vf-icon-success($color-x-light, $color-transparent);
-    }
+  @include deprecate('3.0.0', 'vf-button-white-success-icon mixin will be removed, use @extend %vf-button-white-success-icon instead') {
```
bartaz, line 19: Just a rewording suggestion
```suggestion
  @include deprecate('3.0.0', 'vf-button-white-success-icon mixin will be removed as it is not needed, %vf-button-white-success-icon is part of base styles') {
```

bartaz (approved):
Small suggestion, but LGTM

PR #3446, opened 2020-12-07
Add all missing icons as an "additional" set

### LINKED ISSUE #3039: Make new icons available
Now that there's a final set of icons, we need to make the icons that are not part of vanilla  available. Whether on npm as a separate package or some place else.

sowasred2012:
## Done

Added all icons present in [the SVG icons folder](https://drive.google.com/drive/u/0/folders/1vBymmk6lqtXbySG_4yIPGRgec1-TSMY6).

Also updated the import script to output pattern mixins.

Fixes #3039

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/docs/examples/standalone/patterns/icons/icons-additional, see that you are happy with how the icons look
- Visit http://0.0.0.0:8101/docs/patterns/icons#additional, see that the icons are present, and not broken. Suggest class name changes if any don't make sense.

## Screenshots

Screenshot from 2020-12-07 15-40-01


scss/docs/_patterns_icons.scss
```diff
@@ -46,3 +46,4 @@
 // so include them here for the purposes of
 // presenting the application icons on the docs page.
 @include vf-p-icons--application;
+@include vf-p-icons--additional;
```
bartaz, line 49: As we haven't released the `vf-p-icons--application` I'm tempted to remove it and allow only adding individual icons.

There is no way we can be sure what given product will need.

We could have `vf-p-icons--all` in case someone want them all, but I don't think we should encourage that, as the CSS will be huge.

sowasred2012: Fair point - in that case, how/where should I document them? Should they still have their own example?

bartaz: I think we should put all icons that are not included by default (so previous 'application' and 'additional' together in one example).

I'm not sure about putting them on the docs page each in separate little "card", as this will be a huge list. But on the other hand that seems to be the best way to quickly find what you need and grab class name without looking into source code.

scripts/convert-svgs-to-icon-mixins.js
```diff
@@ -8,39 +8,109 @@ const directory = args[2];
...
+  console.log('//------------------');
+
+  svgs.forEach((svg) => {
+    console.log(`@include vf-p-icon-${svg.name};`);
+  });
+
+  console.log('// Pattern mixins:');
+  console.log('//----------------');
+
+  svgs.forEach((svg) => {
```
bartaz, line 49: I'm a bit concerned about adding these pattern mixins here. I understand their usefulness in terms of just being able to copy them into the source, but they are very tied to the structure of SCSS code in Vanilla.
The base mixins are also Vanilla code, but practically it's just a `background-image` definition in a mixin. Nothing more. Pattern mixins need to be exactly what we have in pattern code in Vanilla, together with extends, correct mixin names, nested selectors...

If we happen to change anything in the icons code we would need to remember to change the script as well.

I don't know what's the best way to approach it. Maybe adding comments both to icons file, mentioning that these were generated by the script and to the script to warn the user in the future to verify if the printed code is correct.

scss/_base_icon-definitions.scss
```diff
@@ -172,3 +172,353 @@ $social-icon-size: map-get($icon-sizes, heading-icon--small);
...
+
+@mixin vf-icon-sort-up($color) {
+  background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill='#{vf-url-friendly-color($color)}'  fill-rule='nonzero' d='M8 2l3.333 5H4.667z'/%3E%3C/svg%3E");
+}
+
+@mixin vf-icon-starred($color) {
+  background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill='#{vf-url-friendly-color($color)}'  d='M8 13l-4.702 2.627.898-5.564-3.804-3.94 5.257-.811L8 .25l2.351 5.062 5.257.811-3.804 3.94.898 5.564z' fill-rule='evenodd'/%3E%3C/svg%3E");
+}
+
+@mixin vf-icon-status-failed-small($color) {
```
bartaz, line 422: I believe all `status-X-small` icons are exactly the same SVG with different color, they could use one definition mixin (`vf-icon-status-small()`) that would define the SVG, and this can be either used directly in pattern mixins (that would define different colors) or we can have icon specific mixins that just calls the SVG mixin with correct color:

```
@mixin vf-icon-status-failed-small($color: $correct-color-for-warning-icon) {
   @include vf-icon-status-small($color);
}
```

This would help remove a bit on unnecessary duplication of SVGs.

templates/docs/examples/patterns/icons/icons-additional.html
```diff
@@ -0,0 +1,190 @@
...
+  <i class="p-icon--timed-out-grey"></i>
+  <i class="p-icon--timed-out"></i>
+  <i class="p-icon--topic"></i>
+  <i class="p-icon--unit-pending"></i>
+  <i class="p-icon--unit-running"></i>
+  <i class="p-icon--unstarred"></i>
+  <i class="p-icon--user-group"></i>
+  <i class="p-icon--video-play"></i>
+  <i class="p-icon--warning-grey"></i>
+  <i class="p-icon--warning-grey"></i>
```
bartaz, line 187: Last one is duplicated.

lyubomir-popov:
You've probably seen this but one icon isn't tinted. Let me know if you need me to look at the svg file.

scripts/convert-svgs-to-icon-mixins.js
sowasred2012: I originally had just the `background-image` definition, and began adding them that way before quickly finding that adding all of these icons meant a lot of repetitive naming of mixins, and creating pattern mixins. Even with just a handful it's time consuming, so I think adding a disclaimer to the printed code is a reasonable solution.

scss/docs/_patterns_icons.scss
sowasred2012: Cool - I've updated so that there is no longer an application set. I haven't created an `vf-p-icons--all` set, I thought if it's something we'd rather people didn't do, then it's best not to give them the ability. What do you think?

bartaz: Yes, I agree that we shouldn't encourage it. Even if it would make our case simpler when we want to include them all for examples.

sowasred2012:
> You've probably seen this but one icon isn't tinted.

@lyubomir-popov Thanks - updated.

lyubomir-popov:
@sowasred2012 approved percy.

scripts/convert-svgs-to-icon-mixins.js
```diff
@@ -8,39 +8,111 @@ const directory = args[2];
...
+    console.log("   [class*='--dark'] &,");
+    console.log('   &.is-light {');
+    console.log(`     @include vf-icon-${svg.name}(${svg.colors.join(', ')});`);
+    console.log('   }');
+    console.log(' }');
+    console.log('}');
+    console.log('');
+  });
+
+  console.log('// **Base and Pattern mixins accurate as of December 2020**');
```
bartaz, line 64: Nice one thanks, may be worth adding Vanilla version for reference as well.

```suggestion
  console.log('// **Base and Pattern mixins accurate as of December 2020, Vanilla v2.21.0**');
```

scss/docs/_patterns_icons.scss
```diff
@@ -44,5 +44,98 @@
 
 // These are not included in Vanilla by default,
 // so include them here for the purposes of
-// presenting the application icons on the docs page.
-@include vf-p-icons--application;
+// presenting the additional icons on the docs page.
+@include vf-p-icon-applications;
```
bartaz, line 48: I'm not sure if we should do it here. While adding couple "application" icons wasn't a huge deal, now we are adding 100 icons to docs CSS, while there are needed on one single page.

Maybe it would be worth trying just adding a separate icons CSS to the icons page of the docs? We already have a standalone CSS with just the icons for the example, so we could reuse it and avoid adding all the icons to the docs CSS. Especially that we wanted to get rid of custom docs styling ideally.

scss/standalone/patterns_icons.scss
```diff
@@ -7,3 +7,100 @@
 
 @import '../patterns_icons';
 @include vf-p-icons;
+
+// these icons are not included in vf-p-icons, so include
```
bartaz, line 11: If we have separate file with additional icons `patterns_icons-additional` why do we need to add them here?
Or if we add them here - do we still need separate file with additional icons?

I guess maybe we need only one file (either one)?

sowasred2012: good point!

scss/docs/_patterns_icons.scss
sowasred2012: In that case - is there a way we can include CSS files on a specific doc page? In this case we're talking about a markdown doc, and I don't think there's front matter available for this purpose. I could drop a style tag in at the top of the document but that's a little dirty.

bartaz: Could we add it to `context` in frontmatter maybe?

```
context:
   title: Icons
   stylesheet: /path/to/styles.css
```

And then in `_root.html` layout check for `{{ stylesheet }}` and add it if needed?
I just quickly checked and it seems we can add custom stuff to `context`.

scss/_base_icon-definitions.scss
```diff
@@ -172,3 +172,337 @@ $social-icon-size: map-get($icon-sizes, heading-icon--small);
 @mixin vf-icon-pin($color) {
   background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M11.547.119l4.412 4.412-1.06 1.06-.366-.365-2.205 2.204-1.385 2.598 1.128 1.129-1.06 1.06-3.012-3.011-4.523 4.524-1.898.916.744-1.882 4.617-4.618-3.01-3.01 1.06-1.06 1.028 1.027 2.819-1.252 2.16-2.161-.51-.51 1.06-1.061zm.381 2.76L9.897 4.912a1.5 1.5 0 01-.329.248l-.123.062L7.16 6.237l2.678 2.678 1.167-2.19a1.5 1.5 0 01.164-.246l.1-.11 2.075-2.076-1.415-1.414z' fill='#{vf-url-friendly-color($color)}' fill-rule='nonzero'/%3E%3C/svg%3E");
 }
+
+// NEW
```
bartaz, line 176: I guess `// NEW` can quickly get outdated ;) Maybe just call them `// ADDITIONAL`?

```suggestion
// ADDITIONAL
// Generated with icon-svgs-to-mixins script
```

scss/_patterns_icons.scss
```diff
@@ -573,3 +563,1049 @@
     }
   }
 }
+
+// NEW
```
bartaz, line 567: Same suggestion as before

```suggestion
// ADDITIONAL
// Generated with icon-svgs-to-mixins script
```

scss/_base_icon-definitions.scss
sowasred2012: Very true

bartaz (approved):
LGTM

templates/_layouts/_root.html
```diff
@@ -42,6 +42,10 @@
 
     {% block custom_head %}
     {% endblock %}
+
+    {% if extra_stylesheet %}
+      <link rel="stylesheet" href="{{ stylesheet }}" />
```
bartaz, line 47:
```suggestion
      <link rel="stylesheet" href="{{ extra_stylesheet }}" />
```

PR #3447, opened 2020-12-09
Hide overflow when nav sidebar is collapsed.

sowasred2012:
## Done

Hides scrollbars on collapsed sidebar, on Windows/Ubuntu (the issue is not present on macOS).

Fixes https://github.com/canonical-web-and-design/vanilla-squad/issues/945

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/docs/examples/layouts/application/default
- Set the width of your browser window to the point where the sidebar collapses automatically but is still visible. See that there are no scrollbars present, and that they appear when you hover over it.

### DIFF
--- a/scss/_layouts_application.scss
+++ b/scss/_layouts_application.scss
@@ -85,6 +85,7 @@ $application-layout--side-nav-width-expanded: 15rem !default;
       );
 
       box-shadow: $panel-drop-shadow-transparent;
+      overflow: hidden;
       transform: translateX(0);
       width: $application-layout--side-nav-width-collapsed;
 
@@ -94,16 +95,20 @@ $application-layout--side-nav-width-expanded: 15rem !default;
       }
     }
 
+    .l-navigation:hover,
+    .l-navigation.is-pinned {
+      overflow-y: auto;
+      width: $application-layout--side-nav-width-expanded;
+    }
+
     .l-navigation:hover {
       box-shadow: $panel-drop-shadow;
-      width: $application-layout--side-nav-width-expanded;
     }
 
     .l-navigation.is-pinned {
       box-shadow: $panel-drop-shadow-transparent;
       grid-area: nav;
       position: static;
-      width: $application-layout--side-nav-width-expanded;
     }
   }
 
@@ -116,6 +121,7 @@ $application-layout--side-nav-width-expanded: 15rem !default;
     .l-navigation {
       box-shadow: none;
       grid-area: nav;
+      overflow-y: auto;
       position: static;
       width: $application-layout--side-nav-width-expanded;
 


scss/_layouts_application.scss
```diff
@@ -89,19 +89,22 @@ $application-layout--side-nav-width-expanded: 15rem !default;
...
       &.is-collapsed {
+        overflow: hidden;
         transform: translateX(0);
         width: $application-layout--side-nav-width-collapsed;
       }
     }
 
     .l-navigation:hover {
       box-shadow: $panel-drop-shadow;
+      overflow-y: scroll;
```
bartaz, line 100: @sowasred2012 We should use overflow `auto` not to force the scrollbar when it's not needed.

```suggestion
      overflow-y: auto;
```

scss/_layouts_application.scss
```diff
@@ -89,19 +89,22 @@ $application-layout--side-nav-width-expanded: 15rem !default;
...
       width: $application-layout--side-nav-width-expanded;
     }
 
     .l-navigation.is-pinned {
       box-shadow: $panel-drop-shadow-transparent;
       grid-area: nav;
+      overflow-y: scroll;
```
bartaz, line 107:
```suggestion
      overflow-y: auto;
```

bartaz (changes requested):
We should use overflow `auto`

bartaz:
When the scrollbar appears it makes the right side of side navigation covered:

Screenshot 2020-12-09 at 15 37 08

It's because the scrollbar is on `l-navigation` (which has 15rem width) and `l-navigation__drawer` inside it has also width fixed to 15rem, so scrollbar doesn't make it shrink.
We can try puttin `max-width: 100%` on `l-navigation__drawer` to make sure it only takes available width (doesn't flow under the scrollbar). But it needs to happen only in expanded navigation (not in collapsed one).

scss/_layouts_application.scss
```diff
@@ -94,16 +95,28 @@ $application-layout--side-nav-width-expanded: 15rem !default;
...
+      max-width: $application-layout--side-nav-width-expanded;
+    }
+
+    .l-navigation:hover,
+    .l-navigation.is-pinned {
+      overflow-y: auto;
+      width: $application-layout--side-nav-width-expanded;
+
+      .l-navigation__drawer {
+        max-width: 100%;
```
bartaz, line 108: @sowasred2012 You probably were right about the transition needing to be only on one direction. This seems to work for me. A bit hard to verify it via browserstack, but seems fine on windows as well. Can you check?

```suggestion
        max-width: 100%;
        transition: max-width 0ms map-get($animation-duration, fast);
```

sowasred2012: Yeah that fixes the transition back we were seeing, the only nagging thing is the slight layout shift when it opens:
navbar

bartaz (approved):
LGTM, thanks!

PR #3450, opened 2020-12-10
Expand sidenav when :focus-within is true.

### LINKED ISSUE #3377: Application layout is not keyboard accessible
**Describe the bug**

The application layout sidebar does not open if a user navigates links within with a keyboard

**To Reproduce**

1. Go to https://codepen.io/barrymcgee/full/RwRpMpj
2. Tab into application layout example
3. See that links are traversed but sidebar does not expand

**Expected behavior**

When I traverse links, I would expect sidebar to open

sowasred2012:
## Done

Added :focus-within support to the expanding sidenav.

Fixes #3377 

## QA

- Pull code
- Run `./run`
- Open http://0.0.0.0:8101/docs/examples/layouts/application/JAAS
- Reduce your browser width until the sidenav automatically collapses, then tab through the page until you see a focus state within the sidenav. The sidenav should expand.


bartaz (approved):
LGTM, thanks!

PR #3454, opened 2020-12-11
Removed redundant MAAS docs example

### LINKED ISSUE #3452: Remove l-docs from mass-docs example
We either need to remove the maas-docs example now that we have supported docs layout examples. Or refactor the template to use the docs layout [instead of `l-docs`](https://github.com/canonical-web-and-design/vanilla-framework/blob/master/templates/docs/examples/templates/maas-docs.html#L10).

sowasred2012:
## Done

Removed MAAS custom docs example, since MAAS is now using the docs layout.

Fixes #3452

## QA

- Visit the [demo]()
- See that there is no link to "MAAS docs / Custom docs layout" (compare to [live Vanilla site](https://vanillaframework.io/docs/examples))
- Try to visit /docs/examples/maas-docs and see that it 404s.


anthonydillon (approved):
LGTM :+1:

PR #3457, opened 2020-12-14
If a standalone template does not exist, return a 404 rather than a 500.

sowasred2012:
## Done

Return a 404 if a standalone template doesn't exist.

## QA

- Open [demo](https://vanilla-framework-3457.demos.haus/docs/examples/standalone/base/code)
- See that you see a 404 page, and not a 500 (compare to [live site](https://vanillaframework.io/docs/examples/standalone/base/code))

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -7,6 +7,7 @@ import random
 # Packages
 import requests
 import flask
+import jinja2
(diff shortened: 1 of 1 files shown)


jpmartinspt (approved):
What a wizard! :mage_man:

PR #3461, opened 2021-01-04
Change %vf-hide-text text indent to IE friendly value

### LINKED ISSUE #3162: In IE text in an icon is partially visible.
## Pattern to amend
`%vf-hide-text`

## Visual
Screenshot 2020-07-09 at 10 13 57

## Context
In IE text in an icon is partially visible. This is an issue with `%vf-hide-text`. IE does not support transitioning values set with `calc()`  therefore a solution could be to switch to using `2vw`. I will tackle that as a separate issue

sowasred2012:
## Done

Updated `text-indent` value on `%vf-hide-text` to `110vw`, from a `calc()` value incompatible with IE.

Fixes #3162 

## QA

- Login to https://www.browserstack.com/
- Within Browserstack, select IE11 and open [demo](https://vanilla-framework-3461.demos.haus)
- Visit https://vanilla-framework-3461.demos.haus/docs/examples/patterns/chip/with-dismiss
- See that the "X" icon is visible, and not obscured by text.

## Screenshots

Screenshot from 2021-01-04 10-47-25

### DIFF
--- a/scss/_base_placeholders.scss
+++ b/scss/_base_placeholders.scss
@@ -116,10 +116,11 @@
   // Utilities
   %vf-hide-text {
     overflow: hidden;
-    // rem value necessary because text indent in % does not work with padding;
-    // 10rem is larger than any padding value on an element you'd hide text in (buttons, icons)
-    // but still smaller than 9999px so should have better performance
-    text-indent: calc(100% + 10rem);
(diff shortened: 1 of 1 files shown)


bartaz (approved):
LGTM, thanks!

PR #3474, opened 2021-01-05
Code snippet pattern

sowasred2012:
## Done

Added new `p-code-snippet` pattern, with two child components: `p-code-snippet__header` and `p-code-snippet__block`.

Fixes #3474 

## QA

- Open [demo](https://vanilla-framework-3474.demos.haus/docs/examples/patterns/code-snippet/code-snippet)
- See that the design matches the one in Zeplin (though without a dropdown, which is coming in a separate PR for #3469): https://app.zeplin.io/project/5fa270d8eee4dbb267e579f1/screen/5fc616c95320fdb32885432e

### DIFF
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -49,12 +49,6 @@
       }
     }
 
-    &.is-dense {
-      margin-bottom: $spv-nudge-compensation;
-      padding-bottom: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
-      padding-top: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
-    }
-
     &.is-small {
       font-size: #{map-get($font-sizes, small)}rem;
       line-height: map-get($line-heights, small);
@@ -62,7 +56,7 @@
       padding: calc(#{map-get($nudges, nudge--small)} - 1px) $sph-inner--small;
     }
 
-    &.is-small.is-dense {
+    &.is-dense {
       margin-bottom: $spv-nudge-compensation;
       padding-bottom: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
       padding-top: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -49,6 +49,12 @@
       }
     }
 
+    &.is-dense {
+      margin-bottom: $spv-nudge-compensation;
+      padding-bottom: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
+      padding-top: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
+    }
+
     &.is-small {
       font-size: #{map-get($font-sizes, small)}rem;
       line-height: map-get($line-heights, small);
@@ -56,7 +62,7 @@
       padding: calc(#{map-get($nudges, nudge--small)} - 1px) $sph-inner--small;
     }
 
-    &.is-dense {
+    &.is-small.is-dense {
       margin-bottom: $spv-nudge-compensation;
       padding-bottom: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
       padding-top: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -59,7 +59,7 @@
     margin-bottom: $spv-outer--scaleable;
     margin-top: 0;
     overflow: auto;
-    padding: map-get($nudges, nudge--p-ubuntumono) $sph-inner;
+    padding: $spv-inner--x-small--scaleable $sph-inner;
     text-align: left;
     text-shadow: none;
     white-space: pre;
--- a/scss/_patterns_code-numbered.scss
+++ b/scss/_patterns_code-numbered.scss
@@ -11,6 +11,7 @@ $sidebar-width: 4.5rem !default;
     display: inline-block;
     padding: $spv-inner--small $sph-inner 0 ($sidebar-width + $sph-inner);
     position: relative;
+    white-space: normal;
     width: 100%;
 
     &:empty {
--- a/scss/_patterns_code-numbered.scss
+++ b/scss/_patterns_code-numbered.scss
@@ -11,7 +11,6 @@ $sidebar-width: 4.5rem !default;
     display: inline-block;
     padding: $spv-inner--small $sph-inner 0 ($sidebar-width + $sph-inner);
     position: relative;
-    white-space: pre-wrap;
     width: 100%;
 
     &:empty {
--- a/scss/_patterns_code-numbered.scss
+++ b/scss/_patterns_code-numbered.scss
@@ -11,7 +11,7 @@ $sidebar-width: 4.5rem !default;
     display: inline-block;
     padding: $spv-inner--small $sph-inner 0 ($sidebar-width + $sph-inner);
     position: relative;
-    white-space: normal;
+    white-space: pre-wrap;
     width: 100%;
 
     &:empty {
--- a/scss/_patterns_code-snippet.scss
+++ b/scss/_patterns_code-snippet.scss
@@ -2,26 +2,10 @@
 
 @mixin vf-p-code-snippet {
   .p-code-snippet {
-    .p-code-snippet__block,
-    .p-code-snippet__block--numbered {
+    .p-code-snippet__block {
       border: 0;
       border-radius: 0;
       margin: 0;
-    }
-
-    .p-code-snippet__block--numbered {
-      counter-reset: line-numbering;
-      padding: $sph-inner--small 0;
-    }
-
-    .p-code-snippet__line {
-      padding-top: 0;
-
-      &::before {
-        background: transparent;
-        border: 0;
-        padding-top: 0;
-      }
     }
 
     .p-code-snippet__header {
@@ -36,25 +20,6 @@
 
       margin-bottom: 0;
       padding-top: 0;
-    }
-
-    .p-code-snippet__dropdowns {
-      flex-shrink: 0;
-      margin-left: $sph-inner--small;
-    }
-
-    .p-code-snippet__toggle {
-      // this comes from the button pattern, but we want to
-      // move it to the base button styles
-      @extend %p-button--base;
-
-      // taken from the .u-table-cell-padding-overlap utility
-      margin-bottom: -#{$table-cell-vertical-padding} !important;
-      margin-top: calc(-#{$table-cell-vertical-padding} - 1px) !important;
-
-      // taken from the .is-dense button modifier
-      padding-bottom: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
-      padding-top: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
     }
   }
 }
--- a/scss/_patterns_code-snippet.scss
+++ b/scss/_patterns_code-snippet.scss
@@ -34,7 +34,9 @@
     .p-code-snippet__title {
       @extend %vf-heading-5;
 
-      margin-bottom: $sph-inner--small;
+      margin-bottom: 0;
+      padding-bottom: $spv-inner--x-small--scaleable;
+      padding-top: $spv-inner--x-small--scaleable;
     }
 
     .p-code-snippet__dropdowns {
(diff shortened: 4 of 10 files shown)


lyubomir-popov:
@sowasred2012 great improvement, thanks!

A couple of things:



- are we changing the default from overflow-x to wrap compared to the current code block

- the contextual menu overflows. can it be right-aligned?

image

sowasred2012:
@lyubomir-popov good shout on menu alignment, updating. Re:

> are we changing the default from overflow-x to wrap compared to the current code block

We spoke about this briefly last year when discussing the code snippet work; if a user has to significantly enlarge the contents of the page for any reason (e.g. their vision is impaired), horizontal scrolling inside an element contained within a page that may or may not also horizontally scroll at that zoom level makes it easier for the user to lose their place on the page and become frustrated.

Also, we currently have [this example](https://vanillaframework.io/docs/examples/patterns/code-numbered), which suggests at least the numbered pattern should support line wrapping (though it currently doesn't and, in the example, the lines below line 4 should all be numbered lines too).

templates/docs/base/code.md
```diff
@@ -18,20 +18,19 @@ When you refer to code inline with other text, use the <code>&lt;code></code> ta
 View example of inline code
 </a></div>
 
-### Block
```
bartaz, line 21: I don't think we should remove the base style documentation and examples.

The documentation and examples of using `<pre>` and `<code>` tags should remain, as we are still going to support it.

We should add `Code snippets` separately as new component.

templates/docs/examples/base/code-block.html
```diff
@@ -4,9 +4,9 @@
 {% block standalone_css %}base{% endblock %}
 
 {% block content %}
-<pre>
-<code>function() {
+<div class="p-code-snippet">
```
bartaz, line 7: This example should stay as it is, because it's example of base styles, not snippet component.

templates/docs/examples/patterns/code-snippet/code-snippet.html
```diff
@@ -0,0 +1,29 @@
...
+  
+  <div class="p-code-snippet__header">
+    <h5 class="p-code-snippet__title">Pre-formatted example</h5>
+  </div>
+
+  <pre class="p-code-snippet__block">This is
+    pre-formatted text.</pre>  
+</div>
+
+<script>
```
bartaz, line 26: This is not needed here, as we don't have dropdowns yet.

templates/docs/examples/patterns/code-snippet/code-snippet.html
```diff
@@ -0,0 +1,29 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Code Snippet{% endblock %}
+
+{% block content %}
+<div class="p-code-snippet">
+  <div class="p-code-snippet__header">
+    <h5 class="p-code-snippet__title">This is the code snippet title</h5>
```
bartaz, line 7: Would be good to have some more 'real' content for this (as we try in most examples).
Maybe some docs or tutorials would have something for "command" + "output" format.

templates/docs/examples/patterns/code-snippet/code-snippet.html
```diff
@@ -0,0 +1,29 @@
...
+
+  <pre class="p-code-snippet__block"><code>something.test();</code></pre>
+
+  <div class="p-code-snippet__header">
+    <h5 class="p-code-snippet__title">Output</h5>
+  </div>
+
+  <pre class="p-code-snippet__block"><code>Test passed!</code></pre> 
+  
+  <div class="p-code-snippet__header">
```
bartaz, line 18: While it is possible to have multiple blocks I think the intended structure is max 2 (code + output), so we shouldn't put more to our examples.

If we want to show some separate use, it should be separate code snippet (in same example) or separate example.

templates/docs/examples/patterns/code-snippet/code-snippet.html
```diff
@@ -0,0 +1,29 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Code Snippet{% endblock %}
+
```
bartaz, line 3: It's the time to add standalone stylesheet for code snippets, as standalone examples for this component are rendered unstyled https://vanilla-framework-3474.demos.haus/docs/examples/standalone/patterns/code-snippet/code-snippet.

scss/_patterns_code-numbered.scss
```diff
@@ -11,6 +11,7 @@ $sidebar-width: 4.5rem !default;
     display: inline-block;
     padding: $spv-inner--small $sph-inner 0 ($sidebar-width + $sph-inner);
     position: relative;
+    white-space: pre-wrap;
```
bartaz, line 14: Do we want to wrap by default?

Even if we do, this PR doesn't seem to be about updating old components, so it shouldn't be part of it.

scss/_base_code.scss
```diff
@@ -59,7 +59,7 @@
     margin-bottom: $spv-outer--scaleable;
     margin-top: 0;
     overflow: auto;
-    padding: map-get($nudges, nudge--p-ubuntumono) $sph-inner;
+    padding: $spv-inner--x-small--scaleable $sph-inner;
```
bartaz, line 62: This changes the base styles (that we haven't updated to new style yet), is this intended here?

bartaz (changes requested):
Couple comments on code and docs

scss/_base_code.scss
sowasred2012: Yes, this one is, @lyubomir-popov and I worked on this yesterday, it's a negligible change for the base style, but means the new snippet code blocks stay on the baseline.

scss/_patterns_code-numbered.scss
sowasred2012: I think we should, for reasons mentioned [above](https://github.com/canonical-web-and-design/vanilla-framework/pull/3474#issuecomment-756071065), but you're right, I'll move it out of this PR.

bartaz: Ok - in this case the standard (not numbered) code block should do this as well by default (if it doesn't).

Also, my small concern is that we have a 'feature' (wrapping) as a default, so now, we will need to have a 'negative' modifier (`is-not-wrapping` ?) to disable it. This can of course be worked around with some good naming.

templates/docs/base/code.md
```diff
@@ -34,6 +34,21 @@ View example of the base pre block
...
+### Code Snippets
+
+A code snippet is composed of one or more related components, contained within a `.p-code-snippet` wrapper. Available components are:
+
+<!-- more to come after this initial example: dropdowns, block--numbered, command prompt etc -->
+
+- `.p-code-snippet__header`, which should briefly describe the content it precedes.
+- `.p-code-snippet__block`, a class added to a `<pre>` element containing your pre-formatted content. If the content is code, use `<pre><code>`.
+
+**Code snippet:**
```
bartaz, line 46: We don't do 'titles' of embeds in other places. I think we can just remove it.

scss/_patterns_code-snippet.scss
```diff
@@ -0,0 +1,25 @@
...
+@mixin vf-p-code-snippet {
+  .p-code-snippet {
+    .p-code-snippet__block {
+      border: 0;
+      border-radius: 0;
+      margin: 0;
+    }
+
+    .p-code-snippet__header {
+      background-color: rgba(0, 0, 0, 0.08);
```
bartaz, line 12: @lyubomir-popov Is the code snippet header background one of our existing colors? Or should we add new one?

This should use a color variable, not inline value.

templates/docs/base/code.md
```diff
@@ -34,6 +34,21 @@ View example of the base pre block
 View example of the base code block
 </a></div>
 
+### Code Snippets
+
+A code snippet is composed of one or more related components, contained within a `.p-code-snippet` wrapper. Available components are:
```
bartaz, line 39: I feel like this sentence makes it feel like code snippet is much more modular than intended.

I wouldn't say 'related components', as 'components' is how we call our higher level building blocks - so 'Code snippet' is a component.
Also it's not intended to just use headers freely. They are optional, but they are meant to be always used with code blocks.

So maybe something along the lines of "The code snippet is used to give more context to code blocks represented on a page. It consist of one or two code blocks with optional headers for their titles."

bartaz: @lyubomir-popov Can you find latest Code snippet spec, so we can make sure our docs reflect the intended uses of it?

scss/_base_button.scss
```diff
@@ -49,20 +49,14 @@
...
       font-size: #{map-get($font-sizes, small)}rem;
       line-height: map-get($line-heights, small);
       margin: 0 0 $input-margin-bottom - $sp-unit 0;
       padding: calc(#{map-get($nudges, nudge--small)} - 1px) $sph-inner--small;
     }
 
-    &.is-small.is-dense {
+    &.is-dense {
       margin-bottom: $spv-nudge-compensation;
       padding-bottom: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
```
lyubomir-popov, line 61: spv-nudge is the nudge on defualt text size. on the small text size it is different (`map-get($nudges, nudge--small)`). I have a feeling the problem with the existing code is that the two have been wrong, it's not a simple case of duplication.

lyubomir-popov (commented):
@bartaz in terms of colour, it should be $colors--light-theme--background-alt but expressed as opacity.

We have a problem - since a decision was made to treat the light-grey as part of the white theme, the only way to make sure this darker background is not lost on strips coloured #f7f7f7f is to use opacity.

However, [that particular colour ](https://github.com/canonical-web-and-design/vanilla-framework/blob/074906bc0c797a6325e9c034be18829b21b0ea99/scss/_settings_colors.scss#L78) is expressed as #f7f7f7, so we either need to convert the darkeness to opacity (e.g. `rgba($color-x-dark, .03)`), or add another variable with this value.


IMO he correct solution is to treat f7f7f7 as a light-grey theme and have it completely separate, so we don't have this kind of in-theme branching...

templates/docs/base/code.md
lyubomir-popov: @bartaz The latest I have is a mention in an email from @wgx - https://miro.com/app/board/o9J_kqQPDak=/?moveToWidget=3074457349016924433&cot=12

Added that to the [brief](https://docs.google.com/document/d/1HwKSetstIGUwZokhaZO6kJDLb1FRDRNQn2tK5td24oI/edit#) as the existing miro boardthere seems seems much older. @wgx Can you confirm this is the latest?

scss/_base_button.scss
bartaz: Ok, but if we want to fix it, we should do it separately, this was just removing duplication.
Changing this may affect other things, and we don't want that in this PR.

lyubomir-popov: jjust want to be careful about removing selectors that most probably should exist.

bartaz: @lyubomir-popov ah, fair point

@sowasred2012 Maybe let's not do it now, seems like Lyubo may be right. While these sections (`.is-dense` and `.is-small.is-dense`) are the same now, they probably shouldn't (because `is-small` uses different paddings (values of nudges).
So let's keep it as it was for now, and maybe add separate issue to check what padding bottom/top should `is-small.is-dense` have.

scss/_patterns_code-snippet.scss
```diff
@@ -1,15 +1,19 @@
 @import 'settings';
 
+$snippet-heading-bg-color: rgba($color-x-dark, 0.08);
```
bartaz, line 3: A bit nit picking, but our color variables usually start with `$color-`, so may be worth for consistency.

sowasred2012: Updated

templates/docs/base/code.md
```diff
@@ -34,6 +34,14 @@ View example of the base pre block
 View example of the base code block
 </a></div>
 
+### Code Snippets
+
+The code snippet pattern should be used to group related code examples, making them easier to find and understand. It should consist of one or, at most, two code blocks with optional headers for their titles.
```
bartaz, line 39: Nice, thanks. We will extend that when we add more features to the component.

bartaz (approved):

templates/docs/base/code.md
wgx: @lyubomir-popov The latest is here: https://miro.com/app/board/o9J_kqQPDak=/?moveToWidget=3074457351246118802&cot=14 and in the public spec here: https://discourse.ubuntu.com/t/code-snippet-pattern/19195

PR #3479, opened 2021-01-07
Remove unnecessary pseudo-element from tables

### LINKED ISSUE #3087: Slight gap between table header and body
There is a slight gap between the table head and body due to an empty `::after` on the thead element. I am not sure that its there for but causing a slight gap only on first render but when you resize the browser sorts it out:

https://github.com/canonical-web-and-design/vanilla-framework/blob/e3e4a919a6df82c5b317db3225096047aff4ad5c/scss/_base_tables.scss#L33-L36

before resize | after resize
-- | --
 | 

Proposed solution:
https://github.com/canonical-web-and-design/jaas-dashboard/pull/503/files

bartaz:
## Done

Removes unnecessary pseudo-element that is causing rendering issues sometimes.

Fixes #3087 

## QA

- Open [demo](https://vanilla-framework-3479.demos.haus/docs/examples)
- Check table examples (and Percy), make sure there are no regressions

### DIFF
--- a/scss/_base_tables.scss
+++ b/scss/_base_tables.scss
@@ -31,10 +31,6 @@
   }
 
   thead {
-    &::after {
-      content: '';
-    }
-
     th {
       @extend %table-header-label;
 


sowasred2012:
LGTM

sowasred2012 (approved):

PR #3484, opened 2021-01-08
Update code copyable style to match code snippet

### LINKED ISSUE #3464: Update code copyable styles

sowasred2012:
## Done

Updated the code copyable to match the code snippet [design](https://app.zeplin.io/project/5fa270d8eee4dbb267e579f1/screen/5fc616c95320fdb32885432e). Notably, this update hides the "copy" button if one is present, without requiring any markup changes. 

Fixes #3464 

## QA

- Open [demo](https://vanilla-framework-3484.demos.haus/docs/examples/patterns/code-copyable)
- See that the pattern matches the design

### DIFF
--- a/scss/_base_icon-definitions.scss
+++ b/scss/_base_icon-definitions.scss
@@ -534,3 +534,16 @@ $social-icon-size: map-get($icon-sizes, heading-icon--small);
 
   display: inline-block;
 }
+
+%leading-linux-prompt-icon {
+  &::before {
+    @extend %icon;
+    @include vf-icon-linux-prompt($color-mid-dark);
+
+    content: '';
+    left: $sph-inner;
+    position: absolute;
+    top: #{$spv-inner--x-small--scaleable + 0.5 * (map-get($line-heights, default-text) - map-get($icon-sizes, default))};
+    width: map-get($icon-sizes, default);
+  }
+}
--- a/scss/_base_icon-definitions.scss
+++ b/scss/_base_icon-definitions.scss
@@ -534,16 +534,3 @@ $social-icon-size: map-get($icon-sizes, heading-icon--small);
 
   display: inline-block;
 }
-
-%leading-linux-prompt-icon {
-  &::before {
-    @extend %icon;
-    @include vf-icon-linux-prompt($color-mid-dark);
-
-    content: '';
-    left: $sph-inner;
-    position: absolute;
-    top: #{$spv-inner--x-small--scaleable + 0.5 * (map-get($line-heights, default-text) - map-get($icon-sizes, default))};
-    width: map-get($icon-sizes, default);
-  }
-}
--- a/scss/_patterns_code-copyable.scss
+++ b/scss/_patterns_code-copyable.scss
@@ -28,6 +28,11 @@ $cc-button-size: 36px;
   }
 
   .p-code-copyable__action {
+    // This pattern is being deprecated in 2.22, but we
+    // want a consistent look among code block patterns, so
+    // code-copyable has been updated to look like a standard
+    // code block. For that reason, we're hiding the "copy"
+    // button on existing implementations of this pattern.
     display: none;
   }
 }
--- a/scss/_patterns_code-copyable.scss
+++ b/scss/_patterns_code-copyable.scss
@@ -3,36 +3,38 @@
 $cc-button-size: 36px;
 
 @mixin vf-p-code-copyable {
-  .p-code-copyable {
-    background-color: $color-code-background;
-    display: flex;
-    margin-bottom: $input-margin-bottom;
-    overflow: hidden;
-    padding: $spv-inner--x-small--scaleable $sph-inner;
-    position: relative;
+  @include deprecate('3.0.0', 'p-code-copyable is being removed, use the code snippet block instead.') {
+    .p-code-copyable {
+      background-color: $color-code-background;
+      display: flex;
+      margin-bottom: $input-margin-bottom;
+      overflow: hidden;
+      padding: $spv-inner--x-small--scaleable $sph-inner;
+      position: relative;
 
-    & + & {
-      margin-top: 0; // overrides p + p
+      & + & {
+        margin-top: 0; // overrides p + p
+      }
     }
-  }
 
-  .p-code-copyable__input {
-    background: transparent;
-    border: 0;
-    font-family: unquote($font-monospace);
-    line-height: map-get($line-heights, default-text);
-    margin-bottom: 0;
-    margin-top: 0;
-    padding: 0;
-    width: 100%;
-  }
+    .p-code-copyable__input {
+      background: transparent;
+      border: 0;
+      font-family: unquote($font-monospace);
+      line-height: map-get($line-heights, default-text);
+      margin-bottom: 0;
+      margin-top: 0;
+      padding: 0;
+      width: 100%;
+    }
 
-  .p-code-copyable__action {
-    // This pattern is being deprecated in 2.22, but we
-    // want a consistent look among code block patterns, so
-    // code-copyable has been updated to look like a standard
-    // code block. For that reason, we're hiding the "copy"
-    // button on existing implementations of this pattern.
-    display: none;
+    .p-code-copyable__action {
+      // This pattern is being deprecated in 2.22, but we
+      // want a consistent look among code block patterns, so
+      // code-copyable has been updated to look like a standard
+      // code block. For that reason, we're hiding the "copy"
+      // button on existing implementations of this pattern.
+      display: none;
+    }
   }
 }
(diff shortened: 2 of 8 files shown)


lyubomir-popov (approved):
LGTM

scss/_settings_colors.scss
```diff
@@ -39,6 +39,7 @@ $active-background-opacity-amount: 0.15;
 
 // NON-SEMANTIC COLOURS
 $color-label-validated: #006b75;
+$color-code-background: rgba($color-x-dark, 0.03);
```
bartaz, line 42: Maybe we should move the other one (code snippet header background) here as well?

scss/_patterns_code-copyable.scss
```diff
@@ -4,58 +4,30 @@ $cc-button-size: 36px;
...
-    border: 0;
-    border-left: $border;
-    border-radius: 0;
-    line-height: map-get($line-heights, default-text);
-    margin: 0;
-    position: absolute;
-    right: 0;
-    top: 0;
-    width: $cc-button-size;
+    display: none;
```
bartaz, line 31: We should add a code comment here explaining why it's hidden.

bartaz (changes requested):
Some small comments on the code.
But I think that if we are hiding the copy button we should explicitly make this component deprecated and document how to use code snippet as a replacement.

lyubomir-popov:
> But I think that if we are hiding the copy button we should explicitly make this component deprecated and document how to use code snippet as a replacement.

Isn't the goal to update the component everywhere it is currently used?

bartaz:
> Isn't the goal to update the component everywhere it is currently used?

Yes, and this will happen with this change. But as we are getting rid of 'copy' functionality of  'code copyable' component it seems like 'code copyable' doesn't make sense as separate component. This will simply be a code snippet.

lyubomir-popov:
Got it, makes sense.

bartaz:
@sowasred2012 @lyubomir-popov 

Code copyable used to have a prompt icon:

Screenshot 2021-01-11 at 11 35 43

Now it doesn't - is that intended?

Screenshot 2021-01-11 at 11 34 40

sowasred2012:
> @sowasred2012 @lyubomir-popov
> 
> Code copyable used to have a prompt icon:
> Now it doesn't - is that intended?

I had assumed so, based on the design - we also have the issue to implement command prompts in the code snippet (https://github.com/canonical-web-and-design/vanilla-framework/issues/3472), if we want it here too, maybe it's better to resolve that issue first, and perhaps introduce a placeholder there that can be included here.

lyubomir-popov:
@bartaz @sowasred2012 sorry I've missed the prompt icon - it should be there, it is in the wireframes. We recently discovered that other icons might be needed, like the link icon  - @anasereijo had it in a cube design. So the icon definitely nees to be there.

bartaz:
@sowasred2012 Was this updated? I don't see prompt in code copyable example.

templates/docs/component-status.md
```diff
@@ -34,6 +34,28 @@ When we add, make significant updates, or deprecate a component we update their
...
+    <tr>
+      <th><a href="/docs/examples/patterns/code-copyable">Code Copyable</a></th>
+      <td><div class="p-label--deprecated">Deprecated</div></td>
+      <td>2.22.0</td>
+      <td><code>.p-code-copyable</code> has been deprecated and will be removed in the v3.0 release. Please use <code>.p-code-snippet .p-code-snippet__block</code> instead.</td>
+    </tr>
+  </tbody>
+</table>
+
+#### Previously in Vanilla
```
bartaz, line 57: This heading is duplicated. And I guess the contents of the table about are supposed to be in "What's new" section, not in "Previously".

templates/docs/base/code.md
```diff
@@ -54,26 +54,11 @@ View example of the code snippet
...
-<div class="embedded-example"><a href="/docs/examples/patterns/code-copyable/" class="js-example">
-View example of the code copyable pattern
-</a></div>
+<span class="p-label--deprecated">Deprecated</span>
+
+The code copyable element is now deprecated and will be removed in a future version of Vanilla.
+
+It is being replaced by the new [code snippet pattern](/docs/base/code#code-snippets).
 
 ### Functionality
```
bartaz, line 63: While we are here, this seems to be a sub-section of 'Copyable', so should be h4 I guess.

```suggestion
#### Functionality
```

templates/_layouts/docs.html
```diff
@@ -47,7 +47,7 @@
 
             <ul class="p-side-navigation__list">
               <li class="p-side-navigation__item--title"><span class="p-side-navigation__text">Base elements</span></li>
-              {{ side_nav_item("/docs/base/code", "Code") }}
+              {{ side_nav_item("/docs/base/code", "Code", "new") }}
```
bartaz, line 50: I tend to put "updated" label in such cases.

While we add a new component, the page itself is not new.

So, I usually mark page as "updated" and in the page content I put "new" label under the section that was added.

Not sure if that's a best approach (because it doesn't it doesn't make it clear in navigation which sections have new stuff and which are just updated), but also doesn't suggest that "Code" is new thing, when it's not.

templates/docs/base/code.md
sowasred2012: Since we're deprecating this pattern, I should remove this section right? I overlooked it previously, but "the example" it mentions is no longer linked to.

scss/_patterns_code-copyable.scss
```diff
@@ -3,59 +3,49 @@
...
+
+      & + & {
+        margin-top: 0; // overrides p + p
+      }
 
-    display: flex;
-    margin-bottom: $input-margin-bottom;
-    overflow: hidden;
-    position: relative;
+      &::before {
```
bartaz, line 19: I feel that we should approach this similarly to code numbered and extract it to common base styles.

Any tweak to icon positioning or something will have to be applied in both places, we need to share these styles as much as possible so they are consistent and up to date in both places.

scss/_patterns_code-copyable.scss
```diff
@@ -3,59 +3,49 @@
...
 @mixin vf-p-code-copyable {
-  .p-code-copyable {
-    @extend %bordered-text-vertical-padding;
-    @extend %vf-bg--x-light;
-    @extend %vf-has-round-corners;
-    @extend %vf-is-bordered;
+  @include deprecate('3.0.0', 'p-code-copyable is being removed, use the code snippet block instead.') {
+    .p-code-copyable {
+      background-color: $color-code-background;
+      display: flex;
```
bartaz, line 9: Is display flex still needed if button is hidden?

sowasred2012: No, good spot.

scss/_patterns_code-copyable.scss
sowasred2012: @bartaz updated

scss/_base_icon-definitions.scss
```diff
@@ -534,3 +534,16 @@ $social-icon-size: map-get($icon-sizes, heading-icon--small);
 
   display: inline-block;
 }
+
+%leading-linux-prompt-icon {
```
bartaz, line 538: It's quite specific to the code components, so I'd expect to find in with base code styles I think.

scss/_patterns_code-copyable.scss
```diff
@@ -3,59 +3,39 @@
...
-
-    display: flex;
-    margin-bottom: $input-margin-bottom;
-    overflow: hidden;
-    position: relative;
+  @include deprecate('3.0.0', 'p-code-copyable is being removed, use the code snippet block instead.') {
+    .p-code-copyable {
+      @extend %leading-linux-prompt-icon;
+
+      background-color: $color-code-background;
```
bartaz, line 10: I wonder if we shouldn't extract the `pre` styling to extend it here.

scss/_patterns_code-copyable.scss
```diff
@@ -3,59 +3,39 @@
...
+      margin-bottom: $input-margin-bottom;
+      overflow: hidden;
+      padding: $spv-inner--x-small--scaleable $sph-inner;
+      position: relative;
 
-    & + & {
-      margin-top: 0; // overrides p + p
+      & + & {
```
bartaz, line 16: Is this doing anything? (I know it was here before). It seems to reset top margin, which doesn't seem to be set anywhere. And the code comment itself is confusing as well (mentions `p` as if it was about paragraphs)

sowasred2012: I couldn't see that it was, but I wasn't confident that removing it wouldn't affect something elsewhere - it was introduced here (funnily, back when "code copyable" was being introduced, after originally being "code snippet"): https://github.com/canonical-web-and-design/vanilla-framework/commit/e29bdfecc8452fcd442470a9ae99d4e3180d0d0f#diff-4cce33a2f0a480da043e242e6c3213fe2f0ca05816fc29719f8500b6d3c65771R16

bartaz: OK. Doesn't hurt to keep it I guess, until we remove it in next major version.

bartaz (approved):
LGTM, thanks!

PR #3486, opened 2021-01-11
Add separator component

### LINKED ISSUE #3437: Proposing `p-separator` for spaced <hr>
In [many sites](https://github.com/search?q=org%3Acanonical-web-and-design+p-separator&type=code), we use an in row border to separate content in the same strip. The way we do that currently is by having a local pattern called `p-separator` which when attached to an `hr` creates a horizontal divider with a few rem padding top and bottom.

## Visual
image

## Context
This is used mainly on marketing sites. The style has been copied or worse inlined on the `hr` element which means it is not resonsive.

## State
None that I can think of.

## Progressive enhancement
The padding should scale with the viewport.

bartaz:
## Done

Adds new separator component

Fixes #3437 

## QA

- Open [demo](https://vanilla-framework-3486.demos.haus/docs/base/separators)
- Check the separator example:
  - https://vanilla-framework-3486.demos.haus/docs/examples/patterns/separator
  - Review new separator documentation page: https://vanilla-framework-3486.demos.haus/docs/base/separators

## Screenshots

Screenshot 2021-01-11 at 15 42 06

### DIFF
--- a/templates/docs/base/separators.md
+++ b/templates/docs/base/separators.md
@@ -28,7 +28,7 @@ View example of the separator component
 
 ### Bordered strips
 
-For the content contain in strips, use [the bordered strip variant](/docs/patterns/strip#bordered-strip).
+For content contained in strips, use [the bordered strip variant](/docs/patterns/strip#bordered-strip).
 
 <div class="embedded-example"><a href="/docs/examples/patterns/strips/is-bordered/" class="js-example">
 View example of the pattern strip is-bordered
--- a/templates/docs/base/separators.md
+++ b/templates/docs/base/separators.md
@@ -1,0 +1,47 @@
+---
+wrapper_template: '_layouts/docs.html'
+context:
+  title: Separators | Base elements
+---
+
+## Separators
+
+<hr>
+
+Vanilla gives you multiple ways to separate parts of the content with a horizontal line.
+
+### Horizontal line
+
+Standard `<hr>` element can be used to separate parts of the longer pieces of text.
+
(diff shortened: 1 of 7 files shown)


anthonydillon:
I wonder if we could inherit the `hr` base styling in this component so the user can use `<div class="p-separator"></div>` without a surprise or do you want that to fail?

templates/docs/base/separators.md
```diff
@@ -0,0 +1,47 @@
...
+
+To separate block sections of the page, use the separator component `<hr class="p-separator">`.
+
+<div class="embedded-example"><a href="/docs/examples/patterns/separator/" class="js-example">
+View example of the separator component
+</a></div>
+
+### Bordered strips
+
+For the content contain in strips, use [the bordered strip variant](/docs/patterns/strip#bordered-strip).
```
sowasred2012, line 31:
```suggestion
For content contained in strips, use [the bordered strip variant](/docs/patterns/strip#bordered-strip).
```

sowasred2012 (commented):
Looks good, just one spelling update.

bartaz:
> I wonder if we could inherit the `hr` base styling in this component so the user can use `<div class="p-separator"></div>` without a surprise or do you want that to fail?

@anthonydillon I thought about it, but feels that it may be better to encourage using semantic elements? Also, using a `<div class="p-separator">` would allow to put some child elements in it, which will easily break. So maybe better to keep it tied to `hr`?

lyubomir-popov (approved):
LGTM

PR #3490, opened 2021-01-13
Add numbered variant to code snippet

### LINKED ISSUE #3471: Implement line numbering for code snippet

sowasred2012:
## Done

Added numbered line variant of code snippet block.

Fixes #3471 
## QA

- Open [demo](https://vanilla-framework-3490.demos.haus/docs/examples/patterns/code-snippet/code-snippet-numbered)
- See that it matches the [design](https://app.zeplin.io/project/5fa270d8eee4dbb267e579f1/screen/5fc616c95320fdb32885432e)
  - Check the updated [documentation](https://vanilla-framework-3490.demos.haus/docs/base/code) makes sense

### DIFF
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -88,17 +88,12 @@ $code-sidebar-width: $sph-inner + (4 * $digit-width) !default;
   white-space: pre-wrap;
   width: 100%;
 
-  &:empty {
-    display: block;
-    min-height: $sp-xx-large;
-  }
-
   &::before {
     color: $color-mid-dark;
     content: counter(line-numbering);
     counter-increment: line-numbering;
     height: 100%;
-    left: 0;
+    left: -$digit-width;
     overflow: hidden;
     padding-right: $sph-inner;
     pointer-events: none;
@@ -110,6 +105,6 @@ $code-sidebar-width: $sph-inner + (4 * $digit-width) !default;
     -ms-user-select: none;
     user-select: none;
     // stylelint-enable property-no-vendor-prefix
-    width: $code-sidebar-width;
+    width: $code-sidebar-width + $digit-width;
   }
 }
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -1,8 +1,8 @@
 @import 'settings';
 
 $color-pre-bg: rgba($color-x-dark, 0.03);
-$digit-width: 0.5rem; // measured width of one character in the monospaced font
-$code-sidebar-width: $sph-inner + (4 * $digit-width) !default;
+$digit-width: 1ch; // measured width of one character in the monospaced font
+$code-sidebar-width: calc(#{$sph-inner} + (4 * #{$digit-width})) !default;
 
 // Base code styles
 @mixin vf-b-code {
@@ -105,6 +105,8 @@ $code-sidebar-width: $sph-inner + (4 * $digit-width) !default;
     -ms-user-select: none;
     user-select: none;
     // stylelint-enable property-no-vendor-prefix
-    width: $code-sidebar-width + $digit-width;
+    // We're adding a negative left position above of $digit-width,
+    // so the width here needs to take that into account
+    width: calc(#{$code-sidebar-width} + #{$digit-width});
   }
 }
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -104,12 +104,7 @@ $code-sidebar-width: calc(#{$sph-inner} + (4 * #{$digit-width})) !default;
     pointer-events: none;
     position: absolute;
     text-align: right;
-    // stylelint-disable property-no-vendor-prefix
-    -webkit-user-select: none;
-    -moz-user-select: none;
-    -ms-user-select: none;
     user-select: none;
-    // stylelint-enable property-no-vendor-prefix
     // We're adding a negative left position above of $digit-width,
     // so the width here needs to take that into account
     width: calc(#{$code-sidebar-width} + #{$digit-width});
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -88,6 +88,11 @@ $code-sidebar-width: calc(#{$sph-inner} + (4 * #{$digit-width})) !default;
   white-space: pre-wrap;
   width: 100%;
 
+  &:empty::after {
+    content: ' ';
+    user-select: none;
+  }
+
   &::before {
     color: $color-mid-dark;
     content: counter(line-numbering);
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -1,5 +1,9 @@
 @import 'settings';
 
+$color-pre-bg: rgba($color-x-dark, 0.03);
+$digit-width: 0.5rem; // measured width of one character in the monospaced font
+$code-sidebar-width: $sph-inner + (4 * $digit-width) + $sph-inner !default;
+
 // Base code styles
 @mixin vf-b-code {
   code,
@@ -74,5 +78,37 @@
     position: absolute;
     top: #{$spv-inner--x-small--scaleable + 0.5 * (map-get($line-heights, default-text) - map-get($icon-sizes, default))};
     width: map-get($icon-sizes, default);
+  }
+}
+
+%code-numbered-line {
+  display: inline-block;
+  padding: 0 $sph-inner 0 $code-sidebar-width;
+  position: relative;
+  white-space: pre-wrap;
+  width: 100%;
+
+  &:empty {
+    display: block;
+    min-height: $sp-xx-large;
+  }
+
+  &::before {
+    color: $color-mid-dark;
+    content: counter(line-numbering);
+    counter-increment: line-numbering;
+    height: 100%;
+    left: 0;
+    padding: 0 $sph-inner;
+    pointer-events: none;
+    position: absolute;
+    text-align: right;
+    // stylelint-disable property-no-vendor-prefix
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
+    // stylelint-enable property-no-vendor-prefix
+    width: $code-sidebar-width;
   }
 }
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -2,7 +2,7 @@
 
 $color-pre-bg: rgba($color-x-dark, 0.03);
 $digit-width: 0.5rem; // measured width of one character in the monospaced font
-$code-sidebar-width: $sph-inner + (4 * $digit-width) + $sph-inner !default;
+$code-sidebar-width: $sph-inner + (4 * $digit-width) !default;
 
 // Base code styles
 @mixin vf-b-code {
@@ -99,7 +99,8 @@ $code-sidebar-width: $sph-inner + (4 * $digit-width) + $sph-inner !default;
     counter-increment: line-numbering;
     height: 100%;
     left: 0;
-    padding: 0 $sph-inner;
+    overflow: hidden;
+    padding-right: $sph-inner;
     pointer-events: none;
     position: absolute;
     text-align: right;
--- a/scss/_patterns_code-numbered.scss
+++ b/scss/_patterns_code-numbered.scss
@@ -1,5 +1,4 @@
 @import 'settings';
-$sidebar-width: 3rem !default;
 
 @mixin vf-p-code-numbered {
   @include deprecate('3.0.0', 'p-code-numbered is being removed, use the code snippet block numbered variant instead.') {
--- a/scss/_patterns_code-numbered.scss
+++ b/scss/_patterns_code-numbered.scss
@@ -1,5 +1,5 @@
 @import 'settings';
-$sidebar-width: 4.5rem !default;
+$sidebar-width: 3rem !default;
 
 @mixin vf-p-code-numbered {
   @include deprecate('3.0.0', 'p-code-numbered is being removed, use the code snippet block numbered variant instead.') {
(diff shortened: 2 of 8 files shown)


templates/docs/base/code.md
```diff
@@ -69,27 +77,13 @@ View example of the code copyable pattern
...
-The code numbered pattern can be used when displaying large blocks of code to enable users to quickly reference a specific line.
-
-<div class="embedded-example"><a href="/docs/examples/patterns/code-numbered/" class="js-example">
-View example of the code numbered pattern
-</a></div>
-
-### Import
-
-To import just code copyable or numbered component into your project, copy either or both snippets below and include it in your main Sass file.
+### Deprecated numbered
```
bartaz, line 80: Keep the title, by put `<span class="p-label--deprecated">Deprecated</span>` below.

bartaz: Ah, you have the label. So keep the title

templates/docs/base/code.md
```diff
@@ -38,10 +38,18 @@ View example of the base code block
```
bartaz, line 38: Just noticed ^ I don't think we use plurals and title case in headings, so probably should be "Code snippet"

templates/docs/examples/patterns/code-snippet/code-snippet-numbered.html
```diff
@@ -0,0 +1,21 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Code Snippet / Numbered{% endblock %}
```
bartaz, line 2: "Code snippet" in the example names as well for consistency (same in other code snippet examples).

scss/_patterns_code-numbered.scss
```diff
@@ -2,43 +2,45 @@
...
-      text-align: right;
-      top: 0;
-      // stylelint-disable property-no-vendor-prefix
-      -webkit-user-select: none;
-      -moz-user-select: none;
-      -ms-user-select: none;
-      user-select: none;
-      // stylelint-enable property-no-vendor-prefix
-      width: $sidebar-width;
+      padding: 0 $sph-inner 0 ($sidebar-width + $sph-inner);
```
bartaz, line 13: Would be nice to get rid of that amount of duplication, as it may be hard to keep them in line if any changes are made.

I'm not exactly sure where to put common parts. We don't want the legacy 'code numbered' to depend on mixins or placeholders from new pattern. Doing it other way around and making new code snippet take styles from legacy component doesn't seem good as well.

Maybe we should have some `%code-numbered` defined in `_base_code.scss`, and extended where needed?

templates/docs/base/code.md
```diff
@@ -81,25 +87,11 @@ Please copy the entire JS in the example, for copy to clipboard functionality.
...
-</a></div>
-
-### Import
-
-To import just code copyable or numbered component into your project, copy either or both snippets below and include it in your main Sass file.
-
-```scss
-@import 'patterns_code-copyable';
-```
+<span class="p-label--deprecated">Deprecated</span>
```
bartaz, line 90: Please add the status change to the 'components status' table as well.

scss/_patterns_code-numbered.scss
sowasred2012: I was in two minds about this - I thought about creating a placeholder, but once this pattern is removed, the only place it'll be used is in the code snippet, defeating the purpose of a placeholder. 

As we're deprecating the older pattern, and it'll ultimately get removed, I thought duplication wasn't so bad in this case and that the older pattern could potentially stay "frozen" this way until it is removed.

bartaz: While we are deprecating it, it's not like we are going to remove it soon, and we want the style consistent, so I don't think we want to keep 2 separate implementations.

Yes, using placeholder means it will technically not be needed later, when there is one component only. But it also doesn't hurt.

I wouldn't mind if it's just a bit of CSS, but there is quite a lot of very specific styling here which we may want to keep up to date properly.

bartaz:
The spacing of the numbers is quite large (even for large numbers), maybe we should reduce it as @lyubomir-popov suggested?

Screenshot 2021-01-14 at 13 39 10

sowasred2012:
@lyubomir-popov could you take a look at the width of the number column? I've reduced it some, it should now fit 4 numbers comfortably whilst occupying less space.

lyubomir-popov:
@sowasred2012 we can safely shave off another .5rem. Since the font is monospaced, all digits occupy the same space. That seems to be almost exactly .5rem per char:
image

So for 4 digits we need 2rem, and we have one rem padding on each side, so 4rem total:

image

scss/_patterns_code-numbered.scss
```diff
@@ -2,43 +2,45 @@
 $sidebar-width: 4.5rem !default;
```
lyubomir-popov, line 2:
```suggestion
$digit-width: .5rem; // measured width of one character in the monospaced font
$sidebar-width: $sph-inner + 4 * $digit-width + $sph-inner !default;
```

lyubomir-popov:
@sowasred2012 maybe add `overflow: hidden;` to be safe? couldn't get `text-overflow:ellipsis` to work though.

scss/_patterns_code-numbered.scss
```diff
@@ -1,44 +1,20 @@
...
-      -webkit-user-select: none;
-      -moz-user-select: none;
-      -ms-user-select: none;
-      user-select: none;
-      // stylelint-enable property-no-vendor-prefix
-      width: $sidebar-width;
+    .p-code-numbered__line {
+      @extend %code-numbered-line;
+
+      &::before {
```
bartaz, line 14: Why is this needed?

scss/_patterns_code-numbered.scss
```diff
@@ -1,44 +1,20 @@
...
-    white-space: pre-wrap;
-    width: 100%;
-
-    &:empty {
-      display: block;
-      min-height: $sp-xx-large;
+  @include deprecate('3.0.0', 'p-code-numbered is being removed, use the code snippet block numbered variant instead.') {
+    .p-code-numbered {
+      counter-reset: line-numbering;
+      padding: $sph-inner--small 0;
```
bartaz, line 8: Shouldn't code numbered have same padding as default code block? we don't need to 0 the padding for the white background anymore.

scss/_patterns_code-snippet.scss
```diff
@@ -44,6 +45,15 @@ $color-snippet-heading-bg: rgba($color-x-dark, 0.08);
       }
     }
 
+    .p-code-snippet__block--numbered {
+      counter-reset: line-numbering;
+      padding: $sph-inner--small 0;
```
bartaz, line 50: Same as with legacy component - does it need to override the padding? I guess we can accomodate the padding on the lines with padding on the block, and avoid overriding this.

bartaz:
@lyubomir-popov Would that be really bad if (in the edge case of exceeding 10000 lines) we let the fifth digit overflow on the left padding (instead of just cropping the last one)?

Screenshot 2021-01-15 at 08 58 30

Basically it's doing `left: -0.5rem` on the number ::before, and bumping its sizes by 0.5rem (up to 5 digits).

Sure, it doesn't look great, but it doesn't break the numbering. And should likely be very very rare (I can't imagine having a code snippet on a page that long, even on things like github such long files should be rare).

scss/_patterns_code-numbered.scss
```diff
@@ -1,44 +1,14 @@
 @import 'settings';
-$sidebar-width: 4.5rem !default;
+$sidebar-width: 3rem !default;
```
bartaz, line 2: we already have `$code-sidebar-width` defined, is it different?

bartaz: Actually, this one doesn't seem to be used at all ATM.

templates/docs/base/code.md
```diff
@@ -81,25 +87,11 @@ Please copy the entire JS in the example, for copy to clipboard functionality.
...
 
-```scss
-@import 'patterns_code-numbered';
-```
+The code numbered pattern is now deprecated and will be removed in a future version of Vanilla.
 
-For more information see [Customising Vanilla](/docs/customising-vanilla/) in your projects, which includes overrides and importing instructions.
+It is being replaced by the new [code snippet pattern](/docs/base/code#code-snippets).
```
bartaz, line 94: This link doesn't work, the heading is not plural
```suggestion
It is being replaced by the new [code snippet pattern](/docs/base/code#code-snippet).
```

templates/docs/component-status.md
```diff
@@ -29,6 +29,12 @@ When we add, make significant updates, or deprecate a component we update their
       <td>2.22.0</td>
       <td>The new <code>p-separator</code> component has been added to be used as a separator between content blocks.</td>
     </tr>
+    <tr>
```
bartaz, line 32: When deprecating and suggesting new component, we should add info about Code snippet being added as well.

lyubomir-popov:
> Would that be really bad if (in the edge case of exceeding 10000 lines) we let the fifth digit overflow on the left padding (instead of just cropping the last one)?

looks ok to me

scss/_base_code.scss
```diff
@@ -1,6 +1,8 @@
 @import 'settings';
 
 $color-pre-bg: rgba($color-x-dark, 0.03);
+$digit-width: 0.5rem; // measured width of one character in the monospaced font
```
anthonydillon, line 4: Could we use `1ch` here?

bartaz:
@sowasred2012 There is something going wrong when there are empty lines in code:

Screenshot 2021-01-15 at 10 17 40

scss/_base_code.scss
sowasred2012: TIL about `ch`... however, unfortunately not, in the next line we're combining `$digit-width` with a `rem` value, `$sph-inner`, and `ch` and `rem` values are incompatible.

lyubomir-popov: that's what calc is for e.g. `calc(2rem + 4cm)` :)

sowasred2012: Ah yes of course! Cool, adding this in now.

scss/_base_code.scss
```diff
@@ -1,5 +1,9 @@
 @import 'settings';
 
+$color-pre-bg: rgba($color-x-dark, 0.03);
+$digit-width: 1ch; // measured width of one character in the monospaced font
+$code-sidebar-width: calc(#{$sph-inner} + (4 * #{$digit-width})) !default;
```
bartaz, line 5: if the 'digit width' uses `1ch` (which literary is 1 character width) I don't think we need separate variable for that.
```suggestion
$code-sidebar-width: calc(#{$sph-inner} + 4ch) !default; // width based on padding + 4 digits width
```

templates/docs/examples/patterns/code-snippet/code-snippet-numbered.html
```diff
@@ -0,0 +1,23 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Code snippet / Numbered{% endblock %}
+
+{% block standalone_css %}patterns_code-snippet{% endblock %}
+
+{% block content %}
+<div class="p-code-snippet">
+  <div class="p-code-snippet__header">
+    <h5 class="p-code-snippet__title">Check you have snap installed</h5>
```
bartaz, line 9: The header doesn't seems to be correct for the code content. Update it (but I don't know what the code is doing) ;) or remove the headier.

templates/docs/base/code.md
```diff
@@ -34,7 +34,7 @@ View example of the base pre block
 View example of the base code block
 </a></div>
 
-### Code Snippets
+### Code snippet
 
 The code snippet pattern should be used to group related code examples, making them easier to find and understand. It should consist of one or, at most, two code blocks with optional headers for their titles.
```
bartaz, line 39: Would be good to add "new" label under Code snippet heading.

templates/docs/base/code.md
```diff
@@ -52,31 +52,27 @@ A code snippet block can include an icon by using the `.p-code-snippet__block--c
 View example of the code snippet
```
bartaz, line 52: Missed that before, there is a typo in class name `.p-code-snippet__block--code`, should be `--icon`.

templates/docs/component-status.md
```diff
@@ -30,11 +30,23 @@ When we add, make significant updates, or deprecate a component we update their
...
+      <td>The new <code>.p-code-snippet</code> component has been added, to be used to display code examples in a number of different formats.</td>
+    </tr>
+    <tr>
+      <th><a href="/docs/examples/patterns/code-copyable">Code copyable</a></th>
       <td><div class="p-label--deprecated">Deprecated</div></td>
       <td>2.22.0</td>
       <td><code>.p-code-copyable</code> has been deprecated and will be removed in the v3.0 release. Please use <code>.p-code-snippet .p-code-snippet__block--icon</code> instead.</td>
     </tr>
+    <tr>
+      <th><a href="/docs/examples/patterns/code-numbered">Code numbered</a></th>
```
bartaz, line 45: Please link to documentation (with #section anchors if needed) rather than examples.

scss/_base_code.scss
sowasred2012: I think we should keep this, we position the number column negative left by the width of one character, I think it helps make that clear?

bartaz: Ah, haven't noticed that we do use negative positioning. Sure thing.

bartaz (approved):
LGTM

scss/_base_code.scss
anthonydillon: Thats @lyubomir-popov that is what I was thinking. 

I believe its clearer to simplify the code like this to show its 4 characters width your calculating:
```scss
code-sidebar-width: calc(#{$sph-inner} + 4ch) !default;
```
Over this:
```scss
$digit-width: 1ch;
$code-sidebar-width: calc(#{$sph-inner} + (4 * #{$digit-width})) !default;
```

PR #3493, opened 2021-01-15
Use correct key in map for x-small text

lyubomir-popov:
## Done

padding-top and margin-bottom on x-small text do not add up to an exact multiple of .5rem, throwing vertical rhythm off.
This is due to a typo, using the nudge for small text instead of the x-small text in a couple of places. 

This pr fixes that
## QA

Inspect the code change, verify the values of padding-top and margin bottom on x-small text add up to a multiple of .5rem (+ the browser rounding compensation).

Here's a screenshot:
image


bartaz (approved):
LGTM

PR #3497, opened 2021-01-18
Apply is-inline button styles at the small breakpoint and up

### LINKED ISSUE #3344: Add styling for SummaryButton from react components
In react components repo we added SummaryButton components (for ModularTable).
It has some minimal custom styling which should be part of Vanilla.

https://github.com/canonical-web-and-design/react-components/pull/300

Screenshot 2020-10-07 at 15 18 45

### LINKED ISSUE #3387: “is-inline” button margin becomes 0 at <875px screen width
Vanilla’s style sheet includes this:
```css
@(min-width: 875px)
[class*="p-button"].is-inline {
    margin-left: 1rem;
    width: auto;
}
```

So when the viewport width is ≥875px, an inline button has a nice left margin:

> image

But at a viewport width of <875px, the button is smooshed against the preceding text:

> image

I can’t think of any reason for this margin to be viewport-dependent.

(Discovered while investigating [ubuntu.com#8633](https://github.com/canonical-web-and-design/ubuntu.com/issues/8633).)

sowasred2012:
## Done
Reduced the breakpoint at which `is-inline` styles are applied to `button` elements, so that the correct margin is present when the button is not full width.

Related to #3344 - once this PR is merged there's another task to remove unnecessary styles from the SummaryButton in React Components.

Fixes #3387

## QA

- Open [demo](https://vanilla-framework-3497.demos.haus/docs/examples/patterns/buttons/small)
- Reduce the width of your browser, see that the second button does not lose its left margin before it becomes full width.

### DIFF
--- a/scss/_patterns_buttons.scss
+++ b/scss/_patterns_buttons.scss
@@ -136,7 +136,7 @@ $hover-background-opacity-percentage: $hover-background-opacity-amount * 100%;
 
 @mixin vf-button-inline {
   [class*='p-button'].is-inline {
-    @media (min-width: $breakpoint-medium) {
+    @media (min-width: $breakpoint-x-small) {
       margin-left: $sph-outer;
       width: auto;
     }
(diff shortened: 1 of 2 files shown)


bartaz (approved):
LGTM, thanks

PR #3500, opened 2021-01-19
Add box-decoration-break: clone to code element

### LINKED ISSUE #3408: <code>: Content wrap results in uneven left margin
## Pattern to amend
`<code>`: Content wrap results in uneven left margin

## Visual
image

sowasred2012:
## Done

Fixed broken box shadow when an inline `code` element wrapped to a second line.

Technically fixes #3408 - in investigating that quite old issue, I discovered that the margin no longer seemed to be an issue, but that there was a related problem.

## QA

- Open [demo](https://vanilla-framework-3500.demos.haus/docs/examples/base/code-inline)
- Reduce the width of the browser to the point that the code element wraps to a second line (or edit the text in your browser dev tools to add a longer piece of text).
- See that the grey background doesn't look as though it has a chunk missing (compare to [live](https://vanillaframework.io/docs/examples/base/code-inline))

## Screenshots

Before:
Screenshot from 2021-01-19 14-43-47

After:
Screenshot from 2021-01-19 14-44-15

### DIFF
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -28,6 +28,7 @@ $code-sidebar-width: calc(#{$sph-inner} + (4 * #{$digit-width})) !default;
     box-shadow: 0 0 0 0.25rem $color-mid-x-light;
     margin-left: 0.25rem;
     margin-right: 0.25rem;
+    word-break: break-all;
   }
 
   code,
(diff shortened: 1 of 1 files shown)


bartaz (approved):
LGTM, would be good for @lyubomir-popov to have a quick look as well

Issue #2457, opened 2019-08-07
Allow wrapping code in code block


bartaz:
## Pattern to amend

Code block was changed in Vanilla 2.0 not to wrap lines by default (uses whitespace `pre` instead of `pre-wrap`).
This caused some issues in snapcraft.io docs that relied on wrapping:

<img width="1672" alt="Screenshot 2019-08-07 at 12 42 39" src="https://user-images.githubusercontent.com/83575/62617777-f0ae1080-b912-11e9-8d6e-20e7bf22d518.png">

It would be good to have at least an option to turn wrapping on.

lyubomir-popov:
Linking the [discussion](https://github.com/canonical-web-and-design/vanilla-framework/issues/1986) that lead to this change.

anthonydillon:
This is a layout issue not a component issue. The component should only consume the space it has available from is parent. This is the same issue I gave fixed in 3 sites now. We really need a central place to resolve layout issues which is the layouts work in Vanilla.

bartaz:
@anthonydillon Yes, it's a layout issue and was fixed in docs, but still, there are cases where using wrapping in code blocks could be useful.

Snapcraft docs often have console outputs with long lines of text. Wrapping made them nicely readable, now you need to scroll horizontally to read text.

anthonydillon:
Interesting, good point. Modifier on the code block pattern?

bartaz:
Vanilla seems to have 2 approaches `--modifier` or `is-modifier`, not sure if there is any formal distinction between the two.
Simply having an option to change the style without need for custom CSS would be welcome.

Unless of course from design point of view we feel it's better not to wrap code blocks in CSS and any wrapping should be done in content for better control.

kwm14:
## WG
- We need to recommend an alternative in our Documentation as a hardcoded markup solution

sowasred2012:
Commented the same on #2184 before spotting this, which is probably the more appropriate place:

I have set white-space: pre-wrap on the new code snippet, but only when using numbered lines (https://github.com/canonical-web-and-design/vanilla-framework/blob/master/scss/_base_code.scss#L88). However, I still think it should be the default on pre, scrolling vertically down the page and then horizontally within a given component presents accessibility issues: https://www.w3.org/WAI/WCAG21/Understanding/reflow

bartaz:
[...]

Issue #2184, opened 2019-03-05
<code> white space set to pre-wrap


lyubomir-popov:
**Describe the bug**
White space inside ```code``` is set to pre-wrap. In the screenshot below, observe how white space is causing the element to wrap. 

![image](https://user-images.githubusercontent.com/2741678/53804343-323f2780-3f3f-11e9-9eb1-4cccad8d89fd.png)

 Shouldn't that behaviour be reserved for the ```<pre>``` element only? Why are we overriding the default?

**To Reproduce**
1. Go to 'http://0.0.0.0:8101/vanilla-framework/examples/base/code/'
2. Inspect the white-space property of the ```<code>``` element

sowasred2012:
@bartaz @lyubomir-popov this doesn't seem to be the case anymore, white space isn't set on code at all now: https://vanillaframework.io/docs/examples/base/code-inline

Loosely related: I have set `white-space: pre-wrap` on the new code snippet, but only when using numbered lines (https://github.com/canonical-web-and-design/vanilla-framework/blob/master/scss/_base_code.scss#L88). However, I still think it should be the default on `pre`, scrolling vertically down the page and then horizontally within a given component presents accessibility issues: https://www.w3.org/WAI/WCAG21/Understanding/reflow

PR #3502, opened 2021-01-20
Add .is-wrapped code snippet utility

### LINKED ISSUE #3478: Implement wrapping variant for code snippet

sowasred2012:
## Done

- Added `.is-wrapped` utility class for code snippets, to allow content to wrap inside code blocks

Fixes #3478, #2457 & #2184

## QA

- Open [demo](https://vanilla-framework-3502.demos.haus/docs/examples/patterns/code-snippet/code-snippet-wrapping)
- If neither block wraps, reduce the browser width
- See that the first block does not wrap, and the second block does
  - [Code](https://vanilla-framework-3502.demos.haus/docs/base/code#code-snippet)
  - [Component status](https://vanilla-framework-3502.demos.haus/docs/component-status)

### DIFF
--- a/scss/_patterns_code-snippet.scss
+++ b/scss/_patterns_code-snippet.scss
@@ -10,6 +10,10 @@ $color-snippet-heading-bg: rgba($color-x-dark, 0.08);
     .p-code-snippet__block--icon,
     .p-code-snippet__block--numbered {
       margin: 0;
+
+      &.is-wrapped {
+        white-space: pre-wrap;
+      }
     }
 
     .p-code-snippet__block--icon {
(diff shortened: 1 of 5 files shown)


lyubomir-popov:
@bartaz @sowasred2012 wondering if the release notes mention is enough for something like this. I remember there were strong feelings and the default was hotly debated in the past.

bartaz:
@sowasred2012 There are some conflicts to resolve

scss/_patterns_code-snippet.scss
```diff
@@ -10,6 +10,10 @@ $color-snippet-heading-bg: rgba($color-x-dark, 0.08);
     .p-code-snippet__block--icon,
     .p-code-snippet__block--numbered {
       margin: 0;
+
+      &.is-wrapped {
```
bartaz, line 14: I wonder if we should let it be applied on just `<pre>`, so it can be used on base element, without the whole code snippet component structure (for example when code blocks are generated, and wrapping class is added by JS).

But maybe it's something we can do later if needed.

bartaz (approved):
LGTM, thanks

PR #3504, opened 2021-01-20
Add .row-breakout to enable layouts that spill into the page margins

lyubomir-popov:
Add a .row-breakout class that allows content to bleed into the page margins (up to a certain point).

QA:
 -Go to /docs/patterns/grid#breaking-out-of-the-grid
- Proof read copy, suggest improvements if something doesn't make sense
- Try importing the plugin with modified arguments, as described in the copy, ensure it works
- QA the following examples on different screen sizes: 
1) left aside: 
docs/examples/patterns/grid/grid-breakout-left-aside

2) right aside:
/docs/examples/patterns/grid/grid-breakout-right-aside

- Try to break it in different ways and report back :)


lyubomir-popov:
Slight misalignment happening at the point font size switches from 16 to 18, something to fix:

image

lyubomir-popov:
@bartaz another way to avoid a wrapper: introduce `row-*` classes same as we have `col-*` classes. This feels the most native way to use css grid:

image

bartaz:
> @bartaz another way to solve the stretching of items in each column without resorting to a wrapper and rigid predefined structures: introduce `row-*` classes same as we have `col-*` classes.

Ok. Yes, that could work. But this means you need to define how many rows you want from the start in HTML, it can be quite limiting. If content changes it may need adjusting.

lyubomir-popov:
> But this means you need to define how many rows you want from the start in HTML, it can be quite limiting. If content changes it may need adjusting.

to an extent, yes - but you don't need to be exact - giving it more rows wouldn't hurt, e.g. you could just give it 999 rows

lyubomir-popov:
- Added the cve table from https://ubuntu.com/security/cve as an example:
image
- Added a grid where the left edge aligns to the fixed width grid, but the content in the aside and the content area can span as much space as it needs. Currently set an overall limit of 120rem:
image

lyubomir-popov:
@bartaz could you take a look at this - I think all the functionality I want is in there, now it's a question of where to move stuff and what to do with the d3 chart

bartaz:
I'm not sure if we should extend the 'default' grid example with this. It was meant to be simple default 12 grid demo, that is also embedded in CodePen on grid docs page.
Throwing all of the 16 wide grid examples there makes it much harder to understand in docs. And included d3 doesn't work in CodePen at all (if we want it embedded we will need to find another way of including it in example).

I would suggest keeping default grid example as it was (just 12 column demo) and create separate example for 16 col grid. Ideally, one basic (just red boxes) that can be embedded, and separate one with exploration of different use cases and specific examples, that will be maybe part of Layout or Templates examples group and not embedded in the docs.

bartaz:
There seem to be some bugs in the example, especially on smaller screen size - the overlapping part is cut:

Screenshot 2021-01-27 at 11 00 33

Also, there is some HTML comment leaked at the end of the page:
Screenshot 2021-01-27 at 11 00 43

And 'S' on the left side of the table:

Screenshot 2021-01-27 at 11 00 49

templates/static/js/chart.js
```diff
@@ -0,0 +1,100 @@
+function responsivefy(svg) {
```
bartaz, line 1: Unfortunately it seems that adding these files in the same folder as example HTML is a bit of a problem.

There are two options here

- we keep them in `/static/js` (maybe just make it `static/js/examples` to know better what they are used for.
- or keep them with example HTML, but include them in HTML with `{% include %}` rather than `<script src="" />`, but this means that it will be inlined in HTML. it may be fine for short scripts, but we likely don't want whole d3 library inlined in the example

So, I guess keeping it in `static/js/examples` would be better for now, unless we have better ideas? @sowasred2012

scss/_patterns_grid.scss
```diff
@@ -179,4 +192,31 @@
       }
     }
   }
+
+  // Ways to break out of the 12-col grid
+
+  // Aside / main aligned to the left edge of the fixed-width grid
+  .row-new {
```
bartaz, line 199: Why is it called `-new`? Does it replace `.row`? Is it a variant?
It may be "new" now, but will not be in couple months.

lyubomir-popov: I actualy renamed it to row-breakout since  Ilast pushed - happy to change it to something else if you prefer

bartaz:
TBH, from the examples, I don't know how this new grid is supposed to work.

I look at this:

Screenshot 2021-01-27 at 16 22 01

and I don't know, why is 'filters' sidebar on top different than from the one on bottom. Why are the other boxes called 'items' not columns? Why the bottom ones are so small as 1 col? (the top ones get so small as well on other screen width) - do we ever want to have anything that small? Can I control width of that? When I add `col-2` to the item it seems to be doing something (making it "2 columns" wide), but these item columns are not the same as standard grid columns.

I know that it's a quite a complicated piece of layout. But it does get me thinking how should we document it, so it can be used in practice. Designers need to know how to use it - what can be built with that, what can't be built with that. And then developers need to know how to use it - and there is a load of new class names here, new concepts of wrapper elements needed around certain parts of the grid, etc. At this point I'm not sure if we should try to reuse 'grid' naming here, because (even that it's kind of extension of the grid) it is build  and behaves so different from standard grid, that it seems that it's more confusing than helpful.

lyubomir-popov:
@bartaz after talking to @spencerbygraves we decided to drop the 16-col grid - as the narrowness of the aside makes it useless. So the bottom variant is gone. I've also renamed he areas to aside (which can go left/right) and main, much simpler. Let me iron out a few things and I'll push again.

scss/_patterns_grid.scss
bartaz: Anything more semantic than 'new' would be better.

On the other hand, as I mentioned in my other comment rant (sorry for making it that long 😉 ) I'm starting to think that maybe trying to reuse the grid naming here is not that well, because it seems to be behaving different from the grid.

lyubomir-popov:
@bartaz moved the combined examples to a template: https://vanilla-framework-3504.demos.haus/docs/examples/templates/grid-breakout-variants

bartaz:
@lyubomir-popov On a large screens the 'items' in new grid seem to have reasonable size (to be a card in store for example), but once you go down the size they all become 1 col with making them very narrow to whatever content they may have.

grid-breakout

Is this expected? Do we intend to ever use such small items in practice, or should it always be used at least as col-2? It doesn't seem clear from the example. And because large screen items seem to have acceptable size, the smaller screen going down to col-1 width looks broken.

lyubomir-popov:
That's a bug, will fi

bartaz:
Does the height of 'main content' with items need to be constrained by the size of the sidebar?

What if sidebar is short and there is lots of items?
Screenshot 2021-01-28 at 10 06 21

What if sidebar is very long, and there is not that many items (filtered search results)?
Screenshot 2021-01-28 at 10 07 38

bartaz:
After this has been distilled down to simply being a list of items with a sidebar it seems a bit confusing to make it part of the grid and call it `row`.

In standard grid we have row, and its children are columns. In this new breakout layout the wrapper (that we call 'row-breakout') can actually be multiple rows of items. There are no real columns (on markup level you don't control the width of the items, they automatically take all available space and have some min-width). It feels to me that trying to call it similar to regular grid may make it more confusing than helpful. We shouldn't make developers or designers fell that it behaves similar to 12 col grid.

I don't know how to call this, but it probably should be a layout king of a thing. `l-breakout`? `l-grid-breakout`?

lyubomir-popov:
> I don't know how to call this, but it probably should be a layout king of a thing. `l-breakout`? `l-grid-breakout`?

That's what I initially had in mind, and I agree. @sowasred2012 @spencerbygraves any ideas for what to call this new layout? It may not always have cards, or a sidebar - for example it mihgt just have a cve table. Does breakout sound fitting? Or maybe something like `fluid`?

bartaz:
> It may not always have cards, or a sidebar - for example it mihgt just have a cve table.

Currently all the examples show sidebar and multiple cards.
If we have other usecases in mind we should show them in examples as well. Maybe keep the existing "sidebar left" "sidebar right" examples for embedding in docs, but have a separate "variants" example that shows different possibilities (like cards + sidebar, large table, etc).

scss/_patterns_grid.scss
```diff
@@ -98,23 +75,23 @@
...
     }
   }
 
   @for $i from 1 through $grid-columns-medium {
-    .#{$grid-medium-col-prefix}#{$i} {
+    .row .#{$grid-medium-col-prefix}#{$i} {
       @extend %span-full-grid-on-mobile;
       @extend %span-full-grid-on-desktop;
-      @extend %display-block;
+      // @extend %display-block;
```
bartaz, line 89: Is this intended? Other columns do extend display block. Why is this one commented?

scss/_patterns_grid.scss
```diff
@@ -27,44 +27,21 @@
...
-
   .row {
     @extend %vf-row;
   }
 
   // mobile grid
   @media (max-width: $threshold-4-6-col) {
     @for $i from $grid-columns-small through 1 {
-      .#{$grid-small-col-prefix}#{$i} {
+      .row .#{$grid-small-col-prefix}#{$i} {
```
bartaz, line 44: I guess the reason to add `.row` here is to make sure `col-X` class names work only in standard row context but not on new breakout layout items?

lyubomir-popov: yes, otherwise it will have unpredictable effect. Just a driveby bug fix.

scss/_settings_grid.scss
```diff
@@ -30,6 +29,7 @@ $grid-margin-widths: (
 $grid-margin-width: $sph-outer--large !default;
 
 // Grid breakout template
-$aside-autofit-width: 13rem !default;
-$main-items-autofit-width: 13rem !default;
-$threshold-12-col-plus: $grid-max-width + $aside-autofit-width;
+$l-fluid-breakout-max-width: $grid-max-width !default;
+$l-fluid-breakout-aside-width: 13rem !default;
+$l-fluid-breakout-main-width: 13rem !default;
```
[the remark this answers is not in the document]
lyubomir-popov: just a misleading var name, this controls the min width of children of main. Would `$l-fluid-breakout-main-child-width` make it clearer?

scss/_patterns_grid.scss
lyubomir-popov: not intended, thanks - fixed

scss/_patterns_grid.scss
```diff
@@ -158,18 +158,26 @@
...
+  $suffix: ''
+) {
   .l-fluid-breakout#{$suffix} {
     grid-gap: 0 0;
 
     @media (min-width: $breakpoint-large) {
       @supports (display: grid) {
         display: grid;
-        grid-template-columns: minmax($aside-autofit-width, 1fr) minmax($main-items-autofit-width, $breakout-autofit-width) minmax($aside-autofit-width, 1fr);
+        grid-template-columns: minmax($l-fluid-breakout-aside-width, 1fr) minmax($l-fluid-breakout-main-width, $l-fluid-breakout-max-width) minmax(
```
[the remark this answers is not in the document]
lyubomir-popov: I'll add some sort of fallback

lyubomir-popov:
@bartaz can you review the fallback stuff?

lyubomir-popov:
Screenshot 2021-02-08 at 15 06 16
Screenshot 2021-02-08 at 15 05 49
Screenshot 2021-02-08 at 14 57 06
Screenshot 2021-02-08 at 14 56 35
Screenshot 2021-02-08 at 14 56 26

scss/_patterns_grid.scss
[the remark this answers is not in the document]
lyubomir-popov: added fallback

bartaz:
@lyubomir-popov I think we should restructure the documentation for it a bit.

Because we technically put it as part of the layouts, instead of adding very long section to already long Grid page, I think it should be a separate "Fluid grid breakout" page in "Layouts" section of the docs.
In similar way we could move examples for it out of the patterns folder to layouts folder and change their titles from "Grid" to "Fluid breakout".

bartaz:
@lyubomir-popov would it be possible to clean up commits a bit? Squash them into a couple bigger commits or just into one. Right now there is almost 40 commits in this PR, many of them have messages like "wip" "typos" etc. If at any time in future we try to figure out where the changes come from it will be a hell to navigate through those.

scss/_patterns_grid.scss
```diff
@@ -180,3 +157,204 @@
     }
   }
 }
+
+@mixin vf-l-fluid-breakout(
```
bartaz, line 161: As this is a separate mixin and a layout outside of the grid it should be in a separate file called `_layouts_something`.

scss/_patterns_grid.scss
```diff
@@ -180,3 +157,204 @@
...
+) {
+  %l-fluid-breakout-ie11-fallback {
+    display: flex;
+
+    @media (max-width: $threshold-4-6-col) {
+      flex-direction: column;
+    }
+  }
+
+  %l-fluid-breakout__full-and-main-ie11-fallback {
```
bartaz, line 175: As it seems that we want to allow including `vf-l-fluid-breakout` multiple times in same SCSS with different params, defining %placeholders inside of it may cause issues. Especially that these placeholders seem to use the parameters of the mixin (`$l-fluid-breakout-aside-width`, `$l-fluid-breakout-main-child-width`), because these placeholders will all be added to the same global scope under the same name.

Not sure if that's supported in SCSS, but worth trying, to do the same with placeholder name as with class name `.l-fluid-breakout#{$suffix}` so placeholder would be `%l-fluid-breakout__full-and-main-ie11-fallback#{suffix}` (but I have no idea if that's possible).

templates/docs/examples/patterns/card/_header-and-footer.html
```diff
@@ -0,0 +1,11 @@
+<div class="p-card">
+  <!-- <div class="p-strip is-shallow u-no-padding--top">
```
bartaz, line 2: Should be removed if not needed.

templates/docs/examples/patterns/card/_header-and-footer.html
```diff
@@ -0,0 +1,11 @@
...
+  </div> -->
+    <img class="p-card__thumbnail" src="https://res.cloudinary.com/canonical/image/fetch/f_auto,q_auto,fl_sanitize,c_fill,w_64,h_64/https://api.charmhub.io/api/v1/media/download/charm_B1WVq07SAhAIBJ72odJQ9fNBCFXk7QYT_icon__ce0ecd4c4fb29e9a8ff84971fbbba8552a822d53717632f2aa5ecbce9dfea108.png" alt="landscape-server" class="p-card__thumbnail" loading="lazy" width="32" height="32"/>
```
bartaz, line 4: Link to some image on asset server, not an external service.

templates/docs/examples/patterns/grid/grid-breakout-no-aside.html
```diff
@@ -0,0 +1,20 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Grid / Wide - no aside{% endblock %}
+
+{% block standalone_css %}patterns_grid-breakout{% endblock %}
+
+{% block content %}
+
+<style>
+  .viz-grid-item, .viz-aside {
```
bartaz, line 9: This seems to be repeated a lot in these examples. Maybe extract it to separate file and `{% include %}` here?

templates/docs/patterns/grid.md
```diff
@@ -80,3 +105,101 @@ For more information see [Customising Vanilla](/docs/customising-vanilla/) in yo
 ### Design
 
 For more information [view the grid design spec](https://github.com/ubuntudesign/vanilla-design/tree/master/Grid), which includes the specification in markdown format and a PNG image.
+
+## Breaking out of the grid
```
bartaz, line 109: As mentioned separately, would be good to split it into separate page.

bartaz: But keep a section on Grid page that links there.

templates/docs/examples/templates/_card-fixed-width.html
```diff
@@ -0,0 +1,33 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Card / Default{% endblock %}
+
+{% block standalone_css %}patterns_card{% endblock %}
+
+{% block test_content %}
+<!-- <div class="p-card">
```
bartaz, line 7: Remove stuff that is not needed.

templates/docs/examples/patterns/side-navigation/_grid.html
```diff
@@ -0,0 +1,83 @@
...
+      <li class="p-side-navigation__item">
+        <a class="p-side-navigation__link" href="#">Deploy machines</a>
+      </li>
+    </ul>
+    <ul class="p-side-navigation__list">
+      <li class="p-side-navigation__item--title">
+        <a class="p-side-navigation__link">Concepts & terms</a>
+      </li>
+    </ul>
+    <!-- <ul class="p-side-navigation__list">
```
bartaz, line 59: Remove if not needed.

templates/docs/examples/patterns/grid/_cve-table.html
```diff
@@ -0,0 +1,539 @@
+<table class="cve-table">
+  <thead>
+    <tr>
+      <th style="width: 10em">ID</th>
+      <th>Priority</th>
+      <th>Package</th>
+      <th>
+        <div class="icon-container__text"> 14.04 ESM </div>
```
bartaz, line 8: I guess that these class names `icon-container__text` `cve-table-cell-priority`, etc are just copy paste from the CVE page, but served as part of Vanilla example they seem a bit confusing as they don't provide any styling so there is no use for them in the example.

lyubomir-popov: I've cleaned up the entire file

templates/docs/examples/patterns/side-navigation/_grid.html
lyubomir-popov: removed entire file

templates/docs/examples/templates/_card-fixed-width.html
lyubomir-popov: removed

scss/_patterns_grid.scss
lyubomir-popov: done

templates/docs/examples/patterns/card/_header-and-footer.html
lyubomir-popov: done

scss/standalone/patterns_grid-breakout.scss
```diff
@@ -0,0 +1,10 @@
+@import '../base';
+@include vf-base;
+
+@import '../patterns_grid';
+@include vf-p-grid;
+@include vf-l-fluid-breakout;
```
bartaz, line 6: You need to `@import '../layouts_fluid-breakout';` because it's moved to separate file now.

scss/standalone/patterns_grid-breakout-full.scss
```diff
@@ -0,0 +1,6 @@
+@import '../base';
+@include vf-base;
+
+@import '../patterns_grid';
+@include vf-p-grid;
+@include vf-l-fluid-breakout(13rem, 100%, 0, '--single-main');
```
bartaz, line 6: You need to `@import '../layouts_fluid-breakout';` because it's moved to separate file now.

scss/standalone/patterns_grid-breakout.scss
lyubomir-popov: done

scss/standalone/patterns_grid-breakout-full.scss
lyubomir-popov: done

scss/_patterns_grid.scss
lyubomir-popov: seems to work fine, nothing breaks

templates/docs/examples/patterns/grid/grid-breakout-no-aside.html
lyubomir-popov: done

templates/docs/patterns/grid.md
lyubomir-popov: done

clagom:
Apart from the comment on the inverted examples showing the same layout, I think this is working very well @lyubomir-popov. I would +1 this UX wise now, and then ask UX to have a look to iterate on it where necessary and to implement UX considerations and guidelines around your work. Great work as usual :)

lyubomir-popov:
Thanks a lot @clagom!

templates/docs/examples/templates/_card-fixed-width.html
bartaz: @lyubomir-popov Doesn't seem removed

lyubomir-popov: oh sorry wrong file, this is actually not used, removed the entire file

templates/docs/examples/patterns/grid/grid-breakout-right-aside.html
```diff
@@ -0,0 +1,27 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Grid / Wide - aside on the right{% endblock %}
```
bartaz, line 2: Please rename it, as it's not Grid example anymore ("Fluid breakout / Wide - aside on the right"?).

Also, because it's not a pattern move the file from `/templates/docs/examples/patterns/grid` to `/templates/docs/examples/layouts/...`.

templates/docs/examples/patterns/grid/grid-breakout-no-aside.html
```diff
@@ -0,0 +1,19 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Grid / Wide - no aside{% endblock %}
```
bartaz, line 2: Please rename it, as it's not Grid example anymore ("Fluid breakout / ..."?).

Also, because it's not a pattern move the file from `/templates/docs/examples/patterns/grid` to `/templates/docs/examples/layouts/...`.

templates/docs/examples/patterns/grid/grid-breakout-left-aside.html
```diff
@@ -0,0 +1,66 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Grid / Wide - aside on the left{% endblock %}
```
bartaz, line 2: Please rename it, as it's not Grid example anymore ("Fluid breakout / ..."?).

Also, because it's not a pattern move the file from `/templates/docs/examples/patterns/grid` to `/templates/docs/examples/layouts/...`.

templates/docs/examples/patterns/grid/grid-breakout-full.html
```diff
@@ -0,0 +1,17 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Grid / Wide - main spanning 3 columns{% endblock %}
```
bartaz, line 2: Please rename it, as it's not Grid example anymore ("Fluid breakout / ..."?).

Also, because it's not a pattern move the file from `/templates/docs/examples/patterns/grid` to `/templates/docs/examples/layouts/...`.

templates/docs/examples/patterns/grid/grid-breakout-full--cve-table.html
```diff
@@ -0,0 +1,34 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Grid / Wide - CVE table{% endblock %}
```
bartaz, line 2: Please rename it, as it's not Grid example anymore ("Fluid breakout / ..."?).

Also, because it's not a pattern move the file from `/templates/docs/examples/patterns/grid` to `/templates/docs/examples/layouts/...`.

templates/docs/examples/patterns/grid/grid-breakout-cards-and-aside.html
```diff
@@ -0,0 +1,37 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Grid / Wide - cards and aside{% endblock %}
```
bartaz, line 2: Please rename it, as it's not Grid example anymore ("Fluid breakout / ..."?).

Also, because it's not a pattern move the file from `/templates/docs/examples/patterns/grid` to `/templates/docs/examples/layouts/...`.

templates/docs/examples/patterns/grid/_cve-table.html
```diff
@@ -0,0 +1,323 @@
+<table>
```
bartaz, line 1: When moving examples from `patterns` to `layouts` this should be moved as well.

templates/docs/examples/patterns/grid/_aside.html
```diff
@@ -0,0 +1,64 @@
+<form>
```
bartaz, line 1: When moving examples from `patterns` to `layouts` this should be moved as well.

bartaz (changes requested):
Please also move examples from "Patterns" to "Layouts"

templates/docs/examples/patterns/grid/grid-breakout-full.html
lyubomir-popov: done

templates/docs/examples/patterns/grid/grid-breakout-right-aside.html
lyubomir-popov: done

templates/docs/examples/patterns/grid/grid-breakout-no-aside.html
lyubomir-popov: done

templates/docs/examples/patterns/grid/grid-breakout-left-aside.html
lyubomir-popov: done

templates/docs/examples/patterns/grid/grid-breakout-full--cve-table.html
lyubomir-popov: done

templates/docs/examples/patterns/grid/grid-breakout-cards-and-aside.html
lyubomir-popov: done

templates/docs/examples/patterns/grid/_cve-table.html
lyubomir-popov: done

templates/docs/examples/patterns/grid/_aside.html
lyubomir-popov: done

templates/docs/patterns/grid.md
```diff
@@ -66,6 +91,10 @@ Please note, specifying a value that exceeds the available number of columns wil
 View example of the incorrect column offset within a nested grid
 </a></div>
 
+### Breaking out of the grid
+
+In some cases, there might be a good reason to break out of the constraints of a 12 column grid and allow content to bleed into the page margins. Vanilla provides a separate [layout](/docs/layouts/fluid-breakout) for this purpose
```
bartaz, line 96: Missing full stop, and more descriptive link.

```suggestion
In some cases, there might be a good reason to break out of the constraints of a 12 column grid and allow content to bleed into the page margins. Vanilla provides a separate [fluid breakout layout](/docs/layouts/fluid-breakout) for this purpose.
```

templates/docs/layouts/fluid-breakout.md
```diff
@@ -0,0 +1,103 @@
...
+
+In some cases, there might be a good reason to break out of the constraints of a 12 column grid and allow content to bleed into the page margins - for example, tables with many columns that would otherwise result in heavy truncation or wrapping, charts with a lot of detail along the x axis, or card layouts that aim to impress with the abundance of available content. The fluid breakout layout allows you to do this.
+
+### Anatomy of the fluid breakout layout
+
+#### Wrapper
+
+The `.l-fluid-breakout` serves as a wrapper, and allows for an aside and a main area. On screens smaller than `$breakpoint--large`, it acts as a single column. On larger screens it switches to a 3 column layout - a central area and two aside areas on either side.
+
+The wrapper aims to align as much as possible with the 12 column grid. On smaller screens, than is limited to ensuring the padding of both layouts match. Once the window width becomes large enough, the left edges of the regular 12 column grid and the central column of the fluid breakout layout also align. For this to be possible, two conditions need to be met:
```
bartaz, line 19: Typo?

```suggestion
The wrapper aims to align as much as possible with the 12 column grid. On smaller screens, that is limited to ensuring the padding of both layouts match. Once the window width becomes large enough, the left edges of the regular 12 column grid and the central column of the fluid breakout layout also align. For this to be possible, two conditions need to be met:
```

templates/docs/layouts/fluid-breakout.md
```diff
@@ -0,0 +1,103 @@
...
+
+- <p>the central area's width needs to match the 12 column grid's width (we achieve this by setting the central area's max-width to the same value as the 12 column grid's max-width)</p>
```
bartaz, line 21: Why is this using paragraphs inside the list?

bartaz (commented):
Couple questions/suggestions on the docs.

scss/_layouts_fluid-breakout.scss
```diff
@@ -0,0 +1,199 @@
...
+      }
+    }
+  }
+
+  %l-fluid-breakout__aside-ie-11-fallback#{$suffix} {
+    flex: 0 0 auto;
+    justify-content: flex-start;
+
+    @media (min-width: $l-fluid-breakout-3-col-threshold) {
+      // display: flex;
```
bartaz, line 41: Remove if not needed
```suggestion
```

scss/_layouts_fluid-breakout.scss
```diff
@@ -0,0 +1,199 @@
...
+          grid-column: 1 / span 3;
+        }
+      }
+    }
+
+    .l-fluid-breakout__main {
+      @extend %l-fluid-breakout__main-and-full-common;
+
+      @media (min-width: $breakpoint-large) {
+        @supports (display: grid) {
```
bartaz, line 138: Would it make sense to wrap everything (that is not IE fallback) into `@supports (display: grid)`?
Right not it seems you separately use `@supports` in every single selector/query.

scss/standalone/patterns_grid-breakout.scss
```diff
@@ -0,0 +1,13 @@
+@import '../base';
```
bartaz, line 1: Is this file still needed or was it renamed to `layouts_fluid-breakout`?

scss/standalone/patterns_grid-breakout-full.scss
```diff
@@ -0,0 +1,12 @@
+@import '../base';
```
bartaz, line 1: Is this file still needed or was it renamed to `layouts_fluid-breakout-full`?

bartaz (commented):
Couple small comments/suggestions on the code.

templates/docs/layouts/fluid-breakout.md
```diff
@@ -0,0 +1,103 @@
...
+
+<div class="embedded-example"><a href="/docs/examples/layouts/fluid-breakout/fluid-breakout-no-aside/" class="js-example">
+View example of he fluid breakout layout without an aside
+</a></div>
+
+### Uses
+
+A couple of examples of where this layout might be useful. Both examples include a section with the regular 12 column grid for comparison.
+
+A large table that would require truncation if fitted within the regular 12 column grid:
```
bartaz, line 66: Because this example relies on custom parameters passed to fluid layout it would be good to mention that in the docs, so it is clear that this example required different params than default built in layout.

scss/_layouts_fluid-breakout.scss
lyubomir-popov: some of the styles - like padding - shouldn't be wrapped in supports, that's why it is wrapped in smaller pieces

templates/docs/layouts/fluid-breakout.md
lyubomir-popov: not needed in narrow layouts, I think I added it when I experimented with the wider area - as list items have no max width. Removing as not currently needed.

templates/docs/layouts/fluid-breakout.md
lyubomir-popov: added

scss/standalone/patterns_grid-breakout-full.scss
lyubomir-popov: no removed

scss/standalone/patterns_grid-breakout.scss
lyubomir-popov: removed

templates/docs/examples/patterns/card/_header-and-footer.html
```diff
@@ -0,0 +1,9 @@
+<div class="p-card">
+  <img class="p-card__thumbnail" src="https://res.cloudinary.com/canonical/image/fetch/f_auto,q_auto,fl_sanitize,c_fill,w_64,h_64/https://api.charmhub.io/api/v1/media/download/charm_B1WVq07SAhAIBJ72odJQ9fNBCFXk7QYT_icon__ce0ecd4c4fb29e9a8ff84971fbbba8552a822d53717632f2aa5ecbce9dfea108.png" alt="landscape-server" class="p-card__thumbnail" loading="lazy" width="32" height="32" />
```
bartaz, line 2: I feel like I commented on this before - please change the image to something from asset server, not claudinary.

lyubomir-popov: fixed

templates/docs/layouts/fluid-breakout.md
lyubomir-popov: reworded

scss/_layouts_fluid-breakout.scss
```diff
@@ -0,0 +1,216 @@
...
+        width: 100%;
+      }
+
+      @media (min-width: $l-fluid-breakout-3-col-threshold) {
+        padding-left: map-get($grid-margin-widths, large);
+        padding-right: map-get($grid-margin-widths, large);
+      }
+    }
+
+    .l-fluid-breakout__full {
```
bartaz, line 144: Do we need to have a separate class name for `__full`? "Full" sounds like a modifier, not an element. It seems that it's just a variant of `__main`. So maybe we could have it as `l-fluid-breakout__main.is-full-width`?

I think with layouts we should aim for a common structure of HTML and modify the layout by modifier classes, not by changing class names on children.

So in this case I think our default structure should be:

```html
<div class="l-fluid-breakout">
   <div class="l-fluid-breakout__aside"></div>
   <div class="l-fluid-breakout__main">
       <div class="l-fluid-breakout__item">
   </div>
</div>
```

Where `__aside` is optional (and can be before or after).
And `__main` is always there, but can be modified. If there is no aside, you add `.no-aside` to it.
If you want it to take full width you add `is-full-width` (instead of replacing it with `__full`).

Also, for an "item" class name we should use one consistent `l-fluid-breakout__item`, and not distinguish between `main-item` and `full-item`. From HTML point of view they are the same, we should be able to change their style based on parent modifier.

Would that be possible?

scss/_layouts_fluid-breakout.scss
```diff
@@ -0,0 +1,207 @@
...
+      }
+
+      .row {
+        max-width: 100%;
+        padding-left: 0;
+        padding-right: 0;
+      }
+    }
+
+    .l-fluid-breakout__main-item {
```
bartaz, line 146: For simplicity, can we rename it to `l-fluid-breakout__item`? I don't think we need to nest under main .

```suggestion
    .l-fluid-breakout__item {
```

bartaz (approved):
Thanks a lot!

PR #3511, opened 2021-01-25
Code snippet dropdowns

### LINKED ISSUE #3469: Implement dropdown menus for code snippet

sowasred2012:
## Done

Added example of using a select within a code snippet

Fixes #3469 

## QA

- Open [demo](https://vanilla-framework-3511.demos.haus/docs/examples/patterns/code-snippet/code-snippet-dropdown)
- Check that the example works when selecting different options from the menu
- Compare to the [Zeplin design](https://app.zeplin.io/project/5fa270d8eee4dbb267e579f1/screen/5ffc2fdcb65e6a99b78f9d47)
  - [Code](https://vanilla-framework-3511.demos.haus/docs/base/code)
  - [Component status](https://vanilla-framework-3511.demos.haus/docs/component-status)

### DIFF
--- a/scss/_patterns_code-snippet.scss
+++ b/scss/_patterns_code-snippet.scss
@@ -77,6 +77,7 @@ $color-snippet-heading-bg: rgba($color-x-dark, 0.08);
       // taken from the .u-table-cell-padding-overlap utility
       margin-bottom: -#{$table-cell-vertical-padding} !important;
       margin-top: calc(-#{$table-cell-vertical-padding} - 1px) !important;
+      min-width: min-content;
 
       // taken from the .is-dense button modifier
       padding-bottom: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
--- a/scss/_patterns_code-snippet.scss
+++ b/scss/_patterns_code-snippet.scss
@@ -63,5 +63,23 @@ $color-snippet-heading-bg: rgba($color-x-dark, 0.08);
       margin-bottom: 0;
       padding-top: 0;
     }
+
+    .p-code-snippet__dropdowns {
+      flex-shrink: 0;
+      margin-left: $sph-inner--small;
+    }
+
+    .p-code-snippet__toggle {
+      @extend %p-button--base;
+      @extend %p-button--with-icon;
+
+      // taken from the .u-table-cell-padding-overlap utility
+      margin-bottom: -#{$table-cell-vertical-padding} !important;
+      margin-top: calc(-#{$table-cell-vertical-padding} - 1px) !important;
+
+      // taken from the .is-dense button modifier
+      padding-bottom: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
+      padding-top: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
+    }
   }
 }
--- a/scss/_patterns_code-snippet.scss
+++ b/scss/_patterns_code-snippet.scss
@@ -69,9 +69,10 @@ $color-snippet-heading-bg: rgba($color-x-dark, 0.08);
       margin-left: $sph-inner--small;
     }
 
-    .p-code-snippet__toggle {
-      @extend %p-button--base;
-      @extend %p-button--with-icon;
+    .p-code-snippet__dropdown {
+      background-color: $color-transparent;
+      border-color: $color-transparent;
+      color: $color-dark;
 
       // taken from the .u-table-cell-padding-overlap utility
       margin-bottom: -#{$table-cell-vertical-padding} !important;
--- a/templates/docs/base/code.md
+++ b/templates/docs/base/code.md
@@ -60,6 +60,14 @@ View example of the code snippet
 View example of the code snippet
 </a></div>
 
+**Code snippet with dropdown**
+
+A [contextual menu](/docs/patterns/contextual-menu) can be included within a `.p-code-snippet__header`, and can be used alongside JS to switch between related code examples. The menu's toggle element should include the `.p-code-snippet__toggle` class.
+
+<div class="embedded-example"><a href="/docs/examples/patterns/code-snippet/code-snippet-dropdown" class="js-example">
+View example of the code snippet
+</a></div>
+
 **Toggling wrapping within a code snippet**
 
 By default, `<pre>` elements do not wrap content, but this can be overridden by adding the `.is-wrapped` utility class to a `.p-code-snippet__block`:
--- a/templates/docs/base/code.md
+++ b/templates/docs/base/code.md
@@ -62,7 +62,7 @@ View example of the code snippet
 
 **Code snippet with dropdown**
 
-A [contextual menu](/docs/patterns/contextual-menu) can be included within a `.p-code-snippet__header`, and can be used alongside JS to switch between related code examples. The menu's toggle element should include the `.p-code-snippet__toggle` class.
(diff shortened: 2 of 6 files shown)


bartaz:
On Zeplin it seems that we should have different background on the button when menu is open:
Screenshot 2021-01-26 at 07 53 59

I don't think we have a 'pressed' state on buttons ATM, so not sure if that's something we should add custom to code snippet button, or can it wait for when buttons have pressed state in Vanilla @lyubomir-popov ?

templates/docs/examples/patterns/code-snippet/_script.js
```diff
@@ -0,0 +1,34 @@
+function setupCodeSnippetOptions(dropdown) {
+  var dropdownLabel = dropdown.querySelector('span');
+  var targetMenu = document.getElementById(dropdown.getAttribute('aria-controls'));
+  var options = [].slice.call(targetMenu.querySelectorAll('a'));
```
bartaz, line 4: We shouldn't target options via tag name. Using class name would be better (we know it as it's part of the pattern), but I guess even better we could check for aria-controls attribute maybe? `[aria-controls]`? Because we rely on this attribute for the functionality anyway.

sowasred2012: good point, updated

sowasred2012:
> On Zeplin it seems that we should have different background on the button when menu is open:
> Screenshot 2021-01-26 at 07 53 59
> 
> I don't think we have a 'pressed' state on buttons ATM, so not sure if that's something we should add custom to code snippet button, or can it wait for when buttons have pressed state in Vanilla @lyubomir-popov ?

Agreed in standup to address in a separate PR that covers buttons in general.

sowasred2012:
After a debate - decided to go with `select`s for the dropdowns, rather than contextual menus.

scss/_patterns_code-snippet.scss
```diff
@@ -63,5 +63,25 @@ $color-snippet-heading-bg: rgba($color-x-dark, 0.08);
...
+      background-color: $color-transparent;
+      border-color: $color-transparent;
+      color: $color-dark;
+
+      // taken from the .u-table-cell-padding-overlap utility
+      margin-bottom: -#{$table-cell-vertical-padding} !important;
+      margin-top: calc(-#{$table-cell-vertical-padding} - 1px) !important;
+      min-width: min-content;
+
+      // taken from the .is-dense button modifier
```
bartaz, line 82: Does this still apply when it's a select rather than a button?

templates/docs/component-status.md
```diff
@@ -23,13 +23,20 @@ When we add, make significant updates, or deprecate a component we update their
...
+      <td>We add the ability to include a select within in a code snippet, allowing users to switch between related code examples.</td>
+    </tr>
     <tr>
       <th><a href="/docs/base/code#code-snippet">Code snippet</a></th>
       <td><div class="p-label--updated">Updated</div></td>
       <td>2.23.0</td>
       <td>We added a utility class for <code>.p-code-snippet__block</code> that allows content to wrap, rather than horizontally scroll: <code>.is-wrapped</code>.</td>
     </tr>
   </tbody>
+  </tbody>
```
bartaz, line 39:
```suggestion
```

Duplicated tbody?

templates/docs/examples/patterns/code-snippet/_script.js
```diff
@@ -0,0 +1,26 @@
...
+    var targetElement = document.getElementById(e.target.value);
+
+    togglePanel(targetElement, dropdown.options);
+  });
+}
+
+function togglePanel(targetPanel, options) {
+  for (var i = 0; i < options.length; i++) {
+    var panel = document.getElementById(options[i].value);
+    panel.classList.add('u-hide');
```
bartaz, line 12: Should we be toggling aria-hidden as well?

templates/docs/examples/patterns/code-snippet/_script.js
```diff
@@ -0,0 +1,26 @@
+function attachDropdownEvents(dropdown) {
```
bartaz, line 1: I think we tend to add a bit of a jsdoc comment on top of those in other examples, to give better idea of what the function does and what the params mean.
Could you have a look at some other example JS and use similar comment style?

templates/docs/base/code.md
```diff
@@ -60,6 +60,14 @@ View example of the code snippet
 View example of the code snippet
 </a></div>
 
+**Code snippet with dropdown**
```
bartaz, line 63: Now, that these are not  really 'dropdowns', does it still make sense to call them like this?

I wouldn't call them 'selects' (as that's HTML jargon), but what are they by function? Dropdowns? Menus? How does the spec call them?

templates/docs/examples/patterns/code-snippet/code-snippet-dropdown.html
```diff
@@ -0,0 +1,38 @@
...
+
+  <pre id="html-panel" class="p-code-snippet__block--numbered is-wrapped"><code>&#60;h1 class="p-heading--two"&#62;How to use code snippets&#60;/h1&#62;</code></pre>
+
+  <pre id="css-panel" class="p-code-snippet__block u-hide"><code>.p-heading--two {
+color: red;
+}</code></pre>
+
+  <pre id="js-panel" class="p-code-snippet__block u-hide"><code>console.log("Example 1");</code></pre>
+
+  <div class="p-code-snippet__header">
```
bartaz, line 28: Do we need the Output part? It doesn't seem to be relevant to the dropdown, so maybe we can drop it to simplify things.

bartaz:
Wow, switching to select made things so much better in terms of accessibility, simplicity of the code, etc. It was definitely a good call.

scss/_patterns_code-snippet.scss
sowasred2012: Yeah, I initially removed these when making the switch to selects, but there was way too much space around the element. Adding these back in solved that.

templates/docs/examples/patterns/code-snippet/code-snippet-dropdown.html
sowasred2012: Not _really_, but it is the output you'd get from the JS? :man_shrugging:

templates/docs/base/code.md
sowasred2012: [Here](https://discourse.ubuntu.com/t/code-snippet-pattern/19195) they're called "version switchers", in [the design](https://app.zeplin.io/project/5fa270d8eee4dbb267e579f1/screen/5ffc2fdcb65e6a99b78f9d47) they're called "dropdowns".

I did initially go with just calling them selects, but it felt inappropriate and like it was tied to a form. "Version switcher" won't suit every use case, but we could also go with "switcher", it's ultimately what the element does. That said, I think they can also be considered as "dropdowns", as it's _also_ what the element does. "Dropdown switcher"?

templates/docs/examples/patterns/code-snippet/_script.js
sowasred2012: Good point

templates/docs/examples/patterns/code-snippet/code-snippet-dropdown.html
bartaz: I find it bit confusing, as it's shown regardless of the lang chosen in the dropdown - it doesn't change, is not affected by the dropdown.

I feel that it would be better to get only what's needed into the example.

But, not a big deal. I don't mind if it says. Just something to consider.

bartaz (approved):
LGMT, thanks.

PR #3513, opened 2021-01-26
Add .is-overlapping utility for contextual menu dropdowns

sowasred2012:
## Done

Add `.is-overlapping` utility class to contextual menu pattern, allowing the menu to overlap and hide the button.

## QA

- Open [demo](https://vanilla-framework-3513.demos.haus/docs/examples/patterns/contextual-menu/overlapping)
- Click the button, see that the menu overlaps the button
- Check updated documentation:
  - [Docs](https://vanilla-framework-3513.demos.haus/docs/patterns/contextual-menu#overlapping)
  - [Component status](https://vanilla-framework-3513.demos.haus/docs/component-status)

### DIFF
--- a/templates/docs/examples/patterns/contextual-menu/overlapping.html
+++ b/templates/docs/examples/patterns/contextual-menu/overlapping.html
@@ -1,0 +1,26 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Contextual menu / Overlapping{% endblock %}
+
+{% block standalone_css %}patterns_contextual-menu{% endblock %}
+
+{% block content %}
+<span class="p-contextual-menu--left is-overlapping">
+  <button class="p-button--positive p-contextual-menu__toggle has-icon" aria-controls="menu-3" aria-expanded="false" aria-haspopup="true"><i class="p-icon--chevron-down is-light p-contextual-menu__indicator"></i><span>Take action</span></button>
+  <span class="p-contextual-menu__dropdown" id="menu-3" aria-hidden="true">
+    <span class="p-contextual-menu__group">
+      <a href="#" class="p-contextual-menu__link">Commission</a>
+      <a href="#" class="p-contextual-menu__link">Aquire</a>
+      <a href="#" class="p-contextual-menu__link">Deploy</a>
+    </span>
+    <span class="p-contextual-menu__group">
+      <a href="#" class="p-contextual-menu__link">Test hardware</a>
+      <a href="#" class="p-contextual-menu__link">Rescue mode</a>
+      <a href="#" class="p-contextual-menu__link">Mark broken</a>
+    </span>
+  </span>
+</span>
+
+<script>
+  {% include 'docs/examples/patterns/contextual-menu/_script.js' %}
+</script>
+{% endblock %}
--- a/scss/_patterns_contextual-menu.scss
+++ b/scss/_patterns_contextual-menu.scss
@@ -30,6 +30,10 @@
     &[aria-hidden='false'] {
       display: block;
     }
+
+    .is-overlapping & {
+      top: 0 !important;
+    }
   }
 
   // Alignment extension to align the menu to the left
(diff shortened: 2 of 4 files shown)


bartaz:
It's a UX thing, but would be good to have a bit of docs about would be good to have mentioned when should one type of contextual menu be used instead of another.

Do we need another option? Are there different use cases for both?
If there are - we should have them documented. If there are not, we should have one consistent style.

anthonydillon:
Is there an issue for this? Interested in the use case.

sowasred2012:
> Is there an issue for this? Interested in the use case.

@anthonydillon this span out of a discussion in Mattermost regarding https://github.com/canonical-web-and-design/vanilla-framework/pull/3511 - @lyubomir-popov didn't like the colors being switched between when hovering/opening the dropdown (particularly against the grey background), so we found this example in the [Material](https://material.io/components/buttons/web#using-buttons) docs (click the dropdown above the "Using Buttons" heading), and thought a similar approach would be a nice solution that could be used elsewhere too.

templates/docs/examples/patterns/contextual-menu/overlapping.html
```diff
@@ -0,0 +1,26 @@
...
+<span class="p-contextual-menu--left is-overlapping">
+  <button class="p-button--positive p-contextual-menu__toggle has-icon" aria-controls="menu-3" aria-expanded="false" aria-haspopup="true"><i class="p-icon--chevron-down is-light p-contextual-menu__indicator"></i><span>Take action</span></button>
+  <span class="p-contextual-menu__dropdown" id="menu-3" aria-hidden="true">
+    <span class="p-contextual-menu__group">
+      <a href="#" class="p-contextual-menu__link">Commission</a>
+      <a href="#" class="p-contextual-menu__link">Aquire</a>
+      <a href="#" class="p-contextual-menu__link">Deploy</a>
+    </span>
+    <span class="p-contextual-menu__group">
+      <a href="#" class="p-contextual-menu__link">Test harware</a>
```
wgx, line 16: Not relevant to the UX, but typo on Hardware

sowasred2012: good catch - fixed

bartaz:
My guess on different use cases is:

- having a contextual menu - list of actions that can be performed on something, it is triggered by some button with a label or icon, menu shows below the button that triggered it
- a replacement for "select" input - when the button that opens the menu shows one of the values, and clicking on one of the options in the menu, changes this value

If we want such distinction, we should make it clear in the docs and by providing a JS to the example to show exactly that this is what we mean.

sowasred2012:
> My guess on different use cases is:
> 
>     * having a contextual menu - list of actions that can be performed on something, it is triggered by some button with a label or icon, menu shows below the button that triggered it
> 
>     * a replacement for "select" input - when the button that opens the menu shows one of the values, and clicking on one of the options in the menu, changes this value
> 
> 
> If we want such distinction, we should make it clear in the docs and by providing a JS to the example to show exactly that this is what we mean.

Ah I see your point - yeah, I'll update the example/docs.

bartaz:
> Ah I see your point - yeah, I'll update the example/docs.

Sure, but I'd still like to have a confirmation from design/UX that this is the case.

Because it seems like we are adding a new variant, without clear use cases, just to 'make it look better' for code snippet.

We tend to have quite long and detailed working group process for proposals from other teams, where we at least try to discuss such changes in terms of why we do it, do we need another variant, what are the use cases. And now we just jump on it. @lyubomir-popov @wgx

lyubomir-popov:
@bartaz yeah I agree we need to discuss this in more detail. In this case it serves to navigate around the problem of not having a pressed state on any of the buttons in vanilla.

sowasred2012:
Closing, no longer needed since we're changing the code snippet dropdown to a select, and there's now a WG proposal for this variant: https://github.com/canonical-web-and-design/vanilla-framework/issues/3514

PR #3527, opened 2021-02-01
Code snippet: multiple dropdowns

### LINKED ISSUE #3470: Implement variant with stacked dropdowns

sowasred2012:
## Done

- Added ability to support multiple dropdowns in `p-code-snippet__header`
- Alternative layout available by adding the `is-stacked` utility to `p-code-snippet__header`

Fixes #3470 

## QA

- Open [demo](https://vanilla-framework-3527.demos.haus/docs/examples/patterns/code-snippet/dropdown-multiple)
- See that the dropdowns behave as expected, and match the [design](https://app.zeplin.io/project/5fa270d8eee4dbb267e579f1/screen/5ffc2fdcb65e6a99b78f9d47)
  - [Docs](https://vanilla-framework-3527.demos.haus/docs/base/code#code-snippet)
  - [Component status](https://vanilla-framework-3527.demos.haus/docs/component-status)


scss/_patterns_code-snippet.scss
```diff
@@ -1,6 +1,7 @@
 @import 'settings';
 
 $color-snippet-heading-bg: rgba($color-x-dark, 0.08);
+$header-v-spacing: $spv-inner--x-small--scaleable;
```
bartaz, line 4: Such variables are global in whole Vanilla, would be good to make the name prefixed by component name to make sure what it refers to.

bartaz:
Not sure if related to this PR, but I noticed that 'wrapping' examples don't have proper headings:

Screenshot 2021-02-03 at 10 06 58

bartaz:
For the 'stacked' example we mention in the docs that is meant for cases that require more horizontal space, I think it would be good to show it by making the stacked example with longer title, to demonstrate the desired use case:

Screenshot 2021-02-03 at 10 10 07

bartaz:
Because title is optional, we should provide the examples with dropdowns and without titles.

templates/docs/base/code.md
```diff
@@ -42,37 +42,43 @@ The code snippet pattern should be used to group related code examples, making t
...
 </a></div>
 
 **Code snippet with numbered lines**
 
-<div class="embedded-example"><a href="/docs/examples/patterns/code-snippet/code-snippet-numbered" class="js-example">
+<div class="embedded-example"><a href="/docs/examples/patterns/code-snippet/numbered" class="js-example">
 View example of the code snippet
 </a></div>
 
 **Code snippet with dropdown**
```
bartaz, line 63: We have quite a few of these subtitles. Maybe it would make sense to actually make them proper headings with `#### Code snippet ...`?

lyubomir-popov:
> For the 'stacked' example we mention in the docs that is meant for cases that require more horizontal space

Also I think it should be the default, as it is the variant that always displays the untruncated title and dropdowns.
The other one should only be used if The title isn't in danger of being overlapped by the dropdowns

bartaz:
> Also I think it should be the default, as it is the variant that always displays the untruncated title and dropdowns.

Should it? I believe in 80-90% of cases only one dropdown will be needed, it's hard to say if the code snippets will really have long titles. It would be a pity to drop the single dropdown to second line by default.

templates/docs/examples/patterns/code-snippet/dropdown-multiple.html
```diff
@@ -0,0 +1,72 @@
...
+        <option value="menu3-select">i386</option>
+      </select>
+
+      <select class="p-code-snippet__dropdown" name="menu2-select" id="menu2-select">
+        <option value="stable-panel">stable</option>
+        <option value="beta-panel">beta</option>
+        <option value="edge-panel">edge</option>
+      </select>
+
+      <select class="p-code-snippet__dropdown u-hide" name="menu3-select" id="menu3-select">
```
bartaz, line 23: If possible I'd like to avoid showing/hiding the selects. Not to complicate the examples too much.

Maybe we could make it work with the same list of options for both architectures?
The real implementation of that in products will need to be quite custom anyway, so I think we should make our example simpler - to focus mostly on showing the HTML structure, not the ways to change the selects with JS.

lyubomir-popov:
> Should it? I believe in 80-90% of cases only one dropdown will be needed, it's hard to say if the code snippets will really have long titles. It would be a pity to drop the single dropdown to second line by default.

Well atm I made the selects completely cover the title on mobile. Given we don't know the title's length, the inline version has to make compromises.

templates/docs/examples/patterns/code-snippet/dropdown-multiple.html
lyubomir-popov: I missed that, why is it hidden in the first place?

bartaz:
> > Should it? I believe in 80-90% of cases only one dropdown will be needed, it's hard to say if the code snippets will really have long titles. It would be a pity to drop the single dropdown to second line by default.
> 
> @bartaz Well atm I made the selects completely cover the title on mobile. Given we don't know the title's length, the inline version has to make compromises.

Ah, maybe on small screens it should always drop to stacked version?

lyubomir-popov:
> Ah, maybe on small screens it should always drop to stacked version?

Not on small screens, but as soon as the two touch - and we need a line in between. I don't think that's possible as there's no way to detect when two variale width elements touch, so you can insert a border in between.

bartaz:
> If we are ok with using JS, then this should really be one variant, not stacked / inline. Depends on where you want to draw the line in terms of complicating things.

I don't think we should provide that much of JS as part of Vanilla CSS. This JS is copied and rarely updated, we can't make sure it will be used in correct way and can't make sure that any changes are applied in projects.

We may think about adding it to React component, but still, feels quite a lot of logic to put into JS - listening to resize events, checking if elements fit next to each other or not. It feels like it may be hard to get it right.

I feel like at this point (especially that code snippets with dropdowns are not used anywhere yet) we should provide CSS to serve both inline and stacked variants and let products decide which is better for their use case.
If we change our mind about what the default should be, etc. We can probably change it later as well.

We will soon have the opportunity to use code snippet with dropdowns for out examples, so hopefully this will let us experiment with real use cases a bit.

templates/docs/base/code.md
sowasred2012: @bartaz should we move it out of the "Base" category, to "Components"? I agree with you and I'm changing these to headings, but maybe it would be good to have these sections visible in the side nav - which wouldn't happen while "Code snippet" is a child of the "Code" section.

templates/docs/examples/patterns/code-snippet/dropdown-multiple.html
sowasred2012: @bartaz @lyubomir-popov fair enough, it's a bit too convoluted as is - showing/hiding a select was a reasonably simple way to have the left most select determine what options were available in the right most select, which in turn would determine what is shown in the panel below.

templates/docs/base/code.md
bartaz: While, as discussed, we should restructure the headings to show variants of code snippet in side nav, I think we should still keep in in 'Base'.

It is a bit confusing, but would be weird to separate pages for 'Code' base styles and 'Code snippet' component.
In similar way 'Form' page contains both base styling for forms and components that we use with forms. Or "Table" page is about all things tables.

scss/_patterns_code-snippet.scss
bartaz: @sowasred2012 This doesn't seem to be addressed.

scss/_patterns_code-snippet.scss
```diff
@@ -51,37 +52,69 @@ $color-snippet-heading-bg: rgba($color-x-dark, 0.08);
...
+      width: min-content;
     }
 
     .p-code-snippet__dropdowns {
+      display: flex;
       flex-shrink: 0;
-      margin-left: $sph-inner--small;
+
+      &:first-child {
+        // if no header is present, use this to
```
bartaz, line 92: Can't the dropdowns always have `margin-left: auto`?

sowasred2012: Good point

templates/docs/examples/patterns/contextual-menu/default.html
```diff
@@ -1,7 +1,7 @@
 {% extends "_layouts/examples.html" %}
 {% block title %}Contextual menu / Default{% endblock %}
 
-{% block standalone_css %}patterns_contextual-menu{% endblock %}
+{% block standalone_css %}patterns_code-snippet{% endblock %}
```
bartaz, line 4: Why the change in contextual menu example?

sowasred2012: too many editor tabs open...

bartaz (approved):
LGTM

PR #3531, opened 2021-02-01
Automate version in title of component status page

bartaz:
## Done

Automate version in title of component status page to avoid updating it manually for release

## QA

- Open [demo](https://vanilla-framework-3531.demos.haus/docs/component-status)
  - make sure that the page heading shows vanilla version from package.json file
  - https://vanilla-framework-3531.demos.haus/docs/component-status

### DIFF
--- a/templates/docs/component-status.md
+++ b/templates/docs/component-status.md
@@ -10,9 +10,9 @@ context:
 
 When we add, make significant updates, or deprecate a component we update their status so that it’s clear what’s available to use. Check back here anytime to see current status information.
 
-### What's new in Vanilla 2.24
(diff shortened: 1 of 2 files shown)


sowasred2012 (approved):
LGTM - though it's a little weird that there's a version disconnect between the heading "What's new in Vanilla X.X" and rows in the table immediately below (in this case, the heading reads 2.23, and the dropdowns item below refers to 2.24).

Then again, that's not the problem this PR is trying to solve, and having the "future" version number in those rows makes our lives easier when it's time to relegate them to the "Previously in Vanilla" section.

PR #3538, opened 2021-02-08
Apply !important on all properties of the .u-vertically-center utuility class

### LINKED ISSUE #3537: Align utility does not have !important on all properties it sets
image

lyubomir-popov:
## Done

Fixes #3537

## QA

- Inspect percy, ensure nothing looks wrong

### DIFF
--- a/scss/_utilities_vertically-center.scss
+++ b/scss/_utilities_vertically-center.scss
@@ -4,8 +4,8 @@
 @mixin vf-u-vertically-center {
   .u-vertically-center {
     align-items: center !important;
-    display: grid;
-    justify-content: flex-start;
+    display: grid !important;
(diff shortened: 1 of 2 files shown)


bartaz (approved):
LGTM

bartaz (changes requested):
Looking at percy it seems to be breaking some things.

Maybe these are some weird edge cases, but we should either update our examples to make them work as expected with this fix, or we should do it a bit differently.

lyubomir-popov:
@bartaz ready for another look. The percy diff is down to this:

image

Seems like a 1px aliasing thing

bartaz (approved):
LGTM, thanks for updates!

PR #3541, opened 2021-02-11
POC replacing CodePen with example JS iframe

### LINKED ISSUE #3522: Investigate updating "example.js" for new examples

bartaz:
## Done

POC demo: https://vanilla-framework-3541.demos.haus/docs/patterns/accordion

please note, in this WIP prototype only first code snippet example on the page fully works (dropdowns to choose language), other just render, but dropdowns will not work

Fixes #3522 

## QA

- [Add QA steps]

### DIFF
--- a/templates/_layouts/docs.html
+++ b/templates/_layouts/docs.html
@@ -152,4 +152,10 @@
     <script src="/static/js/scripts.js"></script>
     <script defer src="https://static.codepen.io/assets/embed/ei.js"></script>
     <script defer src="/static/js/example.js"></script>
+    <script>
+      {% include 'docs/examples/patterns/code-snippet/_script.js' %}
+    </script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/prism/1.23.0/components/prism-core.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/prism/1.23.0/plugins/autoloader/prism-autoloader.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/prism/1.23.0/plugins/keep-markup/prism-keep-markup.min.js"></script>
 {% endblock %}
--- a/templates/static/js/example-codepen.js
+++ b/templates/static/js/example-codepen.js
@@ -52,17 +52,8 @@
     code.appendChild(document.createTextNode(source));
 
     var pre = document.createElement('pre');
-    pre.classList.add('p-code-snippet__block');
-    // FIXME: this doesn't work with multiple examples on a page, we need unique ids or another way of selecting panels to show/hide
-    pre.id = 'panel-' + lang;
-    pre.style.maxHeight = '300px';
-    if (lang !== 'html') {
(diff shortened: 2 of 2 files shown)


anthonydillon:
Looks great! Here are some general comments while playing around with it:
- It's key that it's to easy to copy the code from the example. At the moment, if I double click I select the line and if ctrl+a I get all the content of the page. Meaning I would need to highlight and scroll down over all the code which is cumbersome.
- Could we render the resulting height dynamically based on the height of the rendered component?
- The horizontal lines are distracting in the example result
- Should the result have a border of some sort to see where it ends?
- Example titles should be parsed (I am sure you are aware of this)
- The "Accordion / Headings" example toggle to JS doesn't work

bartaz:
Prototype notes:

- we need to make sure code snippet dropdowns code works with multiple code snippets on the page (they need unique ids or selectors based on parent)
- we need to make sure code snippet dropdowns code and prism is initialized after code snippets are rendered
- iframe size will be hardcoded (may be to small for some examples, or too big for others) - resizing the iframe requires additional work (and caused issues in past)
- we need to separately implement:
  - link to open example in separate window
  - 'edit in codepen' link
  - baseline toggle?

Instead of embedding iframe we may consider embedding examples directly into the document.

Benefits:
 
- no need to deal with separate iframe document, scrolling, trying to calculate height, etc
- we can have baseline grid in example base, but not in embedded example easily

Cons:

- examples that need JS or additional CSS will need to have these added to the document which may collide with rest of the page scripts or styles, so we would need to carefully write any JS or styles for examples

bartaz:
@anthonydillon Thanks for comments
 
> * It's key that it's to easy to copy the code from the example. At the moment, if I double click I select the line and if ctrl+a I get all the content of the page. Meaning I would need to highlight and scroll down over all the code which is cumbersome.

That's unfortunate side-effect of getting rid of copy button. With code snippet being just another element on the page, browser native copy paste handles it.
There are ways around that but they would require JS (to select whole code on double click).
This is not direcly related to CodePen replacement, but is something affecting code snippet in general, so we should address it separately.

> * Could we render the resulting height dynamically based on the height of the rendered component?

Yes, we should be able to do it (we did this before with previous implementation of example-js), but I recall there were some issues of incorrectly sized examples as well. So we need to be aware of that.

> * The horizontal lines are distracting in the example result

Yes, we discussed it with @lyubomir-popov, we are going to try different background for embedded examples

> * Should the result have a border of some sort to see where it ends?

We discussed having gray backround for the examples, which (if example frame is scaled to the size of example) would naturally serve as a border: https://app.zeplin.io/project/60240a4e8ddd37098080b001/screen/6024ef91e461e2086841bd14

> * Example titles should be parsed (I am sure you are aware of this)

Yes, currently we take the `<title>` of the example document, would be good to parse only the relevant part of it.

> * The "Accordion / Headings" example toggle to JS doesn't work

Yes, current quick&dirty solution works well only in first code snippet on a page (as they share some ids)

bartaz:
Work continues in #3570

PR #3543, opened 2021-02-12
Add .is-dark support for inline code element

### LINKED ISSUE #3540: Code inline: darker colour, no gap on wrapping; all code: no border radius
before:
image

after:
image

code snippet for comparisson:
image

sowasred2012:
## Done

- Updated the background color of the inline code element to match the transparent color in use on the new code snippet
- Added support for code elements within an `is-dark` parent

Fixes #3540 

## QA

- Open [demo](https://vanilla-framework-3543.demos.haus/docs/examples/base/code-inline)
- Check that the inline code element looks good, and sits on the baseline.
- Open [dark demo](https://vanilla-framework-3543.demos.haus/docs/examples/base/code-inline-dark)
- See that the code element is nicely visible
- Check the [updated application structure](https://vanilla-framework-3543.demos.haus/docs/examples/layouts/documentation#notifications), see that "l-navigation >l-navigation__drawer" is visible in the side bar.

### DIFF
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -30,10 +30,6 @@ $code-inline-padding: 0.25rem;
     color: inherit;
     line-height: map-get($line-heights, default-text) - $code-inline-padding;
     padding: calc(#{$code-inline-padding} - 1px) $code-inline-padding;
-
-    .is-dark & {
-      background-color: $color-code-background-dark;
-    }
   }
 
   code,
@@ -57,6 +53,7 @@ $code-inline-padding: 0.25rem;
     background: none;
     box-shadow: none;
     display: inline-block;
+    line-height: map-get($line-heights, default-text);
     margin-left: 0;
     margin-right: 0;
     padding: 0;
@@ -74,6 +71,15 @@ $code-inline-padding: 0.25rem;
     text-shadow: none;
     white-space: pre;
   }
+
+  // dark theme
+  // stylelint-disable selector-no-qualifying-type
+  [class*='--dark'] code,
+  code.is-dark {
+    background-color: $color-code-background-dark;
+    color: $colors--dark-theme--text-default;
+  }
+  // stylelint-enable selector-no-qualifying-type
 
   %leading-linux-prompt-icon {
     &::before {
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -3,6 +3,7 @@
 $color-pre-bg: rgba($color-x-dark, 0.03);
 $digit-width: 1ch; // measured width of one character in the monospaced font
 $code-sidebar-width: calc(#{$sph-inner} + (4 * #{$digit-width})) !default;
+$code-inline-padding: 0.25rem;
 
 // Base code styles
 @mixin vf-b-code {
@@ -23,9 +24,11 @@ $code-sidebar-width: calc(#{$sph-inner} + (4 * #{$digit-width})) !default;
   code,
   kbd,
   samp {
-    background-color: $color-mid-x-light;
-    box-decoration-break: clone;
-    padding: $sph-inner--x-small;
+    background-color: $color-code-background;
+    border-radius: $border-radius;
+    box-decoration-break: slice;
+    line-height: map-get($line-heights, default-text) - $code-inline-padding;
+    padding: calc(#{$code-inline-padding} - 0.5px) $code-inline-padding;
   }
 
   code,
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -27,8 +27,13 @@ $code-inline-padding: 0.25rem;
     background-color: $color-code-background;
     border-radius: $border-radius;
     box-decoration-break: slice;
+    color: inherit;
     line-height: map-get($line-heights, default-text) - $code-inline-padding;
     padding: calc(#{$code-inline-padding} - 1px) $code-inline-padding;
+
+    .is-dark & {
+      background-color: $color-code-background-dark;
+    }
   }
 
   code,
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -28,7 +28,7 @@ $code-inline-padding: 0.25rem;
     border-radius: $border-radius;
     box-decoration-break: slice;
     line-height: map-get($line-heights, default-text) - $code-inline-padding;
-    padding: calc(#{$code-inline-padding} - 0.5px) $code-inline-padding;
+    padding: calc(#{$code-inline-padding} - 1px) $code-inline-padding;
   }
 
   code,
--- a/scss/_settings_colors.scss
+++ b/scss/_settings_colors.scss
@@ -40,6 +40,7 @@ $active-background-opacity-amount: 0.15;
 // NON-SEMANTIC COLOURS
 $color-label-validated: #006b75;
 $color-code-background: rgba($color-x-dark, 0.03);
+$color-code-background-dark: rgba($color-x-light, 0.3);
 $color-code-heading-background: rgba($color-x-dark, 0.08);
 
 $states: (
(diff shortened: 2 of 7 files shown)


bartaz:
As far as I understand from catch-up we had, we should also provide the support for inline code in dark strips.

Could we add an example of inline code in `p-strip--dark` and make sure it works as expected?

templates/docs/examples/layouts/application/structure.html
```diff
@@ -19,7 +19,9 @@
         <button class="js-menu-pin is-dense u-no-margin u-hide--small">Pin</button>
         <button class="js-menu-close is-dense u-no-margin u-hide--medium">Close</button>
       </p>
-      <code>l-navigation &gt;<br> l-navigation__drawer</code>
+      <div class="p-card">
```
bartaz, line 22: I'd like to avoid such workarounds. Putting `p-card` here is quite unusual.

I guess the easiest way to avoid would be to get rid of the black background on this element (it's just added there for the sake of example) and use some other (`$color-mid-dark`?) that would work with inline code.

sowasred2012:
@bartaz I thought we had decided to wait to add that change until @lyubomir-popov had done some investigation, [the meeting notes](https://github.com/canonical-web-and-design/meeting-notes/issues/747) under "Inline code on dark background" back this up (I think I was just inspired to step up my meeting note taking game).

templates/docs/examples/layouts/application/structure.html
sowasred2012: Fair point - I tried with `$color-mid-dark`, but it seemed too dark, so I went with `$color-mid-x-light`, let me know what you think.

bartaz:
> @bartaz I thought we had decided to wait to add that change until @lyubomir-popov had done some investigation, [the meeting notes](https://github.com/canonical-web-and-design/meeting-notes/issues/747) under "Inline code on dark background" back this up (I think I was just inspired to step up my meeting note taking game).

Ok, my undestanding was that the mentioned investigation was about code snippet dark theme, not the inline code.
We have two transparent bg colors in dark theme (used for example in side navigation) that we could (at least temporarly) use for the background: https://github.com/canonical-web-and-design/vanilla-framework/blob/master/scss/_settings_colors.scss#L97-L98

But even if we fell it's out of scope for this PR, it would be good to add a dark inline code example, so we can start having Percy screenshoting it and we can be sure we don't introduce regressions.

templates/docs/examples/layouts/application/structure.html
lyubomir-popov: @bartaz @sowasred2012 So it turns out we do actually use inline code on a dark background inside vanilla itself. 
I don't think changing the aside background isa good solution, this makes it look as if that's the default aside colour we recommend - and it doesn't work well with the shadow.

Can we just take the classes out of the inline code and haveit just as text?

image

lyubomir-popov:
@sowasred2012 @bartaz I belive we agreed to let the parent determine the text colour, and flip rgba(0,0,0, .03) to rgba(255,255,255, .03) when the parent has a class `is-dark`. And not worry that this is background hihglight is not very discernable on pure black - as we'll revisit those opacities when properly test a dark theme, and it isn't an accessibility problem anyway.

templates/docs/examples/layouts/application/structure.html
bartaz: @lyubomir-popov we know that we use it, this is the screenshot that allowed us to notice that we have this issue.

Previously we used black, which also is probably not a color we 'recommend' there - so this is really just a placeholder color so we can show the box structure of the layout. We can try using fade-out red as in grid examples to give it more of debugging look.

But yes, because these `<code>` elements are causing us issues here, I'm fine with replacing them with text (of maybe `<h5>` or something.

templates/docs/examples/base/code-inline-dark.html
```diff
@@ -0,0 +1,14 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Code - inline / dark{% endblock %}
+
+{% block standalone_css %}base{% endblock %}
+
+{% block content %}
+<div class="p-strip--dark is-dark">
```
bartaz, line 7: The dark strip uses only `p-strip--dark` class. It shouldn't have `is-dark` as well.

scss/_base_code.scss
```diff
@@ -23,9 +24,16 @@ $code-sidebar-width: calc(#{$sph-inner} + (4 * #{$digit-width})) !default;
...
-    box-decoration-break: clone;
-    padding: $sph-inner--x-small;
+    background-color: $color-code-background;
+    border-radius: $border-radius;
+    box-decoration-break: slice;
+    color: inherit;
+    line-height: map-get($line-heights, default-text) - $code-inline-padding;
+    padding: calc(#{$code-inline-padding} - 1px) $code-inline-padding;
+
+    .is-dark & {
```
bartaz, line 34: Default text color change happens in `p-strip--dark` class, so it should work with that, not `is-dark`. Same as icons currently support `p-strip--dark`.

templates/docs/examples/layouts/application/structure.html
```diff
@@ -13,13 +13,14 @@
     <div class="u-clearfix"><code>l-navigation-bar</code> <button class="u-float-right js-menu-toggle is-dense u-no-margin">Menu</button></div>
   </div>
 
-  <header class="l-navigation">
+  <header class="l-navigation is-dark">
```
bartaz, line 16: This shouldn't be needed, right?

sowasred2012: No it shouldn't, overlooked it - removed!

scss/_settings_colors.scss
```diff
@@ -40,6 +40,7 @@ $active-background-opacity-amount: 0.15;
 // NON-SEMANTIC COLOURS
 $color-label-validated: #006b75;
 $color-code-background: rgba($color-x-dark, 0.03);
+$color-code-background-dark: rgba($color-x-light, 0.3);
```
bartaz, line 43: Is this supposed to be `0.3` or should it be `0.03` as in light version? @sowasred2012 @lyubomir-popov

bartaz: Ok, `0.03` is almost invisible

Screenshot 2021-02-16 at 07 30 23

So I guess 0.3 is better, we can always adjust later

Screenshot 2021-02-16 at 07 30 31

templates/docs/component-status.md
```diff
@@ -29,6 +29,12 @@ When we add, make significant updates, or deprecate a component we update their
       <td>2.24.0</td>
       <td>We added the ability to accommodate multiple selects within a <code>.p-code-snippet__header</code>, either inline, using the <code>.is-stacked</code> utility to class to have the dropdowns below the title.</td>
     </tr>
+    <tr>
+      <th><a href="/docs/base/code#dropdowns">Code inline - dark</a></th>
```
bartaz, line 33: Wrong link
```suggestion
      <th><a href="/docs/base/code#inline">Code inline - dark</a></th>
```

bartaz (approved):
One suggestion for fixing link in docs, but LGTM otherwise, thanks!

templates/docs/examples/base/code-inline-dark.html
```diff
@@ -0,0 +1,14 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Code - inline / dark{% endblock %}
+
+{% block standalone_css %}base{% endblock %}
+
+{% block content %}
+<div class="p-strip--dark">
```
bartaz, line 7: The standalone example for dark code doesn't show dark strip (because it only contains base styles).

And while in case of components we would just add any other needed component to standalone CSS, we don't want to do this in case of base examples (because we want them to work with just the base).

So I guess the easiest way to workaround it would be to add inline background here (to make sure background is dark even without strip CSS):

```suggestion
<div class="p-strip--dark" style="background: #111">
```

PR #3560, opened 2021-02-16
Add small capitalised style, utilities to align small and paragraph text

lyubomir-popov:
## Done

Add small capitalised style, utilities to align small and paragraph text

## QA

paste the following markup into any example:
```
<div class="row">
  <div class="col-1">
    <p>paragraph</p>
  </div>
  <div class="col-1">
    <div class="u-align-text--small-to-default p-text--small">small</div>
  </div>
  <div class="col-9">
    <p class="p-text--x-small-capitalized u-align-text--x-small-to-default">small capitalised text aligned to a paragraph</p>
  </div>
```
- verify you see this:
image

### DIFF
--- a/scss/_base_typography.scss
+++ b/scss/_base_typography.scss
@@ -84,6 +84,13 @@
     @extend %x-small-text;
   }
 
+  .p-text--x-small-capitalized {
+    @extend %x-small-text;
+
+    font-weight: $font-weight-bold;
+    text-transform: uppercase;
+  }
+
   //@section Adjusted spacing for headings (or a p) following a paragraph
   p:not([class*='p-heading--']):not([class*='p-muted-heading']) {
     & + h1,
--- a/scss/_utilities_content-align.scss
+++ b/scss/_utilities_content-align.scss
@@ -58,4 +58,12 @@
       text-align: right !important;
     }
   }
+
+  .u-align-text--small-to-default {
+    padding-top: map-get($nudges, nudge--small) + $sp-unit;
+  }
+
+  .u-align-text--x-small-to-default {
+    padding-top: map-get($nudges, nudge--x-small) + $sp-unit;
+  }
 }


bartaz:
@lyubomir-popov Added this as a follow up to document these class names: https://github.com/canonical-web-and-design/vanilla-framework/issues/3562

bartaz (approved):
LGTM

scss/_utilities_content-align.scss
```diff
@@ -58,4 +58,12 @@
       text-align: right !important;
     }
   }
+
+  .u-align-text--small-to-default {
+    padding-top: map-get($nudges, nudge--small) + $sp-unit;
+  }
+
+  .u-align-text--x-small-to-default {
```
anthonydillon, line 66: What are all these undocumented utility classes? They are like a riddle. What is small to default even mean?

scss/_base_typography.scss
```diff
@@ -84,6 +84,13 @@
     @extend %x-small-text;
   }
 
+  .p-text--x-small-capitalized {
+    @extend %x-small-text;
+
+    font-weight: $font-weight-bold;
+    text-transform: uppercase;
```
anthonydillon, line 91: Why not use create a `u-capitalise` class and put the contents in a small element?

scss/_utilities_content-align.scss
lyubomir-popov: The goal is to make it possible to align small text to a paragraph. Without it, they are slightly misaligned because of the difference in line-heights (first row)

image

Compare with the second row where they are perfectly aligned.

I can change the name to something less cryptic? Just wasn't sure what to call it.

scss/_base_typography.scss
lyubomir-popov: For the smallest text (the one we use as table headers) two things are needed - capitalisation and a change in font-weight, as the light font becomes too unreadable at that size. So if we introduce u-capitalise, we'd have to ask people to always add u-bold or something similar - which seems asking for trouble as it is easy to forget.

PR #3561, opened 2021-02-16
Add light and dense version of hr

lyubomir-popov:
## Done

- add light hr
- add u-no-margin--bottom that compensaes forthe hr 1px height, thereby avoiding a drift from the baseline grid
- add hr centered (and same width) as a row, which means instead of `.row>hr` or `.u-fixed-width>hr` you can just write `hr.is-fixed-width`

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/3510

## QA

- Paste this [code](https://pastebin.canonical.com/p/XYm8qkrSs6/) (pastebin link):
- Verify you see the following:
image

### DIFF
--- a/scss/_base_hr.scss
+++ b/scss/_base_hr.scss
@@ -5,7 +5,7 @@
   hr {
     @extend %hr;
 
-    margin-bottom: calc(#{$spv-inner--scaleable} - 1px);
+    margin-bottom: calc(#{$spv-inner--small} - 1px);
 
     // stylelint-disable-next-line selector-no-qualifying-type
     &.u-no-margin--bottom {
--- a/scss/_base_hr.scss
+++ b/scss/_base_hr.scss
@@ -7,14 +7,10 @@
 
     margin-bottom: calc(#{$spv-inner--scaleable} - 1px);
 
-    & + p {
-      margin-top: -$sp-unit;
-    }
-
     // stylelint-disable-next-line selector-no-qualifying-type
-    &.is-dense {
+    &.u-no-margin--bottom {
       // compensate for hr thickness, to make sure it doesn't drift from baseline grid
-      margin-bottom: -1px;
+      @extend %u-no-margin--bottom--hr;
     }
 
     // stylelint-disable-next-line selector-no-qualifying-type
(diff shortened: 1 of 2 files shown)


scss/_base_hr.scss
```diff
@@ -3,16 +3,28 @@
...
-    position: relative;
-    width: 100%;
 
     & + p {
       margin-top: -$sp-unit;
     }
+
+    // stylelint-disable-next-line selector-no-qualifying-type
+    &.is-dense {
+      margin-bottom: -1px;
```
bartaz, line 16: Adding negative bottom margin seems a bit error prone. Are you sure it's gonna work as expected regardless of what is put after `hr`?

bartaz: Also, is this expected that dense variant has no space below the hr line, but has standard spacing above it?

Screenshot 2021-02-16 at 13 53 46

bartaz:
Could you also rebase this PR? it seems it's on top of old version of master (without code styling) so unrelated changes show up in Percy.

scss/_base_placeholders.scss
```diff
@@ -131,7 +139,7 @@
   }
 
   %u-no-margin--bottom {
-    &:not(h1):not(h2):not(h3):not(h4):not(h5):not(h6):not(p):not(small):not([class*='p-heading']) {
+    &:not(hr):not(h1):not(h2):not(h3):not(h4):not(h5):not(h6):not(p):not(small):not([class*='p-heading']) {
```
bartaz, line 142: I know it was not added in this PR, but wow, didn't know we have something like that...
This feels really arbitrary list. Hard to know if it's up to date. Would be better to have all no-margin--bottom related code in one place, not scattered exceptions around codebase.

Anyway, something for other time maybe.

lyubomir-popov: just wanted to avoid one !important overriding another !important - I gues it is not technically needed, the cascade can take care of that

scss/_base_hr.scss
lyubomir-popov: the space above comes from the p that precedes it, space underneath intentional

bartaz (approved):
LGTM, thanks

PR #3569, opened 2021-02-17
Show ellipsis on buttons that launch modals

### LINKED ISSUE #3566: Buttons launching dialogs should have ellipsis

sowasred2012:
## Done

- Updated buttons in modal and contextual menu examples with an ellipsis
- Documented the reason why the ellipsis should be included

Fixes #3566

## QA

- Open: 
  - [Default modal](https://vanilla-framework-3569.demos.haus/docs/examples/patterns/modal/default)
  - [Scrolling modal](https://vanilla-framework-3569.demos.haus/docs/examples/patterns/modal/modal-scroll)
  - [z-index example](https://vanilla-framework-3569.demos.haus/docs/examples/templates/z-index)
  - [Contextual menu](https://vanilla-framework-3569.demos.haus/docs/examples/patterns/contextual-menu/default)
  - [Contextual menu (dark)](https://vanilla-framework-3569.demos.haus/docs/examples/patterns/contextual-menu/dark)
- See that an ellipsis is present on buttons that launch a dialog on click.
- Review documentation updates:
  - [Modal](https://vanilla-framework-3569.demos.haus/docs/patterns/modal#accessibility)
  - [Contextual menu](https://vanilla-framework-3569.demos.haus/docs/patterns/contextual-menu#accessibility)

### DIFF
--- a/templates/docs/examples/patterns/contextual-menu/dark.html
+++ b/templates/docs/examples/patterns/contextual-menu/dark.html
@@ -5,7 +5,7 @@
 
 {% block content %}
 <span class="p-contextual-menu--left is-dark">
-    <button class="p-contextual-menu__toggle" aria-controls="menu-3" aria-expanded="false" aria-haspopup="true">Take action</button>
+    <button class="p-contextual-menu__toggle" aria-controls="menu-3" aria-expanded="false" aria-haspopup="true">Take action...</button>
     <span class="p-contextual-menu__dropdown" id="menu-3" aria-hidden="true">
         <span class="p-contextual-menu__group">
             <a href="#" class="p-contextual-menu__link">Commission</a>
--- a/templates/docs/examples/patterns/contextual-menu/dark.html
+++ b/templates/docs/examples/patterns/contextual-menu/dark.html
@@ -5,7 +5,7 @@
 
 {% block content %}
 <span class="p-contextual-menu--left is-dark">
-    <button class="p-contextual-menu__toggle" aria-controls="menu-3" aria-expanded="false" aria-haspopup="true">Take action...</button>
+    <button class="p-contextual-menu__toggle" aria-controls="menu-3" aria-expanded="false" aria-haspopup="true">Take action&hellip;</button>
     <span class="p-contextual-menu__dropdown" id="menu-3" aria-hidden="true">
         <span class="p-contextual-menu__group">
             <a href="#" class="p-contextual-menu__link">Commission</a>
--- a/templates/docs/examples/patterns/modal/modal-scroll.html
+++ b/templates/docs/examples/patterns/modal/modal-scroll.html
@@ -4,7 +4,7 @@
 {% block standalone_css %}patterns_modal{% endblock %}
 
 {% block content %}
-<button id="showModal" aria-controls="modal">Show modal</button>
+<button id="showModal" aria-controls="modal">Show modal...</button>
 
 <p>An <em>application</em> is typically a long-running service that is accessible over the network. Applications are the centre of a Juju deployment. Everything within the Juju ecosystem exists to facilitate them.</p>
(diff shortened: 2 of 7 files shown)


anthonydillon:
This is strange behaviour. Why are we adding content to an element that is not in focus? Is this a UX pattern?

sowasred2012:
> This is strange behaviour. Why are we adding content to an element that is not in focus? Is this a UX pattern?

No... I got the wrong end of the stick, I'm updating this now.

templates/docs/examples/patterns/contextual-menu/dark.html
```diff
@@ -5,7 +5,7 @@
 
 {% block content %}
 <span class="p-contextual-menu--left is-dark">
-    <button class="p-contextual-menu__toggle" aria-controls="menu-3" aria-expanded="false" aria-haspopup="true">Take action</button>
+    <button class="p-contextual-menu__toggle" aria-controls="menu-3" aria-expanded="false" aria-haspopup="true">Take action...</button>
```
bartaz, line 8: Make it an ellipsis, not 3 full stops (in all examples plz):

```suggestion
    <button class="p-contextual-menu__toggle" aria-controls="menu-3" aria-expanded="false" aria-haspopup="true">Take action…</button>
```

sowasred2012: Ah yes - done

templates/docs/examples/patterns/modal/modal-scroll.html
```diff
@@ -4,7 +4,7 @@
 {% block standalone_css %}patterns_modal{% endblock %}
 
 {% block content %}
-<button id="showModal" aria-controls="modal">Show modal</button>
```
anthonydillon, line 7: Shouldn't we use `&hellip;`

sowasred2012: Yes it should - updated!

anthonydillon (approved):
LGTM thanks for the encode update

wgx:
I ❤️ this.. the user will have a good idea that there's another step before their action happens. Most users will infer from this that the button is the first part of a multi-part process and there will likely be a second step to confirm or cancel the action.

sowasred2012:
@wgx @lyubomir-popov one thing I should have mentioned - @bartaz and I were discussing this yesterday, and we couldn't agree on whether the contextual menu dropdown needed the ellipsis or not. 

The [WAI advice](https://www.w3.org/TR/wai-aria-practices/#button) suits the modal use case perfectly, but a button + contextual menu, like in the example, is arguably similar in function to a dropdown navigation, so wouldn't they also have ellipses?

lyubomir-popov:
> The [WAI advice](https://www.w3.org/TR/wai-aria-practices/#button) suits the modal use case perfectly, but a button + contextual menu, like in the example, is arguably similar in function to a dropdown navigation, so wouldn't they also have ellipses?

Good question, I wonder in what way a button that triggers a contextual menu is different from a select - other than not supporting a bunch of the accessibility requirements for a select.

sowasred2012:
> > The [WAI advice](https://www.w3.org/TR/wai-aria-practices/#button) suits the modal use case perfectly, but a button + contextual menu, like in the example, is arguably similar in function to a dropdown navigation, so wouldn't they also have ellipses?
> 
> Good question, I wonder in what way a button that triggers a contextual menu is different from a select - other than not supporting a bunch of the accessibility requirements for a select.

We do also have the [contextual menu with indicator pattern](https://vanillaframework.io/docs/examples/patterns/contextual-menu/with-indicator) - the indicator achieves a similar thing as the ellipsis in this case (informing the user something will most likely appear after an interaction), plus makes it seem more like a select. Should this pattern (with either ellipsis or indicator) just be a select, as we settled on for the code snippet dropdowns?

Issue #3557, opened 2021-02-16
Make the height of the embedded example adjust to content


bartaz:

sowasred2012:
Closed by https://github.com/canonical-web-and-design/vanilla-framework/pull/3570

PR #3570, opened 2021-02-17
Use code snippet and iframe for examples instead of CodePen embeds

### LINKED ISSUE #3556: Update example embedding code to replace CodePen

bartaz:
## Done

- Replaces CodePen embeds with code snippet and custom iframe
- Adds Prism to allow syntax highlighting of the examples
- Resizes the embedded iframes to the (estimated) height of the example

Fixes #3556 #3557

## QA

- Open [demo](https://vanilla-framework-3570.demos.haus/docs)
- Browse the docs and check embedded examples
- Make sure to verify various different pages
- Find examples with JS, make sure you can change the code snippet code via dropdown
- Find examples of different heights and verify they render at expected height
- Check in both Chrome and Firefox to make sure embed code works

## Screenshots

Screenshot 2021-02-19 at 13 56 54
Screenshot 2021-02-19 at 13 58 24
Screenshot 2021-02-19 at 13 58 07


bartaz:
@lyubomir-popov There are some issues with our gray background for embedded examples.

If we use CodeSnippet header color, this means code snippet examples 'fade' into the background:

Screenshot 2021-02-18 at 12 10 03

The `body` still has white background (it needs to be white to properly show components that don't define background), so we have gray 'border' and white body inside. In many cases (components with bottom margins), the body white background will be extended by the amount of the margin:
Screenshot 2021-02-18 at 12 10 29
Screenshot 2021-02-18 at 12 10 36

And the 'full screen' layout examples are problematic, because they don't have margin around body (they need to be really full screen), so in these cases they blend with the page because there is no border around iframe:

Screenshot 2021-02-18 at 12 11 54

lyubomir-popov:
@bartaz how about a #e5e5e5 border that wraps the codesnippet and example area as in https://zpl.io/a7jgxOR

lyubomir-popov (approved):
LGTM!

templates/_layouts/docs.html
```diff
@@ -152,6 +152,10 @@
     </div>
 
     <script src="/static/js/scripts.js"></script>
-    <script defer src="https://static.codepen.io/assets/embed/ei.js"></script>
+    <script>
+      window.Prism = window.Prism || {};
+	    Prism.manual = true;
```
sowasred2012, line 157:
```suggestion
      window.Prism = window.Prism || {};
      Prism.manual = true;
```

templates/static/js/example.js
```diff
@@ -1,44 +1,22 @@
...
-
   document.addEventListener('DOMContentLoaded', function () {
     var examples = document.querySelectorAll('.js-example');
 
     // IE11 doesn't support forEach on NodeList, CodePen doesn't support IE as well
     // so we don't want to load CodePen when forEach is not supported
     if (examples.forEach) {
-      examples.forEach(renderExample);
+      examples.forEach(fetchExample);
     }
```
sowasred2012, line 9: Do we still not want to support IE11 here? If we're not using CodePen, we can work around the NodeList/forEach issue with:
```
[].slice.call(examples).forEach(fetchExample(example))
```

sowasred2012 (approved):
LGTM, just one question

templates/static/js/example.js
bartaz: That's true, we can try to support IE here.

I did a quick look, and unfortuanely it's not just this `forEach` - I also use `element.closest()` for the dropdown functionality, which doesn't work in IE. And is a bit more annoying to polyfill than forEach.

So, I'd like to keep IE support out of this PR (just to replace CodePen that didn't have IE support either).
But we can have a look at IE support for this later. My worry is, that the experience will not be really nice anyway, and it may not be worth putting lots of work into it for the very small % of users we may have.

They always have a link to full example.

sowasred2012: Oh true, I didn't spot the `closest` - cool, was just a question, would be interesting to see what it looks like on IE and decide if we want to support it later.

PR #3572, opened 2021-02-18
Focus on first focusable element within a dialog when launching

### LINKED ISSUE #3567: Button launching dialog should move focus inside dialog
If activating the button opens a dialog, the focus moves inside the dialog. ([Ours](https://vanillaframework.io/docs/examples/patterns/contextual-menu/dark) doesn't)

sowasred2012:
## Done

- Added JS to focus on the launched target when a modal or contextual menu is launched.
- Documented recommended behaviour in docs

Fixes #3567 

## QA

- Open [contextual menu demo](https://vanilla-framework-3572.demos.haus/docs/examples/patterns/contextual-menu/default)
- Click "Take action"
- Press tab key - the first item in the launched dialog should be highlighted
- Open [modal demo](/docs/examples/patterns/modal/default)
- Press tab key - the first item in the launched dialog should be highlighted
  - [Contextual menu](https://vanilla-framework-3572.demos.haus/docs/patterns/contextual-menu#accessibility)
  - [Modal](https://vanilla-framework-3572.demos.haus/docs/patterns/modal#accessibility)

### DIFF
--- a/templates/docs/examples/patterns/contextual-menu/_script.js
+++ b/templates/docs/examples/patterns/contextual-menu/_script.js
@@ -17,12 +17,7 @@ function toggleMenu(element, show, top) {
     }
 
     if (show) {
-      var focusableElements = 'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
-      var firstFocusable = target.querySelector(focusableElements);
-
-      if (firstFocusable) {
-        firstFocusable.focus();
-      }
+      target.focus();
     }
   }
 }
--- a/templates/docs/examples/patterns/contextual-menu/_script.js
+++ b/templates/docs/examples/patterns/contextual-menu/_script.js
@@ -15,6 +15,15 @@ function toggleMenu(element, show, top) {
     if (typeof top !== 'undefined') {
       target.style.top = top + 'px';
     }
+
+    if (show) {
+      var focusableElements = 'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
+      var firstFocusable = target.querySelector(focusableElements);
+
+      if (firstFocusable) {
+        firstFocusable.focus();
+      }
+    }
   }
 }
 
--- a/templates/docs/examples/patterns/modal/_script.js
+++ b/templates/docs/examples/patterns/modal/_script.js
@@ -5,7 +5,15 @@
 function toggleModal(modal) {
   if (modal && modal.classList.contains('p-modal')) {
     if (modal.style.display === 'none') {
+      var focusableElements = 'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
+      var firstFocusable = modal.querySelector(focusableElements);
+
       modal.style.display = 'flex';
+
+      if (firstFocusable) {
+        console.log(firstFocusable);
+        firstFocusable.focus();
+      }
     } else {
       modal.style.display = 'none';
     }
--- a/templates/docs/examples/patterns/modal/_script.js
+++ b/templates/docs/examples/patterns/modal/_script.js
@@ -5,15 +5,8 @@
 function toggleModal(modal) {
   if (modal && modal.classList.contains('p-modal')) {
     if (modal.style.display === 'none') {
-      var focusableElements = 'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
-      var firstFocusable = modal.querySelector(focusableElements);
-
       modal.style.display = 'flex';
-
-      if (firstFocusable) {
-        console.log(firstFocusable);
-        firstFocusable.focus();
-      }
+      modal.focus();
     } else {
       modal.style.display = 'none';
     }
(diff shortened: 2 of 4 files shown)


bartaz:
While it does focus when button is clicked to open menu, not when links open menu?

menus

templates/docs/examples/patterns/contextual-menu/_script.js
```diff
@@ -15,6 +15,15 @@ function toggleMenu(element, show, top) {
     if (typeof top !== 'undefined') {
       target.style.top = top + 'px';
     }
+
+    if (show) {
+      var focusableElements = 'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
+      var focusable = target.querySelectorAll(focusableElements);
```
bartaz, line 21: Would `querySelector` just do the same and automatically select first element, so we don't need to get it via index?

templates/docs/examples/patterns/modal/_script.js
```diff
@@ -5,7 +5,14 @@
 function toggleModal(modal) {
   if (modal && modal.classList.contains('p-modal')) {
     if (modal.style.display === 'none') {
+      var focusableElements = 'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
+      var focusable = modal.querySelectorAll(focusableElements);
```
bartaz, line 9: Same as above, maybe `querySelector` would be simpler.

bartaz:
It feels a bit weird that we are focusing first element, but the focus is not visible (because of our use of `:focus-visible`).
But I guess it is just problematic when mixing mouse and keyboard navigation.
When using only keyboard or mouse the experience is consistent.

templates/docs/examples/patterns/contextual-menu/_script.js
sowasred2012: Oh good point, you're right

templates/docs/examples/patterns/contextual-menu/_script.js
```diff
@@ -15,6 +15,15 @@ function toggleMenu(element, show, top) {
     if (typeof top !== 'undefined') {
       target.style.top = top + 'px';
     }
+
+    if (show) {
+      var focusableElements = 'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
+      var firstFocusable = target.querySelector(focusableElements);
+
+      if (firstFocusable) {
+        firstFocusable.focus();
```
anthonydillon, line 24: Won't this be an array? I think we just need to focus on the `target` then any tabbing after that will start in the menu.

anthonydillon (changes requested):

anthonydillon (approved):
Perfect :+1:

lyubomir-popov:
@sowasred2012 The way I understand it, in modals the focus needs to loop within the element. Not sure this applies to contextual menus though. If we make it loop inside, then Esc should close it.

lyubomir-popov:
@sowasred2012 do you think [this](https://www.w3.org/TR/wai-aria-practices/#kbd_general_within) applies to our case here?

lyubomir-popov:
related modal accessibility [issue](https://github.com/canonical-web-and-design/vanilla-framework/issues/3223)

wgx:
@sowasred2012  Doesn't behave as expected in Firefox 85.0.2, MacOS (latest). Tab instead 'breaks out' of the viewport and tabs through browser controls.

sowasred2012:
> @sowasred2012 do you think [this](https://www.w3.org/TR/wai-aria-practices/#kbd_general_within) applies to our case here?

@lyubomir-popov I don't think so - that seems to suggest being able to navigate through the (for example) modal contents with a key other than tab, i.e. arrow keys, but in their documentation on modal keyboard navigation, tabs appear to be recommended? https://www.w3.org/TR/wai-aria-practices/#keyboard-interaction-7

@anthonydillon would like your opinion on this too, the docs linked above, there's a green "Note" section advising on how focus should be set when a dialog opens:

> 1. When a dialog opens, focus placement depends on the nature and size of the content.
>     - In all circumstances, focus moves to an element contained in the dialog.
>     - Unless a condition where doing otherwise is advisable, focus is initially set on the first focusable element.

To me, that reads as though the focus should be set on a focusable element within the dialog, not the dialog itself - though the "condition where doing otherwise is advisable" leaves some wiggle room.

anthonydillon:
Referring to this section in [Inclusive Components](https://inclusive-components.design/) by Heydon Pickering.

> ## aria-controls
> As I wrote in Aria-controls Is Poop, the attribute, intended to help screen reader users navigate from a controlling element to a controlled element, is only supported in the JAWS screen reader. So you simply can’t rely on it. Without a good method for directing users between elements, you should instead make sure one of the following is true:
> 1. The expanded list’s first link is next in focus order after the button(as in the previous code example).
> 2. The first link is focused programmatically upon revealing the list.

> In this case, I would recommend (1). It’s a lot simpler since you don’t have to worry about moving focus back to the button and on which event(s) to do so. Also, there’s currently nothing in place to warn users that their focus will be moved to somewhere different. In the true menus we’ll be discussing shortly, this is the job of.

So, both seem right but for a general use pattern, it's safer to do the simpler option to prevent edge case issues.

Issue #3565, opened 2021-02-16
Ubuntu Mono fontface mismatch 


sowasred2012:
**Describe the bug**

Related to #3564 - in [_base_fontface.scss](https://github.com/canonical-web-and-design/vanilla-framework/blob/master/scss/_base_fontfaces.scss) we use a large if/else conditional based on the `$font-use-subset-latin` variable. By default, Vanilla has this set to false, and within the `else` block we have "Ubuntu Mono" fontfaces for font weights 300 and 400, which both use the same font files: 

- fdd692b9-UbuntuMono-R_W.woff2
- 85edb898-UbuntuMono-R_W.woff

If `$font-use-subset-latin` is true, then different files are used for each font weight:

300:
- fdd692b9-UbuntuMono-R_W.woff2
- 85edb898-UbuntuMono-R_W.woff

400:
- a662364d-UbuntuMono-B-subset.woff2
- 22f97dd9-UbuntuMono-B-subset.woff

**To Reproduce**

- Run the project locally and visit http://0.0.0.0:8001/docs/examples/patterns/code-copyable
- See that the font looks OK
- In dev tools, check the computed styles of the input element, see that font weight is 400
- In _settings_font.scss, change `$font-use-subset-latin` to `true`
- Reload the page
- See that the font now looks bold, but the font weight is still 400

**Expected behavior**

The fonts should appear consistent whether the subset variable is set to true or not (they should not be bold, that is being addressed in #3564).

**Screenshots**

![Screenshot from 2021-02-16 16-49-01](https://user-images.githubusercontent.com/2376968/108094646-f1327600-7076-11eb-96f0-09eaa8aaff64.png)
![Screenshot from 2021-02-16 16-49-19](https://user-images.githubusercontent.com/2376968/108094657-f2fc3980-7076-11eb-8093-7c4f5e254d91.png)

PR #3573, opened 2021-02-18
Ubuntu mono font bugs

### LINKED ISSUE #3564: Code copyable font weight should be explicitly set to 300
**Describe the bug**

Inputs within [code copyable elements](https://vanillaframework.io/docs/examples/patterns/code-copyable) inherit `font-weight: 400;` from browser stylesheets, leading to unintentionally bold font in some situations (in this case, when `$font-use-subset-latin` is set to true, it reveals a related bug in Vanilla: #3565)

**To Reproduce**

- Run the project locally and visit http://0.0.0.0:8001/docs/examples/patterns/code-copyable
- See that the font looks OK
- In dev tools, check the computed styles of the input element, see that font weight is 400
- In _settings_font.scss, change `$font-use-subset-latin` to `true`
- Reload the page
- See that the font now looks bold, but the font weight is still 400

**Expected behavior**

The font weight should explicitly be 300

**Screenshots**

Screenshot from 2021-02-16 16-49-01
Screenshot from 2021-02-16 16-49-19

sowasred2012:
## Done

- Set font-weight to 300 on code copyable component
- Replace font file used in non-subset Ubuntu Mono, font-weight 400 font face with a bold variant, rather than the regular font beign used previously

Fixes #3564 & #3565 

## QA

- Pull this branch, `dotrun` the project and visit http://0.0.0.0:8101/docs/examples/patterns/code-copyable
- See that the font weight looks correct
- In `_settings_font.scss`, change `$font-use-subset-latin` value to `true`
- Reload the code copyable page, see that the font looks the same, and isn't bold

### DIFF
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -91,6 +91,13 @@ $code-inline-padding: 0.25rem;
       position: absolute;
       top: #{$spv-inner--x-small--scaleable + 0.5 * (map-get($line-heights, default-text) - map-get($icon-sizes, default))};
       width: map-get($icon-sizes, default);
+
+      @if ($increase-font-size-on-larger-screens) {
+        // stylelint-disable-next-line max-nesting-depth
+        @media (min-width: $breakpoint-x-large) {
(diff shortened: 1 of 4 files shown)


bartaz (approved):
LGTM, thanks!

PR #3581, opened 2021-02-22
add "Edit on CodePen" link to code snippet examples

### LINKED ISSUE #3558: Add "Edit in CodePen" to embedded examples
CodePen docs on how to prefill the editor: https://blog.codepen.io/documentation/prefill/

sowasred2012:
## Done

Added "Edit on CodePen" link to code examples in Vanilla docs

Fixes #3558 

## QA

- Open [demo](https://vanilla-framework-3581.demos.haus/docs)
- Go to any doc with a code example, see that an "Edit on CodePen" link is present at the bottom of the code snippet.
- Click the link, see that any HTML, CSS and JS is present and correct, and that Vanilla styles have been imported.
- Visit [code snippet - syntax highlighting docs](https://vanilla-framework-3581.demos.haus/docs/base/code#syntax-highlighting)
- Click the codepen link, see that syntax highlighing is applied in the new window (this demonstrates external js is being included)

## Screenshots
Screenshot from 2021-02-22 16-01-32

### DIFF
--- a/templates/static/js/example.js
+++ b/templates/static/js/example.js
@@ -28,16 +28,6 @@
     }
   });
 
-  document.addEventListener('click', function (e) {
-    var isCodePenLink = e.target.classList.contains('js-codepen-link');
-
-    if (isCodePenLink) {
-      var form = e.target.closest('form');
-      e.preventDefault();
-      form.submit();
-    }
-  });
-
   function fetchExample(exampleElement) {
     var link = exampleElement.href;
     var request = new XMLHttpRequest();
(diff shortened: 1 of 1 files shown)


templates/static/js/example.js
```diff
@@ -9,6 +28,16 @@
     }
   });
 
+  document.addEventListener('click', function (e) {
```
anthonydillon, line 31: Why not target `js-codepen-link` class instead of adding a listener to the body?

templates/static/js/example.js
```diff
@@ -9,6 +28,16 @@
...
+    var isCodePenLink = e.target.classList.contains('js-codepen-link');
+
+    if (isCodePenLink) {
+      var form = e.target.closest('form');
+      e.preventDefault();
+      form.submit();
```
anthonydillon, line 37: You should null check the form here

anthonydillon (approved):
Code looks good :+1:

anthonydillon:
QA-wise can you make the middle click work or open on a new tab as it is what I always do.

templates/static/js/example.js
sowasred2012: I'm fairly certain it used to be difficult to attach events to dynamically created elements, and this was a workaround - turns out that's not the case..! I've attached the event at the time the link is appended, removing this block altogether.

templates/static/js/example.js
sowasred2012: Now that the click event is being attached when the element is appended - there's a null check on line 191, if all of those values are empty, the form isn't created, and there's nothing to click.

sowasred2012:
> QA-wise can you make the middle click work or open on a new tab as it is what I always do.

Updated with this - I _never_ use the middle button, so this is the first time addressing something like this, let me know what you think.

anthonydillon (approved):
LGTM :+1:

Issue #3542, opened 2021-02-12
Button space incorrect when using an anchor tag instead of button


carkod:
**Describe the bug**

When using `p-button--neutral` or `p-button--negative` for `<a class="p-button--neutral">` tags, the space shrinks

**To Reproduce**

Here is an [example code pen](https://codepen.io/carkod/pen/qBqqgYe)

**Expected behavior**

Default spacing as if using `<button class="p-button--neutral">`.

bartaz:
This is something we've been struggling to properly fix in Vanilla.

We are using margin-right on buttons (to make sure that when button wraps into new line it doesn't have offset on the left). So when buttons are next to each other (`<button A> <button B>`) the first one (button A) will have margin right, and the second one will have the unnecessary margin removed via `:last-of-type` selector. The issue with `:last-of-type` selector is, that it only works on same type of the element (`button` or `a`), so when links and buttons are mixed together both are treated as last of type and have their margin removed.

There is no `:last-of-class` selector that we could use (but arguabely even that won't fix the issue completely as you could mix `p-button--positive` `p-button--negative` that don't share the same class).

I don't know what would be the best way to fix this.

The only solution that comes to my mind now is to remove `:last-of-type` selector from buttons (so they always have right margin), but add `is-last` state, so manually the margin can be removed. This would be a breaking change though.

Or the other way around, we could add a class that would add the margin back, `is-followed-by-a-button` (or however we can call that).

anthonydillon:
The more I looked into this the simpler the solution became which worries me that I'm missing an edge case:
https://codepen.io/anthonydillon/pen/xxRqBqN

PR #3584, opened 2021-02-24
Fix margin right on multiple buttons of mixed element types

sowasred2012:
## Done

Added a slightly modified version of @anthonydillon's suggested fix (see [CodePen example](https://codepen.io/anthonydillon/pen/xxRqBqN)) for #3542 

## QA

- Open [button alignment example](/docs/examples/patterns/buttons/alignment/)
- See that each button in each row has a margin right, except for the last buttons in each row
- Review Percy changes

**Breaking change:** this fixes the original issue, but in instances where a button is prevented from being the `:last-child` by a hidden element, such as a contextual menu toggle, a right margin is still applied and needs to be removed with the `u-no-margin--right` utility.

### DIFF
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -42,10 +42,11 @@
     }
 
     @media only screen and (min-width: $breakpoint-x-small + 1) {
+      margin-right: $sph-outer;
       width: auto;
 
-      &:not(:last-of-type):not(:only-of-type) {
-        margin-right: $sph-outer;
+      &:last-child {
+        margin-right: 0;
       }
     }
 
@@ -58,7 +59,7 @@
     &.is-small {
       font-size: #{map-get($font-sizes, small)}rem;
       line-height: map-get($line-heights, small);
-      margin: 0 0 $input-margin-bottom - $sp-unit 0;
+      margin-bottom: $input-margin-bottom - $sp-unit;
       padding: calc(#{map-get($nudges, nudge--small)} - 1px) $sph-inner--small;
     }
 
--- a/scss/_patterns_chip.scss
+++ b/scss/_patterns_chip.scss
@@ -35,14 +35,14 @@
       background-color: rgba($color-x-dark, 0.15);
     }
 
-    &__lead,
-    &__value {
+    .p-chip__lead,
+    .p-chip__value {
       margin-bottom: 0;
       overflow: hidden;
       text-overflow: ellipsis;
     }
 
-    &__lead {
+    .p-chip__lead {
       $space-between-lead-and-value: 0.25rem;
       @extend %x-small-text;
 
@@ -59,14 +59,14 @@
       }
     }
 
-    &__value {
+    .p-chip__value {
       @extend %small-text;
 
       line-height: $chip-line-height;
       padding-top: 0.05rem;
     }
 
-    &__dismiss {
+    .p-chip__dismiss {
       @include vf-button-pattern(
         $button-background-color: transparent,
         $button-border-color: transparent,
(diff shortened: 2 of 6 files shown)


scss/_patterns_search-box.scss
```diff
@@ -38,7 +38,7 @@
     @extend %transparent-button;
 
     &:not(:last-of-type):not(:only-of-type) {
-      margin-right: $bar-thickness;
+      margin-right: $bar-thickness !important;
```
anthonydillon, line 41: Does this need important?

sowasred2012: I don't mean for this to be reviewed yet, I'm using Percy to show me what effect changes are having across the examples.

However, it seems like !important will need to be used somewhere - I had them originally on the base button styles like the codepen, but that overrode the search box button margin. I removed it from the base styles, but found that the `:last-child` rule still took precedence, so added !important here to work around that.

sowasred2012: Ugh - turns out this is not where I meant to add the rule anyway...!

templates/docs/examples/patterns/buttons/alignment.html
```diff
@@ -0,0 +1,30 @@
...
+<p>
+  <button class="p-button--neutral">Neutral button</button>
+  <button class="p-button--positive">Positive button</button>
+  <button class="p-button--negative">Negative button</button>
+</p>
+
+<p>
+  <a href="#" class="p-button--neutral">Neutral link</a>
+  <a href="#" class="p-button--positive">Positive link</a>
+  <a href="#" class="p-button--negative">Negative link</a>
```
anthonydillon, line 21: It would be nice to have some mixed examples as that was the issue

scss/_patterns_search-box.scss
```diff
@@ -65,7 +65,7 @@
     border-left-style: solid;
     border-left-width: 1px;
     border-radius: 0 $border-radius $border-radius 0;
-    margin-right: $bar-thickness;
+    margin-right: $bar-thickness !important;
```
anthonydillon, line 68: I think the right fix here is to ensure we are wrapping all pattern elements with the block class as we agreed before.. 

For example: `.p-search-box .p-search-box__button`

This means you won't need the `!important`.

scss/_patterns_chip.scss
```diff
@@ -80,7 +80,7 @@
       left: $sp-unit * 0.5;
       line-height: 1rem;
       margin-bottom: 0;
-      margin-right: -$sp-unit * 0.5;
+      margin-right: -$sp-unit * 0.5 !important;
```
anthonydillon, line 83: Same here

scss/_base_button.scss
```diff
@@ -42,10 +42,11 @@
     }
 
     @media only screen and (min-width: $breakpoint-x-small + 1) {
+      margin-right: $sph-outer;
       width: auto;
 
-      &:not(:last-of-type):not(:only-of-type) {
-        margin-right: $sph-outer;
+      &:last-child {
+        margin-right: 0;
```
anthonydillon, line 49: I think this is much easier to read

anthonydillon (changes requested):
Few suggestions in the code

templates/docs/examples/patterns/buttons/alignment.html
sowasred2012: The block below this one has a mix of links and buttons.

anthonydillon (approved):
LGTM works well and is much simpler. I believe its much better to support mixed button + a then some hidden button groups.

PR #3596, opened 2021-03-01
Rename expanding table

### LINKED ISSUE #2946: Rename `p-table-expanding` to `p-table--expanding`

bartaz:
## Done

Renamed `p-table-expanding` to `p-table--expanding`(and `p-table-expanding__panel` to `p-table__expanding_panel`) for consistency.

Fixes #2946

## QA

- Open [demo](https://vanilla-framework-3596.demos.haus/docs/base/tables#expanding)
- Table expanding example should use new class names and look/work as expected: https://vanilla-framework-3596.demos.haus/docs/examples/patterns/tables/table-expanding
- Old deprecated demo should still work as expected: https://vanilla-framework-3596.demos.haus/docs/examples/patterns/tables/table-expanding-deprecated
  - https://vanilla-framework-3596.demos.haus/docs/base/tables#expanding
  - make sure docs are updated to use new class names
  - make sure old class names are mentioned to be deprecated
- Check component status page to see if relevant changes are mentioned there
- Check if version in package.json is updated accordingly

### DIFF
--- a/scss/_patterns_table-expanding.scss
+++ b/scss/_patterns_table-expanding.scss
@@ -1,7 +1,7 @@
 @import 'settings';
 
 @mixin vf-p-table-expanding {
-  .p-table-expanding {
+  .p-table--expanding {
     display: flex;
     flex-flow: column nowrap;
     justify-content: space-between;
@@ -39,7 +39,7 @@
         grow: 1;
       }
 
-      &.p-table-expanding__panel {
+      &.p-table__expanding-panel {
         flex-basis: 100%;
         max-width: 100%;
 
@@ -59,5 +59,17 @@
       }
     }
     // stylelint-enable selector-no-qualifying-type
+  }
+
+  @include deprecate('3.0.0', 'Use the `.p-table--expanding` instead') {
+    .p-table-expanding {
+      // stylelint-disable-next-line scss/at-extend-no-missing-placeholder
+      @extend .p-table--expanding;
+
+      .p-table-expanding__panel {
+        // stylelint-disable-next-line scss/at-extend-no-missing-placeholder
+        @extend .p-table__expanding-panel;
+      }
+    }
   }
 }
--- a/templates/docs/base/tables.md
+++ b/templates/docs/base/tables.md
@@ -41,15 +41,17 @@ View example of the table sortable pattern
 
 <span class="p-label--updated">Updated</span>
 
-Using `.p-table-expanding` in conjunction with the `<table>` element will allow expanding and hidden table cells which take up the full width of the table row element.
+Using `.p-table--expanding` in conjunction with the `<table>` element will allow expanding and hidden table cells which take up the full width of the table row element.
 
 This pattern should be used when a table requires configuration fields (add, edit). Expandable rows can also be used to supply additional information not visible on the table row.
 
-Using `p-table-expanding__panel` it can be hidden using the `aria-hidden` attribute. The table must contain all table cells required. The expanding panel is implemented as additional cell in each row, so to keep the markup of the table valid, an additional cell is also needed in the table `<thead>`. This placeholder heading cell should be hidden using `aria-hidden="true"` attribute.
+Using `p-table__expanding-panel` it can be hidden using the `aria-hidden` attribute. The table must contain all table cells required. The expanding panel is implemented as additional cell in each row, so to keep the markup of the table valid, an additional cell is also needed in the table `<thead>`. This placeholder heading cell should be hidden using `aria-hidden="true"` attribute.
 
 <div class="embedded-example"><a href="/docs/examples/patterns/tables/table-expanding/" class="js-example">
 View example of the expanding table pattern
 </a></div>
+
+<span class="p-label--deprecated">Deprecated</span> We are deprecating the use of `p-table-expanding` and `p-table-expanding__panel` class names. They will be removed in future version 3.0 of Vanilla. Use `p-table--expanding` and `p-table__expanding-panel` instead.
 
 ### Responsive
 
--- a/templates/docs/examples/patterns/tables/_script-expanding.js
+++ b/templates/docs/examples/patterns/tables/_script-expanding.js
@@ -1,0 +1,47 @@
+/**
+  Toggles the necessary aria- attributes' values on the table panels
+  to show or hide them.
+  @param {HTMLElement} element The tab that acts as the handles.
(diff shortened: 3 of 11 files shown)


scss/_patterns_table-expanding.scss
```diff
@@ -39,7 +39,7 @@
         grow: 1;
       }
 
-      &.p-table-expanding__panel {
+      &.p-table__expanding-panel {
```
sowasred2012, line 42: I get confused with BEM in these sorts of situations - looking at [the BEM faqs](http://getbem.com/faq/#block-modifier-affects-elements), it's obviously OK to style child elements within a modifier block, but I haven't found guidance on when an element is only ever intended to be a child of a modifier. 

In this case, `.p-table__expanding-panel` doesn't feature anywhere other than as a child of `.p-table--expanding`, but its naming might lead you to think that you could use `.p-table__expanding-panel` in `.p-table--sortable`.

I would have thought that would be reason enough for it to be a pattern distinct from the one being modified, but that's what this PR is moving away from, so I'm not sure - is there any call for a `p-table__panel`/`p-table__full-width-cell`/etc in other `p-table` patterns?

templates/docs/examples/patterns/tables/_script-expanding.js
```diff
@@ -0,0 +1,47 @@
...
+  Toggles the necessary aria- attributes' values on the table panels
+  to show or hide them.
+  @param {HTMLElement} element The tab that acts as the handles.
+  @param {Boolean} show Whether to show or hide the expanded row panel.
+*/
+function toggleExpanded(element, show) {
+  var target = document.getElementById(element.getAttribute('aria-controls'));
+
+  if (target) {
+    element.setAttribute('aria-expanded', show);
```
sowasred2012, line 11: Should we also user `aria-pressed` here? @lyubomir-popov ?

sowasred2012 (commented):
Code looks good - but had a couple of questions

lyubomir-popov:
@bartaz Given we have p-table--sortable and now p-table--expanding, does that mean for tables we move away from the (very problematic) single bem modifier rule enshrined in the Vanilla practices? So any new table features will use a modifier, unlike any other pattern where the first to be introduced is a modifier and he rest are `.is-*` classes?

scss/_patterns_table-expanding.scss
lyubomir-popov: Standard BEM has a simple solution - just chain all the classes you need, so `.p-table.p-table--sortable.p-table--expanding` and inside `.p-table__expanding-panel`. In our modified convention I can't think of a satisfying answer.

templates/docs/examples/patterns/tables/_script-expanding.js
lyubomir-popov: I don't think you can use two states. You have to pick one and expanded makes more sense to me:
https://www.w3.org/TR/wai-aria/#button
image

scss/_patterns_table-expanding.scss
bartaz: Yes, this is a weird case. I didn't want to make it `p-table--expanding__panel`, (a child under a modifier) as it seems like a hack and something we don't have anywhere in Vanilla.

I considered `p-table__panel`, but it would only make sense in case if we ever need any kind of panel in other tables. More so, this particular one is used in very specific context, on a cell that expands into whole new row. I doubt we will do anything similar in other tables.
When we implement expanding rows in modular table I believe we should approach it differently.
So, to keep the semantic meaning that this is explicilty a panel that is expanded, I made it `__expanding-panel`.

Current expading table implementation is very specific, it drops table layout for flexbox which has a lots of weird consequences. In longer term, when expanding functionality is impemented as part of modular table, I think we should deprecate and drop current `p-table--expanding`.

templates/docs/examples/patterns/tables/_script-expanding.js
bartaz: It doesn't seem to be mentioned anywhere that states are exclusive. I think the button can have multiple states if they are applicable - it can be `aria-expanded` and `aria-pressed` at the same time, or `aria-busy`. These seem to be independent states that are turned on or off.

On a side note of that - maybe we should style buttons with `aria-expanded` the same way as we style `aria-pressed` to mark they are active? These would save us adding second `pressed` state when `expanded` is already on the button.

lyubomir-popov: > maybe we should style buttons with `aria-expanded` the same way as we style `aria-pressed` to mark they are active

yes, definitely!

templates/docs/base/tables.md
```diff
@@ -41,16 +41,18 @@ View example of the table sortable pattern
...
+Using `.p-table--expanding` in conjunction with the `<table>` element will allow expanding and hidden table cells which take up the full width of the table row element.
 
 This pattern should be used when a table requires configuration fields (add, edit). Expandable rows can also be used to supply additional information not visible on the table row.
 
-Using `p-table-expanding__panel` it can be hidden using the `aria-hidden` attribute. The table must contain all table cells required. The expanding panel is implemented as additional cell in each row, so to keep the markup of the table valid, an additional cell is also needed in the table `<thead>`. This placeholder heading cell should be hidden using `aria-hidden="true"` attribute.
+Using `p-table__expanding_panel` it can be hidden using the `aria-hidden` attribute. The table must contain all table cells required. The expanding panel is implemented as additional cell in each row, so to keep the markup of the table valid, an additional cell is also needed in the table `<thead>`. This placeholder heading cell should be hidden using `aria-hidden="true"` attribute.
```
anthonydillon, line 48: This should be `p-table__expanding-panel`

bartaz: Good catch, thanks!

bartaz:
> @bartaz Given we have p-table--sortable and now p-table--expanding, does that mean for tables we move away from the (very problematic) single bem modifier rule enshrined in the Vanilla practices?

@lyubomir-popov `p-table--sortable` is going to be removed as well. It's gonna be a class on a column, not whole table (because the column is sortable, not the table).

But this is just cleaning up of the existing table patterns. In a longer term for modular table we are probably going to need a more modular approach. Ideally if any feature relates to smaller chunk of the table, the `p-table` root element would not need any modifier. If it does, we will have to use multiple, possibly via `is-something` flags as we do in other places.

sowasred2012 (approved):
rereviewed - looks good!

PR #3601, opened 2021-03-02
Remove margin right from contextual menu

sowasred2012:
## Done

Removed margin-right from contextual menu, which was introduced by this PR https://github.com/canonical-web-and-design/vanilla-framework/pull/3584 - since a contextual menu toggle is never the last child, in the previous setup, it _always_ had a margin right applied, when it should never have one.

## QA

- Open [demo](https://vanilla-framework-3601.demos.haus/docs/patterns/contextual-menu)
- Inspect the examples, see that each instance of a contextual menu toggle does not have a right margin


bartaz (approved):
LGTM, thanks

PR #3606, opened 2021-03-03
Install linkchecker via pip3

bartaz:
## Done

As suggested by @nottrobin we can use pip3 to install linkchecker.

## QA

Make sure CI checks pass (including linkchecker)


sowasred2012 (approved):
Oh nice - easy to review too! :smile:

PR #3607, opened 2021-03-03
Add selector-max-type linting rule

### LINKED ISSUE #2949: Linting: prevent using element selectors

sowasred2012:
## Done

- Added `selector-max-type` linting rule, and set it to zero.
- Added exemptions for existing styles

Fixes #2949 

## QA

- See that this PR's checks pass
- Pull down this branch
- In any scss file, break the `selector-max-type` rule by adding a naughty rule e.g. `div { display: block; }`
- Run `dotrun lint-scss`, it should fail

### DIFF
--- a/scss/_base_background.scss
+++ b/scss/_base_background.scss
@@ -1,5 +1,5 @@
 @mixin vf-base-background {
-  // stylelint-disable-next-line selector-max-type
+  // stylelint-disable-next-line selector-max-type -- base styles can use type selectors
   html {
     background: $color-x-light;
   }
--- a/scss/_base_background.scss
+++ b/scss/_base_background.scss
@@ -1,4 +1,5 @@
 @mixin vf-base-background {
+  // stylelint-disable-next-line selector-max-type
   html {
     background: $color-x-light;
   }
--- a/scss/_base_blockquotes.scss
+++ b/scss/_base_blockquotes.scss
@@ -2,6 +2,7 @@
 
 // Base blockquotes styling
 @mixin vf-b-blockquotes {
+  // stylelint-disable selector-max-type
   blockquote {
     border-left: 2px solid $color-mid-dark;
     margin-bottom: $spv-inner--large;
@@ -20,4 +21,5 @@
       font-style: normal;
     }
   }
+  // stylelint-enable selector-max-type
 }
--- a/scss/_base_blockquotes.scss
+++ b/scss/_base_blockquotes.scss
@@ -2,7 +2,7 @@
 
 // Base blockquotes styling
 @mixin vf-b-blockquotes {
-  // stylelint-disable selector-max-type
+  // stylelint-disable selector-max-type -- base styles can use type selectors
   blockquote {
     border-left: 2px solid $color-mid-dark;
     margin-bottom: $spv-inner--large;
--- a/scss/_base_box-sizing.scss
+++ b/scss/_base_box-sizing.scss
@@ -3,6 +3,7 @@
 // Box-sizing FTW!
 //https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/
 @mixin vf-b-box-sizing {
+  // stylelint-disable-next-line selector-max-type
   html {
     box-sizing: border-box;
   }
--- a/scss/_base_box-sizing.scss
+++ b/scss/_base_box-sizing.scss
@@ -3,7 +3,7 @@
 // Box-sizing FTW!
 //https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/
 @mixin vf-b-box-sizing {
-  // stylelint-disable-next-line selector-max-type
+  // stylelint-disable-next-line selector-max-type -- base styles can use type selectors
   html {
     box-sizing: border-box;
   }
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -2,7 +2,7 @@
 
 // Base button styles
 @mixin vf-b-button {
-  // stylelint-disable selector-max-type
+  // stylelint-disable selector-max-type -- base styles can use type selectors
   button {
     @extend %vf-button-base;
     @include vf-button-pattern;
--- a/scss/_base_button.scss
+++ b/scss/_base_button.scss
@@ -2,6 +2,7 @@
 
 // Base button styles
 @mixin vf-b-button {
+  // stylelint-disable selector-max-type
   button {
     @extend %vf-button-base;
     @include vf-button-pattern;
@@ -84,6 +85,7 @@
         margin-top: $spv-outer--small + $spv-nudge-compensation;
       }
     }
+    // stylelint-enable selector-max-type
   }
 
   %vf-button-white-success-icon {
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -7,6 +7,7 @@ $code-inline-padding: 0.25rem;
 
 // Base code styles
 @mixin vf-b-code {
+  // stylelint-disable selector-max-type
   code,
   kbd,
   pre,
@@ -79,7 +80,7 @@ $code-inline-padding: 0.25rem;
     background-color: $color-code-background-dark;
     color: $colors--dark-theme--text-default;
   }
-  // stylelint-enable selector-no-qualifying-type
+  // stylelint-enable selector-no-qualifying-type, selector-max-type
 
   %leading-linux-prompt-icon {
     &::before {
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -7,7 +7,7 @@ $code-inline-padding: 0.25rem;
 
 // Base code styles
 @mixin vf-b-code {
-  // stylelint-disable selector-max-type
+  // stylelint-disable selector-max-type -- base styles can use type selectors
   code,
   kbd,
   pre,
--- a/scss/_base_code.scss
+++ b/scss/_base_code.scss
@@ -74,13 +74,11 @@ $code-inline-padding: 0.25rem;
   }
 
   // dark theme
-  // stylelint-disable selector-no-qualifying-type
   [class*='--dark'] code,
   code.is-dark {
     background-color: $color-code-background-dark;
     color: $colors--dark-theme--text-default;
   }
-  // stylelint-enable selector-no-qualifying-type, selector-max-type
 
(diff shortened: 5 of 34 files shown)


scss/_patterns_switch.scss
```diff
@@ -50,6 +50,7 @@ $knob-size: $sp-unit * 3;
       width: $knob-size;
     }
 
+    // stylelint-disable-next-line selector-max-type
```
bartaz, line 53: In places like this we should add also some `// FIXME` comment to note that this is not an expected place for element selector and should be fixed.

scss/_patterns_subnav.scss
```diff
@@ -104,6 +104,7 @@
   }
 
   @include deprecate('3.0.0', 'Use .p-navigation__items, .p-navigation__item, .p-navigation__link classes instead.') {
+    // stylelint-disable-next-line selector-max-type
```
bartaz, line 107: In places like this we should add also some `// FIXME` comment to note that this is not an expected place for element selector and should be fixed.

scss/_patterns_side-navigation.scss
```diff
@@ -262,6 +262,7 @@
 
   // Styles for markup in raw HTML docs variant
   .p-side-navigation--raw-html {
+    // stylelint-disable selector-max-type
```
bartaz, line 265: We could add a comment explaining why it is fine here to disable the rule

```suggestion
    // stylelint-disable selector-max-type -- we support raw HTML markup for discourse-generated side nav
```

scss/_patterns_subnav.scss
bartaz: This is actually deprecated, so no need to fix that, maybe just add a comment to disable rule
```suggestion
    // stylelint-disable-next-line selector-max-type -- this is deprecated use, will be removed
```

scss/_patterns_navigation.scss
```diff
@@ -225,6 +225,7 @@ $lightness-threshold: 70;
       &__link {
         @extend %navigation-item;
 
+        // stylelint-disable-next-line selector-max-type
```
bartaz, line 228:
```suggestion
        // stylelint-disable-next-line selector-max-type -- this is deprecated use, will be removed
```

scss/_patterns_navigation.scss
```diff
@@ -397,6 +398,7 @@ $lightness-threshold: 70;
   }
 
   @include deprecate('3.0.0', 'Use .p-navigation__items, .p-navigation__item, .p-navigation__link classes instead.') {
+    // stylelint-disable selector-max-type
```
bartaz, line 401:
```suggestion
    // stylelint-disable selector-max-type -- this is deprecated use, will be removed
```

scss/_patterns_side-navigation.scss
```diff
@@ -542,7 +544,7 @@
   // border color of items list
   $color-sidenav-list-border
 ) {
-  // stylelint-disable selector-no-qualifying-type
+  // stylelint-disable selector-no-qualifying-type, selector-max-type
```
bartaz, line 547: I wonder if we need `selector-no-qualifying-type` now, `selector-max-type` with value of 0 seems to catch all of these as well.

scss/_base_background.scss
```diff
@@ -1,4 +1,5 @@
 @mixin vf-base-background {
+  // stylelint-disable-next-line selector-max-type
```
bartaz, line 2: I think we should comment the places where we disable it
```suggestion
  // stylelint-disable-next-line selector-max-type -- base styles can use type selectors
```

scss/_base_blockquotes.scss
```diff
@@ -2,6 +2,7 @@
 
 // Base blockquotes styling
 @mixin vf-b-blockquotes {
+  // stylelint-disable selector-max-type
```
bartaz, line 5:
```suggestion
  // stylelint-disable selector-max-type -- base styles can use type selectors
```

scss/_base_box-sizing.scss
```diff
@@ -3,6 +3,7 @@
 // Box-sizing FTW!
 //https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/
 @mixin vf-b-box-sizing {
+  // stylelint-disable-next-line selector-max-type
```
bartaz, line 6:
```suggestion
  // stylelint-disable-next-line selector-max-type -- base styles can use type selectors
```

scss/_base_button.scss
```diff
@@ -2,6 +2,7 @@
 
 // Base button styles
 @mixin vf-b-button {
+  // stylelint-disable selector-max-type
```
bartaz, line 5:
```suggestion
  // stylelint-disable selector-max-type -- base styles can use type selectors
```

scss/_base_code.scss
```diff
@@ -7,6 +7,7 @@ $code-inline-padding: 0.25rem;
 
 // Base code styles
 @mixin vf-b-code {
+  // stylelint-disable selector-max-type
```
bartaz, line 10:
```suggestion
  // stylelint-disable selector-max-type -- base styles can use type selectors
```

scss/_base_details.scss
```diff
@@ -1,4 +1,5 @@
 @mixin vf-b-details {
+  // stylelint-disable selector-max-type
```
bartaz, line 2:
```suggestion
  // stylelint-disable selector-max-type -- base styles can use type selectors
```

scss/_base_forms-range.scss
```diff
@@ -7,6 +7,7 @@
   $track-border: $track-border-size solid $colors--light-theme--border-high-contrast;
   $track-radius: $bar-thickness;
 
+  // stylelint-disable selector-max-type
```
bartaz, line 10:
```suggestion
  // stylelint-disable selector-max-type -- base styles can use type selectors
```

scss/_base_forms-range.scss
```diff
@@ -7,6 +7,7 @@
...
   // stylelint-disable-next-line selector-no-qualifying-type
```
bartaz, line 11: Not sure if we still need this.

scss/_base_forms-tick-elements.scss
```diff
@@ -126,6 +126,7 @@ $box-offsets-top: (
 
   // Default form checkbox and radio styles
   %vf-tick-elements {
+    // stylelint-disable-next-line selector-max-type
```
bartaz, line 129:
```suggestion
    // stylelint-disable-next-line selector-max-type -- base styles can use type selectors
```

scss/_base_forms.scss
```diff
@@ -89,6 +89,7 @@
     }
   }
 
+  // stylelint-disable selector-max-type
```
bartaz, line 92:
```suggestion
  // stylelint-disable selector-max-type -- base styles can use type selectors
```

scss/_base_typography.scss
```diff
@@ -11,6 +11,7 @@
     @include vf-b-typography-max-widths;
   }
 
+  // stylelint-disable selector-max-type
```
bartaz, line 14:
```suggestion
  // stylelint-disable selector-max-type -- base styles can use type selectors
```

scss/_base_tables.scss
```diff
@@ -2,6 +2,7 @@
 
 // Tables
 @mixin vf-b-tables {
+  // stylelint-disable selector-max-type
```
bartaz, line 5:
```suggestion
  // stylelint-disable selector-max-type -- base styles can use type selectors
```

scss/_base_media.scss
```diff
@@ -2,6 +2,7 @@
 
 // Media element styles
 @mixin vf-b-media {
+  // stylelint-disable selector-max-type
```
bartaz, line 5:
```suggestion
  // stylelint-disable selector-max-type -- base styles can use type selectors
```

scss/_base_lists.scss
```diff
@@ -1,6 +1,7 @@
 @import 'settings';
 
 @mixin vf-b-lists {
+  // stylelint-disable selector-max-type
```
bartaz, line 4:
```suggestion
  // stylelint-disable selector-max-type -- base styles can use type selectors
```

scss/_base_links.scss
```diff
@@ -2,6 +2,7 @@
 
 // Base styling for links
 @mixin vf-b-links {
+  // stylelint-disable selector-max-type
```
bartaz, line 5:
```suggestion
  // stylelint-disable selector-max-type -- base styles can use type selectors
```

scss/_base_hr.scss
```diff
@@ -2,6 +2,7 @@
 
 // Horizontal rule
 @mixin vf-b-hr {
+  // stylelint-disable selector-max-type
```
bartaz, line 5:
```suggestion
  // stylelint-disable selector-max-type -- base styles can use type selectors
```

bartaz (changes requested):
I think we should try to comment anytime we disable the rule. Either to explain why it is used (in base styles or other expected uses), or to mark places that need fixing (in patterns).

I added couple of suggestions around that, but probably not all.

Also, we may not need `selector-no-qualifying-type` now. It seems to catch the same things.

scss/_patterns_side-navigation.scss
sowasred2012: Good points about the comments, adding more now. I did check whether we needed `selector-no-qualifying-type`, and we do, in every instance I remove it, it raises an error.

`selector-max-type` catches any instance of an html selector, which this is an instance of, but `<selector>.class`/`::psuedo`/`[attribute]`/`etc` is not caught by the same rule.

bartaz:
@sowasred2012 Thanks for adding the comments.

How about `selector-no-qualifying-type` rule? It seems to be redundant now, because if we disallow types in selectors, it also disallows any "qualifying type" selectors, so the second rule is not needed. We seem to disable both in same places.

scss/_patterns_side-navigation.scss
sowasred2012: @bartaz just saw your latest comment, I think my reply here got lost amongst the comments - we do still need the `selector-no-qualifying-type`.

bartaz (approved):
LGTM, thanks!

PR #3608, opened 2021-03-03
Remove the need for `p-table--sortable` class name

### LINKED ISSUE #2948: Create sortable table column in place of `p-table--sortable`

bartaz:
## Done

- Removes `p-table--sortable` class name
- Any table with `aria-sort` on headers can be sorted

Fixes #2948 

## QA

- Open [demo](https://vanilla-framework-3608.demos.haus/docs/examples/patterns/tables/table-sortable)
- Check new sortable example:
  - https://vanilla-framework-3608.demos.haus/docs/examples/patterns/tables/table-sortable
  - make sure it works as expected without any additional class names
- Check old deprecated sortable example:
  - https://vanilla-framework-3608.demos.haus/docs/examples/patterns/tables/table-sortable-deprecated
  - make sure it still works as expected
  - https://vanilla-framework-3608.demos.haus/docs/base/tables#sortable

### DIFF
--- a/templates/docs/examples/patterns/tables/_script-sorting.js
+++ b/templates/docs/examples/patterns/tables/_script-sorting.js
@@ -1,0 +1,99 @@
+/**
+ * Sorts a table by the column specified.
+ * @param {HTMLElement} header Sortable header element that was clicked.
+ * @param {HTMLTableElement} table Table to sort.
+ */
+function sortTable(header, table) {
+  var SORTABLE_STATES = {
+    none: 0,
+    ascending: -1,
+    descending: 1,
+    ORDER: ['none', 'ascending', 'descending'],
+  };
+
+  // Get index of column based on position of header cell in <thead>
+  // We assume there is only one row in the table head.
+  var col = [].slice.call(table.tHead.rows[0].cells).indexOf(header);
+
+  // Based on the current aria-sort value, get the next state.
+  var newOrder = SORTABLE_STATES.ORDER.indexOf(header.getAttribute('aria-sort')) + 1;
+  newOrder = newOrder > SORTABLE_STATES.ORDER.length - 1 ? 0 : newOrder;
+  newOrder = SORTABLE_STATES.ORDER[newOrder];
+
+  // Reset all header sorts.
+  var headerSorts = table.querySelectorAll('[aria-sort]');
+
+  for (var i = 0, ii = headerSorts.length; i < ii; i += 1) {
+    headerSorts[i].setAttribute('aria-sort', 'none');
+  }
+
+  // Set the new header sort.
+  header.setAttribute('aria-sort', newOrder);
+
+  // Get the direction of the sort and assume only one tbody.
+  // For this example only assume one tbody.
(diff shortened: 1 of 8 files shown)


bartaz:
@anthonydillon What do you think about removing `p-table--sortable` altogether, and rely on `aria-sort` to make any table sortable without additional class names? Or should we at least use something like `is-sortable` on `th` to be more explicit?

anthonydillon:
I think it's good to rely on aria where we can.

templates/docs/examples/patterns/tables/_script-sorting.js
```diff
@@ -0,0 +1,99 @@
...
+  // Set the new header sort.
+  header.setAttribute('aria-sort', newOrder);
+
+  // Get the direction of the sort and assume only one tbody.
+  // For this example only assume one tbody.
+  var direction = SORTABLE_STATES[newOrder];
+  var body = table.tBodies[0];
+
+  // Convert the HTML element list to an array.
+  var newRows = Array.prototype.slice.call(body.rows, 0);
```
sowasred2012, line 39: Minor thing, but this could be consistent with the casting being done on line 16 (or vice versa):
```suggestion
  var newRows = [].slice.call(body.rows, 0);
```

Also (and not related to this PR) - maybe it's worth revisiting the IE11 question, we do a lot to accommodate it, but even Microsoft products are beginning to [drop](https://docs.github.com/en/github/getting-started-with-github/supported-browsers) [support](https://techcommunity.microsoft.com/t5/microsoft-365-blog/microsoft-365-apps-say-farewell-to-internet-explorer-11-and/ba-p/1591666) for it.

templates/docs/upgrade-guide-v3.md
```diff
@@ -66,3 +66,5 @@ Use of `.col` classes outside of `.row` is deprecated. If you use `.col-X` class
 ### Tables
 
 We renamed and deprecated `p-table-expanding` and `p-table-expanding__panel`. Use `p-table--expanding` and `p-table__expanding-panel` instead.
+
+We removed the `p-table--sortable` that was previously required to enable sorting functionality in the tables. Currently any table with correctly used `aria-sort` attributes on column headers can be sorted. The `p-table--sortable` class name can be removed from HTML (any relevant JavaScript may need to be updated).
```
sowasred2012, line 70: :scream: I was not aware of this guide! I think there may be one or two things I've deprecated that should be mentioned here (code copyable for instance)

sowasred2012 (approved):
LGTM - 2 non blocking comments/questions in the code

templates/docs/upgrade-guide-v3.md
bartaz: @sowasred2012 No worries, I tend to forget about it as well. when you have a sec give it a look - I may have added what was missing in the meantime.
If not, just go around and compare 'deprecated' list from component status and check if it's mentioned there.

It doesn't have to be a final nicely edited guide yet, but a draft so we have everything in place when we need it.

PR #3609, opened 2021-03-03
Fix bottom margin on last p element in p-matrix__desc

sowasred2012:
## Done

Fixed a bug introduced in the [2.0 refactor](https://github.com/canonical-web-and-design/vanilla-framework/pull/2138/files#diff-02b46287dfcdad2bcacdccc1da64b1f474cb5cfd855d1e604567e68da4ba8df3R108)* where last-child  `p` elements within `.p-matrix__desc` no longer had their bottom margin reduced.

*noticed while working on https://github.com/canonical-web-and-design/vanilla-framework/pull/3607

## QA

- Open [demo](https://vanilla-framework-3609.demos.haus/docs/patterns/matrix)
- See that the margin below the last paragraph in each matrix item has been reduced, compared to [live version](https://vanillaframework.io/docs/examples/patterns/matrix)

## Screenshots

Before:
Screenshot from 2021-03-03 15-13-23

After:
Screenshot from 2021-03-03 15-13-53

### DIFF
--- a/scss/_patterns_matrix.scss
+++ b/scss/_patterns_matrix.scss
@@ -12,10 +12,6 @@
     margin-bottom: $spv-outer--scaleable;
     margin-left: 0;
     padding-left: 0;
-
-    > p:last-child {
-      margin-bottom: $spv-nudge-compensation;
-    }
   }
 
   .p-matrix__item {
@@ -139,8 +135,14 @@
   }
 
   .p-matrix__desc {
+    margin-bottom: $spv-nudge-compensation;
+
     @media (max-width: $breakpoint-heading-threshold) {
       margin-top: -#{$sp-unit};
+    }
+
+    > :last-child {
+      margin: 0;
     }
   }
 }


package.json
```diff
@@ -43,7 +43,7 @@
     "percy": "percy exec -- node snapshots.js",
     "icon-svgs-to-mixins": "node scripts/convert-svgs-to-icon-mixins.js"
   },
-  "version": "2.26.0",
+  "version": "2.26.1",
```
bartaz, line 46: This is gonna be confusing. We haven't released `2.26.0` yet, so no need to bump.

sowasred2012: Hmmm. I took from the meeting yesterday that every PR would update the version to some degree - but then, we're not automatically releasing yet, so fair point.

bartaz:
It seems that even in our example we have different cases. Sometimes `p` is a child of `p-matrix__desc`, but sometimes `p-matrix__desc` itself is the only `p`.

Screenshot 2021-03-03 at 16 59 16


If we are fixing this, it should work in both situations.

sowasred2012:
> It seems that even in our example we have different cases. Sometimes `p` is a child of `p-matrix__desc`, but sometimes `p-matrix__desc` itself is the only `p`.
> If we are fixing this, it should work in both situations.

Good catch - I've updated, moving the margin-bottom to `p-matrix__desc` itself, and removing the bottom margin from its last child element (if it has one). Though, the margin is so small, maybe we don't need it at all?

bartaz:
> Though, the margin is so small, maybe we don't need it at all?

I think it's meant to keep things on baseline grid.

bartaz (approved):
Thanks!

PR #3612, opened 2021-03-03
Add the search and filter component

anthonydillon:
## Done
Add the search and filter component
- Styling for the component
- A basic example of the pattern
- Documentation for the pattern
- Standalone example

Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/3144

## QA

- Open /docs/examples/patterns/search-and-filter/default
- Check when focusing on the input it opens the panel and looks correct.
  - Go to /docs/patterns/search-and-filter and check the content is ok and clear
  - Go to /docs/examples/standalone/patterns/search-and-filter/default and check the stand-alone example renders

## Screenshots
image


bartaz:
@anthonydillon It seems this branch is built on top of older version of Vanilla, could you make sure to rebase to update to current state of master branch?

scss/_patterns_search-and-filter.scss
```diff
@@ -0,0 +1,183 @@
...
+          display: none;
+        }
+      }
+
+      .p-search-box {
+        flex-grow: 1;
+      }
+    }
+
+    .p-icon--search {
```
bartaz, line 68: Why do we have this icon in the markup at all if we are just hiding it with `display: none`?

bartaz:
Should this support also the case when chips are entered into the search?

Screenshot 2021-03-04 at 09 38 35

We don't seem to have example of that. And it's not possible to do it with script in current example.

templates/docs/patterns/search-and-filter.md
```diff
@@ -0,0 +1,73 @@
...
+@include vf-p-search-box;
+@include vf-p-search-and-filter;
+@include vf-u-off-screen;
+```
+
+For more information see [Customising Vanilla](/docs/customising-vanilla/) in your projects, which includes overrides and importing instructions.
+
+### React
+
+For use of the React component you can do that by installing our react-component page and importing via:
```
bartaz, line 47: I agree that linking to React docs is good and we should do it in other places too, but keeping a snippet of react code in the docs may quickly become hard to maintain. We are not going to know if anything here needs to be updated.

templates/docs/examples/patterns/search-and-filter/_script.js
```diff
@@ -0,0 +1,27 @@
...
+    if (panel.getAttribute('aria-hidden') === 'false') {
+      panel.setAttribute('aria-hidden', 'true');
+    } else {
+      panel.setAttribute('aria-hidden', 'false');
+    }
+  }
+}
+
+// Add click handler for clicks on elements with aria-controls
+const patterns = document.querySelectorAll('.p-search-and-filter');
```
bartaz, line 16: We need these scripts to work in IE, so please use `var` instead of const/let, and cast the result of `querySelectorAll` to an array (`[].slice.call(document.querySelectorAll("...")`) to make sure `forEach` is available.

scss/_patterns_search-and-filter.scss
anthonydillon: I think your right, I went back and forward on that a few times.

templates/docs/patterns/search-and-filter.md
anthonydillon: Your right, edited.

templates/docs/examples/patterns/search-and-filter/_script.js
anthonydillon: Done.

anthonydillon:
@bartaz thanks for the review. I believe I have addressed all your comments

scss/_patterns_search-and-filter.scss
```diff
@@ -0,0 +1,168 @@
...
+      width: 100%;
+      z-index: 9999;
+
+      &[aria-hidden='true'] {
+        opacity: 0;
+        pointer-events: none;
+      }
+    }
+
+    .p-search-prompt {
```
bartaz, line 79: I don't see that anywhere in HTML. Is this part of search and filter?

scss/_patterns_search-and-filter.scss
```diff
@@ -0,0 +1,168 @@
...
+      &[aria-expanded='true'] {
+        height: auto;
+        max-height: 100%;
+
+        .p-search-and-filter__selected-count {
+          display: none;
+        }
+      }
+
+      .p-search-box {
```
bartaz, line 57: We have a lot more overrides for `.p-search-box` in line 101.
Why is this one separated? Search box seems to be always in `p-search-and-fliter__search-container`, so there doesn't seem to be a need for different styling in different contexts.

scss/_patterns_search-and-filter.scss
```diff
@@ -0,0 +1,168 @@
...
+      height: 2rem;
+      margin: 0;
+      max-width: 48px;
+      position: absolute;
+      right: 3px;
+      top: 3px;
+      z-index: 9999;
+    }
+
+    .p-search-box {
```
bartaz, line 101: I'm not really sold on relying and overriding external component here.

I understand that we want to reuse as much as possible from search box component, but it seems that we override more than we use:

Screenshot 2021-03-04 at 16 05 51
Screenshot 2021-03-04 at 16 05 33

So maybe this should be simply a part of search and filter? `p-search-and-filter__search` `p-search-and-filter__input`?

And if there are significant common parts between this one and search box, maybe they can be extracted to some %placeholder.

scss/standalone/patterns_search-and-filter.scss
```diff
@@ -0,0 +1,11 @@
...
+@include vf-base;
+
+@import '../patterns_chip';
+@import '../patterns_search-box';
+@import '../patterns_search-and-filter';
+@import '../utilities_off-screen';
+@include vf-p-chip;
+@include vf-p-search-box;
+@include vf-p-search-and-filter;
+@include vf-u-off-screen;
```
bartaz, line 11: I think you also need to include icons here - the X dismiss icon in chips is missing in standalone examples.

templates/docs/examples/patterns/search-and-filter/with-chips.html
```diff
@@ -0,0 +1,121 @@
...
+          <div class="p-chip" aria-pressed="false" role="button" tabindex="0">
+            <span class="p-chip__lead">REGION</span><span class="p-chip__value">us-south3</span>
+          </div>
+        </span>
+        <span>
+          <div class="p-chip" aria-pressed="false" role="button" tabindex="0">
+            <span class="p-chip__lead">REGION</span><span class="p-chip__value">us-north4</span>
+          </div>
+        </span>
+        <span>
```
bartaz, line 62: Are these spans needed? Is this the markup from react component?
I think having div (block element) inside span (inline) is not valid HTML.

templates/docs/examples/patterns/search-and-filter/with-chips.html
```diff
@@ -0,0 +1,121 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Search and filter / Default{% endblock %}
```
bartaz, line 2: You have both examples with same title. 
```suggestion
{% block title %}Search and filter / With chips{% endblock %}
```

scss/_patterns_search-and-filter.scss
```diff
@@ -0,0 +1,168 @@
...
+      cursor: pointer;
+      margin-left: -1rem;
+      margin-right: -1rem;
+      overflow: hidden;
+      padding: 1rem;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+    }
+
+    .p-search-and-filter__clear {
```
bartaz, line 90: This doesn't seem to be anywhere in markup as well.

bartaz: Oh, I think both (prompt and clear) are from the state during typing into the component.

Looking at react example:
Screenshot 2021-03-04 at 16 20 40

We need either JS that adds those elements when typing (seems too much for the scope of the example) or a static example of the component in this state.

scss/_patterns_search-and-filter.scss
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    .p-search-and-filter__panel {
+      background: $color-x-light;
+      border-radius: $border-radius;
+      box-shadow: $box-shadow;
+      opacity: 1;
+      padding: 0.5rem 1rem 0;
+      position: absolute;
+      transition: opacity 0.25s;
```
bartaz, line 69: We should use animation mixin.

scss/_patterns_search-and-filter.scss
```diff
@@ -0,0 +1,168 @@
...
+        flex-grow: 1;
+      }
+    }
+
+    .p-search-and-filter__panel {
+      background: $color-x-light;
+      border-radius: $border-radius;
+      box-shadow: $box-shadow;
+      opacity: 1;
+      padding: 0.5rem 1rem 0;
```
bartaz, line 67: There are a lot of hardcoded numbers all over this file for spacing.
These should be variables from Vanilla spacing settings.

scss/_patterns_search-and-filter.scss
```diff
@@ -0,0 +1,168 @@
...
+      &[data-empty='false'],
+      &[aria-expanded='false'] {
+        height: $height;
+      }
+
+      .p-chip {
+        margin-bottom: 0.5rem;
+      }
+
+      .p-search-and-filter__selected-count {
```
bartaz, line 40: Another one missing from HTML. 

We should have an example with this state of the component:

Screenshot 2021-03-04 at 16 25 25

bartaz: And this? (after clicking on count) and with count in the dropdown:

Screenshot 2021-03-04 at 16 26 51

scss/_patterns_search-and-filter.scss
```diff
@@ -0,0 +1,168 @@
...
+      padding: 1rem;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+    }
+
+    .p-search-and-filter__clear {
+      border: none;
+      height: 2rem;
+      margin: 0;
+      max-width: 48px;
```
bartaz, line 94: All values should be in rem, as the value of rem is different on largest screens.

bartaz (changes requested):
I added comments around different topics:

- spacing values should be from spacing variables, not hardcoded
- there seem to be several elements in SCSS not currently in examples, we should add examples for all the states the component can be in (to be sure we style them correctly)
- we should think if reusing `p-search-box` is needed here, because we seem to override more styles from it than we use

scss/_patterns_search-and-filter.scss
anthonydillon: Its gone

scss/_patterns_search-and-filter.scss
anthonydillon: I have removed the dependency on the search box.

templates/docs/examples/patterns/search-and-filter/with-chips.html
anthonydillon: That would be react, they are gone

templates/docs/examples/patterns/search-and-filter/with-chips.html
anthonydillon: Thanks done

scss/_patterns_search-and-filter.scss
anthonydillon: Yeah I'm not sure about this, maybe we should chat tomorrow

scss/_patterns_search-and-filter.scss
anthonydillon: :+1: done

scss/_patterns_search-and-filter.scss
anthonydillon: I have done what I think makes sense

scss/_patterns_search-and-filter.scss
anthonydillon: Its back

anthonydillon:
@bartaz added some more states and improved the documentation. Ready for another look if you have time.

lyubomir-popov:
Can anything be done about the focus state?
/docs/examples/patterns/search-and-filter/default

Ideally it should be the focus state blue and follow the contour of the element. I know outline can't do this, but maybe a [box-shadow](https://css-tricks.com/having-a-little-fun-with-custom-focus-styles/)? Not crucial, if it's too much hassle then just the correct colour would be enough.

anthonydillon:
@lyubomir-popov Added focus state to the filter chips:
image

lyubomir-popov:
@anthonydillon thanks, looks good. I've added design +1.

templates/docs/examples/patterns/search-and-filter/chip-overflow.html
```diff
@@ -0,0 +1,128 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Search and filter / With chips{% endblock %}
```
bartaz, line 2: Duplicated title. Should it be called 'Chips overflow' or something like that?
```suggestion
{% block title %}Search and filter / Chips overflow{% endblock %}
```

templates/docs/patterns/search-and-filter.md
```diff
@@ -0,0 +1,95 @@
...
+@include vf-u-off-screen;
+```
+
+For more information see [Customising Vanilla](/docs/customising-vanilla/) in your projects, which includes overrides and importing instructions.
+
+## React
+
+For use of the React component you can do that by installing our react-component page and importing.
+
+[Playbook example](https://canonical-web-and-design.github.io/react-components/?path=/docs/search-and-filter--default-story)
```
bartaz, line 91: Did you mean "Storybook" instead of playbook? Also, I'm not sure if it's useful to call it by the name of the documentation tool. Maybe something like "See the React documentation for SearchAndFilter component"?

```suggestion
[See the documentation for our React `SearchAndFilter` component](https://canonical-web-and-design.github.io/react-components/?path=/docs/search-and-filter--default-story)
```

bartaz (commented):
Couple small comments on documentation.

Also, please rebase to current master and add the new component to the table on component status page.

anthonydillon:
@bartaz thanks, done.

bartaz (approved):
LGTM, thanks.

The only remaining thing is test failing due to css file size.
You would need to bump the threshold in /tests/parker.js from 260000 to 27k I guess.

PR #3613, opened 2021-03-04
Add indeterminate state checkbox

### LINKED ISSUE #3293: Tri state checkbox
## Pattern to amend

Not amend, rather add feature to the checkbox pattern

According to the aria spec, two types of checkboxes are supported - two state and tri-state. Not sure whether "supported" means we have to support them, still we already use them in RBAC, and should be using them in the MAAS machine listing header.

## Visual

TBD, something simple - possibly just a dash in place of the tick?

## Context

A checkbox that is parent to other checkboxes and controls / shows the mixed state of the other checkboxes

/// 
From the [spec](https://www.w3.org/TR/wai-aria-practices/#checkbox):
Tri-state: This type of checkbox supports an additional third state known as partially checked.

Tri-state checkboxes are currently used in RBAC I believe, and should be used in maas machine listing - the column header checkbox. Since they are useful in two places, arguably they should be in Vanilla.

Example implementation: has aria-checked:mixed, aria-controls="{list of checkboxes it controls}"

sowasred2012:
## Done

Added indeterminate state checkbox

Fixes #3293 

## QA

- Open [demo](https://vanilla-framework-3613.demos.haus/docs/examples/patterns/forms/checkbox-indeterminate)
- See that the checkbox shows a dash, and cycles through checked, unchecked and mixed/indeterminate when clicked
  - [Checkbox docs](https://vanilla-framework-3613.demos.haus/docs/base/forms#checkbox)
  - [Component status](https://vanilla-framework-3613.demos.haus/docs/component-status)

## Screenshots

Screenshot from 2021-03-04 17-23-17

### DIFF
--- a/templates/docs/base/forms.md
+++ b/templates/docs/base/forms.md
@@ -64,6 +64,16 @@ To use checkbox component in headings use the `.p-checkbox--heading` variant of 
 View example of the heading checkbox components
 </a></div>
 
+#### Indeterminate state checkbox
+
+<span class="p-label--new">New</span>
+
+When a checkbox requires a state between checked and unchecked, `checkbox.indeterminate = true;` can be set via JavaScript, which will cause the checkbox appear in an "indeterminate" state.
+
+<div class="embedded-example"><a href="/docs/examples/patterns/forms/checkbox-indeterminate/" class="js-example">
+View example of the indeterminate checkbox component
+</a></div>
+
 #### Deprecated base checkboxes
 
 <span class="p-label--deprecated">Deprecated</span>
--- a/templates/docs/base/forms.md
+++ b/templates/docs/base/forms.md
@@ -70,6 +70,8 @@ View example of the heading checkbox components
 
 When a checkbox requires a state between checked and unchecked, `checkbox.indeterminate = true;` can be set via JavaScript, which will cause the checkbox appear in an "indeterminate" state.
 
+In cases when JavaScript can't be used for that, set `aria-checked="mixed"` on checkbox to show "indeterminate" state. Please note that it will only change the visual look of the checkbox, it will not affect its state available to the browser via JavaScript.
+
 <div class="embedded-example"><a href="/docs/examples/base/forms/checkbox-indeterminate/" class="js-example">
 View example of the indeterminate checkbox component
 </a></div>
--- a/templates/docs/base/forms.md
+++ b/templates/docs/base/forms.md
@@ -70,7 +70,7 @@ View example of the heading checkbox components
 
 When a checkbox requires a state between checked and unchecked, `checkbox.indeterminate = true;` can be set via JavaScript, which will cause the checkbox appear in an "indeterminate" state.
 
-<div class="embedded-example"><a href="/docs/examples/patterns/forms/checkbox-indeterminate/" class="js-example">
+<div class="embedded-example"><a href="/docs/examples/base/forms/checkbox-indeterminate/" class="js-example">
 View example of the indeterminate checkbox component
 </a></div>
 
--- a/templates/docs/examples/patterns/forms/_indeterminate_script.js
+++ b/templates/docs/examples/patterns/forms/_indeterminate_script.js
@@ -12,6 +12,7 @@ function initIndeterminateCheckboxes(checkboxSelector, triggerSelector) {
   // which doesn't use JavaScript, so we hardcode the `aria-checked='mixed'`
   // attribute in HTML, and remove it when JavaScript _is_ available.
   checkbox.removeAttribute('aria-checked');
+  checkbox.indeterminate = true;
 
   // The example checkbox isn't controlling other checkboxes (as it would be in
   // a real world implementation), so we need a way to show each state here. When
--- a/templates/docs/examples/patterns/forms/_indeterminate_script.js
+++ b/templates/docs/examples/patterns/forms/_indeterminate_script.js
@@ -1,0 +1,33 @@
+/**
+  Sets up indeterminate checkboxes and attaches event handlers to it
+  @param {String} checkboxSelector The CSS selector matching indeterminate checkboxes
+  @param {String} triggerSelector The CSS selector matching the trigger element
+*/
+function initIndeterminateCheckboxes(checkboxSelector, triggerSelector) {
+  var checkbox = document.querySelector(checkboxSelector);
+  var trigger = document.querySelector(triggerSelector);
+
+  // FOR EXAMPLE PURPOSES ONLY:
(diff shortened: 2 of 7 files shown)


templates/docs/base/forms.md
```diff
@@ -64,6 +64,16 @@ To use checkbox component in headings use the `.p-checkbox--heading` variant of
 View example of the heading checkbox components
 </a></div>
 
+#### Tri-state checkbox
```
bartaz, line 67: As it seems to be officially called "indeterminate" or "mixed-state" in the docs, let's use that, as it better describes what this is meant to be used for.

templates/docs/base/forms.md
```diff
@@ -64,6 +64,16 @@ To use checkbox component in headings use the `.p-checkbox--heading` variant of
...
+
+<span class="p-label--new">New</span>
+
+When a checkbox requires a state between checked and unchecked, the HTML attribute `aria-checked` can be set to `mixed`. In JavaScript, `checkbox.indeterminate = true;` will also cause the third state to appear, and should be set along with the `aria-checked` attribute.
```
bartaz, line 71: As far as I understand `aria-checked` is meant to be used on elements that are not native checkbox input.
Native inputs, but default implement accessibility features, so don't require additional aria.

In this case, native checkbox with `indeterminate` set to true, doesn't seem to require also setting `aria-checked`.
In same way, you don't set `aria-checked="true"` when setting `checkbox.checked = true`.

see: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role

I think we should encouraging using native `checkbox.indeterminate`, and only using `aria-checked` in cases when JS is not available.

templates/docs/examples/patterns/forms/_tristate_script.js
```diff
@@ -0,0 +1,31 @@
...
+  tristateCheckboxes.forEach(function (checkbox) {
+    checkbox.addEventListener('change', function () {
+      var currentState = checkbox.getAttribute('aria-checked');
+
+      if (currentState === 'false') {
+        checkbox.setAttribute('aria-checked', 'mixed');
+        checkbox.checked = true;
+        checkbox.indeterminate = true;
+      } else if (currentState === 'mixed') {
+        checkbox.setAttribute('aria-checked', 'true');
```
bartaz, line 17: This is not needed. It's a native browser input, you don't need to set `aria-checked` on it, as setting `checked` does exactly the same.

templates/docs/examples/patterns/forms/_tristate_script.js
```diff
@@ -0,0 +1,31 @@
...
+function initTristateCheckboxes(selector) {
+  var tristateCheckboxes = [].slice.call(document.querySelectorAll(selector));
+
+  tristateCheckboxes.forEach(function (checkbox) {
+    checkbox.addEventListener('change', function () {
+      var currentState = checkbox.getAttribute('aria-checked');
+
+      if (currentState === 'false') {
+        checkbox.setAttribute('aria-checked', 'mixed');
+        checkbox.checked = true;
```
bartaz, line 14: Technically it probably doesn't matter, but in MDN example they set `checked` to false when setting indeterminate to true: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

bartaz (changes requested):

templates/docs/base/forms.md
sowasred2012: Didn't we agree to use `aria-checked` for the sake of Percy? Without this, we rely on JavaScript to get the indeterminate state, so we won't see the impact in Percy.

bartaz: @sowasred2012 Yes, what I mean is that we shouldn't document it as a default way to get this state (using JS to set `checkbox.intederminate = true` should be default. And if it's used `aria-checked` is not needed, so we shouldn't say that it "should be set along with the `aria-checked`".

So, I think in the docs we should say that to get intederminate state JS should be used. In cases when it can't `aria-checked="mixed"` can be used instead

lyubomir-popov:
@sowasred2012 LGTM!

templates/docs/examples/patterns/forms/checkbox-indeterminate.html
```diff
@@ -0,0 +1,17 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Forms / Checkbox indeterminate{% endblock %}
```
bartaz, line 2: This is not limited to `p-checkbox` pattern, so we should move the file to /examples/base/forms instead.

templates/docs/examples/patterns/forms/_indeterminate_script.js
```diff
@@ -0,0 +1,33 @@
...
+*/
+function initIndeterminateCheckboxes(checkboxSelector, triggerSelector) {
+  var checkbox = document.querySelector(checkboxSelector);
+  var trigger = document.querySelector(triggerSelector);
+
+  // FOR EXAMPLE PURPOSES ONLY:
+  // In this example, we want the indeterminate state to be visible to Percy,
+  // which doesn't use JavaScript, so we hardcode the `aria-checked='mixed'`
+  // attribute in HTML, and remove it when JavaScript _is_ available.
+  checkbox.removeAttribute('aria-checked');
```
bartaz, line 14: We want to also make this checkbox indeterminate at the same time:

```suggestion
  checkbox.removeAttribute('aria-checked');
  checkbox.indeterminate = true;
```

templates/docs/examples/patterns/forms/checkbox-indeterminate.html
sowasred2012: Cool - that's done.

templates/docs/base/forms.md
```diff
@@ -64,6 +64,16 @@ To use checkbox component in headings use the `.p-checkbox--heading` variant of
 View example of the heading checkbox components
 </a></div>
 
+#### Indeterminate state checkbox
+
+<span class="p-label--new">New</span>
+
+When a checkbox requires a state between checked and unchecked, `checkbox.indeterminate = true;` can be set via JavaScript, which will cause the checkbox appear in an "indeterminate" state.
+
```
bartaz, line 72: I think I would still mention `aria-checked` as an option, especially that we use it in our HTML example.

```suggestion

In cases when JavaScript can't be used for that, set `aria-checked="mixed"` on checkbox to show "indeterminate" state. Please note that it will only change the visual look of the checkbox, it will not affect its state available to the browser via JavaScript.

```

bartaz (approved):
LGTM, with small suggestion for the docs.

PR #3622, opened 2021-03-09
Accordion accessibility update

### LINKED ISSUE #3318: Accordion: Accessibility issues that may require a breaking change
[The following are either difficult, require discussion, or are breaking changes, so separating from the [original list of validator issues](https://github.com/canonical-web-and-design/vanilla-framework/issues/3217)]

- [Rule:](https://www.w3.org/TR/wai-aria-practices/#wai-aria-roles-states-and-properties) Each accordion header button should be wrapped in an element with role heading (or h1/h2/h3/h4/h5/h6) that has a value set for aria-level that is appropriate for the information architecture of the page.
Each accordion header button is wrapped in an element with role heading that has a value set for aria-level that is appropriate for the information architecture of the page.

[Source](https://www.w3.org/TR/wai-aria-practices/#accordion)

Our accordion buttons are wrapped in `li`s, seems there should be a heasding in between:
image

We agreed not to tackle the above as part of the accessibility audit.

sowasred2012:
## Done

- Deprecated previous accordion default pattern, and `.p-accordion__tab--with-title` pattern.
- Added a more accessible pattern that replaces both previous patterns

Fixes #3318 

## QA

- Visit [default accordion](https://vanilla-framework-3622.demos.haus/docs/examples/patterns/accordion/default)
- Check that each accordion tab functions as expected, and looks good
- Visit [headings accordion](https://vanilla-framework-3622.demos.haus/docs/examples/patterns/accordion/headings)
    - Check that there are no visual changes compared to [live](https://vanillaframework.io/docs/examples/patterns/accordion/headings)
- Verify deprecated patterns still work:
  - Visit [deprecated default accordion](https://vanilla-framework-3622.demos.haus/docs/examples/patterns/accordion/default-deprecated)
    - Check that there are no visual changes compared to [live](https://vanillaframework.io/docs/examples/patterns/accordion/default)
  - [Component status](https://vanilla-framework-3622.demos.haus/docs/component-status)
  - [Accordion](https://vanilla-framework-3622.demos.haus/docs/patterns/accordion)


templates/docs/examples/patterns/accordion/default.html
```diff
@@ -20,16 +22,20 @@
       </section>
     </li>
     <li class="p-accordion__group">
-      <button type="button" class="p-accordion__tab" id="tab2" aria-controls="tab2-section" aria-expanded="false">Networking</button>
-      <section class="p-accordion__panel" id="tab2-section" aria-hidden="true" aria-labelledby="tab2">
+      <div role="heading" aria-level="2" class="p-accordion__heading">
+        <button type="button" class="p-accordion__tab" id="tab4" aria-controls="tab4-section" aria-expanded="false">Networking</button>
```
bartaz, line 26: It's really a nitpicking, but this could be `tab2` and the last one `tab3` (especially that `aria-labelledby` say so).

sowasred2012: ah - good catch! fixing

sowasred2012: fixed

scss/_patterns_accordion.scss
```diff
@@ -59,53 +61,93 @@
...
 
-  .p-accordion__tab--with-title[aria-expanded='true'] .p-accordion__title::before {
-    @include vf-icon-minus($color-mid-dark);
-  }
+  .p-accordion__heading {
+    margin-bottom: 0;
+    max-width: none;
+    padding-top: 0;
+
+    @at-root {
```
bartaz, line 120: While it is a valid use case for `@at-root`, it seems that there is no really much benefit of using that.
It doesn't make the code shorter and it doesn't make it easier to read (quite contrary, the combination of `@at-root` and `#{&}` requires quite a bit of understanding of SASS to process).

I think explicitly writing it as:

```scss
  // stylelint-disable selector-max-type
  h2.p-accordion__heading > .p-accordion__tab,
  p-accordion__heading.p-heading--2 > .p-accordion__tab {
    $icon-top-position: calc(#{$table-cell-vertical-padding + (map-get($line-heights, h2) - $icon-size) * 0.5});

    background-position-y: $icon-top-position;
  }

  h3.p-accordion__heading > .p-accordion__tab,
  .p-accordion__heading.p-heading--3 > .p-accordion__tab {
    $icon-top-position: calc(#{$table-cell-vertical-padding + (map-get($line-heights, h3) - $icon-size) * 0.5});

    background-position-y: $icon-top-position;
  }

  h4.p-accordion__heading > .p-accordion__tab,
  .p-accordion__heading.p-heading--4 > .p-accordion__tab {
    $icon-top-position: calc(#{$table-cell-vertical-padding + (map-get($line-heights, h4) - $icon-size) * 0.5});

    background-position-y: $icon-top-position;
  }
  // stylelint-enable selector-max-type
```
would make it much easier to read and understand in the future.

scss/_patterns_accordion.scss
```diff
@@ -59,53 +61,93 @@
...
+      h2#{&} > .p-accordion__tab,
+      &.p-heading--two > .p-accordion__tab,
```
bartaz, line 122: `--two`, etc class names are deprecated. We don't need to add support for them in new code. If someone is about to use new accordion structure, they should use proper heading elements or class names.

scss/_patterns_accordion.scss
```diff
@@ -59,53 +61,93 @@
...
+        $icon-top-position: calc(#{$table-cell-vertical-padding + (map-get($line-heights, h2) - $icon-size) * 0.5});
 
-  .p-accordion__tab--with-title[aria-expanded='false'] .p-accordion__title::before {
-    @include vf-icon-plus($color-mid-dark);
+        background-position-y: $icon-top-position;
+      }
+
+      h3#{&} > .p-accordion__tab,
+      &.p-heading--three > .p-accordion__tab,
+      &.p-heading--3 > .p-accordion__tab {
```
bartaz, line 131: Do we need to support `p-heading--` class names here or should using heading elements be enough?
I don't think we did support them in previous accordion implementation, we don't have examples for those (at least to test if they work as expected).

Allowing using them requires to mix component classes on one element (`<div class="p-heading--3 p-accordion__heading">`) which ideally we should not recommend.

bartaz (changes requested):

scss/_patterns_accordion.scss
sowasred2012: I would have thought we should - we allow those classes to be applied on standard heading elements, and I've found that ability useful on a number of occasions where, for example, a design requests a `h3` but in the page heirarchy it should semantically be a `h4`.

bartaz: Ok. I would just limit this to new `--3` class names, not the deprecated ones.

sowasred2012: :+1: I've updated with the rest of the feedback

bartaz (approved):
LGTM, thanks

templates/docs/examples/patterns/accordion/default.html
```diff
@@ -7,7 +7,9 @@
 <aside class="p-accordion">
   <ul class="p-accordion__list">
     <li class="p-accordion__group">
-      <button type="button" class="p-accordion__tab" id="tab1" aria-controls="tab1-section" aria-expanded="true">Advanced topics</button>
+      <div role="heading" aria-level="2" class="p-accordion__heading">
```
bartaz, line 10: Tiny thing really, but I guess many people may just copy this blindly without adjusting the `aria-level` and I'm not sure if proposing level 2 is a good default. I think within a page copy h3 or h4 is most likely to be something you expect in accordion.

Would be good for our example be such more common use case.

templates/docs/examples/patterns/accordion/default-deprecated.html
```diff
@@ -0,0 +1,44 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Accordion - deprecated{% endblock %}
```
bartaz, line 2: Seems like we only have the deprecated example for default accordion (not the one with headings).
Would be good to have both covered.

I guess we don't need a separate example, maybe just add second accordion with headings to this one, and rename the file to `deprecated.html`. Just so that Percy has something to screenshot, so we are sure we don't introduce some regressions.

PR #3631, opened 2021-03-17
Remove custom icon styles from browser support page

### LINKED ISSUE #982: Remove navigation margin
Currently, the navigation pattern has `margin-bottom: 2rem`. This causes a white strip between the navigation and the first row.

sowasred2012:
## Done

Removed custom scss relating to icons on /browser-support and replaced the icons with img tags.

Fixes #982

## QA

- Open [demo](https://vanilla-framework-3631.demos.haus/browser-support)
- Check the icons at the bottom of the page, compare to [live site](https://vanillaframework.io/browser-support)


bartaz:
I was hoping that switching to images would not require another set of custom inline styling. Too bad we don't have some `img.is-inline` variant or something.

Maybe we can replace this with some other Vanilla pattern? The closest I found is `p-heading-icon--small` (with paraghaph as heading):

Screenshot 2021-03-17 at 13 40 12

I removed the dividers on the list as they don't work well with heading icon margins.
I guess it looks fine (the icons are just bigger), and doesn't need any custom styling.

sowasred2012:
> Maybe we can replace this with some other Vanilla pattern? The closest I found is `p-heading-icon--small` (with paraghaph as heading):
Oh I was trying to make it look the same as it was, didn't realise I had free reign, haha - cool, I'll update.

bartaz (approved):
LGTM, thanks

PR #3673, opened 2021-03-29
Increase the transparency on the background of the modal

hatched:
## Done

Increased the transparency of the background of the modal. This is done as sometimes it's important to see the context below the modal as per the screenshot below and 0.85 was too dark. 

## QA

- View the modal, is the background lighter now?

## Screenshots

III  App - Unit - Panel Action - Confirm action

### DIFF
--- a/scss/_patterns_modal.scss
+++ b/scss/_patterns_modal.scss
@@ -3,7 +3,7 @@
 @mixin vf-p-modal {
   .p-modal {
     align-items: center;
-    background: transparentize($color-dark, 0.15);
+    background: transparentize($color-dark, 0.5);
     bottom: 0;
     display: flex;
     height: 100vh;


bartaz:
@lyubomir-popov Is such change fine?

It seems to me then in cases where we want user to focus on modal such light background is too light and doesn't give enough contrast. So if we want to have lighter background it would need to be an option, not a default. But in cases when user needs to see the context of what's behind the overlay - should a modal with an overlay be used at all?

lyubomir-popov:
@bartaz I agree with you - The opacity should simply indicate that the context (page you're at) hasn't changed. Anything beyond that calls for a different ux IMO.

anthonydillon:
In my opinion, if being able to see the detail on the page is important, the model pattern is the wrong pattern for this interaction. 

Two reasons;
- why obscure the background even a little, if it's important the user can see it?
- if it's a floating element we cannot guarantee the piece of information, the user wants to see will be visible on all viewports.

It seems to me, a purposeful form confirmation pattern would be best here. Something that is guaranteed to take allocated space in the UI.

hatched:
No problem I'll close this and leave the override.

PR #3681, opened 2021-04-01
Use linkchecker installed on older ubuntu LTS

bartaz:
## Done

Revert installing linkchecker via pip to package on ubuntu@18.04 due to building issues.
Created #3682 to investigate why it was failing and try to find a different way to fix it.

## QA

- All CI checks should pass, including linkchecker

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -35,10 +35,9 @@ jobs:
 
   docker:
     name: Build and run
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


sowasred2012 (approved):
LGTM

anthonydillon:
You could also leave it on latest and pip install.

bartaz:
@anthonydillon That's exactly what we used to have and needed to change because it started failing yesterday.

PR #3695, opened 2021-04-08
Print size of build file in CI logs

bartaz:
## Done

Before we implement better automated ways of doing this we can at least print the size of the CSS file in CI logs, so they are easier to check.

## QA

- Check the 'Test' CI job, it should print build.css file size in logs

Screenshot 2021-04-08 at 14 43 41


sowasred2012 (approved):
Nice!

PR #3761, opened 2021-05-25
Handle text overflow on expanding table

### LINKED ISSUE #2697: Truncation does not show ellipsis with expanding tables
**Describe the bug**
When using u-truncate with a sortable table the text is cut off, but the ellipsis are not show.

The problem can be fixed with something like:
```css
.p-table-expanding td.u-truncate {
  display: block;
}
```

**To Reproduce**
See this codepen, uncomment the css to fix the issue:
https://codepen.io/huwshimi/pen/mdyVPKe

**Expected behavior**
The text should be truncated with an ellipsis.

**Screenshots**
What it looks like now:

Screen Shot 2019-12-09 at 3 48 51 pm

With the fix applied:

Screen Shot 2019-12-09 at 3 49 09 pm

sowasred2012:
## Done

- Updated expanding table to behave more like other table examples: 
  - removed `word-break: break-all` rule
  - added `text-overflow: ellipsis` and `overflow: hidden` to all `th` and `td` elements, so there's no need to use `u-truncate`
  - switched `display` to `block` (`text-overflow` seems to only apply to [block elements](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow#formal_definition))

Fixes #2697 

## QA

- Open [demo](/docs/examples/patterns/tables/table-expanding)
- Reduce the width of the browser, and see that the text in the column both wraps and displays an ellipsis when the word is too long to fit

- [x] Any changes to component class names (new patterns, variants, removed or added features) should be listed on the [component status page](https://github.com/canonical-web-and-design/vanilla-framework/blob/master/templates/docs/component-status.md).
- [x] Documentation side navigation should be updated with the relevant labels


bartaz (approved):
LGTM, seems to correctly replicate our truncating behaviour of regular tables.

Long words are truncated, but text wraps between words.
Applying `u-truncate` to cell disables wrapping.

PR #3762, opened 2021-05-26
Make nav highlight colour overrideable

lyubomir-popov:
## Done

Add !default to the recently introduced $colors--light-theme--nav-highlight so it can be overriden

## QA
- Look ast diff
- Verify the $colors--light-theme--nav-highlight ha sa !default at the end

### DIFF
--- a/scss/_settings_colors.scss
+++ b/scss/_settings_colors.scss
@@ -109,4 +109,4 @@ $colors--dark-theme--border-low-contrast: rgba($colors--dark-theme--text-default
 $color-brand: #333 !default;
 $color-brand-background: $colors--dark-theme--background-default;
 $color-accent: #e95420 !default;
-$color-accent-background: $color-accent !default;
+$color-accent-background: $color-brand !default;
--- a/scss/_settings_colors.scss
+++ b/scss/_settings_colors.scss
@@ -42,7 +42,7 @@ $color-label-validated: #006b75;
 $color-code-background: rgba($color-x-dark, 0.03);
 $color-code-background-dark: rgba($color-x-light, 0.3);
 $color-code-heading-background: rgba($color-x-dark, 0.08);
-$color-navigation-highlight: #e95420;
+$color-navigation-highlight: #e95420 !default;
 
 $states: (
   error: $color-negative,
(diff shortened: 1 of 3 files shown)


scss/_settings_colors.scss
```diff
@@ -42,7 +42,7 @@ $color-label-validated: #006b75;
 $color-code-background: rgba($color-x-dark, 0.03);
 $color-code-background-dark: rgba($color-x-light, 0.3);
 $color-code-heading-background: rgba($color-x-dark, 0.08);
-$color-navigation-highlight: #e95420;
+$color-navigation-highlight: #e95420 !default;
```
anthonydillon, line 45: Why is this not using one the themed colors? I know brand is dark by default by it feels like this should be accent color. The use of accent color is to add a small flourish of colour to the patterns.

lyubomir-popov:
@anthonydillon It should be $color-accent, but then it would become almost invisible in the dark theme. Would it be ok to set  $color-brand to this particular orange? This change would be in line with the push away from a whitelabel framework and towards a design system.

lyubomir-popov:
@anthonydillon I completely removed the ne vars and pointd the navigation mixin to use $color-accent, which is now set to orange.

scss/_settings_colors.scss
```diff
@@ -108,10 +105,8 @@ $colors--dark-theme--border-default: rgba($colors--dark-theme--text-default, 0.2
...
 $colors--dark-theme--border-low-contrast: rgba($colors--dark-theme--text-default, 0.1) !default;
 
-$colors--dark-theme--nav-highlight: $color-navigation-highlight !default;
-
 // Branding colors
 $color-brand: #333 !default;
 $color-brand-background: $colors--dark-theme--background-default;
-$color-accent: $color-brand !default;
+$color-accent: #e95420 !default;
 $color-accent-background: $color-accent !default;
```
anthonydillon, line 112: We probably still want this one as `$color-brand` as that was the previous experience.

anthonydillon (changes requested):
One small request in the code.

anthonydillon:
Can you also update the documentation please?

lyubomir-popov:
@anthonydillon updated the $acent-color swatch to orange.

scss/_settings_colors.scss
lyubomir-popov: do you mean the $color-accent-background?

anthonydillon (approved):
LGTM

PR #3775, opened 2021-06-03
Spacing vars update

lyubomir-popov:
## Done

Review usage and fix some issues around spacing vars.

### DIFF
--- a/scss/_patterns_breadcrumbs.scss
+++ b/scss/_patterns_breadcrumbs.scss
@@ -18,7 +18,7 @@
     @extend %default-text;
 
     display: inline-block;
-    margin-bottom: $spv-outer--small-scaleable - $spv-nudge;
+    margin-bottom: $spv-nudge-compensation;
 
     &:not(:first-of-type) {
       text-indent: $sph-outer;
--- a/scss/_patterns_breadcrumbs.scss
+++ b/scss/_patterns_breadcrumbs.scss
@@ -5,6 +5,7 @@
   .p-breadcrumbs {
     list-style: none;
     margin: 0;
+    overflow-y: auto;
     padding: 0;
     width: 100%;
 
--- a/scss/_patterns_breadcrumbs.scss
+++ b/scss/_patterns_breadcrumbs.scss
@@ -5,7 +5,6 @@
   .p-breadcrumbs {
     list-style: none;
     margin: 0;
-    overflow-y: auto;
     padding: 0;
     width: 100%;
 
--- a/scss/_patterns_pull-quotes.scss
+++ b/scss/_patterns_pull-quotes.scss
@@ -64,7 +64,7 @@
   }
 
   .p-pull-quote {
-    padding: 0 1rem;
+    padding: 0 2rem;
 
     .p-pull-quote__quote {
       @extend %vf-heading-4;
--- a/scss/_patterns_pull-quotes.scss
+++ b/scss/_patterns_pull-quotes.scss
@@ -64,7 +64,7 @@
   }
 
   .p-pull-quote {
-    padding: 0 $sph-outer * 2;
+    padding: 0 1rem;
 
     .p-pull-quote__quote {
       @extend %vf-heading-4;
@@ -77,7 +77,7 @@
   }
 
   .p-pull-quote--small {
-    padding: 0 $sph-outer * 1.5;
+    padding: 0 1.5rem;
 
     .p-pull-quote__quote {
       @extend %default-text;
@@ -90,7 +90,7 @@
   }
 
   .p-pull-quote--large {
-    padding: 0 $sph-outer * 2.5;
+    padding: 0 2.5rem;
 
     .p-pull-quote__quote {
       @extend %vf-heading-3;
--- a/scss/_patterns_pull-quotes.scss
+++ b/scss/_patterns_pull-quotes.scss
@@ -9,12 +9,12 @@
 
 @mixin vf-pull-quote {
   border: 0;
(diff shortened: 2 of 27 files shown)


anthonydillon:
Not sure if this is ready for review yet, but the few places of concern are breadcrumbs and notifications. 

Looks like lists have slightly less space below but that could be expected.

lyubomir-popov:
@anthonydillon thanks for reviewing these. I fixed the notifications and the breadcrumbs.

lyubomir-popov:
@anthonydillon for the lists, there's an invisible margin-bottom (which doesn't dstretch the example white background, but is there):
image

scss/_patterns_pull-quotes.scss
```diff
@@ -65,7 +64,7 @@
   }
 
   .p-pull-quote {
-    padding: 0 $sph-outer * 2;
+    padding: 0 2rem;
```
sowasred2012, line 67: Is it intended that these values are hardcoded, rather than variables?

lyubomir-popov: These are offsets to fit a quote at the size it is used in each pattern, done by eyeballing what looks good basically - and this has nothing to do with keeping spacing consistent across components, as no other components use quotes at those font sizes. I just wanted that to be clear. 

For example, if you want to change that quote size, and respectively the spacing around it, you wouldn't expect anything outside the quote component to also change - therefore it shouldn't use $sph-outer.

And viceversa, if you changed `$sph-outer` so that components like the matrix gert a different spacing of the icon, you wouldn't expect the quote spacing to change as it is unrelated.

scss/_settings_grid.scss
```diff
@@ -26,8 +26,9 @@ $grid-margin-widths: (
   large: $sp-unit * 3,
 ) !default;
 
-$grid-margin-width: $sph-outer--large !default;
-
+@include deprecate('3.0.0', 'Use the `$grid-margin-widths` map instead') {
```
bartaz, line 29: Could you add a note about this to `update-guide-3.0.md` as well?

scss/_settings_colors.scss
```diff
@@ -91,12 +91,9 @@ $colors--light-theme--border-low-contrast: rgba($color-x-dark, 0.1) !default;
 
 // Dark theme
 $colors--dark-theme--text-default: hsl(0, 0%, 100%) !default;
-$colors--dark-theme--nav-link-hover: $color-x-light !default;
-$colors--dark-theme--text-hover: hsl(0, 0%, 76%) !default; // minimum contrast on primary that satisfies contrast checker AA
 $colors--dark-theme--text-disabled: rgba($colors--dark-theme--text-default, 0.55) !default;
 $colors--dark-theme--text-muted: rgba($colors--dark-theme--text-default, $muted-text-opacity-amount) !default;
 $colors--dark-theme--text-inactive: rgba($colors--dark-theme--text-default, $inactive-text-opacity-amount) !default;
-$colors--dark-theme--nav-link-hover: $colors--dark-theme--text-hover !default;
```
bartaz, line 99: This doesn't seem related to spacing. I guess it's just a drive-by clean up - are those not used anywhere? (I see that we don't have those in light theme).

scss/_patterns_pull-quotes.scss
bartaz: If pull quote is a special case in terms of spacing we should comment on it in the code, so we know why vars are not used here.

Maybe worth adding some internal vars for this component only to be used instead?

scss/_settings_colors.scss
lyubomir-popov: nowhere in org:canonical-web-and-design

scss/_patterns_pull-quotes.scss
lyubomir-popov: My goal initially was to decouple this from the vars as it is an incorrect coupling. Didn't use vars as it isn't used in more than one place. The correct way to do this is to measure the width of the qutaiton mark in ems, add a spacing var and use that. I'll add an issue for that.

lyubomir-popov: Issue created here https://github.com/canonical-web-and-design/vanilla-framework/issues/3821

scss/_settings_colors.scss
lyubomir-popov: @bartaz ready for another look.

scss/_patterns_breadcrumbs.scss
```diff
@@ -5,6 +5,7 @@
   .p-breadcrumbs {
     list-style: none;
     margin: 0;
+    overflow-y: auto;
```
bartaz, line 8: Why the change in overflow?

lyubomir-popov: to make sure the margin-bottom of children is covered by the height of the parent. Otherwise it still applies, but it looks like it doesn't - the background of the example doesn't extend to cover it.

lyubomir-popov: This should also be applied to the body, otherwise many components look like they have no margin-bottom, as @anthonydillon pointed out further up in this thread.

bartaz: I don't think we should apply it for the sake of making examples look "better", we would need to apply this to every component in Vanilla and it can have unexpected consequences (scrollbars appearing).

If we want to make margins visible in examples we need to find different solution - maybe by applying some styles to body only in the examples? But that's something for another PR.

I'd remove this overflow from here as it doesn't seem to be needed for correct styling of the component.

bartaz (approved):
LGTM

PR #3837, opened 2021-06-30
Update webpack build steps

andesol:
## Done

Minor documentation fixes on how to build with Webpack.

### DIFF
--- a/templates/docs/building-vanilla.md
+++ b/templates/docs/building-vanilla.md
@@ -86,7 +86,7 @@ yarn add node-sass css-loader sass-loader style-loader
 Make a folder `src/`, create a file inside called `style.scss` and import Vanilla:
 
 ```
-@import vanilla-framework/scss/vanilla;
+@import 'vanilla-framework/scss/vanilla';
 @include vanilla;
 ```
 
@@ -97,7 +97,7 @@ var path = require('path');
 
 module.exports = {
   mode: 'development',
-  entry: './src/index.js',
+  entry: './src/app.js',
   output: {
     filename: 'bundle.js',
     path: path.resolve(__dirname, 'dist')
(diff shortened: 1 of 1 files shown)


templates/docs/building-vanilla.md
```diff
@@ -118,10 +118,6 @@ module.exports = {
           {
             // Loads a SASS/SCSS file and compiles it to CSS
             loader: 'sass-loader',
-            options: {
```
bartaz, line 121: @andesol Are you sure this can be removed? I think `includePaths` was needed to make sure `.scss` files from within Vanilla files in node_modules are properly loaded as well.

bartaz (commented):
Thanks @andesol for taking time to review and update this page.
I just have one question about your changes below.

templates/docs/building-vanilla.md
andesol: It's possible to pass options nesting them inside `sassOptions`, like this:
```            
loader: "sass-loader",
options: {
   sassOptions: {
      includePaths: ["absolute/path/a", "absolute/path/b"],
   }
```
But I found that to be unnecessary since webpack heads to `node_modules` when resolving imports, as stated in the [documentation](https://webpack.js.org/loaders/sass-loader/#resolving-import-at-rules): 

> The loader will first try to resolve @import as a relative path. If it cannot be resolved, then the loader will try to resolve @import inside node_modules.

bartaz: Great, thanks for detailed answer @andesol, when using Vanilla internally we rarely need to set up new webpack project with it, so our docs here got a bit rusty :)

bartaz (approved):
LGTM, thanks!

PR #3899, opened 2021-08-03
Add password toggle pattern

steverydz:
## Done
Added a pattern for showing and hiding the value of password fields

## QA
- Open [demo](https://vanilla-framework-3899.demos.haus/docs/examples/patterns/forms/password-toggle)
- Check that the pattern matches [the spec](https://docs.google.com/document/d/1Q5i2DI801rKA4XO8MZv0dldUUGGbB6zPSWsRVLf8mk0/edit)

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "percy": "percy exec -- node snapshots.js",
     "icon-svgs-to-mixins": "node scripts/convert-svgs-to-icon-mixins.js icons"
   },
-  "version": "2.33.1",
+  "version": "2.34.0",
   "files": [
     "/scss",
     "!/scss/docs"
--- a/scss/_patterns_form-password-toggle.scss
+++ b/scss/_patterns_form-password-toggle.scss
@@ -1,0 +1,30 @@
+@import 'settings';
+
+@mixin vf-p-form-password-toggle {
+  .p-form-password-toggle {
+    display: flex;
+    justify-content: space-between;
+
+    .p-form-password-toggle__button {
+      @include vf-button-pattern(
+        $button-background-color: $color-transparent,
+        $button-border-color: $color-transparent,
+        $button-hover-border-color: $color-transparent,
+        $button-active-border-color: $color-transparent,
+        $button-disabled-border-color: $color-transparent
+      );
+
+      font-size: #{map-get($font-sizes, small)}rem;
+      line-height: map-get($line-heights, small);
+      padding: calc(#{map-get($nudges, nudge--small)} - 1px) $sph-inner--small;
+      width: auto !important;
+
+      @extend %u-no-margin--bottom--default-text;
+    }
+  }
+
+  .p-form-password-toggle__label {
+    display: inline-block;
+    margin-right: $sp-xx-small;
+  }
+}
--- a/scss/_patterns_form-password-toggle.scss
+++ b/scss/_patterns_form-password-toggle.scss
@@ -2,24 +2,14 @@
 
 @mixin vf-p-form-password-toggle {
   .p-form-password-toggle {
+    align-items: flex-start;
     display: flex;
     justify-content: space-between;
 
-    .p-form-password-toggle__button {
-      @include vf-button-pattern(
-        $button-background-color: $color-transparent,
-        $button-border-color: $color-transparent,
-        $button-hover-border-color: $color-transparent,
-        $button-active-border-color: $color-transparent,
-        $button-disabled-border-color: $color-transparent
-      );
-
-      font-size: #{map-get($font-sizes, small)}rem;
-      line-height: map-get($line-heights, small);
-      padding: calc(#{map-get($nudges, nudge--small)} - 1px) $sph-inner--small;
-      width: auto !important;
-
-      @extend %u-no-margin--bottom--default-text;
+    .p-button--base {
+      @media (max-width: $breakpoint-small) {
+        width: auto !important;
+      }
     }
   }
 
--- a/templates/docs/base/forms.md
+++ b/templates/docs/base/forms.md
@@ -206,6 +206,21 @@ In contexts where vertical space is limited, e.g. inside a table row, you might 
 View example of the dense form elements
 </a></div>
 
+### Password toggle
+
+When using a password field, use this pattern to allow the user to toggle the password visibility.
+
+<div class="p-notification--information is-inline">
+  <div class="p-notification__content">
+    <h5 class="p-notification__title">Note:</h5>
+    <p class="p-notification__message">Do not use this pattern in inline forms</p>
+  </div>
+</div>
+
+<div class="embedded-example"><a href="/docs/examples/patterns/forms/password-toggle/" class="js-example">
+View example of the password toggle
+</a></div>
+
 ### Import
 
 To import just this base element into your project, copy the snippet below and include it in your main Sass file.
--- a/templates/docs/component-status.md
+++ b/templates/docs/component-status.md
@@ -22,7 +22,29 @@ When we add, make significant updates, or deprecate a component we update their 
     </tr>
   </thead>
   <tbody>
-    <!-- 2.33 -->
+    <!-- 2.34 -->
+      <tr>
+        <th><a href="/docs/base/forms#password-showhide">Forms / Password toggle</a></th>
+        <td><div class="p-label--new">New</div></td>
+        <td>2.34.0</td>
+        <td>Password fields now have a show/hide toggle.</td>
+      </tr>
+  </tbody>
+</table>
+
+#### Previously in Vanilla
+
+<table>
+  <thead>
+    <tr>
+      <th style="width: 20%">Component</th>
+      <th style="width: 15%">Status</th>
+      <th style="width: 10%">Version</th>
+      <th style="width: 55%">Notes</th>
+    </tr>
+  </thead>
+  <tbody>
+  <!-- 2.33 -->
     <tr>
       <th><a href="/docs/patterns/links#skip-link">Links / Skip Link</a></th>
       <td><div class="p-label--new">New</div></td>
@@ -41,21 +63,6 @@ When we add, make significant updates, or deprecate a component we update their 
       <td>2.33.0</td>
       <td>The following notification classes have been deprecated: <code>.p-notification__response</code>, <code>.p-notification__status</code></td>
     </tr>
-  </tbody>
-</table>
-
-#### Previously in Vanilla
-
-<table>
-  <thead>
-    <tr>
-      <th style="width: 20%">Component</th>
-      <th style="width: 15%">Status</th>
-      <th style="width: 10%">Version</th>
-      <th style="width: 55%">Notes</th>
-    </tr>
-  </thead>
-  <tbody>
     <!-- 2.32 -->
     <tr>
       <th><a href="/docs/patterns/labels#default">Labels / Default</a></th>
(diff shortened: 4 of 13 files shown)


ziheliu214:
@steverydz 
Hola thanks for your work!
On mobile size the position of the hide/show button is off, could you please check?
Screenshot 2021-08-03 at 11 55 02

package.json
```diff
@@ -43,7 +43,7 @@
     "percy": "percy exec -- node snapshots.js",
     "icon-svgs-to-mixins": "node scripts/convert-svgs-to-icon-mixins.js icons"
   },
-  "version": "2.33.1",
+  "version": "2.34.1",
```
sowasred2012, line 46:
```suggestion
  "version": "2.34.0",
```

templates/docs/examples/patterns/password-toggle/default.html
```diff
@@ -0,0 +1,50 @@
...
+      <i class="p-icon--show"></i>
+      <span class="p-password-toggle__label">
+        Show
+      </span>
+    </button>
+  </div>
+  <input type="password" name="password" id="password" value="letmein">
+</div>
+
+<script>
```
sowasred2012, line 20: We try to keep JS separate from HTML examples, see the [tabs example](https://github.com/canonical-web-and-design/vanilla-framework/blob/master/templates/docs/examples/patterns/tabs/content.html#L54).

templates/docs/examples/patterns/password-toggle/default.html
```diff
@@ -0,0 +1,50 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Password toggle field{% endblock %}
+
+{% block standalone_css %}patterns_password-toggle{% endblock %}
+
+{% block content %}
+<div class="p-password-toggle">
+  <div class="p-password-toggle__container">
+    <label for="password">Password</label>
+    <button class="p-password-toggle__button" aria-controls="password">
```
sowasred2012, line 10: Accessibility Man strikes again! I think we need assistive technology to announce when this has changed, which we'd get with the `aria-live` attribute:

```suggestion
    <button class="p-password-toggle__button" aria-live="polite" aria-controls="password">
```

sowasred2012 (commented):
Looking good! Had a few comments, and we need to get your [question about inline/stacked form inputs](https://docs.google.com/document/d/1Q5i2DI801rKA4XO8MZv0dldUUGGbB6zPSWsRVLf8mk0/edit?disco=AAAANmsnZPI) resolved too. @ziheliu214 @lyubomir-popov what do you think (see link)?

steverydz:
@sowasred2012 thanks - I've made all of the above changes

bartaz:
@ziheliu214 @lyubomir-popov 

It seems it was already asked in the spec doc, but maybe worth repeating here too. How this should work with inline/stacked forms? Should we just say: only use that in standard forms where label is above the input?

Screenshot 2021-08-04 at 09 30 01
Screenshot 2021-08-04 at 09 30 08


Also, from visual point of view. The "show/hide" button doesn't have any hover styles and because it's right aligned the icon shifts a bit when toggling (because "hide" is narrower than "show"), which looks bit weird. What should we do about that?

form-password-toggle

templates/docs/examples/patterns/password-toggle/default.html
```diff
@@ -0,0 +1,23 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Password toggle field{% endblock %}
+
+{% block standalone_css %}patterns_password-toggle{% endblock %}
+
+{% block content %}
+<div class="p-password-toggle">
```
bartaz, line 7: @steverydz I guess you wanted to use it as a root of a component, but I think we don't need this root element here on this level, as it doesn't seem to have any styling at all.

It seems that the only changes needed are with the label + button area, so instead of having:

```html
<div class="password-toggle">
   <div class="password-toggle__container">
      <label></label>
      <button class="p-password-toggle__button>Show</button>
    </div>
</div>
```

we could drop the root element and just make the `__container` the root of this component:

```html
<div class="p-password-toggle">
   <label></label>
   <button class="p-password-toggle__button>Show</button>
</div>
```

We already have a wrapper for form fields when they are validated `p-form-validation`, so let's not add another one. Also, as a password field, it would be good to have an example of this toggle working with validation styles as well.

bartaz: The other thing I'm not sure about is the name `p-password-toggle`.

Truth is, that naming consistency in Vanilla forms is a bit messy. We have `p-form` `p-form__group` `p-form__control` `p-form-validation` all mixed up without good structure. But adding `p-password-toggle` on top of that just adds to the mess.

It at least should be `p-form-password-toggle`, or `p-form-visibility-toggle`? I don't know, but having `p-form` prefix at least suggests that it's part of form components.

lyubomir-popov:
@bartaz how about flip the text and icon to avoid the movement on show/hide?

lyubomir-popov:
Can we just use a base button for the toggle? this way we get the consistent button styling, and the small indent on the right isn't much of an issue.

steverydz:
@bartaz @lyubomir-popov I've addressed both of your comments.

I agree that maybe we just don't use this pattern on inline forms?

lyubomir-popov:
@steverydz I tend to agree. It would be highly unusual to use an inline form for password authentication in my opinion. 
cc @davegoddard42 who did the ux here I believe.

davegoddard42:
@steverydz @lyubomir-popov It was myself and @rayito paired for this one

Just to be clear, are we suggesting that we don't use the password toggle at all on inline fields, therefore a password input can never be an inline field? I would be happy with that personally. I don't know your thoughts Edén?

steverydz:
@davegoddard42 yes, we are suggesting we don't use the toggle pattern if the password is inline, however we can still use the password input independently.

lyubomir-popov:
We're trying to simplify the problem by elliminating an unlikely scenario. So the question is:
Is an inline password field a valid need? If not then we're golden.

davegoddard42:
Sounds good to me.

bartaz:
The CI test failure because of file size troubles me a bit:

```
✘ - Stylesheet size (327017) exceeds threshold (310000)
```

While usually we just bump the threshold it feels strange that the difference is so large. This PR doesn't seem to add that much of CSS. I'll try to look into it in more detail tomorrow.

scss/_patterns_form-password-toggle.scss
```diff
@@ -0,0 +1,34 @@
+@import 'settings';
+@import 'base_placeholders';
+@import 'patterns_icons';
+@include vf-b-placeholders;
+@include vf-p-icon-hide;
+@include vf-p-icon-show;
```
bartaz, line 6: @steverydz  Are these imports/includes needed here?

We don't import or include anything in patterns CSS, because it will be already provided by Vanilla in base styles.
I guess these imports/includes may be the reason the file size is so much larger in this PR.

bartaz:
On a mobile screen size buttons get 100% width by default, messing up with positioning.
I guess adding `width: auto` should solve that.

Screenshot 2021-08-04 at 16 49 46

Also, I don't think we need all off different validation states here in this example (it's covered in validation example where it's needed), so just one of them would be enough, thanks!

scss/_patterns_form-password-toggle.scss
steverydz: @bartaz I can remove placeholders as they were only needed when I was using the button link styles. The icon patterns are needed though, otherwise we don't have the mixins for the show and hide icons.

templates/docs/base/forms.md
```diff
@@ -206,6 +206,21 @@ In contexts where vertical space is limited, e.g. inside a table row, you might
 View example of the dense form elements
 </a></div>
 
+### Password show/hide
```
sowasred2012, line 209:
```suggestion
### Password toggle
```

templates/docs/base/forms.md
```diff
@@ -206,6 +206,21 @@ In contexts where vertical space is limited, e.g. inside a table row, you might
...
+
+<div class="p-notification--information is-inline">
+  <div class="p-notification__content">
+    <h5 class="p-notification__title">Note:</h5>
+    <p class="p-notification__message">Do not use this pattern in inline forms</p>
+  </div>
+</div>
+
+<div class="embedded-example"><a href="/docs/examples/patterns/forms/password-toggle/" class="js-example">
+View example of the password show/hide
```
sowasred2012, line 221:
```suggestion
View example of the password toggle
```

templates/docs/component-status.md
```diff
@@ -22,7 +22,29 @@ When we add, make significant updates, or deprecate a component we update their
     </tr>
   </thead>
   <tbody>
-    <!-- 2.33 -->
+    <!-- 2.34 -->
+      <tr>
+        <th><a href="/docs/base/forms#password-showhide">Forms / Password show/hide</a></th>
```
sowasred2012, line 27:
```suggestion
        <th><a href="/docs/base/forms#password-showhide">Forms / Password toggle</a></th>
```

templates/docs/examples/patterns/forms/_script.js
```diff
@@ -0,0 +1,25 @@
+(function () {
+  var passwordToggleButtons = Array.prototype.slice.call(document.querySelectorAll('.p-form-password-toggle__button'));
+
+  passwordToggleButtons.forEach(function (passwordToggleButton) {
+    var passwordToggleLabel = passwordToggleButton.querySelector('.p-form-password-toggle__label');
```
sowasred2012, line 5: It's a little thing, but I'd be tempted to make this target the element rather than the class, we generally try to keep classes out of JS examples as much as possible, and it'd be more consistent with how you're getting the icon element:
```suggestion
    var passwordToggleLabel = passwordToggleButton.querySelector('span');
```

scss/_patterns_form-password-toggle.scss
```diff
@@ -0,0 +1,30 @@
...
+        $button-active-border-color: $color-transparent,
+        $button-disabled-border-color: $color-transparent
+      );
+
+      font-size: #{map-get($font-sizes, small)}rem;
+      line-height: map-get($line-heights, small);
+      padding: calc(#{map-get($nudges, nudge--small)} - 1px) $sph-inner--small;
+      width: auto !important;
+
+      @extend %u-no-margin--bottom--default-text;
```
sowasred2012, line 22: `@extend` should come before `@include`

sowasred2012 (commented):
Sorry - changed my mind on the "show/hide"/"toggle" thing! Added a couple of other minor comments

steverydz:
@sowasred2012 all points addressed

sowasred2012:
Looking good - I've added a design needed label, I think it'd be good to get @lyubomir-popov's thoughts

scss/_patterns_form-password-toggle.scss
```diff
@@ -0,0 +1,29 @@
+@import 'settings';
+
+@mixin vf-p-form-password-toggle {
+  .p-form-password-toggle {
+    display: flex;
+    justify-content: space-between;
+
+    .p-form-password-toggle__button {
```
lyubomir-popov, line 8: I propose using the standard p-button--base with a u-no-margin--bottom class on it to keep things on   the baseline grid. THen you only need to tell it to no go full-width on mobile.

lyubomir-popov (commented):
Some comments in the code.

scss/_patterns_form-password-toggle.scss
lyubomir-popov: needs the `has-icon` class too

lyubomir-popov:
Nice work!

lyubomir-popov (approved):
LGTM

sowasred2012 (approved):
:tada: LGTM! :tada:

PR #3910, opened 2021-08-05
WIP: Allow card content to bleed into margins for backgrounds

steverydz:
## Done
Added negative margins to card content so that it can have a background that fills the whole card

## QA
- Check the card examples:
  - https://vanilla-framework-3910.demos.haus/docs/examples/patterns/card/default
  - https://vanilla-framework-3910.demos.haus/docs/examples/patterns/card/header
  - https://vanilla-framework-3910.demos.haus/docs/examples/patterns/card/highlighted
  - https://vanilla-framework-3910.demos.haus/docs/examples/patterns/card/overlay

## Issue
Fixes https://github.com/canonical-web-and-design/vanilla-squad/issues/971

### DIFF
--- a/scss/_base_placeholders.scss
+++ b/scss/_base_placeholders.scss
@@ -41,6 +41,11 @@
     padding: $spv-inner--large;
   }
 
+  %vf-card-content {
+    margin: -#{$spv-inner--large};
+    padding: $spv-inner--large;
+  }
+
   //accordion, table
   %single-border-text-vpadding--scaling {
     padding-bottom: $table-cell-vertical-padding;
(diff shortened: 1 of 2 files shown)


steverydz:
@lyubomir-popov @bartaz @sowasred20 Here's a WIP of the card content bleed ticket. It seems to work, although obviously something's off with the spacing according to Percy.

bartaz:
While negative margin probably works technically it feels more like a hack. It seems that we may need to break current use cases and create updated card component to fit out new needs.

Upcomming 3.0 may be a good opportunity for that.  But the cards are so very common on our sites, that if we break it, we need to provide smooth way to update (ideally without HTML structure changes, just a class rename), because it may be really hard to find and update all cards on our sites if they need HTML structure changes.

steverydz:
is this something we want to park for now then?

Issue #3897, opened 2021-07-30
Tab content accessibility docs are vague


sowasred2012:
https://vanillaframework.io/docs/patterns/tabs#content should be more explicit that keyboard events (specifically arrow key presses) are needed for this pattern to be accessible. The example JS includes these keyboard events, but it's not specifically called out in the copy, and easily missed.

Noticed in: https://github.com/canonical-web-and-design/ubuntu.com/pull/10061

PR #3921, opened 2021-08-11
`p-tab-buttons` pattern

### LINKED ISSUE #3887: Segmented control - build
## UX work:
https://docs.google.com/document/d/1__Ee_7s1jLJdNBEyrA-fUKNHx5Jbxz158IZuWtb6zEA/edit#heading=h.shysd5ek3g27

## UX issue:
https://github.com/canonical-web-and-design/vanilla-framework/issues/3571

## Link to Zeplin:
[Zeplin](https://zpl.io/2vDWD1Q)

steverydz:
Co-authored-by: Scott Mason Nash <scott.mason.nash@canonical.com>

## Done

Added the tab buttons pattern

Fixes #3887 & #3897

## QA

- Open [demo](https://vanilla-framework-3921.demos.haus/docs/examples/patterns/tab-buttons/default)
- Check that the pattern looks and behaves as expected
  - https://vanilla-framework-3921.demos.haus/docs/patterns/tab-buttons

- [x] Any changes to component class names (new patterns, variants, removed or added features) should be listed on the [component status page](https://github.com/canonical-web-and-design/vanilla-framework/blob/master/templates/docs/component-status.md).


scss/_patterns_segmented-control.scss
```diff
@@ -0,0 +1,60 @@
...
+      border-right: 0;
+    }
+  }
+
+  .p-segmented-control {
+    @extend %vf-has-round-corners;
+
+    border: 1px solid $color-mid-dark;
+    display: inline-flex;
+    margin-bottom: $spv-outer--small-scaleable;
```
lyubomir-popov, line 18:
```suggestion
    margin-bottom: $spv-outer--scaleable;
```

scss/_patterns_segmented-control.scss
```diff
@@ -0,0 +1,60 @@
...
+        outline: $bar-thickness solid $color-focus;
+        position: relative;
+        z-index: 1;
+      }
+    }
+
+    &.is-dense {
+      button,
+      [class*='p-button'],
+      [type='radio'] + span {
```
lyubomir-popov, line 53: can we avoid element selectors? It's a new component so why not a class instead of the span selector?

scss/_patterns_segmented-control.scss
```diff
@@ -0,0 +1,60 @@
...
+        padding-bottom: calc(#{$spv-nudge - $sp-unit * 0.5} - 1px);
```
lyubomir-popov, line 54: this si repeated in _base_buttons.scss; can you move to a placeholder and extend it from here and from the base buttton?

scss/_patterns_segmented-control.scss
```diff
@@ -0,0 +1,60 @@
+@mixin vf-p-segmented-control {
+  %vf-button-group-borders {
+    border-bottom: 0;
```
lyubomir-popov, line 3: A bit worried that we're moving the borders like this. Why not keep them but make them transparent if needed?

scss/_patterns_segmented-control.scss
```diff
@@ -0,0 +1,60 @@
...
+
+      margin-bottom: 0;
+      padding-top: 0;
+    }
+
+    [type='radio'] {
+      @extend %vf-hidden-tick-input;
+
+      &:checked + span {
+        background-color: $color-mid-x-light;
```
lyubomir-popov, line 40: The buttons are themed, this component shjould be able to leverage the theming. Have you tested with dark theme buttons? This colour would need to come from the themese from example.

scss/_patterns_segmented-control.scss
```diff
@@ -0,0 +1,60 @@
...
+
+    &:last-child {
+      border-right: 0;
+    }
+  }
+
+  .p-segmented-control {
+    @extend %vf-has-round-corners;
+
+    border: 1px solid $color-mid-dark;
```
lyubomir-popov, line 16: This will interfere with the theming. Ideally applying is-dark on every button should immediately turn the thing dark without any work needed on this component - although  it would be nice to be able to do it from the top level with an is-dark classs

lyubomir-popov (changes requested):

scss/_patterns_segmented-control.scss
```diff
@@ -0,0 +1,60 @@
...
+    display: inline-flex;
+    margin-bottom: $spv-outer--small-scaleable;
+
+    // stylelint-disable selector-max-type
+    button,
```
bartaz, line 21: We shouldn't style class-less elements inside components, and we should avoid overriding style of other class names `p-button` inside components. This creates situations when elements or components look and behave differently in different places.

This should be something like `p-segmented-control__button` and no other external class name should be needed.

scss/_patterns_segmented-control.scss
bartaz: Yes, we should use class names instead of element selectors.

templates/docs/examples/patterns/segmented-control.html
```diff
@@ -0,0 +1,72 @@
...
+  <button><i class="p-icon--information"></i> Two</button>
+  <button>Three</button>
+  <a class="p-button" href="#">Link</a>
+</div>
+
+<br><br>
+
+<div class="p-segmented-control">
+  <label class="p-segmented-control__radio">
+    <input type="radio" name="radio" aria-labelledby="radio-one-input">
```
bartaz, line 19: Do we need both `button` and `radio` implementations? If so, what are the usecases for them? Documentation doesn't mention it, so which one should developer choose?

scss/standalone/patterns_segmented-control.scss
```diff
@@ -0,0 +1,6 @@
+@import '../vanilla';
+@include vf-base;
+
+@include vf-p-segmented-control;
+@include vf-p-buttons;
```
bartaz, line 5: Does it need to rely on `p-buttons`? New components should be independent from each other and only rely on base styles ideally.

scss/_patterns_segmented-control.scss
bartaz: Theming here is a tricky question. We didn't have situation like this before that parent component would have child component that may be themed differently.

Now that we added themes to buttons it is technically possible, but should it be allowed?
I guess we don't want to have light/dark buttons mixed inside one segmented control, so it's the segmented control that should control the theme here, not the buttons.

I would say theming is out of scope of this PR for now, we need to think on Vanilla level how we want to approach such situations.

scss/_patterns_segmented-control.scss
bartaz: I'm also bit worried about taking the border out of the buttons. It seems to be affecting focus a bit (buttons are smaller) and overall will make it harder to keep this component aligned with buttons.

It seems that maybe we need to refactor buttons styles a bit to make it easier to reuse their styling in other places.

templates/docs/examples/patterns/segmented-control.html
sowasred2012: Existing implementations (Charmhub etc) are just buttons, not radios, so we went with buttons at first, and thought a wrapper around existing button pattern would be easy to understand.

However, the designs show a selected state, which we don't need in the button use case, but we could see that this pattern could easily be used as an alternative to content tabs, so we wanted to show we could support both. If we had to drop one, I'd think it would be radios since it hasn't been created in other projects already - what do you think?

scss/_patterns_segmented-control.scss
lyubomir-popov: Yes, not implying we should add themeing support here - but we should implement new components in a way that fascilitates theming in the future. Adding the is-dark classes on every button here is just a way to uncover the hidden problems of the approach taken here. We should avoid moving borders from one element to another, especialy when borders and backgrounds have multiple states.

Once implemented cleanly, it should be possible to add is-dark classes and simulate theming here - although of course not something we advertise or recommend, just a sanity check.

scss/standalone/patterns_segmented-control.scss
lyubomir-popov: This is a bit of a gray area IMO.  This segmented control is a collection of buttons. What would be the alternative - move most of the button code into base?

sowasred2012: We looked at it as we were just adding a wrapper around buttons, but dropping their border styles - I think it'd be best to chat it out after standup.

lyubomir-popov:
Can this margin be 0 instead of -.5rem? This way thespacing on the outside remains consistent whether you have an icon or not:
image

templates/docs/patterns/tabs.md
```diff
@@ -28,12 +28,32 @@ View example of the tabs navigation pattern
...
+  </div>
+</div>
 
 <div class="embedded-example"><a href="/docs/examples/patterns/tabs/content" class="js-example">
 View example of the tabs content pattern
 </a></div>
 
+### Segmented
+
+Segmented tabs can be used for secondary tabs or view switchers. There is an `is-dense` modifier for condensed buttons.
```
lyubomir-popov, line 44: Have we defined when to use the dense and when to use the regular version? @danielmutis anything we can copy from your spec into the documentation here?

scss/_patterns_tabs.scss
```diff
@@ -93,4 +93,37 @@
...
     }
   }
+
+  .p-tabs-segmented {
+    &.is-dense .p-tabs-segmented__button {
+      @extend %vf-button-dense-vertical-spacing;
+    }
+
+    .p-tabs-segmented__list {
+      display: flex;
```
lyubomir-popov, line 103: without this, the buttons are stretched to fill the parent. If for whatever reason something is different (e.g. if you apply `.is-dense` on one of the buttons, not that you should) then unexpected stuff happens.
```suggestion
      display: flex;
      align-items: flex-start;
```

petermakowski:
Looks like _segmented_ is only a stylistic variation of the base pattern - they appear to behave identically and have they same meaning. "Segmented" is not a bad name, but it seems to indicate some core/structural differences. I had some other ideas, just throwing them out there:

- "tabs-secondary" - fairly well known convention for distinguishing different variations of the same pattern (primary, secondary etc.). Not a good choice if you expect this pattern to further diverge from regular tabs component in the future.
- "tabs-bordered" / "tabs-bar" / - emphasising the stylistic difference - bordered style / that it forms a single bar

scss/_patterns_tabs.scss
```diff
@@ -93,4 +93,37 @@
       }
     }
   }
+
+  .p-tabs-segmented {
+    &.is-dense .p-tabs-segmented__button {
```
lyubomir-popov, line 98: Every component should push things that come after it by the same amount. Also, to ensure child elements' margins are inside the component, we need overflow: auto;
```    margin-bottom: $spv-outer--scaleable;
         overflow: auto;
```

lyubomir-popov (changes requested):
Sorry, still a few things to address - and a quesiton to @danielmutis. Unless we can come up with clear guidance when to use dense/regular, we should just choose one of the options and drop the other.

sowasred2012:
> Can this margin be 0 instead of -.5rem? This way thespacing on the outside remains consistent whether you have an icon or not:

@lyubomir-popov we extended the `vf-button-has-icon` mixin for this, which does the same thing on other buttons, e.g.

Screenshot from 2021-08-13 12-05-49

Is that intended? Or should it be different for this pattern?

scss/_patterns_tabs.scss
sowasred2012: Ah, we were following `is-dense` button modifier, which reduces the bottom margin.

templates/docs/examples/patterns/tabs/segmented.html
```diff
@@ -0,0 +1,99 @@
...
+      class="p-tabs-segmented__button"
+      role="tab"
+      aria-selected="false"
+      aria-controls="charmhub-tab"
+      id="charmhub"
+      tabindex="-1"
+    >Charmhub</button>
+  </div>
+
+  <div tabindex="0" role="tabpanel" id="olm-tab" aria-labelledby="olm">
```
bartaz, line 38: Are these panels part of the component, do they need to be inside the same wrapper?
In real use cases any panel controlled by segmented buttons will probably in different part of the app screen.

sowasred2012:
> Looks like _segmented_ is only a stylistic variation of the base pattern - they appear to behave identically and have they same meaning. "Segmented" is not a bad name, but it seems to indicate some core/structural differences. I had some other ideas, just throwing them out there:
> 
>     * "tabs-secondary" - fairly well known convention for distinguishing different variations of the same pattern (primary, secondary etc.). Not a good choice if you expect this pattern to further diverge from regular tabs component in the future.
> 
>     * "tabs-bordered" / "tabs-bar" / - emphasising the stylistic difference - bordered style / that it forms a single bar

Thanks for these suggestions! I had a few thoughts:

- `tabs-secondary` might confuse it with the use case of the tabs navigation pattern, which is occasionally used as a secondary (or tertiary) nav
- I like `tabs-bar`, it feels appropriate, and distinguishes it from the other variants better than "segmented" does.

@lyubomir-popov @bartaz what do you think?

templates/docs/examples/patterns/tabs/segmented.html
sowasred2012: No, they don't need to be, I'll update

bartaz:
> * I like `tabs-bar`, it feels appropriate, and distinguishes it from the other variants better than "segmented" does.

Isn't our standard tabs component a "tabs bar" as well? :)

Maybe "tabs button" (in place of segmented button)? It makes this both adjacent to "tabs" which it represents, and "button" that its style is based on.

sowasred2012:
> > * I like `tabs-bar`, it feels appropriate, and distinguishes it from the other variants better than "segmented" does.
> 
> Isn't our standard tabs component a "tabs bar" as well? :)
> 
> Maybe "tabs button" (in place of segmented button)? It makes this both adjacent to "tabs" which it represents, and "button" that its style is based on.

lol, it's easier to name children than it is patterns/components :rofl:  - I would maybe have described the standard one as more of a strip, but yeah, it could also be a bar. So, `p-tabs-buttons`?

scss/_patterns_tabs.scss
sowasred2012: @lyubomir-popov I reverted the `overflow: auto` and `align-items: flex-start` changes, I noticed they broke the pattern at smaller widths:
buttons

lyubomir-popov:
> @lyubomir-popov we extended the `vf-button-has-icon` mixin for this, which does the same thing on other buttons, e.g.
You're right, the space on theright threww me off. It is currently bigger than in the button with icon, can we make it the same?

image

lyubomir-popov:
@sowasred2012 There are two  options, I would vote for the second one here - truncation. Tabs shouldn't wrap, text should be kept as short as possible IMO. If you need to write a sentence then the tab is not the correct component for this case.

image

what is the issue with overflow: auto? margins shouldn't leak outside the parent container, maybe just `overflow-y: auto` would be better?

lyubomir-popov:
@sowasred2012 @bartaz  `p-tab-buttons`? Double plural sounds strange to me.

templates/docs/examples/patterns/tabs/tab-buttons.html
```diff
@@ -0,0 +1,97 @@
...
+</div>
+
+<div tabindex="0" role="tabpanel" id="charmhub-tab" aria-labelledby="complex" hidden="hidden">
+  <p>A repository for charms - from Observability to Data to Identity and more.</p>
+</div>
+
+<div class="p-tab-buttons is-dense">
+  <div class="p-tab-buttons__list" role="tablist" aria-label="Juju technology">
+    <button
+      class="p-tab-buttons__button has-icon"
```
bartaz, line 53: It says `has-icon`, but this button doesn't have the icon.
The next one has the icon, but doesn't have `has-icon` class. 

Is this class needed here at all?

templates/docs/examples/patterns/tabs/tab-buttons.html
```diff
@@ -0,0 +1,97 @@
...
+
+<div tabindex="0" role="tabpanel" id="operator-tab" aria-labelledby="operator" hidden="hidden">
+  <p>A set of tools to help you write Charmed Operators and to package them as Charms.</p>
+</div>
+
+<div tabindex="0" role="tabpanel" id="charmhub-tab" aria-labelledby="complex" hidden="hidden">
+  <p>A repository for charms - from Observability to Data to Identity and more.</p>
+</div>
+
+<div class="p-tab-buttons is-dense">
```
bartaz, line 50: I guess we should maybe keep this in separate example as different variant ("kind", "size" or something 😉).

templates/docs/examples/patterns/tabs/tab-buttons.html
```diff
@@ -0,0 +1,97 @@
...
+<div class="p-tab-buttons">
+  <div class="p-tab-buttons__list" role="tablist" aria-label="Juju technology">
+    <button
+      class="p-tab-buttons__button has-icon"
+      role="tab" 
+      aria-selected="true" 
+      aria-controls="olm-tab" 
+      id="olm"
+    >
+      <i class="p-icon--information"></i><span>Charmed Operator Lifecycle Manager</span>
```
bartaz, line 17: The spec suggests that you should only use icons if you can put icons on all buttons. I think we should make our examples also consistent with this recommendation.

So probably we need a separate example with icons on all buttons?

templates/docs/examples/patterns/tabs/tab-buttons.html
sowasred2012: Ahhhh - you're right. We extended the `vf-button-has-icon` mixin and mistakenly assumed the class was needed, but you're right, it's not.

scss/standalone/patterns_tab-buttons.scss
```diff
@@ -0,0 +1,5 @@
+@import '../vanilla';
+@include vf-base;
+
+@include vf-p-tabs;
+@include vf-p-icon-info;
```
bartaz, line 5: It would be good to clearly separate what is a source of given component styles and what is just needed for the examples.

Usually I was grouping them properly and commenting in code - there should be some examples in other standalone stylesheets.

scss/_patterns_tab-buttons.scss
```diff
@@ -0,0 +1,37 @@
+@mixin vf-p-tab-buttons {
+  .p-tab-buttons {
+    &.is-dense .p-tab-buttons__button {
```
bartaz, line 3: Doesn't linter warn about this line here? It seems like something more specific that `.p-tab-buttons__button`, so probably could be defined after it. Also to make sure the styles are overridden?

sowasred2012: @bartaz hmmm - it _did_ warn about this being after buttons in a previous version of this, so we moved it here. It doesn't warn currently, but I've moved it after the button, and it no longer warns, so, everyone wins!

templates/docs/patterns/tab-buttons.md
```diff
@@ -0,0 +1,62 @@
+---
+wrapper_template: '_layouts/docs.html'
+context:
+  title: Tab Buttons | Components
+---
+
+## Tab Buttons
```
bartaz, line 7: I think we could still keep this as part of existing "Tabs" documentation page, same as on "Navigation" we have different components that do navigation. Especially that `p-tabs` and `p-tab-buttons` are functionally very similar.

If we want to keep them on separate pages (I'm fine with that, it is a separate component) we should maybe link between each other. So "Tabs" page would mention and link to "Tab buttons" and vice versa.

sowasred2012:
@long-chung would you mind giving this a UX review in @danielmutis's absence?

sowasred2012:
@spencerbygraves could you give this a design review please?

ziheliu214:
Works well from UX 👍

long-chung:
- [ ] The divider lines overlap the border which looks a bit weird. Not a blocker but would be nice to fix this so its sits inside the wrapper rather than outside.

- [ ] Mobile behaviour is supposed to be a dropdown if we can't fit the content. In the demo everything is truncated and unreadable. (Granted there should be better guidance on label lengths/descriptors)

sowasred2012:
>     * [ ]  Mobile behaviour is supposed to be a dropdown if we can't fit the content. In the demo everything is truncated and unreadable. (Granted there should be better guidance on label lengths/descriptors)

Thanks @long-chung - could you take another look? I've addressed the border problem, but for the above point, we can't make it into a dropdown with CSS, so it would have to be another component that is revealed hidden/shown at smaller breakpoints, which is outside the scope of a single component example. That said, I've updated the docs with some guidance around using another component (like the select) at smaller breakpoints: https://vanilla-framework-3921.demos.haus/docs/patterns/tabs#tab-buttons

PR #3934, opened 2021-08-17
Remove some of the deprecated patterns

bartaz:
## Done

This PR removes some of the deprecated patterns, related examples and SCSS code. This includes:

- `p-table-expanding` - which was renamed to `--expanding`
- `p-table--sortable` - which is redundant and uses only `aria-sort` now
- notification class names - which were renamed when new notification component was implemented
- deprecated accordion markup - which was updated to new HTML structure and 'title' related class names are not used anymore
- code numbered and code copyable - which were replaced by code snippet

Part of #3744 

## QA

- Open [demo](https://vanilla-framework-3934.demos.haus)
  - Check if documentation and examples of all affected components works as expected
  - There should be no mention of deprecation of these components in the docs
  - There should be no deprecated examples
  - Check the code to see if deprecated styles were removed


sowasred2012 (approved):
LGTM!

PR #3943, opened 2021-08-20
Remove deprecated styles for headings, buttons and grid

bartaz:
## Done

Remove deprecated styles related to:

- headings (`--one`, `--two`, etc class names), updated examples that still used those
- neutral button was removed, updated any examples that still used them
- removed is-active button
- removed unnecessary button icon mixin
- removed deprecated grid settings

Part of #3744 

## QA

- Check the code to see which examples were affected
- Make sure affected examples render as expected
- Make sure relevant doc pages don't include deprecated docs anymore


sowasred2012 (approved):
LGTM! Out of interest, I ran the parker checks, filesize is already down just below 300kb, we could adjust the limit back down?

PR #3953, opened 2021-08-25
Move card padding to `p-card__inner` 

### LINKED ISSUE #971: ARIA navigation update

sowasred2012:
## Done

- Removed padding from `p-card` elements, and added it to a new `p-card__inner` element
- Created "Content Bleed" card example
- Updated card markup across the site, including examples
  - Refactored icon docs to make it easier to do the above for 100s of icons

Fixes #971

## QA

- Open [demo](https://vanilla-framework-3953.demos.haus/docs/examples/patterns/card/default)
- See that the pattern looks the same as on live (https://vanillaframework.io/docs/examples/patterns/card/default)
- Open [content bleed example](https://vanilla-framework-3953.demos.haus/docs/examples/patterns/card/content-bleed)
- See that it looks good
  - https://vanilla-framework-3953.demos.haus/docs/upgrade-guide-v3
  - https://vanilla-framework-3953.demos.haus/docs/patterns/card

- [x] Any changes to component class names (new patterns, variants, removed or added features) should be listed on the [what's new page](https://github.com/canonical-web-and-design/vanilla-framework/blob/master/templates/docs/whats-new.md).

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -82,3 +82,5 @@ dropdown
 dropdowns
 playbook
 timestamp
+div
+u-no-padding--top
--- a/templates/docs/examples/templates/typographic-spacing.html
+++ b/templates/docs/examples/templates/typographic-spacing.html
@@ -726,433 +726,6 @@
   </div>
 </div>
 <hr>
-<!--
-<div class="p-strip is-shallow">
-  <div class="row p-divider">
-    <div class="col-6 p-divider__block">
-      <h1>h1</h1>
-      <div class="col-1 u-no-margin--left">
-        <h1>h1</h1>
-        <h1>h1</h1>
-      </div>
-      <div class="col-1">
-        <h1>h1</h1>
-        <h2>h2</h2>
-      </div>
-      <div class="col-1">
-        <h1>h1</h1>
-        <h3>h3</h3>
-      </div>
-      <div class="col-1">
-        <h1>h1</h1>
-        <h4>h4</h4>
-      </div>
-      <div class="col-1">
-        <h1>h1</h1>
-        <h5>h5</h5>
-      </div>
-      <div class="col-1">
-        <div class="col-6">
-          <h1>h1</h1>
-          <h6>h6</h6>
-        </div>
-        <div class="col-6">
-          <h1>&nbsp;</h1>
-          <p>p</p>
-        </div>
-      </div>
-    </div>
-    <div class="col-6 p-divider__block">
-      <p class="p-heading--1">p-heading--1</p>
-      <div class="col-1 u-no-margin--left">
-        <p class="p-heading--1">h1</p>
-        <p class="p-heading--1">h1</p>
-      </div>
-      <div class="col-1">
-        <p class="p-heading--1">h1</p>
-        <p class="p-heading--2">h2</p>
-      </div>
-      <div class="col-1">
-        <p class="p-heading--1">h1</p>
-        <p class="p-heading--3">h3</p>
-      </div>
-      <div class="col-1">
-        <p class="p-heading--1">h1</p>
-        <p class="p-heading--4">h4</p>
-      </div>
-      <div class="col-1">
-        <p class="p-heading--1">h1</p>
-        <p class="p-heading--5">h5</p>
-      </div>
-      <div class="col-1">
-        <div class="col-6">
-          <p class="p-heading--1">h1</p>
-          <p class="p-heading--6">h6</p>
-        </div>
-        <div class="col-6">
-          <p class="p-heading--1">&nbsp;</p>
-          <p>p</p>
-        </div>
-      </div>
-    </div>
-  </div>
-  <div class="row">
-    <hr>
-  </div>
-  <div class="row p-divider">
-    <div class="col-6 p-divider__block">
-      <h2>h2</h2>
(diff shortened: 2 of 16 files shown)


templates/docs/patterns/icons.md
```diff
@@ -7,6 +7,29 @@ context:
 
 # Icons
 
+{% macro icon_section(icon_set, theme) -%}
```
bartaz, line 10: Nice! Thanks for taking time to refactor it.

One thing I'd suggest is moving it above the `#Icons` header to have clearer distinction between content and "meta" definitions.

templates/docs/examples/templates/typographic-spacing.html
```diff
@@ -1223,43 +1237,57 @@ <h6 class="p-card__title">Title</h6>
     <div class="col-2">
       <h1>H1</h1>
       <div class="p-card">
-        <h1 class="p-card__title">Title</h1>
+        <div class="p-card__inner">
```
bartaz, line 1240: If that's commented out I'd suggest dropping it completely. Seems like we don't need these examples. And if we need them we should uncomment them.

bartaz:
Looking at the size of this PR made me question our approach here yet again.

Cards are probably one of most used patterns right now, most of the time they are quite simple: just a card with a bit of content in it, and current markup is perfect for that:

```html
<div class="p-card">
   <-- content -->
</div>
```

To support the flexibility that we want for more complex "modular" cards (that may have different sections, borders, etc), that are currently needed in just a couple places, we are complicating the markup for our most common case. Because something as simple as `<div class="p-card">`, becomes 2 nested arbitrary divs:

```html
<div class="p-card">
  <div class="p-card__inner">
      <!-- content -->
  </div>
</div>
```

With the sole purpose of the `__inner` is to provide the padding. This seems like overkill for simple content cards (that we have the most of).

Sure, 3.0 is a good moment to introduce such breaking changes, but so we want people to need to wrap all their card content in 2 divs just for the sake of it? Just to be able to occasionally have a border from side to side?

I don't know... Maybe better approach would be to introduce separate second "modular" card pattern - but how to we call it, so it's not confused with existing card?

Any thoughts @sowasred2012 @lyubomir-popov @anthonydillon ?

I know we want and need those more flexible "paddingless" cards, but we should try to find a way to make the most common padded card case still working if we can.

lyubomir-popov:
I agree, if we can avoid nesting for the 90% case that would be better so maybe `.p-card--modular`?

anthonydillon:
What about we have a variant in the docs to show how you can achieve this design. For example:
```html
<div class="p-card u-no-padding">
    <img src="..." alt="" />
    <div class="p-card__inner">
        <!-- content -->
    </div>
</div>
```
This would leave the existing "basic" card pattern to remain functional as is?

sowasred2012:
I see your point @bartaz, but if we were introducing the cards pattern for the first time with the knowledge/use cases we have now, I think this would be seen as a fairly simple way to achieve the goal of allowing content to bleed to the borders AND have padded content where desirable.

That's my worry with introducing a variant - e.g. `p-card--modular` (though I don't think that should be the name, it's no more modular than the standard card) - that we're doing it to save a lot of legwork when 3.0 lands, despite that release being where we're introducing a number of breaking changes.

I'll admit to having a certain amount of bias towards the approach currently in this PR given the amount of back and forth we've had on this already (the negative margins approach etc) - maybe there's another way? We could have `p-card__inner` (or a more suitable name), and use `u-no-padding` on parent `p-card` elements where content bleeding is needed?

```html
<div class="p-card u-no-padding">
  <img src="my-full-bleed-image.png" />
  <div class="p-card__inner">
    <p>Some text</p>
  </div>
</div>
</div>
```

templates/docs/patterns/icons.md
sowasred2012: good shout - will do

sowasred2012:
@bartaz this is ready for review

templates/docs/patterns/icons.md
bartaz: Please do ;)

templates/docs/patterns/card.md
```diff
@@ -18,6 +18,14 @@ The purpose of the default card is to display information, without user interact
 View example of the default card pattern
 </a></div>
 
+## Content Bleed
```
bartaz, line 21: I think we don't do title case in docs, this feels like a name of something

```suggestion
## Content bleed
```

templates/docs/patterns/icons.md
sowasred2012: Rargh, rebase/cherry picking snafu...

.spelling
```diff
@@ -81,4 +81,6 @@ specs
 dropdown
 dropdowns
 playbook
-timestamp
\ No newline at end of file
+timestamp
+div
```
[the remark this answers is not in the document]
sowasred2012: Yep, they're used by the icon macro

bartaz (approved):
LGTM

PR #4028, opened 2021-10-12
Update renovate config to group non-major dependencies monthly

bartaz:
## Done

Update renovate config  to group non-major dependencies monthly

Driveby: rename old 'master' branch references to 'main'

## QA

- Review updated config file

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -3,7 +3,7 @@ name: Release Drafter
 on:
   push:
     branches:
-      - master
+      - main
 
 jobs:
   update_release_draft:


bartaz:
@hatched Any thoughts on managing dependency updates? How do you do it for JAAS? You think if vanilla/react-components do updates on less regular pace it may affect downstream projects?

huwshimi:
> @huwshimi Oh, I thought your comment is on react-components repo, not Vanilla framework. Do you think dependencies of Vanilla framework would affect other projects?

Oh sorry, I clicked on both PRs but commented on the wrong one, I don't know about Vanilla itself, the only time we've run into the issue has been with the mis-matched sass builder (which is only a very minor annoyance of having lots of warnings).

> How are you dealing with the dependencies in MAAS/JAAS?

Renovate runs weekly on MAAS and JAAS, but we could probably drop this to monthly.

> I can understand that for React projects maybe we need a more frequent schedule, are there any specific dependencies that we want updated frequently? Types? React itself?

At the very least I think we'd need to update everything in `dependencies` and `peerDependencies` as well as typescript, babel and node-sass. Those are the dependencies we've had trouble with in the past.

Because projects include vanilla-framework from node_modules I'd also recommend keeping `dependencies` up to date for the vanilla-framework repo as well.

> It would be nice to have some shared renovate config for React projects.

We have a task to look into that, but haven't scheduled anything:  https://github.com/canonical-web-and-design/app-squad/issues/203

PR #4110, opened 2021-11-01
Inline labels

### LINKED ISSUE #4104: Investigate adding inline form labels in 3.0
Labels in Vanilla are block. With removal of base checkboxes this causes a problem that native browser checkboxes can't easily be used with Vanilla labels.

We could explore making Vanilla labels inline (or inline-block) by default to make it easier, but should verify what impact it will have on forms.

Initial exploration PR: https://github.com/canonical-web-and-design/vanilla-framework/pull/4073

sowasred2012:
## Done

- Set display of labels to `inline-block`, rather than `block`.
- Added an example with base tick elements

Fixes #4104 

## QA

- Visit https://vanilla-framework-4110.demos.haus/docs/examples/base/forms/tick-elements
- See that the elements look good
- Check Percy, see that other examples are not affected

### DIFF
--- a/templates/docs/examples/patterns/switch.html
+++ b/templates/docs/examples/patterns/switch.html
@@ -9,16 +9,19 @@
     <input type="checkbox" class="p-switch" id="checkbox1" checked/>
     <span class="p-switch__slider"></span>
 </label>
+<br />
 <label for="checkbox2">
     Off
     <input type="checkbox" class="p-switch"  id="checkbox2"/>
     <span class="p-switch__slider"></span>
 </label>
+<br />
 <label for="checkbox3">
     On disabled
     <input type="checkbox" class="p-switch"  id="checkbox3" checked disabled/>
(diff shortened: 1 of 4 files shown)


templates/docs/examples/patterns/switch.html
```diff
@@ -9,16 +9,19 @@
     <input type="checkbox" class="p-switch" id="checkbox1" checked/>
     <span class="p-switch__slider"></span>
 </label>
+<br />
```
bartaz, line 12: Probably not for this PR, but seems this may be a breaking change for some if they relied on switch being block.

But also, seems like switches are these weirdly structured components - root label element doesn't have class name, main component class name is on input but there is also separate slider element outside of it...
Maybe we should refactor it as part of 3.0 as well while we are at it.

bartaz (approved):
LGTM

Issue #4159, opened 2021-11-29
Expose $text-max-width as a class 


lyubomir-popov:
Somethign we used to have but under a confusing high-brow name (.measure--p). It would be very useful, for example to make a list that contains text only as wide as a paragraph. (Lists don't have a max-width as anything could be implemented as a list, there's no way of knowing if it is going to contain text only at framework level). My proposal: `.u-text-max-width`.

anthonydillon:
What is the use case for this?

bartaz:
I'm also not sure what it is solving - if we want lists (`<ul>`, `<ol>`) to have max width same as paragraphs shouldn't they have it by default? To make sure they are consistent and work whenever they are used (for example in docs from discourse, where class can't be added).

Sure, we use lists a lot in components (for navigation, etc), but in component we can easily remove the max-width, because we control the styling via class name.

lyubomir-popov:
The use case is to be able to make anything match the size of text. e.g. Images, code snippets, lists. 

@bartaz If we did that navs and all sorts of other things that use the ul markup would get max-width. This was how I originally did it, but we had to revert.

lyubomir-popov:
Thinking some more about it, the framework is in a much better state now than it was at v 1.6 - so maybe we can try setting it as default to `ul`s and `ol`s again? 

Regardless I think it is still very usefull to be able to apply a max-width to other elements like code snippets or images - not by default, but when those elements are single instances and appear to depart from the overall layout for no good reason. We currently use the grid for that, e.g. wrapping everything in a `.col-8`, but that causes unnecessary reflow when resizing, or too narrow a width when on a smaller resolution that is still above the desktop threshold.

bartaz:
Yes, it would be good if we could try adding it by default and remove in components that use lists.

As for the usefulness of the utility - sure, it seems we can find use cases for it, but it seems that it would be most useful in places where we can't add class name - discourse docs.

For example: https://snapcraft.io/docs - should the tables have limited max-width here? Should the notification on the bottom?
Code blocks here? https://snapcraft.io/docs/channels - they are not code snippets, just `<pre>` blocks - should we limit those by default?

I'm worried about making these decisions now, when we want to release 3.0 really soon, and these seem like quite significant  breaking changes that we will have to QA extensively.

lyubomir-popov:
Well we did have it before under a different name - `.measure--p`. So technically it's just giving it a more user friendly name.

PR #4160, opened 2021-11-29
Add text max width util

bethcollins92:
## Done

- Added a new utility `u-text-max-width` which uses the '$text-max-width' value. This can be applied to list elements composed of text. 

#4159 

## QA

- Inspect the list in the `Text element max-width` section [here](), Add the new `u-text-max-width` class to the `<ul>` element and check the text max with changes to match the rest of the text. 
  - Not sure if I need to add this to the docs? I can update the upgrade guide in the other [PR](https://github.com/canonical-web-and-design/vanilla-framework/pull/4119)

### DIFF
--- a/templates/docs/examples/utilities/text-max-width.html
+++ b/templates/docs/examples/utilities/text-max-width.html
@@ -3,10 +3,14 @@
 
 {% block content %}
 
+<p>This is an example paragraph to show where paragraphs wrap because of the max-width applied to them.</p>
+
+<ul class="p-list">
+  <li class="p-list__item">Here is an example of an li element without the .u-text-max-width class applied to the ul, where the text doesn't wrap.</li>
+</ul>
+
 <ul class="p-list u-text-max-width">
-  <li class="p-list__item">List item 1</li>
-  <li class="p-list__item">List item 2</li>
-  <li class="p-list__item">This is a sentence long enough to show the <code>u-text-max-width</code> class adding a max width to text within a list element</li>
+  <li class="p-list__item">Here is an example of an li element with the .u-text-max-width class applied to the ul, where the text wraps.</li>
 </ul>
 
 {% endblock content %}
--- a/templates/docs/examples/utilities/text-max-width.html
+++ b/templates/docs/examples/utilities/text-max-width.html
@@ -1,0 +1,12 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Text max width{% endblock %}
+
+{% block content %}
+
+<ul class="p-list u-text-max-width">
+  <li class="p-list__item">List item 1</li>
+  <li class="p-list__item">List item 2</li>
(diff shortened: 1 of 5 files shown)


lyubomir-popov (changes requested):
Can you please add an example (using a `ul` maybe? - it's one of those things that obviously needs a max width).

bethcollins92:
Thanks @lyubomir-popov, example added

templates/docs/examples/utilities/text-max-width.html
```diff
@@ -0,0 +1,12 @@
...
+
+{% block content %}
+
+<ul class="p-list u-text-max-width">
+  <li class="p-list__item">List item 1</li>
+  <li class="p-list__item">List item 2</li>
+  <li class="p-list__item">This is a sentence long enough to show the <code>u-text-max-width</code> class adding a max width to text within a list element</li>
+</ul>
+
+{% endblock content %}
```
lyubomir-popov, line 12: We'll also need this included in a section under "Utilities" in the side navigation of the docs. To make it obvious, I would also add a paragraph to this example and a before/after comparison.
The text could say something like this: 

"
By default, unordered/ordered lists do not have a max-width, as they may be used for various purpouses (e.g. navigation). When they contain text only, we recommend adding the `.u-text-max-width` utility to the `ul/ol` element to ensure the list's width matches that of a paragraph:

This is an example paragraph to show where paragraphs wrap because of the max-width applied to them.
{List example here without the class}
{List example here with the class}
" 
Or something along those lines.

lyubomir-popov (changes requested):

lyubomir-popov:
Thinking some more about it, the framework is in a much better state now than it was at v 1.6 - so maybe we can try setting it as dsefault to `ul`s and `ol`s again?

templates/docs/examples/utilities/text-max-width.html
bethcollins92: Not at all! Thanks for the help. Just still getting used to everything I need to add. Added it to the docs now, let me know if I've missed anything else

bethcollins92:
> Thinking some more about it, the framework is in a much better state now than it was at v 1.6 - so maybe we can try setting it as default to `ul`s and `ol`s again?
> 
> Regardless I think it is still very usefull to be able to apply a max-width to other elements like code snippets or images - not by default, but when those elements are single instances and appear to depart from the overall layout for no good reason. We currently use the grid for that, e.g. wrapping everything in a `.col-8`, but that causes unnecessary reflow when resizing, or too narrow a width when on a smaller resolution that is still above the desktop threshold.

Issue created here - https://github.com/canonical-web-and-design/vanilla-framework/issues/4161

bethcollins92:
Closing this PR as discussed in sync with Ant that we shouldn't be using a utility for something which should work out of the box. It could be solved by applying max-width to ul/ol, but not sure what this will break. We need to have a discussion after 3.0 release.

PR #4173, opened 2021-12-02
Prevent renovate from upgrading python in CircleCi image

bartaz:
## Done

Prevent renovate from upgrading python in CircleCi image, which fails the build (as in #4165)


sowasred2012 (approved):
lgtm2

PR #4181, opened 2021-12-06
Attempt to force node LTS version in renovate

bartaz:
## Done

Attempt to freeze node version in renovate and only allow LTS releases

## QA

- Just a code review, we need to merge it to verify if renovate picks it up


bethcollins92 (approved):
code LGTM

PR #4188, opened 2021-12-08
Update the docs and examples for chips

### LINKED ISSUE #4187: Remove non-interactive chips
As part of chips work we decided not to use chips for non-interactive elements and keep label component for this purpose. We need to make sure we remove any examples and documentation around non-interactive chips.

The only chips should be interactive `<button>` or a chip with dismiss button in it.

bartaz:
## Done

Remove non-interactive chips from docs and examples.
Make sure all chips are using `<button>` if they are meant to be interacted with, and `<span>` is only used when chip contains a dismiss button.

Fixes #4187 
## QA

- Open [demo](https://vanilla-framework-4188.demos.haus/docs/patterns/chip)
- Make sure documentation and examples show "button" chips by default and only chips with dismiss button use span
- Check ALL of the chips examples to make sure they are correct


templates/docs/upgrade-guide-v3.md
```diff
@@ -171,9 +171,9 @@ The `aria-label` attribute on table cells of the responsive table has been repla
 
 ## Chips
 
-By default, chips are not interactive anymore (they don't have hover, focus or active state). Apply the class to a button element to get interactive behaviour, e.g.: `<button class="p-chip">` instead of `<span class="p-chip">`.
+As chips are meant to be interactive by default they are now using `<button>` element. Any chips that don't have dismiss button should be built using `<button class="p-chip">` instead of `<span class="p-chip">`. Only the chips that have the dismiss button are still using span as their main element.
```
bethcollins92, line 174:
```suggestion
As chips are meant to be interactive by default, they now use the `<button>` element. Any chips that don't have a dismiss button should be built using `<button class="p-chip">` instead of `<span class="p-chip">`. Only chips that have the dismiss button still use span as their main element.
```

templates/docs/upgrade-guide-v3.md
```diff
@@ -171,9 +171,9 @@ The `aria-label` attribute on table cells of the responsive table has been repla
...
 
-The dismiss button in the chip now provides its own icon, so it should not include separate icon element. Please remove any icons from the chips, and only keep the `Dismiss` text in the button.
+Also the dismiss button in the chips has been updated. Now provides its own icon, so it should not include separate icon element. Please remove any icons from the chips, and only keep the `Dismiss` text in the button.
```
bethcollins92, line 176:
```suggestion
Also, the dismiss button in chips has been updated. It now provides its own icon, so it should not include a separate icon element. Please remove any icons from the chips, and keep only the `Dismiss` text in the button.
```

bethcollins92 (approved):
What a lot of buttons to check 😅. LGTM with a couple of rewording suggestions in the docs

elioqoshi:
@bartaz Not sure how Lyubo prepared it in the design specs, though it would be best if the hover state of the dismiss icon would be round, and not a square. Similar to how Material does it: https://m3.material.io/components/chips/specs

grafik

elioqoshi:
As @bartaz addressed this in https://github.com/canonical-web-and-design/vanilla-framework/issues/4189 this is good to go, thanks!

PR #4236, opened 2022-01-11
Add buttons accessibility notes

### LINKED ISSUE #4039: [Accessibility Docs] Buttons
Using [this template](https://docs.google.com/document/d/1IwFMfJrTfJPXDcVJfLWXYkw-pjgWhBWo4GX8X098Q64/edit#), create accessibility docs for [Buttons](https://vanillframework.io/docs/patterns/buttons)

bethcollins92:
## Done

- Add buttons accessibility notes. copy doc [here](https://docs.google.com/document/d/11H6xPtBRxB1a8KCmK7XFH1FeZDduQhSJXxq-wNJ_CZ8/edit#heading=h.qxwy2zql3yy7)

Fixes #4039 

## QA

- Check button accessibility section


templates/docs/patterns/buttons.md
```diff
@@ -123,6 +123,32 @@ In some contexts, it may be necessary to indicate to the user that a button is i
 View example of the contextual menu pattern
 </a></div>
 
+## Accessibility
+
+{# copy doc: https://docs.google.com/document/d/11H6xPtBRxB1a8KCmK7XFH1FeZDduQhSJXxq-wNJ_CZ8/edit#heading=h.qxwy2zql3yy7 #}
```
bartaz, line 128: Overall I'd prefer to have one source of truth, so once we merge it lets keep all the edits here. So I'm not sure if we need to link to the doc.

templates/docs/patterns/buttons.md
```diff
@@ -123,6 +123,32 @@ In some contexts, it may be necessary to indicate to the user that a button is i
...
+The button component is used to trigger an action or event - this could be opening or closing a modal, navigating to the next page, or cancelling an action. Rather than adding `role=button` to links, it’s always advisable to use the native HTML button element, as native HTML buttons provide keyboard and focus requirements by default and are best supported by assistive technologies.
+
+### Considerations
+
+This component strives to follow [WCAG 2.1 (level AA) guidelines](https://www.w3.org/TR/WCAG21/), and care must be taken to ensure this effort is maintained when the component is implemented across other projects. This section offers advice to that effect:
+
+- A button, once focussed should be activated by using the `Space` or `Enter` keys.
+- The button should have an accurate description. This can be as text in the button, or by using `aria-label` or `aria-labelledby`.
+- If there is a description of the button, the button should have an `aria-describedby` which matches the ID of the description.
+- If the action of the button is not available, add `aria-disabled=true`
```
bartaz, line 141: Isn't `disabled` on a button enough?

bethcollins92: I remembered reading something a while back about how using disabled isn't ideal, as with this attribute the VoiceOver just completely ignores the button and if the user was blind they wouldn't even know it was there. I've just noticed we use `disabled` on our examples.

here is a resource explaining: https://a11y-101.com/development/aria-disabled

I could file an issue on Vanilla to change the buttons from using `disabled` to `aria-disabled="true"` and change them before I merge this?

bartaz: Interesting. The issue is - we don't use `aria-disabled` anywhere currently and we don't support it in Vanilla, because we thought that recommended way is to use native `disabled`. And we probably have many places that do use `disabled` that won't update to `aria-disabled`.

Also, it seems that browsers dont treat `aria-disabled` in any special way, so the button is not disabled - you can still click it. Which is the main point of having it `disabled`. Otherwise we would have to explicitly detect `aria-disabled` in click handler and ignore it. Which is a really a big change.

Long story short, for now I'd suggest:
- remove this from the docs, because it's now how we do it
- bring this up in Accessibility Guild to discuss and verify what is the best way to do it both for screenreaders, but also to have the button actually disabled
- once we know what we want to do, have an issue in Vanilla to update our examples/docs.

templates/docs/patterns/buttons.md
bartaz: @bethcollins92 Did you remove the link from the others?

bethcollins92: Ahh sorry I thought I'd done it from them all! Doing it now

bartaz (approved):
LGTM

PR #4260, opened 2022-01-19
Add role=switch to switch component

bethcollins92:
## Done

- Add `role=switch` to switch component. Without this the screen reader describes this as a checkbox being checked and unchecked. 
- We mention using aria-checked in the description but don't include it in the example. Should we include it as default? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/switch_role

## QA

- Open [demo](https://vanilla-framework-4260.demos.haus/docs/patterns/switch)
- Use VoiceOver and check the switch component. It should read out the label, then the status, then switch. It sounds a little strange with our example but that's because our labels are on and off.

### DIFF
--- a/templates/docs/examples/patterns/switch.html
+++ b/templates/docs/examples/patterns/switch.html
@@ -5,22 +5,22 @@
 
 {% block content %}
 <label class="p-switch">
-    <input type="checkbox" class="p-switch__input" checked/>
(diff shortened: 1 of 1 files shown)


petermakowski:
I think you should include `aria-checked` in the example as it's required for the aria switch role.

petermakowski (approved):
Definitely an improvement no matter if you add an `aria-checked` or not 👍

PR #4267, opened 2022-01-21
Added min-height to close button

mtruj013:
## Done

- Added min height to `.p-notification__close` to fix positioning issue when missing button text

- Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/4034

## QA

- Open [demo](https://vanilla-framework-4267.demos.haus)
- Go to patterns > notification > actions > buttons
- Edit on codepen or checkout current branch
- Remove close button text and ensure positioning is unchanged 

### DIFF
--- a/scss/_patterns_notifications.scss
+++ b/scss/_patterns_notifications.scss
@@ -101,6 +101,7 @@ $notification-text-margin-bottom: $spv--large - $spv-nudge;
       background-repeat: no-repeat;
       background-size: unset;
       border: 0;
+      min-height: $spv--large	* 2;
       position: absolute;
       right: $sph--large * 0.5;
       top: $spv--small;
--- a/scss/_patterns_notifications.scss
+++ b/scss/_patterns_notifications.scss
@@ -101,7 +101,8 @@ $notification-text-margin-bottom: $spv--large - $spv-nudge;
       background-repeat: no-repeat;
       background-size: unset;
       border: 0;
-      min-height: $spv--large * 2;
+      // set the height of the button to be size of an icon + padding on top and bottom
+      height: calc(2 * $spv--small + $default-icon-size);
       position: absolute;
       right: $sph--large * 0.5;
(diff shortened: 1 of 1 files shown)


scss/_patterns_notifications.scss
```diff
@@ -101,6 +101,7 @@ $notification-text-margin-bottom: $spv--large - $spv-nudge;
       background-repeat: no-repeat;
       background-size: unset;
       border: 0;
+      min-height: $spv--large * 2;
```
bartaz, line 104: @mtruj013 That kinda works, but we don't want to hardcode "random" values like that.

The height that we need is based on icon size plus button padding + button border.

Icon size you have as `$default-icon-size`; The button padding/border thing is bit harder to dig out, but it seems to be `$sp-nudge` value (on top and bottom), so I think what we need is:

```suggestion
      // set the height of the button to be size of an icon + padding on top and bottom
      height: calc(2 * $sp-nudge + $default-icon-size);
```

Also use `height` we don't want this button to change height based on content, so we don't need min-height.
And use `calc` to make sure the math works even if units inside the variables don't match.

bartaz: BTW, I'm not 100% sure this code will work, it's just a suggestion how I think it could look like.

bethcollins92:
LGTM code wise. Just want to double check something with @lyubomir-popov 

Although the calculation makes sense (2 x padding + icon size), the `x` icon is alignment is very slightly different on live. See screen shots below. 

Live: 

Screenshot 2022-01-24 at 10 27 17

Demo:

Screenshot 2022-01-24 at 10 27 04

lyubomir-popov:
@bethcollins92 @mtruj013 it's not a problem, the icon buttons aren't perfectly square, hence the small difference. 
Making the button square as you've done here is an improvement.

bethcollins92 (approved):
LGTM - yay

PR #4273, opened 2022-01-24
WIP - Add native form validation

mtruj013:
## Done

- Added styling for native input validation

- Fixes https://github.com/canonical-web-and-design/vanilla-framework/issues/3262

## QA

- Check that invalid entry is correctly flagged 

- [ ] Any changes to component class names (new patterns, variants, removed or added features) should be listed on the [what's new page](https://github.com/canonical-web-and-design/vanilla-framework/blob/main/templates/docs/whats-new.md).

## Screenshots

### DIFF
--- a/scss/_patterns_form-validation.scss
+++ b/scss/_patterns_form-validation.scss
@@ -41,22 +41,36 @@
     }
   }
 
-  .p-form-validation, .is-error {
-    //used to be p-form-validation__input
-    input:invalid{
+  .is-error {
+    .p-form-validation__input {
       border-color: $color-negative;
     }
 
-    .p-form-validation__message{
+    .p-form-validation__message {
       @include vf-icon-error($color-negative);
       @extend %vf-validation-wrapper;
 
       color: $color-negative;
     }
   }
-  
-  input:valid + p {
+
+  // Native browser form validation styles
+
+  input:invalid,
+  input:not(:placeholder-shown):invalid {
+    border-color: $color-negative;
+  }
+
+  input:invalid + p {
+    @extend .p-form-validation__message;
+    @include vf-icon-error($color-negative);
+    @extend %vf-validation-wrapper;
+
+    color: $color-negative;
+  }
+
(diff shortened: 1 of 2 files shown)


sowasred2012:
@mtruj013 @bethcollins92 I think you're right. When building the first version of the /advantage form, we tried using native validation, and the feedback from UX was immediate - it's too aggressive, we should at least wait for the user to move focus somewhere else, which requires javascript.

There is a psuedo-selector that seems to be on the way that would help, named [`:user-invalid`](https://drafts.csswg.org/selectors-4/#user-pseudos). Browser support is terrible for it at the moment (just [Firefox so far](https://caniuse.com/mdn-css_selectors_user-invalid)), but if you play around with it in FF, you should see that it's way less aggressive. It's not a solution though, not yet. Have we talked about handling it in JS? I think individual projects do, but I imagine they'll all be quite different.

The other thing to mention is the validation error message doesn't disappear in the example when the input is valid, some of the styling is removed but the message remains - is it possible to hide it with CSS?

bethcollins92:
Thanks for your feedback @sowasred2012.

I also read the articles on `:user-invalid`, it's a shame it isn't something we can implement just yet. There was no mention of using JS when we went through the issue with Bartek, so I'm not sure what his thought's are / if anyone would use it? As we have the validation class names already. 

I just double checked the example on Firefox and realised the error message doesn't disappear, however it does on chrome: 

Screenshot 2022-01-26 at 08 40 04

We could leave the PR open for now to check with Bartek what he thinks about adding some JS, but I'm leaning towards maybe not implementing it for now.

bethcollins92:
A meeting has been scheduled with Elio to discuss the UX of this

bethcollins92:
We're closing this issue for now. It's out of our control when the browser marks a field as invalid, so it's difficult to add the styles with the correct UX without a lot of hacky error prone code. When the pseudo selector`:user-invalid` is compatible with all browsers and reliable it could be a solution,  but for now we can't use this.

PR #4274, opened 2022-01-25
Added empty state - 1st variant

ilayda21:
## Done

-  Variant 1 empty state is added.
- Empty state is added to docs.

Fixes #1010

## QA

- Open empty state [example](https://vanilla-framework-4274.demos.haus/docs/examples/layouts/empty-state/no-content)
- Verify its implemented according to [design](https://discourse.ubuntu.com/t/variant-a-no-content-empty-state/24619)
- Open [docs](https://vanilla-framework-4274.demos.haus/docs/base/empty-state)
- Verify empty state's no content variant is documented correctly.
- Open [whats new](https://vanilla-framework-4274.demos.haus/docs/whats-new) section.
- Verify "empty state" is added.


templates/docs/examples/base/empty-state/no-content-page-view.html
```diff
@@ -0,0 +1,26 @@
...
+        <th>Vendor ID</th>
+        <th>Product</th>
+        <th>ProductID</th>
+        <th>Driver</th>
+        <th>Huma Node</th>
+        <th>PCI Address</th>
+      </tr>
+    </thead>
+    <caption>
+        <div class="p-strip">
```
bethcollins92, line 19: I think the formatting is a bit off in the HTML, there should be only 1 indentation

bethcollins92 (commented):
The examples look great! I've spoken to Scott and we both aren't sure of the placement in the base section. I'm sure Bartek has a reason for this but we will wait until he's back to check his reasoning. Nice work :)

templates/docs/examples/base/empty-state/no-content-page-view.html
ilayda21: Fixed :+1:

lyubomir-popov:
Hi, 2 comments from me:

1) Could you please add `u-no-margin--bottom` to all `h4` headings, and change the small text to regular paragraphs please?
Should look like this:

Before:
image
 
After:
image

2) The [documentation ](https://discourse.ubuntu.com/t/variant-a-no-content-empty-state/24619) page shows one left-aligned example (as per the brief and benchmarking, and consistent with everything else we do), and 2 central alignment examples (not consistent with anything else we do, but strongly favoured in the original design). Unless there's a section of the documentation explaining when to use which, I'd suggest dropping the central alignment examples. cc @spencerbygraves

spencerbygraves:
I'd agree it seems unnecessary to have the centred version.

bethcollins92 (commented):
Looking great! - as we discussed if you could move the example into the layout section, and I'll double check with @bartaz on Monday this is an appropriate place for it. Thanks!

bethcollins92:
@lyubomir-popov could you take another look please and design +1 if you're happy

bethcollins92:
LGTM but let's wait for Bartek to check it's ok in the layout section

bartaz:
Thanks for implementing this. I tried to clean up this PR from unnecessary duplicated commits from `main` branch, hopefully I didn't mess up something.

As for the placement of the Empty State page - we don't seem to have a good place for it at the moment. "Empty state" is a design pattern, built out of Vanilla components. It's definitely not a layout, it's also not a base element, and technically not a component (but probably closest to be one). So I guess for now we either put it in "Components" or create a new section "Patterns" for it (but this means we only have Empty State there for a while).

templates/docs/examples/layouts/empty-state/no-content.html
```diff
@@ -0,0 +1,33 @@
...
+        <tr>
+            <th>Vendor ID</th>
+            <th>Product</th>
+            <th>ProductID</th>
+            <th>Driver</th>
+            <th>Huma Node</th>
+            <th>PCI Address</th>
+        </tr>
+    </thead>
+    <caption>
```
bartaz, line 18: @bethcollins92 Do you know if that's fine from accessibility point of view or for screen readers? Seems a lot to put into table 'caption', but maybe it's fine. It is a valid HTML and seems OK from semantics point of view I guess.

bethcollins92: I checked it with voice over and it seems to only read the caption and not the column headers, which is what we want/makes sense as there is no point reading them if there is no content. I can't see anything that would make this incorrect semantics wise either

bartaz (approved):
I moved the empty state page to components section.

LGTM now, thanks!

PR #4302, opened 2022-02-09
Update list examples to use list elements

### LINKED ISSUE #4225: List example doesn't use list HTML elements
**Describe the bug**

This example of a list https://vanillaframework.io/docs/patterns/lists#responsive-divider doesn't use the correct `<ul>` and `<li>` elements like the rest of the list examples. It should either use the elements or last resort `role=list` and `role=list-item`

bethcollins92:
## Done

- Ensure all list examples use list elements

Fixes #4225 

## QA

- Open [demo](https://vanilla-framework-4302.demos.haus/)
- Check responsive divider lists still work as expected

### DIFF
--- a/templates/docs/patterns/lists.md
+++ b/templates/docs/patterns/lists.md
@@ -164,25 +164,6 @@ To import list patterns into your project, copy the snippet below and include it
 @include vf-p-lists;
 ```
 
-To add dividers into your project, copy the snippet below and include it in your main Sass file.
-
-```scss
-// import Vanilla and include base mixins
-// this only needs to happen once in a given project
-@import '../vanilla';
-@include vf-base;
-
-@include vf-p-divider;
-
-// grid column classes are used within divider component, so you need to include grid pattern as well
-// import Vanilla and include base mixins
-// this only needs to happen once in a given project
-@import '../vanilla';
-@include vf-base;
-
-@include vf-p-grid;
-```
-
 To include individual list patterns you need to include the `vf-p-list-placeholders` mixin first.
 
 ```scss
--- a/templates/docs/patterns/lists.md
+++ b/templates/docs/patterns/lists.md
@@ -127,6 +127,8 @@ View example of the patterns list split
 
 To separate non-list content with a responsive divider, see our [Divider component](/docs/patterns/divider).
 
+For a more interactive layered lists style [view the list tree component](/docs/patterns/list-tree).
+
 ## Accessibility
 
 ### How it works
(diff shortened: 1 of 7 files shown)


bartaz:
Overall I'm not sure about this change.

Technically I agree that we should implement list components as HTML lists, but this doesn't seem like a list component. The `p-divider` seems more like a layout kind of component not a "list of things". And we don't implement grid as list of rows or columns.

TBH I'm not sure why divider is even part of the "Lists" component page, it's so different from all other lists visually and doesn't even have a "list" in the name, it's just a "Divider". It's not a variant of a list component, it's a separate one.

I guess I'd vote for removing keeping divider as divs, and possibly removing it from "lists" page, but I don't think we have other place for it to live, so it would need to be a separate "Divider" component? (maybe it was before?) - I don't know where is it easier to find it? Would you expect to see it as a list or component on it's own? It seems to me that "Divider" component name by itself also doesn't say much about what it is… So having a "Divider" component page may be as confusing as having it on lists page…

bartaz:
Looks good but couple small things:
- we should update the titles of the examples, they still say "List / Divider" and "List / Divider dark", so it looks weird on the list:

Screenshot 2022-02-16 at 10 54 49

- We should also still mention divider on "Lists" page for reference if anyone may be looking for it there. In new template we have "Related components" section, maybe worth adding it somewhere on the page (after all list variants?)

bethcollins92:
@bartaz Thanks for the review, I've made those changes

templates/docs/patterns/lists.md
```diff
@@ -131,20 +123,9 @@ If you wish to split the items in a list into two columns above `$breakpoint-sma
...
-By default, the responsive divider uses the light theme. To change the global default, set `$theme-default-p-divider` to `dark`.
-
-To change the appearance of an individual instance of the responsive divider pattern, you can use the `is-dark` class.
-
-For more details about themes in Vanilla refer to the [Color theming](/docs/settings/color-settings#color-theming) section of the documentation.
-
-<div class="embedded-example"><a href="/docs/examples/patterns/lists/divider-dark/" class="js-example">
-View example of the divider list with an is-dark class
-</a></div>
+To separate non-list content with a responsive divider, see our [Divider component]('/docs/patterns/divider').
```
bartaz, line 128: Typo in URL:

```suggestion
To separate non-list content with a responsive divider, see our [Divider component](/docs/patterns/divider).
```

bartaz (approved):

PR #4365, opened 2022-03-17
WIP - Generate react docs using typedoc

ClementChaumel:
## Done

Added vanilla react components as a dependency and then ran
```sh
npx typedoc node_modules/@canonical/react-components/dist/index.d.ts
```

it generated a docs site under ./docs

image

It appears to have all the data that we need for building a props table.

You can check it yourself by cloning this branch and then opening `/docs/modules.html` in a browser.
(I've committed the generated doc site for now so you can easily see it but it will of course be removed from the actual PR)

=========================

Next steps are:

- Export this data as a JSON instead
- Maybe add some filtering to only keep relevant info
- Create a script that will populate a page's with a table listing all the props
- Make all of this part of the build process for every component page


bartaz:
Closing this as proof of concept, we will get back to it when properly implementing React docs for new Vanilla site.

PR #4371, opened 2022-03-24
Ensure sidenav is only tabable when in view

### LINKED ISSUE #4134: A11Y: Side navigation is focusable when off-screen
Side navigation (in docs layout) on small screen gets moved off screen and can be toggled back to view.
But even off screen when using tab to move focus, or screen reader, it is accessible and may be confusing.

We should make sure it's either completely hidden (not focusable), or gets into view when focused via keyboard.

bethcollins92:
## Done

- Added functionality using tabindex to ensure the side navigation is only tab-able when it's in view (aria-hidden can't be used on focusable elements)
- Ensure on page load depending on screen size the side is either tab-able or not 
- Update the tab index when the screen size changes and the drawer is hidden or in view
- Update tab index when the toggle side nav button is clicked 
- Added accessibility notes

Fixes #4134 

## QA

- Open [demo](https://vanilla-framework-4371.demos.haus/docs)
- First, load the page on a screen size bigger than 1035 when the side nav is visible by default
- Check you can tab through the links
- Then change the screen size so the side nav is hidden. Check when you tab through the page it ignores the side nav links
- Make the screen bigger again to reveal the side nav and check you can tab through the links
- Make it smaller again, and use the "Toggle side navigation" button to open the side nav drawer. Check you can tab through the links
- Close the side nav with the button, and then make sure the links return to being non tab-able. 
- When opening and closing toggle drawer button using the keyboard, make sure the focus lands on the "toggle side navigation" button inside and outside the drawer


bartaz:
I wonder if we should do it another way around. Keep the side navigation focusable, but make sure it shows up on the screen when it's focused? Not sure which one would be a better accessibility practice.

anthonydillon:
If it helps, I would expect it to be a well labelled (using `aria-controls` and `aria-expanded`) user action to open and enter the side navigation. Then programmaticly focus on the first navigation item.

bethcollins92:
We will discuss in the Accessibility guild to get some more opinions before moving forward

bethcollins92:
I've updated the PR to ensure the focus lands on the CTA/first element in the side nav. I couldn't decide, when the drawer opens, whether the focus should land on the first element (which is the toggle side navigation button) or the "Get started" list item, what do you think @sowasred2012 @bartaz?

bartaz:
Something strange is happening when tabbing.
When focus is on "Toggle button" on the page (the one that opens the nav) and you hit Tab focus "disappears" somewhere offscreen and only the next Tab lands on some link after the "Toggle button".

Interestingly when the focus is "away" hitting Enter opens the navigation. So it seems after tabbing out of "Toggle button" the focus is still on it? Or focus moves to "Toggle button" inside the navigation?

bethcollins92:
@bartaz Ahh yes I forgot about this little bug, I'll need to remove the tab focus from the button as well as all the links

bartaz:
Would be nice to have ESC button close the side nav as well.

templates/docs/examples/patterns/side-navigation/_toggle_script.js
```diff
@@ -6,16 +6,32 @@
...
   }
 }
 
+/**
+  Ensures links are only focusable when side nav is in view
+  @param {HTMLElement} sideNavigation The side navigation element.
+  @param {Boolean} sideNavCollapsed Whether the side nav is expanded or collapsed.
+*/
+function toggleTabindex(sideNavigation, sideNavCollapsed) {
+  const links = sideNavigation.querySelectorAll('.p-side-navigation__link');
```
bartaz, line 31: If we are doing this, I wonder if we can just completely remove the side nav content from the document?
I mean 'detach' it from DOM when it is collapsed (hidden) and attach again when it is expanded. But I guess it may conflict with toggle animation.

templates/docs/examples/patterns/side-navigation/_toggle_script.js
```diff
@@ -6,16 +6,32 @@
 */
 function toggleDrawer(sideNavigation, show) {
   if (sideNavigation) {
+    const outsideDrawer = sideNavigation.querySelector('.p-side-navigation__toggle');
+    const insideDrawer = sideNavigation.querySelector('.p-side-navigation__toggle--in-drawer');
```
bartaz, line 10: While we are using these buttons directly I guess it would be good for them to have `aria-controls` and `aria-expanded` set to correct values.

templates/docs/examples/patterns/side-navigation/_toggle_script.js
bartaz: I guess that's the same problem as if we would set `display: none` in CSS. We don't do that, because we animate the closing of the sidebar, so we can only remove it after the animation is finished, but this is handled by CSS.

templates/docs/examples/patterns/side-navigation/_toggle_script.js
```diff
@@ -24,9 +40,15 @@ function setupSideNavigation(sideNavigation) {
   var toggles = [].slice.call(sideNavigation.querySelectorAll('.js-drawer-toggle'));
 
   toggles.forEach(function (toggle) {
+    var sideNav = document.getElementById(toggle.getAttribute('aria-controls'));
+    var drawerEl = document.querySelector('.p-side-navigation__drawer');
+    var drawerPosition = window.getComputedStyle(drawerEl).position;
+    if (drawerPosition == 'fixed') {
```
bartaz, line 46: That seems quite specific and tied to CSS styles, why are we checking this? Is it to detect the initial collapsed state when the site is first loaded on small screen (with navigation off-screen?)

templates/docs/examples/patterns/side-navigation/_toggle_script.js
bethcollins92: I'm taking a look at this now with JS (both buttons already have the aria-controls attribute). I'm wondering if we could add the `aria-expanded` attributes with HTML, as the `.p-side-navigation__toggle--in-drawer` button will only be visible when the drawer is expanded and the `.p-side-navigation__toggle` will only be visible when the drawer is collapsed

bethcollins92:
I'm a few steps closer to the end! The reason `transitionend` event listener was being flakey and not working on the drawer element is because it's an animation not a transition. I've checked it here and I think it's good https://caniuse.com/?search=animationend ? (don't really know what I'm looking at though). Could you give it a QA @bartaz

bartaz:
@bethcollins92 Not sure why but on Percy some of the side navigation examples seem to load hidden.

I don't know if that's some issue with Percy or our JS, but we should investigate why it may happen. We don't want it happening to people regardless if they have JS on or off.

bethcollins92:
Just jotting down what I've noticed QA'ing incase you haven't seen:

If on smaller screens the side nav drawer is expanded, then the screen size is increased so the drawer is no longer in view, when the screen size is reduced back the side nav button still looks active, but the drawer isn't in view. I feel like this is such an edge case - why on earth would someone be changing their screen size so much? 😅

Screenshot 2022-04-14 at 16 14 10

bartaz:
@bethcollins92 It seems that "Toggle side navigation" button inside the side nav doesn't have hover styles now? It does in current live version

image

bethcollins92:
> https://vanilla-framework-4371.demos.haus

Ahh right, I think it must be because we did some overrides of the aria-expanded styles now we've added that. We'll need to add the hover styles back too

bartaz:
With new class name added and other renamed we should update the docs. "Class reference" and "Javascript Functionaliy" sections need to be updated with new class names. We should add information about deprecated class names to "What's new" page and maybe in "Class reference" as well?

bartaz (approved):
LGTM, lets get this thing merged! 🍾 🎆 🎉

PR #4372, opened 2022-03-24
add javascript module to the vanilla package

goulinkh:
## Done

- Add JS module to the Vanilla framework NPM package (and `/build` for JS CDN)
- Implement the `Accordion` component


bartaz:
@goulinkh Thanks for setting this up, looks promising.

Couple of thoughts, but I know that you may not have time to address those:

- I'm not sure if we need the "Component" class abstraction, it would probably be enough to have a "setup" function for components
- We also want to use this code in the examples in Vanilla docs and get rid of the code used there
- if possible we want to provide the component both in a way they can be imported (via `import` in JS), but also some way that doesn't require building JS, having components available in global scope (similar to `global-nav` that exports both in ESM format, and IIFE format: https://github.com/canonical-web-and-design/global-nav)

But overall great work, we will definitely take some parts of it when we decide to do it.

goulinkh:
@bartaz concerning the last point, this also support multiple types of import:
- Directly into the HTML with the `script` tag:
```html
<script src="JSCDN/vanilla/accordion|index.js"></script>
```
The JS for this will be generated to `/build/js`
- As ESM, with NPM:
```ts
import components from "vanilla-framework"
//or
import {Accordion} from "vanilla-framework"
``` 
The JS for the NPM library will be generated to `/dist`

goulinkh:
For the `component` abstraction, I wasn't sure about it a lot, it could provide better maintainability of the project on the long run but surely is more complex to write a new component and maintain the abstraction.
So I would agree with removing the base component class if you feel its unnecessary!

goulinkh:
> * We also want to use this code in the examples in Vanilla docs and get rid of the code used there

Good idea!, this is a good way to test the component, where they will be used directly in the project's doc :+1:

PR #4414, opened 2022-04-14
Add component menu

albertkol:
## Done

Add `component_tabs.yaml` file that stores the /docs tab menu. It will be the source of truth for when the menu shows up on each docs page and which tab items get displayed (design/accessibility).

Structure of the `component_tabs.yaml`:
```yaml
patterns/accordion:
  accessibility: True
  design_guidelines: False 
```

This PR add the menu only for the accordion docs page, future pages will be added in a separate PR.

## QA

- Open https://vanilla-framework-4414.demos.haus/docs 
- Check the side menu is highlighting the right link
- Go to https://vanilla-framework-4414.demos.haus/docs/accordion
- Check the side menu is highlighting the right link
- Is the tab menu there?
- Go to https://vanilla-framework-4414.demos.haus/docs/patterns/accordion/accessibility
- Check the side menu is highlighting the right link
- Check the tab menu is highlighting the right link

## Screenshots

image


bethcollins92 (approved):
Woooo nice job Albert! I'm not the most qualified to review the python but it LGTM!

bethcollins92:
@bartaz do you want to have a look too?

templates/_layouts/docs.html
```diff
@@ -28,7 +28,7 @@
 
             {% macro side_nav_item(url, title, label, type) -%}
                 <li class="p-side-navigation__item">
-                  <a class="p-side-navigation__link" href="{{ url }}" {% if path == url %}aria-current="page"{% endif %}>
+                  <a class="p-side-navigation__link" href="{{ url }}" {% if (slug and slug in url) or (url == path) %}aria-current="page"{% endif %}>
```
bartaz, line 31: Are both parts of this conditional needed? Would only `(slug and slug in url)` work or does it need the `url == path` for some other cases?

bartaz:
All the pages have duplicated titles:

image

I'm not sure if that's something to be addressed later? Is it just a matter or removing title from markdown and keeping it in meta data only?

templates/_layouts/docs.html
albertkol: Initially I had only `slug in url` then highlight side menu item. It was broken for the home page, where slug is `""` making the condition true for all menu items, so everything was highlighted. 

It's not the prettiest condition.. I agree.

templates/_layouts/docs.html
```diff
@@ -155,7 +155,13 @@
       </aside>
 
       <main class="col-9" id="main-content">
-        {% block content %}{{ content | safe }}{% endblock content %}
+        {% block content %}
+          <h1>{{ title.split('|')[0] | safe }}</h1>
```
bartaz, line 159: I think we can drop `I safe` here, we don't want any HTML code passed from title to the page (there shouldn't be any HTML there anyway), so this should be enough:

```suggestion
          <h1>{{ title.split('|')[0] }}</h1>
```

webapp/app.py
```diff
@@ -146,7 +147,24 @@ def global_template_context():
     version_parts = VANILLA_VERSION.split(".")
     version_minor = f"{version_parts[0]}.{version_parts[1]}"
 
-    return {"version": VANILLA_VERSION, "versionMinor": version_minor, "path": flask.request.path}
+    docs_slug = flask.request.path.replace("/docs/", "")
+    docs_slug = docs_slug.replace("/design/", "")
```
andesol, line 151: I'm sure there's a good reason not to... but why not chaining the replace method instead of reassigning the variable?

albertkol: The max 80 char per line rule.

templates/_layouts/docs.html
albertkol: Pushed a fix for the double titles problem and removed `safe`.

webapp/app.py
andesol: I think something like this would do the trick: (It's not particularly pretty, though... Feel free to leave as it is)
```py
docs_slug = (
    flask.request.path.replace("/docs/", "")
    .replace("/design/", "")
    .replace("/accessibility", "")
)
```

albertkol: Good suggestion! I will try it.

PR #4424, opened 2022-04-20
Align accordion's icon with heading

### LINKED ISSUE #4418: Misaligned accordion icons
Accordion icons are misaligned on smaller screens (below 1033px)

andesol:
## Done

~~- Add a media query to change the alignment of the icon when next to a heading, so we use a different line-height to calculate the icon's position depending on the screen size.~~

Update the styling of the accordion.

Fixes #4418 

## QA

- Open demo
- Go to the example with headings (https://vanilla-framework-4424.demos.haus/docs/examples/patterns/accordion/headings) and check that the icon looks aligned no matter the screen size. (The problem we had was that the icon was misaligned in small screens).

## Screenshots

### DIFF
--- a/scss/_patterns_accordion.scss
+++ b/scss/_patterns_accordion.scss
@@ -33,18 +33,12 @@
     z-index: 2;
 
     &::before {
-      $vertical-offset: 0.5px;
       @include vf-icon-chevron($color-mid-dark);
       @include vf-animation($property: transform, $duration: fast);
+      @extend %icon;
 
-      background-position: center;
       content: '';
-      display: inline-block;
-      font-size: inherit;
-      height: $icon-size;
       margin-right: $sph--large;
-      vertical-align: calc(#{$vertical-offset} + #{0.5 * $cap-height} - #{0.5 * $default-icon-size});
-      width: $icon-size;
     }
 
     // aria-selected controls the open and closed state for the accordion tab
@@ -84,12 +78,8 @@
   }
 
   .p-heading--3 > .p-accordion__tab::before,
-  h3.p-accordion__heading > .p-accordion__tab::before {
-    @include vf-icon-size($default-icon-size);
-    vertical-align: 0;
-  }
-
   .p-heading--4 > .p-accordion__tab::before,
+  h3.p-accordion__heading > .p-accordion__tab::before,
   h4.p-accordion__heading > .p-accordion__tab::before {
     vertical-align: 0;
   }
--- a/scss/_patterns_accordion.scss
+++ b/scss/_patterns_accordion.scss
@@ -1,5 +1,11 @@
 @import 'settings';
 
+@mixin align-icon($icon-size: 1rem, $line-height: default-text) {
+  // calculate icon top position based on top padding (from %single-border-text-vpadding--scaling)
+  // and offset to align it with baseline of accordion tab text element
+  top: $table-cell-vertical-padding + 0.5 * (map-get($line-heights, $line-height) - $icon-size);
+}
+
 @mixin vf-p-accordion {
   $icon-size: map-get($icon-sizes, default);
 
@@ -16,10 +22,6 @@
   }
 
   .p-accordion__tab {
-    // calculate icon top position based on top padding (from %single-border-text-vpadding--scaling)
-    // and offset to align it with baseline of accordion tab text element
-    $icon-top-position: calc(#{$table-cell-vertical-padding + (map-get($line-heights, default-text) - $icon-size) * 0.5});
-
     @extend %single-border-text-vpadding--scaling;
     @include vf-focus;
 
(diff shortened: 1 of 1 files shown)


scss/_patterns_accordion.scss
```diff
@@ -1,5 +1,11 @@
 @import 'settings';
 
+@mixin align-icon($icon-size: 1rem, $line-height: default-text) {
```
bartaz, line 3: Just a little suggestion to maybe rename it to `accordion-align-icon` to be a bit safer about not causing any name collisions.

scss/_patterns_accordion.scss
```diff
@@ -1,5 +1,11 @@
...
+  // calculate icon top position based on top padding (from %single-border-text-vpadding--scaling)
+  // and offset to align it with baseline of accordion tab text element
+  top: $table-cell-vertical-padding + 0.5 * (map-get($line-heights, $line-height) - $icon-size);
```
bartaz, line 6: It will be a bit safer to wrap it all in `calc` (https://sass-lang.com/documentation/values/calculations) so that SCSS can be smart about it if at any point these variables are not all using the same units.

```suggestion
  top: calc($table-cell-vertical-padding + 0.5 * (map-get($line-heights, $line-height) - $icon-size));
```

scss/_patterns_accordion.scss
andesol: With the new update I'm no longer using a mixin, but good point :+1:

scss/_patterns_accordion.scss
andesol: :+1:

andesol:
I updated the styling trying to follow what @lyubomir-popov shared yesterday. Now the icon's alignment is done from the baseline and we don't need media-queries.

image


Side effect 1: since the icon is part of the "flow" now (previously it was `position: absolute`), the text in a `h2` is slightly missaligned (due to the icon being a bit bigger and pushing to the right). I don't know if having different headings in the same accordion is a real use-case.

Side effect 2: since icon's size scales with font-size, the icon next to an `h2` is slightly bigger on large screen but slightly smaller than the rest in small screens. I don't know if that's expected, but it happens here as well: https://vanillaframework.io/docs/examples/utilities/font-metrics

scss/_patterns_accordion.scss
```diff
@@ -78,22 +77,21 @@
...
+    vertical-align: 0;
   }
 
-  h3.p-accordion__heading > .p-accordion__tab::before,
-  .p-heading--3 > .p-accordion__tab::before {
-    $icon-top-position: calc(#{$table-cell-vertical-padding + (map-get($line-heights, h3) - $icon-size) * 0.5});
-    top: $icon-top-position;
+  .p-heading--3 > .p-accordion__tab::before,
+  h3.p-accordion__heading > .p-accordion__tab::before {
+    @include vf-icon-size($default-icon-size);
```
bartaz, line 88: Is this needed when it's default icon size? Or is the size different in standard accordion?

scss/_patterns_accordion.scss
```diff
@@ -38,13 +33,17 @@
     z-index: 2;
 
     &::before {
```
bartaz, line 35: Since we seem to copy quite a bit of spacing from icon here, maybe `@extend %icon` would work instead?

Or at least we should have a comment in code mentioning that the vertical-align calculation is based on the one from `%icon`.

andesol: Good point! I didn't realise I was repeating that many properties. Added `@extend %icon`

scss/_patterns_accordion.scss
andesol: I was about to remove it but I thought maybe it was to make sure that the icon next to an `h3` have the right size no matter the base size. But you're right that in this case it's unnecessary. I just deleted it.

scss/_patterns_accordion.scss
```diff
@@ -78,22 +71,17 @@
...
-    $icon-top-position: calc(#{$table-cell-vertical-padding + (map-get($line-heights, h3) - $icon-size) * 0.5});
-    top: $icon-top-position;
-  }
-
-  h4.p-accordion__heading > .p-accordion__tab::before,
-  .p-heading--4 > .p-accordion__tab::before {
-    $icon-top-position: calc(#{$table-cell-vertical-padding + (map-get($line-heights, h4) - $icon-size) * 0.5});
-    top: $icon-top-position;
+  h4.p-accordion__heading > .p-accordion__tab::before {
+    vertical-align: 0;
```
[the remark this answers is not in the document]
andesol: I assumed it was an exception to align the icon to the heading. It is used in the [font metrics example](https://vanillaframework.io/docs/examples/utilities/font-metrics) and defined here: https://github.com/canonical-web-and-design/vanilla-framework/blob/main/scss/_patterns_icons.scss#L62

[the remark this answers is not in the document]
bartaz: @lyubomir-popov 

current (with `vertical-align: 0`):

Screenshot 2022-04-21 at 16 43 57


without `vertical-align 0`:

Screenshot 2022-04-21 at 16 44 53

lyubomir-popov:
@bartaz it tries to align with the top of small letter - known as x-height. It might need a bit of finessing, but it shouldn't just be reset to 0 IMO.

bartaz:
Created #4437 for further adjustments. We can merge this as it is.

PR #4427, opened 2022-04-20
Add menu to all patterns

albertkol:
## Done

Add tabs to all other patterns that had accessibility sections:
- patterns/breadcrumbs
- patterns/buttons
- patterns/card
- patterns/contextual-menu
- patterns/grid
- patterns/heading-icon
- patterns/icon
- patterns/links
- patterns/lists
- patterns/logo-section
- patterns/modal
- patterns/navigation
- patterns/notification
- patterns/pagination
- patterns/search-box
- patterns/status-labels
- patterns/strip
- patterns/switch
- patterns/tabs
- patterns/tooltips

## QA

- Check all other patterns with accessibility pages, they should have the menu and work just as well as accordion did on the previous PR.

### DIFF
--- a/component_tabs.yaml
+++ b/component_tabs.yaml
@@ -1,3 +1,63 @@
 patterns/accordion:
   accessibility: True
   design_guidelines: False
+patterns/breadcrumbs:
+  accessibility: True
+  design_guidelines: False
+patterns/buttons:
+  accessibility: True
+  design_guidelines: False
+patterns/card:
+  accessibility: True
+  design_guidelines: False
+patterns/contextual-menu:
+  accessibility: True
+  design_guidelines: False
+patterns/grid:
+  accessibility: True
+  design_guidelines: False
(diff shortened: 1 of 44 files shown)


component_tabs.yaml
```diff
@@ -1,3 +1,63 @@
 patterns/accordion:
   accessibility: True
   design_guidelines: False
+patterns/breadcrumbs:
+  accessibility: True
+  design_guidelines: False
```
bartaz, line 6: Nit picking a bit - but could it simply default to false? so if it's not there in config, it means there is no tab for it?

bartaz (approved):
LGTM, with a little change to consider

component_tabs.yaml
albertkol: Turns out it was already working that way. I removed all the lines that init the value to False.

bartaz (approved):

PR #4432, opened 2022-04-20
Fix page jumping after codepen loads

### LINKED ISSUE #3017: Loading of embedded examples moves the page when loading specific section
Follow up to #3015

When navigating to specific section of the docs (for example https://vanillaframework.io/docs/patterns/navigation#side-navigation) on slower connection the CodePens loading later make page move and given section is no longer on top of the screen.

**Expected behavior**

When navigating to specific section it should be scrolled to regardless of CodePen loading.

We could make the wrapper div around CodePen reserve the space before the CodePen loads.

bethcollins92:
## Done

- Add a fix scroll function to ensure the codepen loading doesn't move the title out of view

Fixes #3017 

## QA

- Open [demo](https://vanilla-framework-4432.demos.haus/docs/patterns/navigation#side-navigation)
- Check that the "Side navigation" title is in view. 
- Compare to [main](https://vanillaframework.io/docs/patterns/navigation#side-navigation)

### DIFF
--- a/templates/static/js/example.js
+++ b/templates/static/js/example.js
@@ -270,6 +270,15 @@
     return scripts;
   }
 
+  function fixScroll() {
+    const titleId = window.location.hash;
+    const title = document.querySelector(`${titleId}`);
+    var rect = title.getBoundingClientRect();
+    if (title) {
+      window.scrollTo(0, rect.top);
+    }
+  }
+
   /**
     Attaches change event listener to a given select.
     @param {HTMLElement} dropdown Select element belonging to a code snippet.
@@ -293,4 +302,5 @@
       }
     });
   }
+  fixScroll();
 })();
--- a/templates/static/js/example.js
+++ b/templates/static/js/example.js
@@ -166,6 +166,7 @@
           var style = iframe.contentWindow.getComputedStyle(iframe.contentDocument.body);
           iframe.height = frameHeight + 32 + 'px'; // accommodate for body margin
           clearInterval(resizeInterval);
+          fixScroll();
         }
       }, 100);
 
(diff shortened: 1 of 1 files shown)


templates/static/js/example.js
```diff
@@ -270,6 +270,20 @@
     return scripts;
   }
 
+  function fixScroll() {
+    const titleId = window.location.hash;
+    if (titleId) {
+      const title = document.querySelector(`${titleId}`);
+      var rect = title.getBoundingClientRect();
```
bartaz, line 277: This doesn't seem needed?

bethcollins92: Oh no it isn't - I was trying different techniques using scrollTo or scrollIntoView. It still seems flakey though! What's the demo looking like to you?

templates/static/js/example.js
```diff
@@ -293,4 +307,5 @@
       }
     });
   }
+  fixScroll();
```
bartaz, line 310: Around line 160+ we have a section that does resizing of the iframes after they load.
I guess this fixScroll() could be called in there? So it triggers after iframe loads, not when page loads.

bethcollins92: Thanks - that seems to work a treat

templates/static/js/example.js
```diff
@@ -270,6 +271,14 @@
     return scripts;
   }
 
+  function fixScroll() {
+    const titleId = window.location.hash;
+    if (titleId) {
+      const title = document.querySelector(`${titleId}`);
```
andesol, line 277: `titleId` seems to be a string, so do we need the backticks?

andesol (approved):
Such an elegant solution :100: 

Just a minor detail...

templates/static/js/example.js
bethcollins92: Ahh - good spot! I was trying a few diff solutions and forgot to change this back. Thanks!

PR #4434, opened 2022-04-21
Generate id from heading

andesol:
## Done

- Search for `h2`s without an id and generate one from the heading.

(I think in HTML5 any character is allowed except whitespace, so I kept it simple. Just lower cased it for consistency)

## QA

- Open demo
- Go to the "Segmented Control" docs (or "Get Started", for the matter)
- Check side nav is generated from the headings

## Screenshots
image

### DIFF
--- a/templates/static/js/scripts.js
+++ b/templates/static/js/scripts.js
@@ -87,6 +87,14 @@
 
 // Add table of contents to side navigation on documentation pages
 (function () {
+  // Generate id from H2s content when it does not exist
+  document.querySelectorAll('main h2:not([id])').forEach(function (heading) {
+    var id = heading.textContent
+      .toLowerCase()
+      .replaceAll(/\s+/g, "-")
+    heading.setAttribute("id", id)
+  })
+
   // get all headings from page and add it to current highligted item in side navigation
   var list = document.createElement('ul');
   list.classList.add('p-side-navigation__list');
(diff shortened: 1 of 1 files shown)


templates/static/js/scripts.js
```diff
@@ -97,7 +103,7 @@
   var anchor = document.createElement('a');
   anchor.classList.add('p-side-navigation__link');
 
-  // Add all H3s with IDs to the table of contents list
+  // Add all H2s with IDs to the table of contents list
```
bartaz, line 106: Good catch, thanks for reading the comments ;)

templates/static/js/scripts.js
```diff
@@ -87,6 +87,12 @@
 
 // Add table of contents to side navigation on documentation pages
 (function () {
+  // Generate id from H2s content when it does not exist
+  document.querySelectorAll('main h2:not([id])').forEach(function (heading) {
+    var id = heading.textContent.toLowerCase().replaceAll(/\s+/g, '-');
```
bartaz, line 92: This may fail with some special characters in headers.

Maybe worth adding some more rules in here. That's what I found in one markdown implementation:

https://github.com/markedjs/marked/blob/30e90e5175700890e6feb1836c57b9404c854466/src/Slugger.js#L19

bartaz (approved):
LGTM, with a little suggestion to make it more bulletproof

PR #4492, opened 2022-06-14
Docker: Make pip install slightly faster

jkfran:
On startup, pip checks if you’re running the latest version or not, and print a warning if you’re not.

This saves about 0.2-0.3s, not a very significant improvement; the actual improvement probably depends on your network speed and other factors.


bartaz (approved):
LGTM

PR #4496, opened 2022-06-20
Add u.c blog card example and necessary styles

lyubomir-popov:
## Done
Add a card flavour similar to the u.c blog article cards: (for reference - this is in response to this [issue](https://github.com/canonical-web-and-design/vanilla-framework/pull/4496))
image
- header, main, footer parts, with the footer sticking to the bottom in case it is used in an equal heights scenario (which is the case on u.c)
## QA

- Open [demo](https://vanilla-framework-4496.demos.haus/docs/examples/patterns/card/article-preview)
- review the article-preview example `/docs/examples/patterns/card/article-preview`

- [ ] Any changes to component class names (new patterns, variants, removed or added features) should be listed on the [what's new page](https://github.com/canonical-web-and-design/vanilla-framework/blob/main/templates/docs/whats-new.md).

## Screenshots

### DIFF
--- a/scss/_patterns_card.scss
+++ b/scss/_patterns_card.scss
@@ -1,6 +1,7 @@
 @import 'settings';
 
 @mixin vf-p-card {
+  @include vf-p-card-article-preview;
   @include vf-p-card-default;
   @include vf-p-card-highlighted;
   @include vf-p-card-overlay;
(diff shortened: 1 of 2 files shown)


lyubomir-popov:
@sowasred2012 would something like this  work?

I had to use a span instead of an h5 because of some weird overrides on margin-bottom on p, h5, h6, if those are fixed then this can use an h5.

Putting do not merge label as I'm hoping to pass this onto a developer to finish, the pr here is just a prototype.

sowasred2012:
I think this could work, I had one or two comments:

- the tag/category that sits in the header has a lot more whitespace around it now, it feels too prominent to me, as though a lot of space is being given to something that ultimately is not super important.
- I think a span is fine for this use case because I'm not convinced the tag/category should be within a heading element at all, it doesn't make sense for it to be one within the heirarchy of the page. In any case, u.c/blog heading levels are all over the place, so this more of a comment for that project (which I'll raise an issue for), but I think for dev on this there's no need to make special provisions for heading elements, at least not for the uses we currently have.
- On my browser (Chrome, Linux), there's a scroll bar visible:
Screenshot from 2022-06-21 16-00-53

anthonydillon:
The entire purpose of the [content bleed component](https://vanillaframework.io/docs/patterns/card#content-bleed) was to replace the blog cards we have copied all over our code bases. This proposal is far too specific to the use case (`.p-card--article-preview`). We should be very careful and mindful when adding varients to any component. Ensuring they are truly required and generic enough to be used in multiple situations.

I believe we should use the current content-bleed pattern and move the category information into the card body. The card's footer should stick to the bottom of the card in situations such as the use of equal heights. 

Can we do something simple in this case, like this:
image

lyubomir-popov:
This has now been unblocked with the design solution outlined in https://github.com/canonical-web-and-design/vanilla-framework/issues/4495. cc @bartaz @sowasred2012

lyubomir-popov:
@sowasred2012 Closing this, please make a new issue when you get to implementing the decisions outlined in https://github.com/canonical-web-and-design/vanilla-framework/issues/4495

PR #4536, opened 2022-08-17
feat: expandable side navigation

petermakowski:
## Done

- add expandable side navigation

Fixes https://github.com/canonical/vanilla-squad/issues/1420

## QA

- Open https://vanilla-framework-4536.demos.haus/docs/examples/patterns/side-navigation/expandable
- Open https://vanilla-framework-4536.demos.haus/docs/examples/patterns/side-navigation/expandable-dark

## Screenshots

### DIFF
--- a/scss/_patterns_side-navigation.scss
+++ b/scss/_patterns_side-navigation.scss
@@ -607,13 +607,15 @@
     top: 0;
     margin: 0;
     border-radius: 0;
-    // TODO: set the same height as the link
+    padding-top: $spv--x-small;
+    padding-bottom: $spv--x-small;
+    line-height: inherit;
+    font-size: inherit;
 
     &::before {
       @extend %icon;
       content: '';
       @include vf-icon-chevron;
-
       transform: rotate(-90deg);
       transition: transform 100ms;
     }
--- a/scss/_patterns_side-navigation.scss
+++ b/scss/_patterns_side-navigation.scss
@@ -621,6 +621,17 @@
     }
   }
 
+  .fade-in {
+      @include vf-animation();
+      opacity: 0;
+      transform: translateY(-20px);
+  }
+
+  .fade-in--after {
+      opacity: 1;
+      transform: translateY(0);
+  }
+
   .p-side-navigation__item[aria-expanded='true'] > .p-side-navigation__expand::before {
     transform: rotate(0deg);
   }
--- a/scss/_patterns_side-navigation.scss
+++ b/scss/_patterns_side-navigation.scss
@@ -320,13 +320,14 @@
     position: relative;
   }
 
-  .p-side-navigation.is-expandable {
-    // extra padding to accomodate for `.p-side-navigation__expand` button
-    padding-right: 1rem + (2 * $spv--x-small);
-  }
-
   .p-side-navigation__link {
     @extend %side-navigation__link;
+  }
+
+  .p-side-navigation__link.is-expandable,
+  .p-side-navigation__text.is-expandable {
+    // extra padding to accomodate for `.p-side-navigation__expand` button
+    padding-right: 3rem;
   }
 
   .p-side-navigation__heading,
--- a/scss/_patterns_side-navigation.scss
+++ b/scss/_patterns_side-navigation.scss
@@ -627,11 +627,11 @@
     }
   }
 
-  .p-side-navigation__item[aria-expanded='true'] > .p-side-navigation__expand::before {
+  .p-side-navigation__item > .p-side-navigation__expand[aria-expanded='true']::before {
     transform: rotate(0deg);
   }
 
-  .p-side-navigation__item[aria-expanded='false'] > .p-side-navigation__list {
+  .p-side-navigation__item > [aria-expanded='false'] + .p-side-navigation__list {
     display: none;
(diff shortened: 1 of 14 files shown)


lyubomir-popov:
LGTM 👍

templates/docs/examples/patterns/side-navigation/_expand_script.js
```diff
@@ -0,0 +1,43 @@
...
+
+const handleToggle = (e) => {
+  const item = e.currentTarget.closest('.p-side-navigation__item');
+  const button = item.querySelector('.p-side-navigation__expand');
+  const link = item.querySelector('.p-side-navigation__link');
+  const nestedList = item.querySelector('.p-side-navigation__list');
+  [button, link, nestedList].forEach((el) => el.setAttribute('aria-expanded', el.getAttribute('aria-expanded') === 'true' ? 'false' : 'true'));
+};
+
+navigationLinks.forEach((link) => {
```
bartaz, line 29: We should remove this part - we don't want links to toggle the dropdowns.

templates/docs/examples/patterns/side-navigation/_expand_script.js
```diff
@@ -0,0 +1,43 @@
+// Setup expand toggles for all side navigations on the page.
+
+var expandToggles = document.querySelectorAll('.p-side-navigation__expand');
+var navigationLinks = document.querySelectorAll('.p-side-navigation__link');
+
+const setup = (toggle) => {
```
bartaz, line 6: This setup part is a bit tricky depending on how backend is implemented.

If backend only sets `aria-expanded` on the button, it causes an edge case where the button may have `aria-expanded=true`, but link and list will be set to false by this setup function.

So I think ideally we want setup to:
- check if the `__expand` button has aria-expanded, if it doesn't set to false
- set aria-expanded on link and list to the same value as the expand button

This way the setup should end up with consistent state.

templates/docs/examples/patterns/side-navigation/_expand_script.js
```diff
@@ -0,0 +1,43 @@
...
+  if (!nestedList?.hasAttribute('aria-expanded')) {
+    nestedList.setAttribute('aria-expanded', false);
+  }
+};
+
+const handleToggle = (e) => {
+  const item = e.currentTarget.closest('.p-side-navigation__item');
+  const button = item.querySelector('.p-side-navigation__expand');
+  const link = item.querySelector('.p-side-navigation__link');
+  const nestedList = item.querySelector('.p-side-navigation__list');
```
bartaz, line 25: Is `querySelector` going to always return first closest element in the tree? The lists can be nested even more, so technically any item may have multiple `.p-side-navigation__list` nested in it. We want the one closest to the item.

Is `querySelector` going to return that by definition?

petermakowski: Good point. Even though technically `querySelector` doesn't guarantee to return the first closest element in the tree (as it's using depth-first traversal) this isn't a problem in this case.

That's because by definition - any nested `p-side-navigation__list` will always be contained in outer list which will be returned first as the browser traverses down the tree.

templates/docs/examples/patterns/side-navigation/_expand_script.js
petermakowski: Good catch! Updated.

templates/docs/examples/patterns/side-navigation/_expand_script.js
bartaz: Yes, that's also my understanding, so let's hope it works like that in all implementations :)

scss/_patterns_side-navigation.scss
```diff
@@ -375,6 +388,49 @@
     padding: 0; // padding will come from the link in heading
   }
 
+  .p-side-navigation__expand {
```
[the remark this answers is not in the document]
petermakowski: Yes, sorry, this is a duplicate, forgot to push. Updated now.

scss/_patterns_side-navigation.scss
```diff
@@ -644,18 +588,6 @@
     color: $color-sidenav-text-active;
   }
 
-  .p-side-navigation__expand {
```
bartaz: These color theme styles seem to be missing now from the new extracted files.

It won't be trivial to move them though, I guess the only reliable way would be to create a minimal theme in the expandable SCSS file as well.

bartaz (changes requested):

bartaz:
I think we should remove the "New" label and warning icon from this example.

We don't know if and how it would work with expand button, and we don't need to support that. Let's not give an impression that we do.
image

scss/_patterns_side-navigation.scss
petermakowski: Done

petermakowski: I added a new docs page for the dark variant [/docs/examples/patterns/side-navigation/expandable-dark](https://vanilla-framework-4536.demos.haus/docs/examples/patterns/side-navigation/expandable-dark) to make it easier to test.

bartaz (approved):

PR #4584, opened 2022-10-13
Stepped counter size

### LINKED ISSUE #4561: Stepped lists: Change counter font-size to root font size
## Pattern to amend

Proposal: 
1) make counters (numbers in circles) paragraph size, regardless of the heading style of the titles
2) replace gray-filled circles with ones that are white with a black outline. This way it will be consistent with the new stepper designs

Counter circles become too large and prominent when they contain a digit in the same size as the heading: 

image

Can we instead stick to paragraph size like so:

## Visual

image

huwshimi:
## Done

- Update stepped lists to use a fixed font size for the counters.

Fixes #4561.

## QA

- Open all the list examples:
  - https://vanilla-framework-4584.demos.haus/docs/examples/patterns/lists/lists-stepped
  - https://vanilla-framework-4584.demos.haus/docs/examples/patterns/lists/lists-stepped-without-headings
  - https://vanilla-framework-4584.demos.haus/docs/examples/patterns/lists/lists-stepped-heading-classes
  - https://vanilla-framework-4584.demos.haus/docs/examples/patterns/lists/lists-stepped-detailed
- Check that the font size is always the same.

## Screenshots

Screen Shot 2022-10-13 at 2 18 58 pm

### DIFF
--- a/scss/_patterns_lists.scss
+++ b/scss/_patterns_lists.scss
@@ -17,12 +17,15 @@ $list-status-icon-nudge: 0.3rem; // 0.3rem seems to be a magic number that posit
 
     &::before {
       align-self: start;
-      background-color: $color-mid-x-light;
+      background-color: $color-x-light;
+      border: 1px solid $color-x-dark;
       border-radius: 100%;
       content: counter(li);
       counter-increment: li;
       direction: rtl;
       display: block;
+      font-size: map-get($base-font-sizes, base);
+      font-weight: $font-weight-bold;
       left: 0;
       position: absolute;
       text-align: center;
@@ -276,9 +279,13 @@ $list-status-icon-nudge: 0.3rem; // 0.3rem seems to be a magic number that posit
     &::before {
       margin-right: $sph--large;
       width: $bullet-width-mobile;
+      // Need to account for the 1px border:
+      line-height: calc(#{$bullet-width-mobile} - 2px);
 
       @media (min-width: $breakpoint-heading-threshold) {
         width: $bullet-width;
+        // Need to account for the 1px border:
+        line-height: calc(#{$bullet-width} - 2px);
       }
     }
   }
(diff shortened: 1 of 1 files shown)


lyubomir-popov:
@huwshimi I would do this: 
```
line-height: calc(1.5rem - 2px);
width: 1.5rem;
margin-top: 0.5rem;
```
also weight: 300, sorry for the confusion I know I said 400 previously.
(margin-top .25rem) on the h4

And we also need to bring text closer as the padding on it is taking the icon width into account

(Please use variables to express the above things, I've just done it quick and dirty here)

End result:

image

huwshimi:
Thanks @lyubomir-popov I've updated it so it now looks like this:

dev local_8101_docs_examples_patterns_lists_lists-stepped

scss/_patterns_lists.scss
```diff
@@ -295,10 +302,16 @@ $list-status-icon-nudge: 0.3rem; // 0.3rem seems to be a magic number that posit
...
+    $bullet-margin-mobile: 0;
 
     @if $i < 5 {
-      $bullet-width: map-get($line-heights, h#{$i});
-      $bullet-width-mobile: map-get($line-heights, h#{$i}-mobile);
+      // To align the bullet with the title we need to use half
+      // the height difference between the bullet and the title and then add the
+      // nudge to position it with the lower case text.
+      $bullet-margin: ((map-get($line-heights, h#{$i}) - $bullet-width) / 2) + map-get($nudges, h#{$i});
+      $bullet-margin-mobile: ((map-get($line-heights, h#{$i}-mobile) - $bullet-width-mobile) / 2) + map-get($nudges, h#{$i});
```
lyubomir-popov, line 314:
```suggestion
      $bullet-margin-mobile: ((map-get($line-heights, h#{$i}-mobile) - $bullet-width-mobile) / 2) + map-get($nudges, h#{$i}-mobile);
```

lyubomir-popov (changes requested):
you need the mobile nudges in one of the calculations, suggestion in the code.

scss/_patterns_lists.scss
lyubomir-popov: also, if you do the above you need to add h5 as an entry in the map and copy the value from h4. Otherwise you'd need to make an exception in the for loop logic.

huwshimi: I initially tried with the mobile nudges but for some reason the counters get moved quite significantly out of alignment:

dev local_8101_docs_examples_patterns_lists_lists-stepped

Whereas using the non-mobile nudges they maintain their position for both mobile/non-mobile. I wonder if that's because the counter is inside the heading element which already applies the nudge to the padding. It looks like the counter would be in the right place if it was a sibling of the heading element.

bartaz (approved):
LGTM, thanks!

Issue #4100, opened 2021-10-27
Create example of a grouped table row


anthonydillon:
## Pattern to amend
Table

## Visual
![image](https://user-images.githubusercontent.com/1413534/139076786-c34519ca-943a-4055-a554-8534e7ed2436.png)
_Can be found here: https://ubuntu.com/cpu-compatibility_

## Context
We often want to group the some roles in the table around a common column (I assume the first one). We can solve this with `span-row` but then needs some border styling on the table. 

I am [proposing a "patterned" version](https://github.com/canonical-web-and-design/ubuntu.com/pull/10731/files#diff-1af0881f75e9b3c4d8a260f9498a6da0637bdb18fd7fd246adcc8fea010a3f1cR42) of this which will work with the mobile-card variant.

lyubomir-popov:
The row borders are important, they prevent cell contents from appearing vertically grouped from a visual point of view. We actually have a solution for this in the modular table spec under the grouping scenario: https://ubuntucommunity.s3.dualstack.us-east-2.amazonaws.com/original/2X/f/f7acb578236f7ec0b6b4053d8f4ba827b856cfcc.png https://discourse.ubuntu.com/t/grouping/19857

lyubomir-popov:
@clagom this seems ready to build to me. The version in @anthonydillon 's screenshot matches the modular table spec so nothing to add from visual design perspective. @bartaz do you agree? Anything missing that you'd like to see in the spec as it relates to this issue?

bartaz:
If it's fine from design perspective it can go to build. I'm not sure exactly from the top of my head how we would implement that into table pattern, but it's something to figure out when building.

clagom:
Should it be a variant of the table pattern? So that we can pick one or the other case by case.

lyubomir-popov:
@clagom it already is,  the image [here](https://ubuntucommunity.s3.dualstack.us-east-2.amazonaws.com/original/2X/f/f7acb578236f7ec0b6b4053d8f4ba827b856cfcc.png) is from the discourse article on table grouping.

clagom:
Yes sorry I meant a variant implementation wise, e.g. class name, so that we can set which one we want to use every time.

bartaz:
@huwshimi If you wanna look into this, I guess for start adding an example of a table with grouped cells (via row span) would be worth to see if it renders correctly or does it need some amendments.

bartaz:
[...]

PR #4588, opened 2022-10-14
Add example of grouped rows

huwshimi:
## Done

- Add example of grouping table rows using existing classes.

Fixes: #4100.

## QA

- Open [/docs/examples/patterns/tables/table-grouped-rows](https://vanilla-framework-4588.demos.haus/docs/examples/patterns/tables/table-grouped-rows) (you might need to do this locally as the demo doesn't seem to be working).
- Check that the AMD and Intel columns span multiple rows (the borders get inset for the subsequent rows).
- Resize to medium and small screens and check that the vendor is displayed on all cards.

## Screenshots

Screen Shot 2022-10-18 at 10 41 05 am
Screen Shot 2022-10-18 at 10 41 24 am
Screen Shot 2022-10-18 at 10 41 45 am

### DIFF
--- a/templates/docs/examples/patterns/tables/table-grouped-rows.html
+++ b/templates/docs/examples/patterns/tables/table-grouped-rows.html
@@ -1,0 +1,47 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Table / Grouped rows{% endblock %}
+
+{% block content %}
+<table class="p-table--mobile-card" aria-label="Table of CPU compatibility matrix for Ubuntu">
+  <thead>
+    <tr>
+      <th>Vendor</th>
+      <th>CPU family</th>
+      <th>Code name</th>
+      <th class="u-align--right">Initial support in Ubuntu LTS</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td data-heading="Vendor" rowspan="3">AMD</td>
+      <td data-heading="CPU family">EPYC 7xx3</td>
+      <td data-heading="Code name">Milan (Zen 3)</td>
+      <td data-heading="Initial support in Ubuntu LTS" class="u-align--right">18.04.5</td>
+    </tr>
+    <tr>
+      <td data-heading="Vendor" class="u-hide--large">AMD</td>
+      <td data-heading="CPU family">EPYC 7xx2</td>
+      <td data-heading="Code name">Rome (Zen 2)</td>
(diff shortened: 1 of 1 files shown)


huwshimi:
I've added an example of grouped rows here:
https://vanilla-framework-4588.demos.haus/docs/examples/patterns/tables/table-grouped-rows

In my attempt I've used `u-hide--large` so that the cards have headings when the mobile cards appear, however when applying `u-hide--large` to a td or th it gets applied at both medium and large sizes as it uses `min-width: $breakpoint-small`: https://github.com/canonical/vanilla-framework/blob/main/scss/_utilities_hide.scss#L60

All the other `--large` cases in that file use `min-width: $breakpoint-large` so is that possibly a bug?

huwshimi:
> > All the other `--large` cases in that file use `min-width: $breakpoint-large` so is that possibly a bug?
> 
> Definitely a bug! It should consistently be applied to anything above the large breakpoint.

Ah great, thanks. I've got a PR to fix this: https://github.com/canonical/vanilla-framework/pull/4590.

huwshimi:
This one is blocked until #4590 lands.

bartaz:
#4590 is merges, so this is unblocked I guess

huwshimi:
> #4590 is merged, so this is unblocked I guess

Thanks! This is ready for review now.

bartaz:
Seems like cells in some rows are shifted by one column or something?

image

I guess it's because when first cell has `row-span` you should not included it in rows where it spans to.

bartaz:
> This may cause an issue with mobile view though, as there will be no cell to show in mobile card view… Not sure if there is an easy way around this.

@huwshimi I don't know if you will be able to figure out how to make it work with mobile table but without additional columns on desktop. I'm not sure if `u-hide` will help with that. Feel free to explore whatever you can and leave the rest for us.

huwshimi:
@bartaz You're totally right, I missed that. It works if you apply `display: none` but `u-hide` on table cells sets `display: table-cell` and then sets `opacity: 0` (presumably this was done this way as normally you wouldn't want it to change the order of the cells, only hide its contents).

I can't see a way to do this without adding a new class, either as a generic way to remove table cells or a specific grouped rows class.

bartaz:
Thanks for the exploration @huwshimi. I'm gonna close this PR for now as it requires a bit more exploration and thought of how we want to implement this.

PR #4616, opened 2022-12-06
Update imports to the variable font beta

ClementChaumel:
## Done

Update imports to the variable font beta

The change seems not to be enough, when I look at the network tab I can see that the static font still gets downloaded instead of the variable one 

image

Not sure where else I should make the change


ClementChaumel:
Ok it imports correctly now but I'm still not sure it uses the variable font as the 
`font-variation-settings: "wght" 350, "wdth" 73;` declaration has no effect whatsoever

scss/_base_fontfaces.scss
```diff
@@ -1,188 +1,11 @@
...
-          font-weight: $font-weight-regular-text;
-          src: url('#{$assets-path}317bd676-Ubuntu-L-latin-subset.woff2') format('woff2'), url('#{$assets-path}c09862e1-Ubuntu-L-latin-subset.woff') format('woff');
-          unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2212, U+2215;
-        }
-      }
+    @font-face {
+      font-display: $font-display-option;
+      font-family: 'Ubuntu';
+      font-style: normal;
+      font-weight: $font-weight-regular-text;
```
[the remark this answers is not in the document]
bartaz: Or we could specify range between 100 and 600?
https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-weight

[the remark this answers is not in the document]
bartaz: It is set on each heading, we need to let browser know which weights are supported by this particular font.

ClementChaumel:
Hey, I added a little tester to display the difference in behaviour between a variable font from google font and ours
https://vanilla-framework-4616.demos.haus/docs/examples/base/variable-font-test

image


One thing I noticed is that to import the font like that from GF you need to explicitly ask for a range using `..` like so `wght@200..800`

I don't know why the ubuntu font ignores the `font-variation-settings` property entirely though

fitojb:
Fantastic! It finally solves the fake-weight issue when using Bold (700), and it renders crisp and clear on Ubuntu. It's just missing hinting for Windows and interpolation to Thin (100) in the Italic style.

fitojb:
In <https://vanilla-framework-4616.demos.haus/docs/base/typography#:~:text=Edit%20on%20CodePen-,Strong%20text,-Strong%20text>
I think the weight difference when using <strong&gt; needs to be a bit more prominent than it is now. Not an issue that manifested using the static fonts
image

lyubomir-popov:
@fitojb good catch, we need to point the [$font-weight-bold ](https://github.com/canonical/vanilla-framework/blob/7ffe9b8555d9f0d316a1b812b3cea55ff464644a/scss/_settings_font.scss#L17) variable to the 500 weight, it is currently set to 400. cc @ClementChaumel

scss/_base_code.scss
```diff
@@ -12,7 +12,7 @@ $code-inline-padding: 0.25rem;
   kbd,
   pre,
   samp {
-    font-family: unquote($font-monospace);
+    font-family: 'Ubuntu';
```
bartaz, line 15: This should use monospace font. I guess it should be reverted?

scss/_base_fontfaces.scss
```diff
@@ -1,185 +1,88 @@
...
 
-        @font-face {
-          font-display: $font-display-option;
-          font-family: 'Ubuntu Mono';
-          font-style: normal;
-          font-weight: $font-weight-bold;
-          src: url('#{$assets-path}dd4acb63-UbuntuMono-B.woff2') format('woff2'), url('#{$assets-path}e8e36b19-UbuntuMono-B.woff') format('woff');
-        }
+      @font-face {
+        font-family: 'font-family: ' Ubuntu Mono ';';
```
bartaz, line 22: This looks broken, duplicates "font-family" in font name

scss/_base_forms.scss
```diff
@@ -26,7 +26,7 @@
     border-bottom: $input-border-thickness solid $colors--light-theme--text-default;
     border-radius: 0;
     color: $color-dark;
-    font-family: unquote($font-base-family);
+    font-family: 'Ubuntu';
```
bartaz, line 29: Why changing this from variable?

templates/docs/examples/base/variable-font-test.html
```diff
@@ -0,0 +1,46 @@
...
+{% block standalone_css %}base{% endblock %}
+
+
+
+
+{% block content %}
+<style>
+@import url('https://fonts.googleapis.com/css2?family=Open+Sans:ital,wdth,wght@0,75..100,300..800;1,75..100,300..800&display=swap');
+
+.mono{
```
bartaz, line 13: do we need custom class names?

I guess we can simply use built-in stuff, like `<em>` for italic and `<code>` for mono?

templates/docs/examples/base/variable-font-test.html
```diff
@@ -0,0 +1,46 @@
...
+</style>
+<p class="ubuntu">This is a paragraph of Ubuntu variable font.</p>
+<p class="ubuntu-italic">This is a paragraph of Ubuntu variable font in italic.</p>
+
+<p class="mono">This is a paragraph of Ubuntu Mono variable font.</p>
+<label>Weight</label>
+<input id="weight" type="range" min="200" max="800" value="300" step="1" style="appearance: auto" oninput="doTheThing()">
+<label>Width</label>
+<input id="width" type="range" min="75" max="100" value="100" step="1" style="appearance: auto" oninput="doTheThing()">
+<label>Italic</label>
```
bartaz, line 29: This doesn't seem to do anything, can we drop that?

bartaz:
TODO:

- [ ] update docs in typography
- [ ] update docs in settings fonts
- [ ] bump version to 3.10
- [ ] update what's new page with new font information
- [ ] update status labels in side bar navigation (remove current ones, add "updated" to typography page - unless there is any better page about font)

scss/_base_fontfaces.scss
```diff
@@ -1,187 +1,80 @@
...
-          font-style: normal;
-          font-weight: $font-weight-regular-text;
-          src: url('#{$assets-path}fdd692b9-UbuntuMono-R_W.woff2') format('woff2'), url('#{$assets-path}85edb898-UbuntuMono-R_W.woff') format('woff');
-        }
+      @font-face {
+        font-family: 'Ubuntu';
+        font-stretch: 100%; /*  min and max value for the width axis, expressed as percentage */
+        font-style: italic;
+        font-weight: 100 800; /*  min and max value for the weight axis */
+        src: url('#{$assets-path}c1b12cdf-Ubuntu-Italic[wdth,wght]-latin (1).woff2') format('woff2-variations');
```
bartaz, line 18: Could we avoid having unnecessary stuff in font URLs?

Having spaces and `(1)` caused by downloading same name multiple times looks messy.
I know that right now it means uploading it again probably, unless someone from backend guys can help us renaming those on asset server.

anthonydillon: It's important these URLs look clean and clear.

ClementChaumel: Oh my bad I didn't think it mattered. Do we want to agree on a format before I reupload them?

templates/docs/examples/base/variable-font.html
```diff
@@ -0,0 +1,28 @@
...
+{% block standalone_css %}base{% endblock %}
+
+
+{% block content %}
+<p id="roman">This is a paragraph of Ubuntu variable font.</p>
+<p id="italic"><em>This is a paragraph of Ubuntu variable font in italic.</em></p>
+<p><code id="mono">This is a paragraph of Ubuntu Mono variable font.</code></p>
+
+<label>Weight</label>
+<input id="weight" type="range" min="200" max="800" value="300" step="1" style="appearance: auto" oninput="UpdateFontVariations()">
```
bartaz, line 13: Shouldn't it go down to 100 min value?

lyubomir-popov:
Well done @ClementChaumel and @bartaz

bartaz (approved):
Good to go!

fitojb:
Very happy with the improved weight distribution and sizing of the Mono font. Excellent overall

rink-grey:
@fitojb may i ask how can i get the fixed Ubuntu Mono font file, thanks

bartaz:
@rink-grey Ubuntu fonts (including Mono) are available on Google fonts: https://fonts.google.com/?query=Ubuntu (although it may not be the latest variable font version yet)

Latest available font files are in our repos https://github.com/canonical/Ubuntu-Sans-fonts https://github.com/canonical/Ubuntu-Sans-Mono-fonts

rink-grey:
@bartaz  thanks very much!

PR #4623, opened 2022-12-13
Temporary script to fix the upload assets job

bartaz:
This fixes issues with uploading assets to asset server when we release Vanilla.

Let's not merge it for now, but when we will need to next release Vanilla, and uploading to assets server is not fixed yet, we could merge this to be able to release.

### QA

- just code review… we won't know if it works until we run it in GH actions…


bartaz:
This is needed because of https://github.com/canonical/canonicalwebteam.upload-assets/issues/42

ClementChaumel (approved):
LGTM

PR #4636, opened 2023-01-31
Update list styles

ClementChaumel:
## Done

https://vanilla-framework-4636.demos.haus/docs/examples/patterns/lists/lists-dividers-all

Updated the divided list styles and added a dotted variant

## Screenshots

image


lyubomir-popov:
Hi, looking good! Can we ensure the bullets  are positioned as in the design? 
Can we have a way to remove the lines via a utility maybe, or in components like the nav - automatically? (we are styling element qualifiers, so need to be careful to not break anything that is semantically a list, but has a different styling already).
What would the icon versions look like?

bartaz:
What worries me is the number of places we need to reset these styles in Vanilla itself (because of quite complex styles we are adding to native lists). Which likely means we may break any list elements projects have.

It's highly unlikely we will be able to release it as a non-breaking change.

ClementChaumel:
> What worries me is the number of places we need to reset these styles in Vanilla itself (because of quite complex styles we are adding to native lists). Which likely means we may break any list elements projects have.
> 
> It's highly unlikely we will be able to release it as a non-breaking change.

Agreed, the reset is quite simple to apply and the utility class could help in a few instances but updating the styling for the raw tags is bound to create some inconsistencies

scss/_patterns_navigation.scss
```diff
@@ -172,7 +172,6 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
   .p-navigation__item--dropdown-toggle {
     @media (min-width: $breakpoint-navigation-threshold) {
       $nav-link-max-width: 20em !default;
-
```
bartaz, line 175: Seems unrelated unnecessary edit ;)

scss/_base_icon-definitions.scss
```diff
@@ -74,6 +74,10 @@
   background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='16' height='16'%3E%3Cg fill='none' fill-rule='nonzero'%3E%3Cpath fill='#{vf-url-friendly-color($bg-color)}' d='M8 1a7 7 0 110 14A7 7 0 018 1zm2.83 3.502L6.863 9.884 5.174 8.096l-1.09 1.03 2.92 3.096 5.034-6.83-1.208-.89z'/%3E%3Cpath fill='#{vf-url-friendly-color($fg-color)}' d='M10.83 4.502l1.208.89-5.033 6.83-2.922-3.096 1.091-1.03 1.689 1.789z'/%3E%3C/g%3E%3C/svg%3E");
 }
 
+@mixin vf-icon-dot($bg-color: $color-positive, $fg-color: $color-x-light) {
```
bartaz, line 77: This is not really "an icon", and we only use it in one place, so I'm not sure if it needs to be exposed as mixin now.

Even as mixin, I guess it should live in the list component for now, as that's the only place where it is used.

scss/_patterns_lists.scss
```diff
@@ -133,6 +119,11 @@ $list-status-icon-nudge: 0.3rem; // 0.3rem seems to be a magic number that posit
     .p-list__item {
       @extend %vf-list-item;
       @include vf-list-item-divided;
+
+      &.is-ticked,
+      .is-dotted {
```
bartaz, line 124: Is this correct that `&.is-ticked` has `&` and `.is-dotted` doesn't?

scss/_patterns_lists.scss
```diff
@@ -154,6 +145,20 @@ $list-status-icon-nudge: 0.3rem; // 0.3rem seems to be a magic number that posit
   }
 }
 
+%vf-list-item-dotted {
+  background-position-x: -1rem;
```
bartaz, line 149: What are these rem values based on? Can they be expressed as our standard spacing variables or do they need to be hardcoded as values?

scss/_base_icon-definitions.scss
lyubomir-popov: it is a unicode symbol, not an icon - needs to come from the font, this way it will use the correct weight if we ever decide to change it, etc

scss/_patterns_lists.scss
```diff
@@ -73,24 +73,10 @@ $list-status-icon-nudge: 0.3rem; // 0.3rem seems to be a magic number that posit
...
-    right: 0;
-  }
-
-  &:last-of-type,
-  .last-item {
-    &::after {
-      border-bottom: 0;
-    }
-  }
+  box-shadow: inset 0px 1px 0px 0rem #e5e5e5;
```
bartaz, line 76: Do we have this color as variable? We should use one of the available ones.

bartaz:
@ClementChaumel Could you add examples with nested lists (ideally to all combinations), because it seems some styles are broken there with dots:

image

ClementChaumel:
> image these bullets are quite large and seem a little further from the text than in the designs. Can we get them to match the design a little closer please?

Hey, can you be a little more precise? All I have for reference is this 
image
 and no size for the dot. If you have handy the spacing units I need to use and exact size for the dot that would be very helpful so I don't have to eyeball it (and get it wrong) thanks!

lyubomir-popov:
@ClementChaumel Sorry for the unclear feedback. I went in to check, I think I'm using a different bullet, which appears smaller at the same font-size: Can you copy paste it into the `content`: 
`•`

and then the offset of the text from the left edge should be 2rem (32px) - this should be the same regardless of whether you are using an icon, bullet or numeral:
image

ClementChaumel:
Hey, just wanted to make sure this time around, does this look ok?
image

ClementChaumel:
PTAL @bartaz @lyubomir-popov I also updated the example pages to display nested lists!
https://vanilla-framework-4636.demos.haus/docs/examples/patterns/lists/lists-dividers-dotted

ClementChaumel:
Oh wait I misunderstood your comment, no wonder the spacing looks off!

ClementChaumel:
Here's how it looks now 
image

templates/docs/examples/patterns/lists/lists-dividers-dotted.html
```diff
@@ -0,0 +1,16 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Lists / Dotted with horizontal divider{% endblock %}
+
+{% block standalone_css %}patterns_lists{% endblock %}
+
+{% block content %}
+<ul class="p-list--divided">
+  <li class="p-list__item is-dotted">Balance compute load in the cloud</li>
+  <ul class="p-list--divided">
```
bartaz, line 9: That is not valid markup.

Only `<li>` can be a child of ul/ol. So the nested list needs to be inside `<li>`.

```
<ul class="p-list--divided">
  <li class="p-list__item is-dotted">Balance compute load in the cloud
     <ul class="p-list--divided">
     </ul>
   </li>
</ul>
 ```

File: scss/_patterns_lists.scss [hunk 7]
```diff
@@ -133,11 +119,21 @@ $list-status-icon-nudge: 0.3rem; // 0.3rem seems to be a magic number that posit
...
     }
 
     &.is-split .p-list__item:last-of-type {
       border-bottom: 1px solid $color-mid-light;
     }
+
+    ul,
+    ol {
+      margin-bottom: 0;
+      margin-left: 2rem;
```
bartaz, line 135: This should come from spacing variables.

scss/_patterns_lists.scss
```diff
@@ -133,11 +119,21 @@ $list-status-icon-nudge: 0.3rem; // 0.3rem seems to be a magic number that posit
     .p-list__item {
       @extend %vf-list-item;
       @include vf-list-item-divided;
+
+      &.is-ticked {
+        background-position-y: 0.85rem;
```
bartaz, line 124: Could this be expressed as variables as well?

Not sure what this value is, but if it's based of icon size it should be calculated from it.

scss/_patterns_lists.scss
```diff
@@ -133,11 +119,21 @@ $list-status-icon-nudge: 0.3rem; // 0.3rem seems to be a magic number that posit
...
+      }
     }
 
     &.is-split .p-list__item:last-of-type {
       border-bottom: 1px solid $color-mid-light;
     }
+
+    ul,
```
bartaz, line 132: Looking at HTML code it seems that nested lists also have a class `p-list--divided`, so we could target class name here instead of element to avoid element selector.

Would the component even work if we nest a list without this class name? Or with different list class name?

bartaz:
Ticked variant has icons aligned from left, not "centered" like bullets:

image

lyubomir-popov:
@bartaz @ClementChaumel the idea is to have 16px for icons/bullets, then 16px gap: 
image
two numerals (e.g. 12) require 19px width, because it includes some kernign white space, but visually it looks spaced the same way as a 16 px icon followed by a 167px gap, or a bullet. Note that the bullet and numbers have text-align: right

bartaz:
> two numerals (e.g. 12) require 19px width, because it includes some kernign white space, but visually it looks spaced the same way as a 16 px icon followed by a 167px gap, or a bullet. Note that the bullet and numbers have text-align: right

@lyubomir-popov what if numbers get longer? Obviously getting over 100 will not be very likely, but we have nested numbering variant that adds decimals like 11.1, 11.2, … I don't think it's currently supported in divided list (only in standard) but eventually we would need it here as well.

This applies only to nested lists, so I guess we have some more indent there to work with.

templates/docs/examples/patterns/lists/lists-dividers-dotted.html
ClementChaumel: That's what I did at first but then I couldn't manage to style it correctly without using the `:has()` pseudo-selector

scss/_patterns_lists.scss
ClementChaumel: This can go away! I just forgot to remove it

templates/docs/examples/patterns/lists/lists-dividers-dotted.html
bartaz: We can't build on top of invalid markup, so we have to figure out the way to do it.

What we can do in the worst case it to have a class name `is-nested` or `has-nested-list` or something like that. But hopefully this would not be needed.

scss/_patterns_lists.scss
ClementChaumel: Targetting the classname would work in the example because both lists are `p-list--divided` but let's say in the wild someone uses a different style for the inner list. It would not be indented. this is why I targeted the tag instead, as a failsafe. What do you think?

scss/_patterns_lists.scss
```diff
@@ -133,11 +135,45 @@ $list-status-icon-nudge: 0.3rem; // 0.3rem seems to be a magic number that posit
...
+      &:has(ol),
+      &:has(ul) {
+        box-shadow: none;
+        padding-bottom: 0;
+        padding-top: 0;
+
+        &::before {
+          display: none;
+        }
+      }
```
ClementChaumel, line 159: Something funky is happening here:
If I put it all in the same section like this
```css
      &:has(ol),
      &:has(ul), 
      &.nested-list-wrapper {
        box-shadow: none;
        padding-bottom: 0;
        padding-top: 0;

        &::before {
          display: none;
        }
      }
```

Then the rule gets ignored by browsers that don't support the `:has()` pseudo selector.

I also tried 

```css
      &.nested-list-wrapper {
         @extends %some-placeholder
      }

      &:has(ol),
      &:has(ul) {
        @extends %some-placeholder
      }
```

But for some reason I don't understand, it gets ignored too. I suspect it's in the way placeholders are compiled

templates/docs/examples/patterns/lists/lists-dividers-dotted.html
```diff
@@ -0,0 +1,18 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Lists / Dotted with horizontal divider{% endblock %}
+
+{% block standalone_css %}patterns_lists{% endblock %}
+
+{% block content %}
+<ul class="p-list--divided">
+  <li class="p-list__item is-dotted">Balance compute load in the cloud</li>
+  <li class="p-list__item nested-list-wrapper">
```
ClementChaumel, line 9: Here I added a class to specify that this list element is just a wrapper for the nested list and shouldn't have a top border or a dot

ClementChaumel:
@bartaz here is a page with all variations displayed in the same place PTAL

https://vanilla-framework-4636.demos.haus/docs/examples/patterns/lists/lists-dividers-all

templates/docs/examples/patterns/lists/lists-dividers.html
```diff
@@ -5,8 +5,14 @@
 
 {% block content %}
 <ul class="p-list--divided">
-  <li class="p-list__item">Single-command install and upgrade</li>
-  <li class="p-list__item">Postgres database for reliable state</li>
-  <li class="p-list__item">High availability for all services</li>
+  <li class="p-list__item">Balance compute load in the cloud</li>
+  <li class="p-list__item nested-list-wrapper">
+    <ul class="p-list--divided">
```
bartaz, line 10: Why it's nested in separate list element?

Nested lists are within the same list element. You have a list element and it has a child element list with more list elements.

```
  <li class="p-list__item">
    Balance compute load in the cloud
    <ul class="p-list--divided">
        <li> …
     </ul>
  </li>
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul

ClementChaumel: :facepalm: 

Well... I did wonder why I didn't have to do that the first time around. It should be ok now

scss/_patterns_lists.scss
```diff
@@ -138,22 +144,59 @@ $list-status-icon-nudge: 0.3rem; // 0.3rem seems to be a magic number that posit
...
+      margin-left: $sp-x-large;
+      padding-top: $sph--large;
+
+      .p-list__item:last-child {
+        padding-bottom: 0;
+      }
+    }
+  }
+
+  ol.p-list--divided {
```
bartaz, line 159: Fix linting here, we need to disable targeting element in this line (we do want to target ordered lists specifically).

scss/_patterns_lists.scss
```diff
@@ -64,33 +64,23 @@ $list-status-icon-nudge: 0.3rem; // 0.3rem seems to be a magic number that posit
 
   %vf-list-item-state-base {
     @extend %vf-list;
-
-    background-position-y: $spv--x-small + $list-status-icon-nudge;
     background-repeat: no-repeat;
     background-size: $list-status-icon-height;
```
bartaz, line 68: Are the background props still needed on this one?
My guess is these were needed when ticked icon was on item itself, but now we moved it into `:before`, right?

bartaz:
@ClementChaumel Nice work, I think we are getting there. Couple things to address:

- [ ] fix linter issues
- [ ] bump version (new feature)
- [ ] add label to side navigation and new feature to "what's new page"
- [ ] I guess we could update docs a little more, moving the new divided list to top and propose it as a default, mention that it should be used for any lists unless they really need a different styling and possibly mark standard li/ol as deprecated

templates/docs/patterns/lists/index.md
```diff
@@ -7,11 +7,40 @@ context:
...
+
+## Dotted with horizontal divider
+
+You can combine both `.is-dotted` and `.p-list--divided` to style a
+list with horizontal dividers and dots.
+
+<div class="embedded-example"><a href="/docs/examples/patterns/lists/lists-dividers-dotted/" class="js-example">
+View example of the dotted divided list pattern
+</a></div>
+
```
bartaz, line 27: Would be good to have an example and docs for ordered list with lines as well, now that we ask people to use this style for everything.

So please add another example with numbered list and some short docs section for it, thanks!

templates/docs/examples/patterns/lists/lists-dividers-all.html
```diff
@@ -0,0 +1,35 @@
...
+    </ul>
+  </li>
+  <li class="p-list__item">
+    Balance compute load in the cloud
+    <ul class="p-list--divided">
+      <li class="p-list__item">Balance compute load in the cloud</li>
+    </ul>
+  </li>
+</ul>
+<ol class="p-list--divided p-list--nested-counter">
```
bartaz, line 27: Is the `p-list--nested-counter` needed here? It seems that without it the counter doesn't change.

Also, this seems a bit inconsistent with current lists (but maybe we want that).
Right now if you have nested `<ol>`, you will get:

```
1. asfasdf
2. asdfasdf
    1. sdffd
    2. sdfsfd
3. sadfafs
```

and you need the `p-list--nested-counter` to achieve:
```
1. asfasdf
2. asdfasdf
    2.1. sdffd
    2.2. sdfsfd
3. sadfafs
```

with new list it seems that the "nested" style is default, you don't need another class name. But I guess it's better to have it by default and not via additional class name, as it's probably better this way.

bartaz: So, I think that we can just remove nested counter class name from this example as it seems it's not doing anything.

ClementChaumel: Yeah the new ordered list has the nested behaviour by default. Is that ok you think?

bartaz: Yes, let's keep it this way.

bartaz:
```suggestion
<ol class="p-list--divided">
```

templates/docs/examples/patterns/lists/lists-dividers-all.html
```diff
@@ -0,0 +1,35 @@
...
+  <li class="p-list__item">
+    Balance compute load in the cloud
+    <ol class="p-list--divided p-list--nested-counter">
```
bartaz, line 30:
```suggestion
    <ol class="p-list--divided">
```

bartaz (approved):
SHIP IT 🚢 🚀 💯 🎆 🎉

PR #4648, opened 2023-02-03
Update font files and use small caps in table headers

lyubomir-popov:
## Done

Use the latest (0.864) beta of the Ubuntu Variable font, which now includes true small caps.

## QA

- Open [demo](https://vanilla-framework-4648.demos.haus/docs/examples/base/table)
- Check the table erxamples, see that the headings use true small caps instead of the previous faux small caps.

### DIFF
--- a/scss/_base_typography-definitions.scss
+++ b/scss/_base_typography-definitions.scss
@@ -205,10 +205,9 @@
 
   %table-header-label {
     @extend %common-default-text-properties;
-    font-variant: small-caps;
+    font-variant-caps: all-small-caps;
     letter-spacing: 0.07em;
     margin-bottom: map-get($sp-after, default-text) - map-get($nudges, p);
-    text-transform: lowercase;
   }
 
   %muted-heading {
--- a/scss/_base_typography-definitions.scss
+++ b/scss/_base_typography-definitions.scss
@@ -206,7 +206,7 @@
   %table-header-label {
     @extend %common-default-text-properties;
     font-variant-caps: all-small-caps;
-    letter-spacing: 0.07em;
+    letter-spacing: 0.05em;
     margin-bottom: map-get($sp-after, default-text) - map-get($nudges, p);
   }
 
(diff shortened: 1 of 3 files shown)


scss/_base_typography-definitions.scss
```diff
@@ -204,9 +204,10 @@
   }
 
   %table-header-label {
-    @extend %x-small-text;
-
-    text-transform: uppercase;
+    @extend %common-default-text-properties;
+    font-variant-caps: all-small-caps;
```
bartaz, line 208: @lyubomir-popov should these styles be for table header only?  or should they be used in other places where we capitalise text (like the p-text-x-small--capitalised)?

bartaz:
@lyubomir-popov could you have a look at Percy, there is a bit of a shift in tables caused by this change (the text seems bigger than before): https://percy.io/bb49709b/vanilla-framework/builds/25148134/changed/1403546825?browser=firefox&browser_ids=33&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=original&width=1280&widths=375%2C800%2C1280

lyubomir-popov:
> @lyubomir-popov could you have a look at Percy, there is a bit of a shift in tables caused by this change (the text seems bigger than before): https://percy.io/bb49709b/vanilla-framework/builds/25148134/changed/1403546825?browser=firefox&browser_ids=33&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=original&width=1280&widths=375%2C800%2C1280

hmm this might break dense tables where things are taylored to the pixel. Can we leave this out of the release? I feel we need to test in maas before making this a global change. Also, since we're changing font-sizes, we need to also adjust the nudging to put it onto the baseline grid - my guess is that it needs to use the ones from the paragraph.

bartaz:
@lyubomir-popov I just reverted the change in the table and only change small caps util, so that's what we want, only affect the util as in here: https://percy.io/bb49709b/vanilla-framework/builds/25181320/changed/1405449053?browser=firefox&browser_ids=33&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=1280&widths=375%2C1280

bartaz (approved):

Issue #4560, opened 2022-08-31
Side navigation drawer on small screens doesn't trap the focus


bartaz:
via @petermakowski comment via @petermakowski comment https://github.com/canonical/juju.is/pull/431#issuecomment-1231665391

> [...] the side navigation is missing a focus trap (if you keep pressing `Tab` you'll eventually start tabbing through elements on the page outside of it).

**Describe the bug**

The side navigation drawer "popup" on small screen doesn't trap the focus. Ideally it should move back up to "Toggle" side navigation button on top once you focus out of last link in the side nav (but only on small screens where side nav is collapsible).

PR #4680, opened 2023-03-03
Trap focus within the sidenav on small screens [WD-2277]

lorumic:
## Done

Added an example of how the focus should be controlled within a sidenav drawer to the already existing JS code of the sidenav toggle script.

Fixes https://github.com/canonical/vanilla-framework/issues/4560

## QA

- Open [a sidenav example](https://vanilla-framework-4680.demos.haus/docs/examples/patterns/side-navigation/expandable).
- Resize the window so that its width becomes lower than `1036px`. The sidenav switches from "pinned" to "drawer" mode.
- Hit the `Tab` key multiple times to check that the focus is trapped within the sidenav.


ClementChaumel:
In terms of QA, it's working on the example page but not on the main/docs one. Is that expected?

lorumic:
> In terms of QA, it's working on the example page but not on the main/docs one. Is that expected?

I am just checking and I expected it to work there as well - just like any other place where the `docs/examples/patterns/side-navigation/_toggle_script.js` is included - but you're right, it doesn't.

Could it be that there is something with the `embedded-example` mechanism that prevents it from working in the docs page? I see that the same happens for the modal, from which the same behaviour (focus trap) was imported: it is working as expected in the [example page](https://vanilla-framework-4680.demos.haus/docs/examples/patterns/modal/default), but not in the [docs page](https://vanilla-framework-4680.demos.haus/docs/patterns/modal).

bartaz:
I'm not sure if it works as expected… When using `Tab` to go through the items on the side navigation (in mobile mode) the tab focus still doesn't return to top of sidenav but leaves side navigation and focuses on first thing on the page.

lorumic:
> I'm not sure if it works as expected… When using `Tab` to go through the items on the side navigation (in mobile mode) the tab focus still doesn't return to top of sidenav but leaves side navigation and focuses on first thing on the page.

If you're talking about the docs page, I haven't changed anything yet. I just rebased.
In the examples page, it works as intended in mobile mode as well.

bartaz:
I'm not sure if it does:

nav-focus

When focus leaves the last link in the navigation it disappears (it is on a toggle button that is below the navigation) instead of going back to top of navigation.

Maybe it's casued by the fact that this button on the page is part of `p-side-navigation` component as well? Inside the same parent? The focus needs to be trapped inside the `p-side-navigation__drawer` I think.

lorumic:
@bartaz Sorry for the misunderstanding about which element should the focus be trapped into. It is now trapped inside the `p-side-navigation__drawer`.

templates/docs/examples/patterns/side-navigation/_toggle_script.js
```diff
@@ -1,10 +1,70 @@
+// This is not a production ready code, just serves as an example
+// of how the focus should be controlled within the sidenav drawer.
+var sidenavContainer = null;
```
bartaz, line 3: Just a small change, but I would rename it to `expandedSidenavContainer` to make it a bit clearer.

bartaz:
@lorumic Looks good (just small variable rename suggestion), could you apply this to Vanilla docs as well?

bartaz (approved):
Well done, thanks!

Issue #4486, opened 2022-06-08
"Search" patterns selected in side nav on search results page


bartaz:
Go to any search results page. In side nav search patterns are highlighted as if this was the current page:

<img width="535" alt="image" src="https://user-images.githubusercontent.com/83575/172610651-3574cf24-2f71-402d-937f-ad2604736976.png">

PR #4682, opened 2023-03-03
Removed highlight of side nav items on search results page [WD-2276]

lorumic:
## Done

Removed highlight of side nav items on the search results page.

Fixes https://github.com/canonical/vanilla-framework/issues/4486

## QA

- Open [an example search page](https://vanilla-framework-4682.demos.haus/docs/search?q=tooltip).
- Check that no item is selected in  side nav.

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -151,13 +151,17 @@ def global_template_context():
     version_parts = VANILLA_VERSION.split(".")
     version_minor = f"{version_parts[0]}.{version_parts[1]}"
 
-    docs_slug = (
-        flask.request.path.replace("/docs/", "")
-        .replace("/design/", "")
-        .replace("/accessibility", "")
-    )
+    # Add an exception for the /docs/search path
+    if flask.request.path == "/docs/search":
+        docs_slug = ""
+    else:
+        docs_slug = (
+            flask.request.path.replace("/docs/", "")
+            .replace("/design/", "")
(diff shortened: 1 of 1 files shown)


webapp/app.py
```diff
@@ -151,13 +151,17 @@ def global_template_context():
...
+        docs_slug = ""
+    else:
+        docs_slug = (
+            flask.request.path.replace("/docs/", "")
+            .replace("/design/", "")
+            .replace("/accessibility", "")
+        )
 
-    docs_slug = "" if docs_slug == "/docs" else docs_slug
+        docs_slug = "" if docs_slug == "/docs" else docs_slug
```
bartaz, line 164: Why not add the exception here, where we already have a `"/docs` exception

Not sure if that would work but something like:
```
docs_slug = "" if docs_slug == "/docs" or docs_slug == "/docs/search" else docs_slug
```

lorumic: That was the first thing that I tried, but actually the `/docs/` part of the path gets already replaced one instruction before:
```
        docs_slug = (
            flask.request.path.replace("/docs/", "")
            .replace("/design/", "")
            .replace("/accessibility", "")
        )
```
so this condition: `docs_slug == "/docs/search"` will never be matched.

bartaz: Ah, right!

PR #4691, opened 2023-03-08
Updated application layout to use new tagged logo style [WD-2451]

lorumic:
## Done

Added new class `p-panel__tagged-logo` to style a tagged logo for panels in application layouts.
Updated application layout examples to use this tagged logo style.
Added an example of the default application layout using the legacy logo.

Fixes [WD-2451](https://warthogs.atlassian.net/browse/WD-2451)

## QA

- Open the [examples page](https://vanilla-framework-4691.demos.haus/docs/examples).
- Open any of the application layout examples (except the "Structure" one). Check the style of the logo in the side navigation panel.
- Open the [Application / Default with legacy logo](https://vanilla-framework-4691.demos.haus/docs/examples/layouts/application/default-legacy-logo) example.
- Check the legacy logo in the side navigation panel.
  - [Application layout > Panels](https://vanilla-framework-4691.demos.haus/docs/layouts/application#panels)

[WD-2451]: https://warthogs.atlassian.net/browse/WD-2451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vanilla-framework",
-  "version": "3.12.1",
+  "version": "3.12.2",
   "author": {
     "email": "webteam@canonical.com",
     "name": "Canonical Webteam"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vanilla-framework",
-  "version": "3.12.2",
+  "version": "3.13.0",
   "author": {
     "email": "webteam@canonical.com",
     "name": "Canonical Webteam"
--- a/releases.yml
+++ b/releases.yml
@@ -1,4 +1,4 @@
-- version: 3.12.2
+- version: 3.13.0
   features:
     - component: Tagged logo for application layouts
       url: /docs/layouts/application#panels-example
--- a/releases.yml
+++ b/releases.yml
@@ -1,3 +1,9 @@
+- version: 3.12.2
+  features:
+    - component: Tagged logo for application layouts
+      url: /docs/layouts/application#panels-example
+      status: Updated
+      notes: We've introduced a new tagged logo style for application layouts, with class name <code>p-panel__tagged-logo</code>.
 - version: 3.12.1
   features:
     - component: Text with icon utility class
--- a/scss/_layouts_application-panels.scss
+++ b/scss/_layouts_application-panels.scss
@@ -52,6 +52,54 @@
     }
   }
 
+  .p-panel__tagged-logo {
+    @extend %navigation-link;
+
+    padding-left: $sph--x-large;
+
+    &:hover {
+      background-color: transparent !important;
+    }
+
+    &::before {
+      content: none;
+    }
+
+    .p-panel__logo-tag {
+      background-color: $color-accent;
+      height: $navigation-logo-tag-height;
+      left: 0;
+      position: absolute;
+      top: 0;
+      width: $navigation-logo-tag-width;
+
+      @media (min-width: $breakpoint-navigation-threshold) {
+        height: $navigation-logo-tag-height-desktop;
+      }
+
+      .p-panel__logo-icon {
+        bottom: $spv-navigation-logo-bottom-position;
+        height: $navigation-logo-size;
+        left: 50%;
+        position: absolute;
+        transform: translate(-50%, 0);
+        width: $navigation-logo-size;
+      }
+    }
+
+    .p-panel__logo-name {
+      @extend %vf-heading-4;
+
+      color: $colors--light-theme--text-default;
+      font-size: #{map-get($font-sizes, h4)}rem;
+      line-height: map-get($line-heights, x-small);
+
+      .p-panel.is-dark & {
+        color: $colors--dark-theme--text-default;
+      }
+    }
+  }
+
   .p-panel__title {
     @extend %vf-heading-4;
 
@@ -63,6 +111,11 @@
   .p-panel__controls {
     margin-left: auto;
     padding-top: $spv--large;
+  }
+
+  .p-panel__controls-tagged-logo {
+    margin-left: auto;
+    padding-top: $spv--small;
   }
 
   .p-panel__toggle {
--- a/scss/_layouts_application-panels.scss
+++ b/scss/_layouts_application-panels.scss
@@ -110,12 +110,7 @@
 
   .p-panel__controls {
     margin-left: auto;
-    padding-top: $spv--large;
-  }
-
-  .p-panel__controls-tagged-logo {
-    margin-left: auto;
-    padding-top: $spv--small;
+    padding-top: $spv--medium;
   }
 
   .p-panel__toggle {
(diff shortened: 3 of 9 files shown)


bartaz:
Please bump the threshold to 350000:

https://github.com/canonical/vanilla-framework/blob/a72ca52561f688deec00091ffd97ff5aaf30dcf0/tests/parker.js#L15

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vanilla-framework",
-  "version": "3.12.1",
+  "version": "3.12.2",
```
bartaz, line 3: Since we are adding new "feature" (tagged logo) and new CSS class names, this should be 3.13.0.

Sorry, I know that versioning is quite unclear and arbitrary 😉

scss/_layouts_application-panels.scss
```diff
@@ -65,6 +113,11 @@
     padding-top: $spv--large;
   }
 
+  .p-panel__controls-tagged-logo {
```
bartaz, line 116: It would be good if we can avoid this. Adding new class name is fine, but hopefully it should not be necessary to have new class name for controls because of new logo.

Current styles of `.p-panel__controls` are not ideal actually, so maybe we can improve those to work fine with old and new logo. (Making it work with new logo would be a priority, I guess it's fine if controls may be slightly misaligned with old logo - we likely don't have a lot apps depending on it).

scss/_layouts_application-panels.scss
```diff
@@ -65,6 +113,11 @@
     padding-top: $spv--large;
```
bartaz, line 113: This already seems bit too large (buttons are positioned lower than header text) so maybe reducing it down to `$spv--medium` would help.

package.json
lorumic: Right, sorry about that. Set it to 3.13.0 now.

releases.yml
```diff
@@ -1,3 +1,9 @@
+- version: 3.12.2
```
bartaz, line 1: This version needs to be updated as well:
```suggestion
- version: 3.13.0
```

lorumic: Ow! Well spotted. Fixed that as well.

bartaz:
Looks good on large screens:

image

But on medium/small needs some more adjustments:

image

image

image

@lyubomir-popov we haven't previously updated application layout to use new logo and it's already being used in some apps. Can you have a look and decide how we want new logo to align on different screen sizes and in expanded/collapsed mode?

I guess we want the orange tag centered above side navigation icons (so they align in collapse mode), and the panel pin/close buttons need to better align with logo/headings.

bartaz:
@lyubomir-popov sure - the thing is that the tag height is different on smaller screen sizes - should the panel header padding follow that as well? or should logo not get smaller in application layout?

lyubomir-popov:
@bartaz that's a good point - where ca nI see it with the new logo, and does it ever align with the headings? 
Here for example the headings are underneath so it doesn't seem to be an issue. 
image

image

bartaz:
@lyubomir-popov 

So, here is large screen - tag on the logo is full height, headings need to be aligned:
image

When you go down to medium breakpoint, side nav is collapsed, logo tag is already smaller. Headings are also smaller in font size, but still need to be slightly adjusted:

image

But do we need logo tag to get smaller in this case? (it does that in our current top navigation component on sites)

Another thing is horizontal alignment between tag and icons in side nav. Currently tag is has same padding on left as icons, but is wider. I guess it should be centred, right? The icon inside tag, should align with icons in side nav below?

image

On small screen alignment between logo and headers doesn't matter, but still the top bar seems to high for the size of the logo:

image

Another aspect is alignent of logo/headings and buttons (pin, close) currently these buttons don't align very well. We added them for the sake of example, but I'm not sure how should they align (as buttons have padding around the icon as clickable area making them larger than text in logo/headings)

image


I guess we should address all of these issues?

lyubomir-popov:
@bartaz my comments inline


So, here is large screen - tag on the logo is full height, headings need to be aligned:
image

**yes**

When you go down to medium breakpoint, side nav is collapsed, logo tag is already smaller. Headings are also smaller in font size, but still need to be slightly adjusted:

image

But do we need logo tag to get smaller in this case? (it does that in our current top navigation component on sites)

**I would keep this consistent with the normal navigation. I do have one concern here - the jaas icon in particular is too tiny -  but I think that's a problem with the svg.** 

Another thing is horizontal alignment between tag and icons in side nav. Currently tag is has same padding on left as icons, but is wider. I guess it should be centred, right? The icon inside tag, should align with icons in side nav below?

**I noticed that today while looking at LXD. We need to align it exactly as you are describing.**

image

On small screen alignment between logo and headers doesn't matter, but still the top bar seems to high for the size of the logo:

**Yes, why does it become so tall? It should look exactly like the main navigation at that screen size**

image

Another aspect is alignent of logo/headings and buttons (pin, close) currently these buttons don't align very well. We added them for the sake of example, but I'm not sure how should they align (as buttons have padding around the icon as clickable area making them larger than text in logo/headings)

**This is why I wanted the logo text to be exactly like the h4s - if it were, then properly aligning the close icon with a regular h4 will result in perfect alginment with the logo as well. Reducing the top padding as mentioned previously will get us most of the way there I think. And unless there's text in the button, I would propose using the base button variant that doesn't have outlines or color different from the background. The icon can simply be in the same color as the text, in this case white**

image


I guess we should address all of these issues?

**Yes please**

lorumic:
So, summing up:

1. Use `p-heading--4` for the logo text.
2. Increase the height of the tag on medium/large screens - otherwise aligning it with the headings is impossible.
3. Left-align the icon inside the tag with the icons in the side nav below.
4. Fix (= decrease) the height of the top bar in the app layout examples so that it matches that used for the main navigation.
5. Check that using `p-heading--4` for the logo text also solves the alignment problem with the pin/close icon. (as for the style changes that you proposed Lyubo, there's another PR awaiting review: https://github.com/canonical/vanilla-framework/pull/4694)

Did I forget or misunderstand anything? @bartaz @lyubomir-popov

lyubomir-popov:
@lorumic
1. yes
2. no, decreae the padding top. It is very wrong at the moment - we need to take the existing padding on h4s, which ensures the baseline stays on the baseline grid, and add a value to it - e.g. .5rem. This will get you most of  the way to perfect alignment wiht the logo. Then adjust the logo to finalise the alignment.
3. 
image

4. precisely
5. yes

bartaz:
Side navigation for apps is being redesigned which will likely affect how the logo needs to be displayed. So until it's decided how it is going to look like I'm going to close this one. Sorry 😢

lorumic:
> Side navigation for apps is being redesigned which will likely affect how the logo needs to be displayed. So until it's decided how it is going to look like I'm going to close this one. Sorry 😢

No problem! Thanks for the feedback.

PR #4693, opened 2023-03-09
Fix icon override on the main page

ClementChaumel:
## Done

Fix the social links section at the bottom of the index page where inline styles were causing a bug.

## QA

- Check that the twitter and github logos look ok at the bottom

## Screenshots

I tried to used the default icon size of `1.5rem` which makes the icons bigger than they used to 
image

We could also set it to `1rem` as it was before
image

In both cases the alignment is a bit different. I don't know if it's something we want to fix inline or in the `has-icon` utility.

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -318,10 +318,10 @@
         <h2>Get involved</h2>
         <ul class="p-list--divided">
           <li class="p-list__item">
-            <a href="https://twitter.com/vanillaframewrk" class="p-link--soft u-has-icon"><i class="p-icon--twitter"></i>Twitter&nbsp;&rsaquo;</a>
+            <a href="https://twitter.com/vanillaframewrk" class="p-link--soft">Twitter&nbsp;&rsaquo;</a>
           </li>
           <li class="p-list__item">
-            <a href="https://github.com/canonical/vanilla-framework" class="p-link--soft u-has-icon"><i class="p-icon--github"></i>GitHub&nbsp;&rsaquo;</a>
+            <a href="https://github.com/canonical/vanilla-framework" class="p-link--soft">GitHub&nbsp;&rsaquo;</a>
           </li>
         </ul>
       </div>
--- a/templates/index.html
+++ b/templates/index.html
@@ -317,11 +317,11 @@
       <div class="col-3">
         <h2>Get involved</h2>
         <ul class="p-list--divided">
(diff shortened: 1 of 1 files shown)


bartaz:
It looks more aligned with smaller icons

lyubomir-popov:
The icons are still not aligned, please follow the alignment process I've implemented in this example - https://vanillaframework.io/docs/examples/utilities/font-metrics

ClementChaumel:
> I'm affraid the icons are still not aligned. We've solved this issue before - for example here: https://vanillaframework.io/docs/examples/utilities/font-metrics
> 
> Hopefully this is something you can reuse in this case, and maybe generalise it so it works that way wit hall heading sizes, as in the above example?

Idk, it seems to be aligned in roughly the same way
image

The main difference seems to be that the tick icon is not actually 1rem big but has some built-in padding: 
image

There is a `vertical-align: calc(.5px + .3465em - .5rem)` rule on the tick one but that's just to offset the built-in padding

If we apply that to the Twitter one (or any other one with a perceived height that matches its actual height) then the icon starts to be aligned _under_ the base line.
image


Now if that's something that we want, I can add it to the `u-has-icon` utility but I feel like this nudge is just here to fix the fact that the tick icon is slightly off so maybe we'd be better off fixing that instead?

lyubomir-popov:
@ClementChaumel @bartaz I think the best thing to do is to center any size icon with the x-height (lowercase) letters. We have variables for where the baseline is and how big the x-height is derived from the font itself, [here](https://github.com/canonical/vanilla-
framework/blob/a72ca52561f688deec00091ffd97ff5aaf30dcf0/scss/_settings_font.scss#L22)

I would leave the tick  icon as it is, but maybe update the calculations to be more precise? If the center of the icon is aligned with the vertical center of lowercase letters, any white space within the icon will not make a difference, as it is symmetrically applied top and bottom.

```
// Ubuntu Font metrics
$baseline-position: 0.932em !default; // HHead Ascent from the font file's metrics. Viewable in Fontforge Font Information>OS/2
$cap-height: 0.693em !default;
$x-height: 0.517em !default;
```

bartaz:
@lyubomir-popov OK, let's keep it off this PR. We need to fix home page now, so I'll just removed these icons.

I created https://github.com/canonical/vanilla-framework/issues/4695 for alignment adjustments.

bartaz (approved):

PR #4703, opened 2023-03-16
Layout system proof of concept

lyubomir-popov:
Quick exploration of layouts and some updated typography (far from final). This is just to discuss the layout approach with @bartaz please do not merge.


bartaz:
@lyubomir-popov it's a bit hard to reason about without visual examples - could you throw a few screenshots of what we try to achieve here?

As with the "split layout" component: show couple examples of this in practice, what are the layout variants that we want to cover

lyubomir-popov:
What I also want to do is maybe rule out all other layouts, e.g. 5x7, 3x3x3, etc - and recommend alternatives from the ones above. so for example, a 3x3x3 can become empty/25%/25%/25% etc.

anthonydillon:
We could define row/strip level grids such as 50/50, 25/75, 50/25/25, etc. But leave the current grid there for ultimate flexibility but to be used sparingly.

bartaz:
We need to break this down into separate PRs:

- [ ] typography
- [ ] wider grid
- [ ] new layout components

scss/_settings_breakpoints.scss
```diff
@@ -1,6 +1,6 @@
 // Global breakpoints
 $breakpoint-x-small: 460px !default;
-$breakpoint-small: 620px !default;
+$breakpoint-small: 800px !default;
```
bartaz, line 3: @lyubomir-popov Why is small breakpoint changing?

It makes the range for "tablet" only between 800 and 1036px, it's just 200 pixels. Seems quite small range to maintain.

Now everything below 800 will end up in "mobile" view.

carkod:
@bartaz what's the status the top borders, are we moving forward with this?
image

bartaz:
We have that in backlog, we may have something this or next week (https://warthogs.atlassian.net/browse/WD-2885). But if you need them now, we don't have them yet.

There is an `vf-highlight-bar` mixin in Vanilla that we use internally to create such borders in various components, you can have a look at how it is used in Vanilla and try to reuse that in some custom CSS.

bartaz:
Replaced by #4734 and #4727

PR #4714, opened 2023-03-23
WD-2617 - Add sliding animation to sub navigation on mobile

pedoch:
## Done

- Add styles for the sliding navigation
- Add new example for the sliding navigation

## Fixes/Issues

- [WD-2617](https://warthogs.atlassian.net/browse/WD-2617)

## QA

- Open [demo](https://vanilla-framework-4714.demos.haus/docs/examples)
- Check out the `Navigation / Sliding` under components and make it behaves as expected.
- Check out the other navigation examples as well and make sure they're still working like they used to.

## Screenshots

[WD-2617]: https://warthogs.atlassian.net/browse/WD-2617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scss/_patterns_navigation.scss
```diff
@@ -550,75 +575,81 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
...
-
-      @media (min-width: $breakpoint-navigation-threshold) {
-        right: calc($sph--small + 1px); // 1px for the border on selects. this aligns it with any selects underneath
-        transform: rotate(0deg);
-      }
-    }
-
     &.is-active {
-      &::after {
+      > .p-navigation__link::after {
```
bartaz, line 610: Why are the direct child `>` selectors needed now?

scss/_patterns_navigation.scss
```diff
@@ -651,6 +682,152 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
     }
   }
 
+  // SLIDING NAVIGATION
+  .p-navigation.is-sliding {
+    overflow-x: hidden; // stops overflow scrolling on the page on smaller screens
```
bartaz, line 687: Do we need overflow hidden both here and on p-navigation__nav below?

bartaz (commented):
Couple comments/questions so far.

scss/_patterns_navigation.scss
```diff
@@ -93,6 +93,18 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
...
 
+  %navigation-slide {
+    @include vf-transition(transform);
+
+    height: calc(100vh - $navigation-logo-banner-height);
+    overflow-y: auto;
+    position: absolute;
+    top: $navigation-logo-banner-height;
+    transform: translateX(100vw);
+    width: 100%;
```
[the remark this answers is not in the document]
pedoch: I did this because the body tag on vanilla examples has margins. If I use `100vw` wouldn't look clean because the navigation will be overlapping the space created by the margin.

scss/_patterns_navigation.scss
pedoch: Since there are nested dropdown now, the style should only be applied to the immediate dropdowns and navigation links in the tree

scss/_patterns_navigation.scss
pedoch: So in order for the sliding animation to work properly, the the dropdown has to the visible ie: can not have `display: none`. This then causes the parent elements to overflow. That's why I added it to the parents of the sliding elements.

scss/_patterns_navigation.scss
```diff
@@ -297,17 +309,31 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
...
     list-style: none;
     margin: -1px 0 0 0;
     padding: 0;
     position: relative;
 
-    // remove navigation separator for the first navigation items list
-    &:first-child::before {
-      content: none;
+    // remove navigation separator for the last navigation items in last navigation items list
+    &:last-child {
```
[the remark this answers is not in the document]
pedoch: Hmm okay, I'll try to see if I can make this work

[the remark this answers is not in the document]
pedoch: Upon trying to implement this I realised that the situation for the override is very specific: 
- the last `.p-navigation__items`
  - the last `.p-navigation__item / .p-navigation__item--dropdown-toggle`
    - the last `.p-navigation__link`
  - the last `.p-navigation__item--dropdown-toggle`
    - the last `li`
      - `.p-navigation__dropdown-item`

And that implementing this would require restructuring what I think would be a great chunk of code in the file.

bartaz:
The arrow next to back button is a bit too low, it should be on the same baseline line as the bottom of the text:

image


There is also some glitch with focus when focus moves out of last element in dropdown, the next level "appears", and then disappears when focus returns to dropdown:

nav-focus-wrap

pedoch:
> The arrow next to back button is a bit too low, it should be on the same baseline line as the bottom of the text:
> 
> image
> 
> There is also some glitch with focus when focus moves out of last element in dropdown, the next level "appears", and then disappears when focus returns to dropdown:
> 
> nav-focus-wrap [ nav-focus-wrap ](https://user-images.githubusercontent.com/83575/228562044-5f0e11a9-e212-4725-87b3-59ada4b24208.gif) [ ](https://user-images.githubusercontent.com/83575/228562044-5f0e11a9-e212-4725-87b3-59ada4b24208.gif)

I'll push a fix for the arrow. But as for the focus issue, it happens because the focus moves to the back button of the last drop down before I move it manually to the top. I could hack it and make it move to the first screen first before taking it to the top.

bartaz:
@pedoch It seems it's getting quite complex to get the focus work correctly in this tree... Managing focus manually this way will be very tricky and error-prone.

I looked at the side navigation, and it seems that we fixed it using "display: none" there. We have class `is-drawer-hidden` that sets display none on whole side navigation. And then we have two additional class names `is-drawer-expanded` `is-drawer-collapsed` that handle animation. When we want to show navigation we first remove `is-drawer-hidden` (to remove display block) and then add `is-drawer-expanded` to trigger the animation. And to hide it we listen to `animationend` event.

Here are releavant code blocks:
https://github.com/canonical/vanilla-framework/blob/main/templates/docs/examples/patterns/side-navigation/_toggle_script.js#L118-L126
https://github.com/canonical/vanilla-framework/blob/main/templates/docs/examples/patterns/side-navigation/_toggle_script.js#L139-L142

Maybe something similar could be done with mobile nav dropdowns?
Even if we use `inert` property, we would need to add/remove it when sliding panels in and out, so similar code would be needed (regardless if that's display none or inert), I think.

ClementChaumel:
@bartaz @pedoch oops I updated this branch directly instead of creating a PR targeting it. Let me know if it's ok.
If we're happy with the js export here I think it's fine but if there's more discussion to be had, I'll revert it and open it in its own PR so as to not pollute this one.

As a side, I tried to use rollup to mimic what we do in the cookie policy repo but I couldn't get it working so I opted for esbuild instead.

bartaz:
> @bartaz @pedoch oops I updated this branch directly instead of creating a PR targeting it. Let me know if it's ok. If we're happy with the js export here I think it's fine but if there's more discussion to be had, I'll revert it and open it in its own PR so as to not pollute this one.

I just reverted your commit, it was causing some CI checks fail, and demo didn't run, so make a separate PR and make sure it runs (I guess docker config may need updating to include the built JS file).

> here I add the `slidingNav` function directly to the window object. I think moving forward we'd want to maybe bundle everything under a `vanillaFramework` property to be called like
> 
> ```js
> vanillaFramework.slidingNav('element')
> ```
> 
> As a side, I tried to use rollup to mimic what we do in the cookie policy repo but I couldn't get it working so I opted for esbuild instead.

I think we should do it from the start. Once we settle on JS API and expose it we don't want to change it. So good to make it future-proof from the start.

So something like:

```js
vanillaFramework.setupMobileNavigation();
vf.setupMobileNavigation();
```

We can discuss details in your separate PR I guess.

scss/_patterns_navigation.scss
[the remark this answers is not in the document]
bartaz: If the margin get's in the way we could remove it from this example. We do it in some places by adding `margin: 0` on body.

Like this:
https://github.com/canonical/vanilla-framework/blob/d40b5d1e6d52466666ae4dbe0512c7a0f5bcad74/templates/docs/examples/layouts/sticky-footer.html#L4-L8

scss/_patterns_navigation.scss
```diff
@@ -93,19 +93,49 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
...
+  }
+
   %vf-reset-horizontal-padding {
     padding-left: 0;
     padding-right: 0;
   }
 
   %vf-navigation-separator {
     &::before {
+      bottom: 0;
```
bartaz, line 133: Why do we need to move it to bottom from top?

It seems that this complicates the rules on when to remove this separator (last item of the last list… all that long nested CSS below).

If the border is on top, we can easily hide it from first element if needed.

scss/_patterns_navigation.scss
```diff
@@ -651,6 +695,158 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
...
+    .p-navigation__dropdown--right {
+      @extend %navigation-slide;
+
+      box-shadow: none;
+      display: block;
+      overflow-x: hidden;
+      top: 0;
+
+      // reset navigation dropdown styles on larger screens
+      @media (min-width: $breakpoint-navigation-threshold) {
```
bartaz, line 786: I wonder if we can avoid doing that or make it less repetitive.

Right not it seems that these styles are defined for dropdowns, then overridden by sliding styles, just to be "reset" again to the same original values.

Especially that some of the values we reset are not visibly set here (height, position, transform, width…) - I guess they come from %navigation-slide, but it's not very clear.

I guess it would be good to jump on a call and discuss some not obvious parts.

bartaz (commented):
Couple of comments added. I guess some maybe easier to address, other are more discussion points, so would be good to catch up on those.

bartaz:
Some visual glitches I noticed:

### Opened menu background

I think the "menu" button and dropdown should have slightly different background when open, at least that how it works on the newest example with search:
image
https://vanilla-framework-4714.demos.haus/docs/examples/patterns/navigation/search-dark

Original navigation that worked with CSS only doesn't do it, but I think sliding navigation should, especially that it uses JS and applies `has-menu-open` class name that should be adding this background.

Other issues are related to animation:

### Chevron moving when sliding 

When you click on an item with a dropdown the chevron icon moves (rotates to bottom), but is quickly covered by sliding menu. (GIF only barely catches is when clicked on last element on the bottom).
Chevron icon should not move at all.

chevron-move

### Panels moving when dropdown is open on desktop

I guess real users will rarely change screen size a lot, but if you open a dropdown in desktop view, and then make screen smaller, weird things happen with panels moving in and out:

panels-moving

scss/_patterns_navigation.scss
pedoch: I tried extracting the styles being reset but the nav and dropdown only had about 2 similar styles that could be extracted which means I'd have had to create 2 different mixins for both of them. I don't think all that is necessary so I left a comment stating some of the styles are being reset are from `%navigation-slide`

bartaz: OK, thanks, that's fine.

bartaz:
Thanks for all the updates. Two small issues remaining:

When you open a dropdown (in mobile mode) and resize screen, dropdown gets closed (but menu remains open). But some class names must be incorrect, because you have to click TWICE on the menu item to reopen this dropdown again.


panels-hiding


Second issue is when screen is smaller than menu (for example in mobile landscape mode). When you open dropdown with small number of items you can scroll down and reveal the previous panel.

panels-landscape

pedoch:
@bartaz I've pushed fixes for both the scroll and resize bugs. You can take a look now.

bartaz:
@pedoch Good work on the animation!

Few small issues I noticed:

1. It is possible now to open multiple dropdowns on desktop view, this should not happen, opening one should close the others:

nav-multiple-dropdowns


2. There is some weird issue with disappearing navigation items.

nav-hiding

I think a way to reproduce it is:
- go to mobile view, open the navigation dropdown
- make screen bigger into desktop view
- click on any dropdown
- all navigation disappears

I guess some class names are not removed and mobile styles are applied?

juanruitina:
Discussing [on Jira](https://warthogs.atlassian.net/browse/WD-2617).

templates/docs/examples/patterns/navigation/_script-sliding-search.js
```diff
@@ -0,0 +1,67 @@
+function initNavigationSearch(element) {
```
bartaz, line 1: I haven't looked into details, but how different it is from existing script for search? https://github.com/canonical/vanilla-framework/blob/main/templates/docs/examples/patterns/navigation/_script-search.js

This looks to be share a lot, would it be possible for one script to handle both existing version and sliding version?

ClementChaumel:
@juanruitina @lyubomir-popov would something like that be ok? (I slowed it down to demonstrate better what's happening but it would be quicker ofc)

slidingnav

PR #4717, opened 2023-03-29
Add unified spacing utils

yurii-vasyliev:
## Done

- Added common spacing classes to be used between elements.

Fixes [WD-2884](https://warthogs.atlassian.net/browse/WD-2884)

## QA

- Open [demo](https://vanilla-framework-4714.demos.haus/docs/examples/utilities/vertical-spacing-common)
- Check the new example page
- Review [updated documentation](https://vanilla-framework-4714.demos.haus/docs/utilities/vertical-spacing)

[WD-2884]: https://warthogs.atlassian.net/browse/WD-2884?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scss/_settings_spacing.scss
+++ b/scss/_settings_spacing.scss
@@ -75,8 +75,9 @@ $spv--small: $sp-unit !default;
 $spv--medium: $sp-unit * 1.5 !default;
 $spv--large: $sp-unit * 2;
 $spv--x-large: $sp-unit * 3 !default;
+$spv--strip-shallow: $sp-unit * 3 !default;
 $spv--strip-regular: $sp-unit * 8 !default;
-$spv--strip-deep: $sp-unit * 12 !default;
+$spv--strip-deep: $sp-unit * 12 !default; // will be changed to 16
 
 $sph--x-small: $sp-unit * 0.5 !default; // to be used in place of an inline space between characters/words
 $sph--small: $sp-unit !default;
--- a/scss/_utilities_vertical-spacing.scss
+++ b/scss/_utilities_vertical-spacing.scss
@@ -1,3 +1,11 @@
+@mixin vf-section-spacing-vertical($spacing) {
+  margin-bottom: calc($spacing / 2 - 1px) !important;
+
+  @media (min-width: $breakpoint-large) {
+    margin-bottom: calc($spacing - 1px) !important;
+  }
+}
+
 @mixin vf-u-vertical-spacing($start: -3, $end: 3) {
   %u-vertical-spacer {
     content: '';
@@ -16,5 +24,17 @@
         }
       }
     }
+  }
+
+  .u-sv--shallow {
+    @include vf-section-spacing-vertical($spv--strip-shallow);
+  }
+
+  .u-sv--regular {
+    @include vf-section-spacing-vertical($spv--strip-regular);
+  }
+
+  .u-sv--deep {
+    @include vf-section-spacing-vertical($spv--strip-deep);
   }
 }
--- a/templates/docs/utilities/vertical-spacing.md
+++ b/templates/docs/utilities/vertical-spacing.md
@@ -8,6 +8,28 @@ context:
 
 <hr>
 
+When desired vertical spacing between sections of the content can't be achieved using the default margins and paddings that are built into components, you can the bottom spacing of an element is using the following spacing utilities: `.u-sv--shallow`, `.u-sv--regular` and `.u-sv--deep`.
+
+| Class            | Formula <sup>\*</sup> | Default Value |
+| ---------------- | --------------------- | ------------- |
+| `.u-sv--shallow` | `$sp-unit * 3`        | `1.5rem`      |
+| `.u-sv--regular` | `$sp-unit * 8`        | `4rem`        |
+| `.u-sv--deep`    | `$sp-unit * 12`       | `6rem`        |
+
+<sup>\*</sup> The values are provided for large screens. They go twice smaller on medium and small screens.
+
+<div class="embedded-example"><a href="/docs/examples/utilities/vertical-spacing/" class="js-example">
+View example of the vertical spacing utility
+</a></div>
+
+Use `shallow` gap for spacing under siblings within a section.
+
+Use `regular` gap for spacing under each section, spacing under big headings.
+
+Use `deep` gap for spacing reserved for special cases, like making a large heading / CTA stand out, or at the end of the main content of a page, just before the dark footer.
+
+## More granular spacing adjustments
+
 The `.u-sv` set of utility classes adjusts the space after an element in positive or negative multiples of the baseline grid unit. A baseline grid unit is equivalent to `.5rem`.
 
 For example, to pull the next element `1rem` closer, apply `.u-sv-2`. To push it `.5rem` away from the current element, apply `u-sv1`.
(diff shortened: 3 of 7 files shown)


scss/_settings_spacing.scss
```diff
@@ -75,8 +75,9 @@ $spv--small: $sp-unit !default;
 $spv--medium: $sp-unit * 1.5 !default;
 $spv--large: $sp-unit * 2;
 $spv--x-large: $sp-unit * 3 !default;
-$spv--strip-regular: $sp-unit * 8 !default;
-$spv--strip-deep: $sp-unit * 12 !default;
+$spv--strip-shallow: $sp-unit * 3 !default;
+$spv--strip-regular: $sp-unit * 8 !default; // will be changed to 6
```
bartaz, line 79: I don't think this one is meant to be changed. Regular spacing should remain at 4rem, so value of 8 spacing units is correct.

templates/docs/utilities/vertical-spacing.md
```diff
@@ -8,6 +8,24 @@ context:
...
+<div class="embedded-example"><a href="/docs/examples/utilities/vertical-spacing-common/" class="js-example">
+View example of the vertical spacing utility
+</a></div>
+
+Use `shallow` gap for spacing under siblings within a section.
+
+Use `regular` gap for spacing under each section, spacing under big headings.
+
+Use `deep` gap for spacing reserved for special cases, like making a large heading / CTA stand out, or at the end of the main content of a page, just before the dark footer.
+
```
bartaz, line 28: Would be good to have some subheading before the "old" section of the docs.

Maybe something like "More granular spacing adjustments"?
```suggestion

## More granular spacing adjustments
```

templates/docs/utilities/vertical-spacing.md
```diff
@@ -8,6 +8,24 @@ context:
 
 <hr>
 
+A preferable way to control bottom spacing of an element is using Vanilla pre-defined classes: `.u-sv--shallow`, `.u-sv--regular` and `.u-sv--deep`:
```
bartaz, line 11: The "preferable" way is still using spacing coming from components, so would be good to reword it a bit.

```suggestion
When desired vertical spacing between sections of the content can't be achieved using the default margins and paddings that are built into components, you can the bottom spacing of an element is using the following spacing utilities: `.u-sv--shallow`, `.u-sv--regular` and `.u-sv--deep`.
```

templates/docs/utilities/vertical-spacing.md
```diff
@@ -8,6 +8,24 @@ context:
...
+
+| Class            | Value                                    |
+| ---------------- | ---------------------------------------- |
+| `.u-sv--shallow` | Increases space by 3 baseline grid units |
```
bartaz, line 15: I think it would be more useful to define the "value" in rems rather than spacing units. Spacing units make sense in table below (where the number in class name corresponds to number of units directly)

So I think this table here may follow the Figma file in the structure:
image

So let's have "Class" (same as current), "Height" (expressed as rem), and "Use" column that defines when to use each size. I think we can just use the content from Figma (but edit out the parts that refer to other sections of Figma file).

scss/_utilities_vertical-spacing.scss
```diff
@@ -17,4 +17,42 @@
...
+  .u-sv--regular,
+  .u-sv--deep {
+    &::after {
+      @extend %u-vertical-spacer;
+    }
+  }
+
+  .u-sv--shallow {
+    &::after {
+      margin-top: calc($spv--strip-shallow / 2 - 1px); // subtracting the element height, which is set to 1px above.
```
bartaz, line 31: the "substracting element height…" is explained in first occurence in the file, so maybe we don't need to copy the comment every time?

scss/_utilities_vertical-spacing.scss
```diff
@@ -17,4 +17,42 @@
...
+
+      @media (min-width: $breakpoint-large) {
+        margin-top: calc($spv--strip-shallow - 1px); // subtracting the element height, which is set to 1px above.
+      }
+    }
+  }
+
+  .u-sv--regular {
+    &::after {
```
bartaz, line 40: These sections are quite repetitive (I guess the only change is the variable used), maybe it would be possible to extract it to some mixin and pass variable as a parameter, to avoid copying these sections of the code?

templates/docs/examples/utilities/vertical-spacing-common.html
```diff
@@ -0,0 +1,19 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Vertical common spacing{% endblock %}
```
bartaz, line 2: On a second thought "common" sounds weird in here. It doesn't add much.

I think it would be best to call this one simply "Vertical spacing" (I guess the same for file name).
And the old example we could rename to "Vertical spacing by unit" (`vertical-spacing-unit.html` for the file).
When renaming files the links for embedded examples will need to be updated as well.

bartaz (changes requested):
Looks good, thanks! Mostly some cosmetic comments.

bartaz:
Could you also add an entry in `releases.yaml` file: https://github.com/canonical/vanilla-framework/blob/main/releases.yml ?

It should reference the part that changed: "Vertical spacing", "New" - because we add new class names, and link to the updated docs page.

templates/docs/utilities/vertical-spacing.md
yurii-vasyliev: @bartaz , but the `$sp-unit-ratio` is available for overriding, and if somebody does it, these sizes will be changed automatically. So absolute values in rems might be not accurate.

bartaz: @yurii-vasyliev Good point. I don't think we expect someone to do it, and it may have some breaking consequences if someone does, but still we may take that into account.

On our [spacing documentation](https://vanillaframework.io/docs/settings/spacing-settings) page we mention spacing unit mulitplier and default value, maybe we can do something similar here:

image

So, keep the value expressed by sp-unit, but add a column with default value in rem.

scss/_utilities_vertical-spacing.scss
yurii-vasyliev: I don't think it's possible, because we use exact SCSS vars in our code. I can't run an eval command to figure out a variable name by adding another variable into it.

bartaz: I was thinking about something like:

```
@mixin vf-section-spacing-vertical($spacing) {
    &::after {
      margin-top: calc($spacing / 2 - 1px);

        margin-top: calc($spacing - 1px);
      }
    }
}

.u-sv--regular {
   @include vf-section-spacing-vertical($spv-strip--regular);
}
```

I think something like that should work - you don't have to evaluate variable name, just pass it as a parameter to mixin.

yurii-vasyliev: Should it be a part of `vf-u-vertical-spacing` or create a new file for that mixin so people would have to include it separately?

bartaz: For now we only need it here I think, so just make it part of this file. We can always extract it if we start needing it elsewhere.

bartaz: As it is only needed in this util for now let's keep it here. It can always be extracted to some common file if we need to reuse it elsewhere.

scss/_utilities_vertical-spacing.scss
```diff
@@ -17,4 +17,28 @@
       }
     }
   }
+
+  @mixin vf-section-spacing-vertical($spacing) {
```
bartaz, line 21: Mixins can't be defined within other mixins, so this would need to be extracted to top level in this file (above `@mixin vf-u-vertical-spacing($start: -3, $end: 3)` I guess.

This is causing build and test errors. Locally it should fail too if you run `dotrun build` or `dotrun test`.

bartaz (approved):

PR #4718, opened 2023-03-30
Update releases.yml

yurii-vasyliev:


bartaz:
Sorry if there was some misunderstanding, this should be part of the #4717 I believe, as that's where the change is introduced.

PR #4721, opened 2023-04-03
Add husky and lint-staged to run linters on every commit

yurii-vasyliev:
## Done

Add husky and lint-staged to run linters on every commit

It runs `stylelint` against all `{css,scss}` files and `prettier --check` against `{js,css,scss,html,md,json}` added to the commit.


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged
```
bartaz, line 4: I don't think we should assume people have npx locally to run it.
Project runs via dotrun, so it's not necessary to have local node.

Also I guess I would prefer to run simply the built-in lint commands (like `dotrun test` in case of Vanilla) rather than having a separate config for linting only the staged files.

bartaz:
While I like how this allows to catch errors early, the way postcommit hooks interact with Git GUIs (like UI for commits in VSCode) is not great. It doesn't give enough feedback on why commit failed and user needs to understand where to look for it.

It works great on CLI, but I don't think we can assume everyone uses CLI for git.

In the end this should be developers choice when they want to run linter or tests. (Ideally linting should be done automatically by editor).

PR #4730, opened 2023-04-05
Add sections and blocks components

yurii-vasyliev:
## Done

Added three new classes: `.p-block`, `.p-section` for managing spacing between different types of elements on the page.

Fixes [WD-3019](https://warthogs.atlassian.net/browse/WD-3019)

## QA

- Open [demo](https://vanilla-framework-4730.demos.haus/docs/patterns/section)
- Review updated examples:
  - [blocks](https://vanilla-framework-4730.demos.haus/docs/examples/patterns/section-and-block/block)
  - [sections](https://vanilla-framework-4730.demos.haus/docs/examples/patterns/section-and-block/section)
  - [sections and blocks](https://vanilla-framework-4730.demos.haus/docs/patterns/section)
- Check if brochure site docs and examples use new section/block components:
  - https://vanilla-framework-4730.demos.haus/docs/layouts/brochure
  - https://vanilla-framework-4730.demos.haus/docs/examples/layouts/brochure-site/index
 
[WD-3019]: https://warthogs.atlassian.net/browse/WD-3019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bartaz:
@yurii-vasyliev Could you separate removing your previous utils to separate branch? This one will need design review, so may need more time to be merged, and I don't want to block release on having the utils in main branch. I'd like to be able to release whatever else we have.

templates/docs/patterns/container.md
```diff
@@ -0,0 +1,50 @@
+---
+wrapper_template: '_layouts/docs.html'
+context:
+  title: Containers | Components
```
bartaz, line 4: I think it would be clearer if we call it "Section and block", this way the title will align with component names making it easier to understand and search for.

This would affect the heading on this page, file name (and therefore URL) as well.

And possibly in the code, the mixin instead of `vf-p-container` should be called `vf-p-section` (because there is no `.p-container`). I think it's fine for `vf-p-section` to include both `.p-section` and `.p-block` as they are closely related.

scss/_patterns_section-and-block.scss
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  .p-section {
+    @include vf-container-spacing-vertical($spv--strip-regular);
+  }
+
+  .p-section--deep {
+    @include vf-container-spacing-vertical($spv--strip-deep);
+  }
+}
```
lyubomir-popov, line 21: We need to enumerate all elements that have a padding nudge from the top, and a margin-bottom compensation - that's `p, h6, h5, h4, h3, h2, h1, label, ...` etc - and adjust the values to be the current nudged values + the values above.

lyubomir-popov (changes requested):

scss/_patterns_section-and-block.scss
bartaz: @lyubomir-popov I'm not sure what do you mean, this is supposed to be a wrapper around the content, not a class name you put on heading or paragraph. 

So it's meant to be used as:

```
<section class="p-section">
   <div class="p-block">
      <h2>heading</h2>
      <p>some text…</p>
   </div>
   <div class="p-block">
       <p>another sibling block…</p>
    </div>
</section>
```

And NOT like:

```
<h2 class="p-block">Heading block…</h2>
```

bartaz:
> How are we handling the case where people start aplying this to headings and override the baseline grid alignment?

We are not, they are not supposed to do that. It's a container element, not text element.

If people start to put class names on wrong elements we can't support whatever comes out of it. The best we can do is properly document and provide examples.

By principle you are NEVER supposed to put multiple component class names on single element. So if section/block is a component you don't put this on a heading. But on a wrapper element around the content.

scss/_patterns_section-and-block.scss
lyubomir-popov: oh sorry I thought I saw u-sv--deep

lyubomir-popov:
@bartaz sorry I think I got confused - I think I saw u-sv--shallow, u-sv--deep, etc - that's what I was talking about. Looking at the code now it is p-section, and I fully agree with you about those cases - noone will put that on text. Did the pr change? Or did I see u-sv-sdeep in some other part and leave the comment in the wrong place

bartaz:
From talking with Lyubo it seems that the "deep" sections will always need a top and bottom padding, so should be implemented using existing strips. 

The new section component will only be used instead of strip (or inside the strip) when section follow each other without background change or emphasis.

So the `--deep` variant can be removed.

scss/_patterns_section-and-block.scss
```diff
@@ -0,0 +1,17 @@
+@mixin vf-container-spacing-vertical($spacing) {
+  margin-bottom: calc($spacing / 2) !important;
+
+  @media (min-width: $breakpoint-large) {
+    margin-bottom: calc($spacing) !important;
+  }
```
ClementChaumel, line 6: What is the `!important` for?

ClementChaumel (approved):

ClementChaumel:
I just noticed that the margin in the example is not quite right (I added an outline for clarity)
image

They are spaced more than they should because the `<p>`'s margin-bottom is bigger than the block's
image

lyubomir-popov:
overflow:auto fixes the above ^

@bartaz  Separately, if I wanted to make this section dark or light grey, how would I do it? Asking as the spac eat the bottom is done with a margin, so  that part would not be tinted by any background applied to the p-section or any of its children.

ClementChaumel:
> overflow:auto fixes the above ^
> 
> @bartaz Separately, if I wanted to make this section dark or light grey, how would I do it? Asking as the spac eat the bottom is done with a margin, so that part would not be tinted by any background applied to the p-section or any of its children.

According to the docs you shouldn't do it! :see_no_evil: 
image

lyubomir-popov:
@ClementChaumel ah sorry I forgot. Still think padding-bottom would be more predictable and intuitive given strips use padding-bottom

scss/_patterns_section-and-block.scss
bartaz: I think it was a leftover from utilities that were used before, should be removed.

ClementChaumel:
I thought that maybe instead we could remove the margin bottom from the last element?

I added some more content to visualise it better:

Here it is now with the big spacing: 
image

And here it is with the margin of the last `<p>` removed
image


Which one do you reckon looks best?

lyubomir-popov:
@ClementChaumel I've often thought of doign this but refrained as it might backfire. Let's try it and see what happens. We can always drop it if it  causes issues. Now's the time to experiment with things like this.

bartaz:
> @ClementChaumel I've often thought of doign this but refrained as it might backfire. Let's try it and see what happens. We can always drop it if it causes issues. Now's the time to experiment with things like this.

@ClementChaumel @lyubomir-popov the issue unfortunately is also with how to implement it. We can't just `margin-bottom: 0` to last child, because (unfortunately) it would mess up baseline grid spacing. We would need to apply `u-no-margin--bottom` styles to any last element, not knowing what type it is, etc. I don't see a reliable way to do it right now without rethinking how bottom spacing is applied.

lyubomir-popov:
> the issue unfortunately is also with how to implement it.

let's not do it then. we don't want to make it more complicated.

ClementChaumel:
> @ClementChaumel @lyubomir-popov the issue unfortunately is also with how to implement it. We can't just `margin-bottom: 0` to last child, because (unfortunately) it would mess up baseline grid spacing. We would need to apply `u-no-margin--bottom` styles to any last element, not knowing what type it is, etc. I don't see a reliable way to do it right now without rethinking how bottom spacing is applied.

What do you mean, what happens if we just set it to 0?

PR #4731, opened 2023-04-05
Add rule component (WD-2885)

andesol:
## Done

- Add new component: `p-rule`

## QA

- Open [demo](https://vanilla-framework-4731.demos.haus/docs/patterns/rule)
- Review rule component docs and examples: https://vanilla-framework-4731.demos.haus/docs/patterns/rule


bartaz:
I think in between all the discussions and name changes we missed the scope of this component.

I'd like to end up in a place where we have one component to serve both as "thin" divider line, and as "thicker" divider line.

We discussed "divider", but this one is taken, so is "separator", and by getting back to "highlight-bar" we again reduced the scope of it to be just the thicker one.

My proposal is a "divider line" component (to have some distinction with the "divider" that I hope we will deprecate), which makes it a `p-divider-line`, that by default should look like current `<hr class="u-no-margin--bottom" >`. With an optional variant `p-divider-line--highlighted` that will use the highlight bar thickness.

lyubomir-popov:
@bartaz just checked the typographic naming, "rule" or sometimes "ruler" seems to be the correct name. so p-rule or p-ruler maybe?

andesol:
I renamed it to `p-rule` and `p-rule--highlight`

PS: For reference, bootstrap has a [vertical rule](https://getbootstrap.com/docs/5.1/helpers/vertical-rule/)

bartaz:
> I renamed it to `p-rule` and `p-rule--highlight`

One rule to rule them all!

> PS: For reference, bootstrap has a [vertical rule component](https://getbootstrap.com/docs/5.1/helpers/vertical-rule/)

Good find, thanks!

ClementChaumel:
What is the status on this?

bartaz:
Update examples to include something like that:

image
https://canonical.com/careers/company-culture

ClementChaumel:
@bartaz I also added the grid changes to the brochure site examples on this branch. I know it wasn't part of it initially. Let me know if you'd rather review them separately!

releases.yml
```diff
@@ -3,6 +3,13 @@
     - component: Common grid patterns
       url: /docs/patterns/grid#common-grid-patterns
       status: New
+    - component: Rule
+      url: /docs/patterns/rule
+      status: New
+      notes: We've added a new horizontal bar component.
+    - component: Vertical spacing
```
bartaz, line 10: This again is creeping back. Vertical spacing has been removed from `main` branch already.

bartaz:
There is an issue with `p-rule` that's visible on the test page: http://localhost:8101/docs/examples/layouts/brochure-site/index

It takes the whole width of the screen, while it should be limited to max width of the grid.

Hopefully it should be possible to apply max-width to it by default? I wanted to avoid the need of having additional `is-fixed-width` class names like with current `hr` implementation…

Technically we can put it inside the row, but again, I don't want to make that a requirement.

Hopefully we can make it work that it takes the width of the parent up to maximum of max width of the grid?
(I guess the issue may be the fact that the grid has some paddings)…

releases.yml
ClementChaumel: Damnit!

bartaz:
@ClementChaumel One little thing I noticed - the highlight rule then used "on top" of normal rule appears above it, but by design should appear below it I think.

This is how it looks (https://vanilla-framework-4731.demos.haus/docs/examples/layouts/brochure-site/index):
image


This is how it's meant to look (https://canonical.com/careers/company-culture) [nevermind grid differences]:

image

lyubomir-popov:
coreect shoule be under the line

scss/_patterns_rule.scss
```diff
@@ -0,0 +1,21 @@
...
+    @extend %u-no-margin--bottom--hr;
+    max-width: calc($grid-max-width - 2 * $small-offset);
+
+    @media (min-width: $threshold-4-6-col) {
+      max-width: calc($grid-max-width - 2 * $default-offset);
+    }
+
+    &--highlight {
+      @extend %u-no-margin--bottom--hr;
+      @include vf-highlight-bar($color-dark, bottom, false);
```
bartaz, line 18: @ClementChaumel I guess all we need is change the position here to top (which is default) - but I'm not 100% sure, so please verify.

```suggestion
      @include vf-highlight-bar($color-dark);
```

scss/_patterns_rule.scss
```diff
@@ -0,0 +1,21 @@
...
+  .p-rule {
+    @extend %hr;
+    @extend %u-no-margin--bottom--hr;
+    max-width: calc($grid-max-width - 2 * $small-offset);
+
+    @media (min-width: $threshold-4-6-col) {
+      max-width: calc($grid-max-width - 2 * $default-offset);
+    }
+
+    &--highlight {
```
bartaz, line 16: I think at some point we discussed not using such extrapolations of class names as a best practice, as it make it confusing and harder to find.

Could you move it below `.p-rule` simply as `.p-rule--highlight`?

releases.yml
```diff
@@ -3,7 +3,10 @@
     - component: Common grid patterns
       url: /docs/patterns/grid#common-grid-patterns
       status: New
-      notes: "We've added new classes: <code>.row--50-50</code>, <code>.row--25-75</code>, <code>.row--25-25-50</code>, <code>.row--25-25-25-25</code> as an alias for common grid layouts."
```
bartaz, line 6: @ClementChaumel Well… this should not get removed. I guess we need another fix to this yaml file ;)

scss/_patterns_rule.scss
```diff
@@ -0,0 +1,20 @@
+@import 'settings';
+
+$small-offset: map-get($grid-margin-widths, small);
+$default-offset: map-get($grid-margin-widths, default);
+
+@mixin vf-p-rule {
+  .p-rule {
+    @extend %hr;
+    @extend %u-no-margin--bottom--hr;
+    max-width: calc($grid-max-width - 2 * $small-offset);
```
bartaz, line 10: This is not working well unfortunately.

It's fine on large screens, bigger than max-width:

image

But on a screens smaller than max-width, it still takes the whole width of the screen (because it is smaller than max width), while at the same time grid has padding:

image


This makes it trickier to implement, because we need to take that into account into width, similar to what `is-fixed-width` variant of current `hr` does:

https://github.com/canonical/vanilla-framework/blob/main/scss/_base_hr.scss#L20-L40

But, it needs to happen ONLY if it's outside of the grid containers. If it is inside row. Which `is-fixed-width` also takes into account.

So I guess porting whole `is-fixed-width` functionality into `p-rule` should solve it hopefully?

lyubomir-popov: @bartaz so why not just put that in the hr directly? the code is already there.

also what ius $small-offset? this sounds very cryptic

scss/_patterns_rule.scss
```diff
@@ -0,0 +1,20 @@
+@import 'settings';
+
+$small-offset: map-get($grid-margin-widths, small);
+$default-offset: map-get($grid-margin-widths, default);
```
bartaz, line 4: When defined like this it will pollute global context (variables will be available inside whole vanilla and wherever its imported) which is not expected from such internal aliases.

And also, you seem to need them in one place only in the file, so not really needed I think?
And if we do reuse `is-fixed-width` from `hr` as suggested in other comment they will likely not be needed at all I guess?

scss/_patterns_rule.scss
bartaz: > @bartaz so why not just put that in the hr directly? the code is already there.

I guess you mean using the styles of `is-fixed-width` directly on `<hr>` element by default, without a class name? That could work. My only worry is if someone is using `<hr>` outside of standard grid environment (like application layout). it may not be expected to have max width and auto margins.

This also reminds me… In the new Vanilla layout (that we have on design site and Vanilla), it will also not work out of the box, because it's not a centered standard grid anymore. You would need to put it inside strip or row anyway…

scss/_patterns_rule.scss
ClementChaumel: Ah yes I didn't think of the scope issue! It's just that formulas like `max-width: calc($grid-max-width - 2 * map-get($grid-margin-widths, default));` can get unwieldy and hard to read very quickly!

lyubomir-popov: we just put `.u-no-max-width` in that case

bartaz:
@ClementChaumel Based on the discussions on mattermost today, I guess we need to do it a bit differently than discussed in standup, because we can't remove bottom margin from `hr` without breaking stuff. 

So let's keep the scope of `p-rule` as is currently. So;

- [ ] we don't change anything about existing base `hr` styles
- [ ] `p-rule` is a thin line, works like `hr.u-no-margin--bottom`, doesn't have any max-width (adapts to parent width)
- [ ] `p-rule--highlighted` is a thicker line

This we can release as non-breaking addition in 3.0, and during component refresh before 5.0 we can verify how we can adapt that if needed.

lyubomir-popov:
+1, but please use the text from the brochure guide at the top, as the current one is not very accurate. It's important to mention that rules indicate the beginning of a group of elements, and that they visually tie together elemenents that might feel separate because they are in different columns

bartaz (approved):
LGTM!

PR #4732, opened 2023-04-06
Skip installing Chromium for production on jenkins

bartaz:
## Done

After moving dependencies and installing them all when building docker image, our jenkins builds started to fail.
This should fix that.

## QA

- Open [demo](https://vanilla-framework-4732.demos.haus/)
- Make sure all CI checks pass
- Make sure demo runs


yurii-vasyliev (approved):
LGTM

PR #4735, opened 2023-04-17
Align prefix for stepped lists

ClementChaumel:
## Done

Align stepped lists prefixes

[/docs/examples/patterns/lists/lists-stepped-detailed](https://vanilla-framework-4735.demos.haus/docs/examples/patterns/lists/lists-stepped-detailed)

## Screenshots

image

### DIFF
--- a/scss/_patterns_lists.scss
+++ b/scss/_patterns_lists.scss
@@ -375,7 +375,13 @@ $list-status-icon-height: $default-icon-size;
     $bullet-margin: 0;
     $bullet-margin-mobile: 0;
 
+    // if i is 3 or 4, we need to use only the nudge
     @if $i < 5 {
+      $bullet-margin: map-get($nudges, h#{$i});
+      $bullet-margin-mobile: map-get($nudges, h#{$i}-mobile);
+    }
+
+    @if $i < 3 {
       // To align the bullet with the title we need to use half
       // the height difference between the bullet and the title and then add the
       // nudge to position it with the lower case text.
--- a/scss/_patterns_lists.scss
+++ b/scss/_patterns_lists.scss
@@ -1,7 +1,6 @@
 @import 'settings';
 $list-status-icon-height: $default-icon-size;
-$list-step-bullet-size-offset: 0.25rem;
-$list-step-bullet-margin: $sph--large;
+$list-step-bullet-margin: $sph--x-large;
 
 @mixin vf-p-list-placeholders {
   %numbered-step-container {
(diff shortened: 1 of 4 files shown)


bartaz:
As discussed with Lyubo:

- font size of number same as heading, on same baseline, same font style
- circle size of line height (or .5rem bigger)
- spacing between circle and heading fixed (1.5rem?)

lyubomir-popov:
@bartaz mockups [here](https://www.figma.com/file/H6MSsN3taoXXOJCPUbIImQ/Design-System-Library?node-id=5744%3A12484&t=blA6N4yLsHWIDOhI-1), to confirm circles == line-height + .5rem

ClementChaumel:
Updated: 
0 0 0 0_8101_docs_examples_patterns_lists_lists-stepped-detailed

lyubomir-popov:
I wonder whether we even need the circle. Thjis looks good to me (increased margin-right to 1.5rem):

image

bartaz:
> I wonder whether we even need the circle. Thjis looks good to me (increased margin-right to 1.5rem):

@lyubomir-popov No separator at all? Dot at least for consistency with default ordered list styling?

1. something
2. something

bartaz:
@lyubomir-popov And don't forget about default "stacked" version, also on small screens, where content is below, and needs to be aligned. Circle gives us consistent alignment, numbers will differ when every step has different width:

image

scss/_patterns_lists.scss
```diff
@@ -13,21 +14,20 @@ $list-status-icon-height: $default-icon-size;
...
+    padding-left: calc(map-get($line-heights, default-text) + $list-step-bullet-margin);
+    position: relative;
 
     &::before {
-      align-self: start;
-      background-color: $color-x-light;
-      border: 1px solid $color-x-dark;
-      border-radius: 100%;
-      content: counter(li);
+      aspect-ratio: 1 / 1;
```
ClementChaumel, line 21: This doesn't seem to do anything

ClementChaumel (commented):
LGTM! I can't approve since it's originally my PR but I think it's in a good state!

scss/_patterns_lists.scss
bartaz: Ah, true. I added it before I had to switch to use both width and height.

bartaz:
```suggestion
```

lyubomir-popov:
@bartaz just one thing, can we please make the stepped list example use h2s? This way people who copy-paste from here will get a better, non-bold default. I mostly use h2s for these so chances are it will be the right choice.

lyubomir-popov:
+1 for an easy merge after the above hasbeen adressed

PR #4741, opened 2023-04-20
Fix grid offsets

ClementChaumel:
## Done

Restored the prefix to the original value
explicitely set the target to `.col` in the common grid patterns


ClementChaumel:
> @ClementChaumel small thing - we have this `.grid-demo` styling that makes columns redish to make them visible in example, this stopped working now as I guess it relies on having "grid prefix" in the classname.
> 
> image
> 
> I guess we need to extend the `.grid-demo` to work with `.col` as well.

Ah yes nice catch!

ClementChaumel:
It should be fixed, PTAL @bartaz

scss/_patterns_grid.scss
```diff
@@ -37,7 +37,7 @@
 @mixin vf-p-grid {
   // FIXME: this should be removed from framework SCSS
   // (see https://github.com/canonical/vanilla-framework/issues/3199)
-  .grid-demo [class*='#{$grid-column-prefix}'] {
+  .grid-demo [class*='col'] {
```
bartaz, line 40: This would match ALL class names containing "col" string, this may lead to very unexpected results.

While it's a similar case with `col-` it at least is a bit safer as it has the structure of col- column, but for a name like `.col` we should explicitly only match that class name.


```suggestion
  .grid-demo .col,
  .grid-demo [class*='#{$grid-column-prefix}'] {
```

scss/_patterns_grid.scss
```diff
@@ -232,18 +243,45 @@
...
     @extend %vf-row;
 
-    .#{$grid-column-prefix} {
+    > .col {
       @include vf-grid-column($grid-columns-small);
 
       @media (min-width: $threshold-4-6-col) {
-        &:nth-child(-n + 2) {
-          @include vf-grid-column(calc($grid-columns-medium / 2));
+        &:nth-of-type(-n + 2) {
```
bartaz, line 250: Why changing to `nth-of-type`? This would make it match nth `div` or whatever else element type is used. This does not mean nth element with `.col` class name unfortunately.

So I think `nth-child` is actually safer there, as it's closer to what CSS grid does by default (each child is put into one grid area).

templates/docs/examples/layouts/brochure-site/_25-25-50.html
```diff
@@ -1,16 +1,16 @@
 <section class="p-strip u-no-padding--top">
-  <hr class="is-fixed-width u-no-margin--bottom">
-  <div class="row">
-    <div class="col-3 col-medium-3">
+  <hr class="p-rule">
```
bartaz, line 2: `p-rule` is not part of this branch and doesn't do anything.

bartaz: So for this PR we should revert to previous `hr` with class names in all occurences of `p-rule`, and use it only it PR where rule component is introduced
```suggestion
  <hr class="is-fixed-width u-no-margin--bottom">
```

scss/_patterns_grid.scss
bartaz: So I'd suggest reverting back to `nth-child` in all occurences of this, unless there is some issue with nth-child in this case?

```suggestion
        &:nth-child(-n + 2) {
```

ClementChaumel: But `nth-child` would break if even a `<hr>` was present within the row. I think, even if not perfect this is actually more robust!

bartaz: But you should not put anything that is not a column inside the row I think. Direct child of row should always be a column, and only in column should be anything else.

Even if technically it works right now, it may be unsafe to do it, as row expects column children, not anything else. And if we change the implementation of row at some point, having other children may break.

Anyway, doesn't matter to much, both work the same when markup is as expected, so I guess the question is what should happen if it's not as expected. And indeed `nth-of-type` is more flexible in this regard.

Let's leave it like that.

scss/_patterns_grid.scss
```diff
@@ -190,40 +191,51 @@
   }
 
   // Common grid patterns
+  $col-50-medium: $grid-columns-medium / 2;
```
bartaz, line 194: This throws a warning when building sass, should be wrapped in `calc`
```suggestion
  $col-50-medium: calc($grid-columns-medium / 2);
```

scss/_patterns_grid.scss
```diff
@@ -190,40 +191,51 @@
...
+
+  $col-25: $grid-columns / 4;
```
bartaz, line 196:
```suggestion
  $col-25: calc($grid-columns / 4);
```

scss/_patterns_grid.scss
```diff
@@ -190,40 +191,51 @@
...
+  $col-50: $grid-columns / 2;
```
bartaz, line 197:
```suggestion
  $col-50: calc($grid-columns / 2);
```

scss/_patterns_grid.scss
```diff
@@ -190,40 +191,51 @@
...
+  $col-75: ($grid-columns / 4) * 3;
```
bartaz, line 198:
```suggestion
  $col-75: calc(($grid-columns / 4) * 3);
```

bartaz (commented):
Some SCSS warnings to fix

scss/_patterns_grid.scss
```diff
@@ -190,40 +191,51 @@
...
+        }
+
+        &:nth-of-type(2) {
+          @include vf-grid-column($col-75);
         }
 
-        &:nth-child(2) {
-          @include vf-grid-column(calc($grid-columns / 4 * 3));
+        // if there is only one column we offset it to the right
+        &:only-of-type {
```
bartaz, line 236: Just noticed that, this is pretty cool!

I'm not super sure it's clear that it does that, but before we have any better implementation of the offset we can keep that. I like the simplicity of it, even if not super explicitly clear that this is gonna work this way. Especially that no other do this. But maybe they can?

bartaz (approved):
LGTM, thanks!

PR #4744, opened 2023-04-24
POC: Simplify common grid patterns

ClementChaumel:
## Done

Make the parent define the grid layout and let children fall into place

## QA

- Go there https://vanilla-framework-4744.demos.haus/docs/examples/layouts/brochure-site/index
- Verify that the common grid patterns render correctly

### DIFF
--- a/scss/_base_grid-definitions.scss
+++ b/scss/_base_grid-definitions.scss
@@ -25,7 +25,7 @@
     max-width: $grid-max-width;
   }
 
-  %vf-row {
+  %vf-row-base {
     @extend %fixed-width-container;
 
     & & {
@@ -41,6 +41,23 @@
       margin-right: auto;
       max-width: $grid-max-width;
 
+      // set static gutter width per breakpoint
+      @media (min-width: $threshold-4-6-col) {
+        grid-gap: 0 map-get($grid-gutter-widths, default);
+        grid-template-columns: repeat($grid-columns-medium, minmax(0, 1fr));
+      }
+
+      @media (min-width: $threshold-6-12-col) {
+        grid-gap: 0 map-get($grid-gutter-widths, default);
+        grid-template-columns: repeat($grid-columns, minmax(0, 1fr));
+      }
+    }
+  }
+
+  %vf-row {
+    @extend %vf-row-base;
+
+    @supports (display: grid) {
       & > * {
         grid-column-end: span $grid-columns-small;
       }
@@ -49,20 +66,13 @@
         grid-column-start: auto;
       }
 
-      // set static gutter width per breakpoint
       @media (min-width: $threshold-4-6-col) {
-        grid-gap: 0 map-get($grid-gutter-widths, default);
-        grid-template-columns: repeat($grid-columns-medium, minmax(0, 1fr));
-
         & > * {
           grid-column-end: span $grid-columns-medium;
         }
       }
 
       @media (min-width: $threshold-6-12-col) {
-        grid-gap: 0 map-get($grid-gutter-widths, default);
-        grid-template-columns: repeat($grid-columns, minmax(0, 1fr));
-
         & > * {
           grid-column-end: span $grid-columns;
         }
--- a/scss/_patterns_grid.scss
+++ b/scss/_patterns_grid.scss
@@ -191,29 +191,12 @@
   }
 
   %common-row-variants {
-    @extend %vf-row;
+    @extend %vf-row-base;
     grid-template-columns: 1fr;
-
-    & > * {
-      grid-column-end: span 1;
-    }
 
     // undo line 160 from _base_typography-definitions.scss
     > :not([class*='p-heading--']):not([class*='p-muted-heading']) + p {
       margin-top: 0;
-    }
-
-    // set static gutter width per breakpoint
-    @media (min-width: $threshold-4-6-col) {
-      & > * {
-        grid-column-end: span 1;
-      }
-    }
-
-    @media (min-width: $threshold-6-12-col) {
-      & > * {
-        grid-column-end: span 1;
-      }
     }
   }
 


anthonydillon:
@ClementChaumel can you improve the QA steps here please?

anthonydillon (commented):
Why would we switch from declarative to imperative? Dropping the `col` class means we are applying grid styling to any direct child of the grid parent. Which could lead to unexpected behaviours.

bartaz:
> Why would we switch from declarative to imperative? Dropping the `col` class means we are applying grid styling to any direct child of the grid parent. Which could lead to unexpected behaviours.

@anthonydillon Yes, we wanted to experiment what would that mean for HTML and SCSS code. The declarative approach we normally use causes quite deep nesting of element (layout > strips/sections > rows > cols > content blocks > content), and also having the grid fixed with specific layout of specific number of columns felt like it was an area that we may use to reduce number of wrappers needed.

But looking at the CSS code itself and the examples I do agree it's quite confusing. From the code itself it's very hard to understand what will end up in one column and what in the other. And the grid implementation in CSS with all those `& > *` seems a bit hacky. 

Anyway, that's just experiment, we haven't decided to change it.

ClementChaumel:
The idea was to "Let the parent handle the layout" and rely on the browser to automatically place items within the grid.
That way we don't have to make any assumptions about the nature of the children and we can remove one nesting level by getting rid of `col`

All the 
```css
    & > * {
      grid-column-end: span 1;
    }
```
Are to cancel the rules set in `%vf-row`

```scss
  %vf-row {
    ...

      & > * {
        grid-column-end: span $grid-columns-small;
      }

      ...

      @media (min-width: $threshold-4-6-col) {
        & > * {
          grid-column-end: span $grid-columns-medium;
        }
      }

        & > * {
          grid-column-end: span $grid-columns;
        }
      }
    }
  }
```
It was made under the assumption that the grid would always be 12 columns and children, should, by default, span it all.

I debated whether I should have inherited this `%vf-row`, but I think it's still better than redefining the grid spacings here.
It could be improved by making `%vf-row` slightly smarter.

ClementChaumel:
> And the grid implementation in CSS with all those `& > *` seems a bit hacky.

I updated it @bartaz What do you think now?

bartaz:
> > And the grid implementation in CSS with all those `& > *` seems a bit hacky.
> 
> I updated it @bartaz What do you think now?

It is simpler now, but I guess to fully benefit from simplification we would need to drop the 12 col grid from row, and redefine it with something like `grid-column-template: 1fr 3fr` for 25/75 (of course keeping the correct width and gap we have in current grid).

But it's not so much an issue in CSS, as it is in HTML. 

Having individual elements, without any class name "magically" snap into grid seems a bit confusing and unexpected. Even in simple example like this you need a classless div to wrap couple paragraphs. I think in reality, more often than not we will need such wrapper divs, because there will be rarely a single content element to place in a column.

```html
<div class="row--25-25-50">
  <h4>25/25/50 split</h4>
  <p>Use 25/25/50 split on section with 3 columns where 2 columns have less content than the last large column.</p>
  <div>
    <p>On medium screens it becomes 2 rows of 50/50 + 100 split.</p>
    <p>To achieve that use <code>row--25-25-50</code> class name on the row.</p>
  </div>
</div>
```

Comparing the code above, to current, more explicit version:

```html
<div class="row--25-25-50">
  <div class="col">
      <h4>25/25/50 split</h4>
  </div>
  <div class="col">
      <p>Use 25/25/50 split on section with 3 columns where 2 columns have less content than the last large column.</p>
  </div>
  <div class="col">
    <p>On medium screens it becomes 2 rows of 50/50 + 100 split.</p>
    <p>To achieve that use <code>row--25-25-50</code> class name on the row.</p>
  </div>
</div>
```

While may seem a bit redundant, it is way clearer from the code what is what, it's easier to add/remove content from individual columns, etc.

Yes, the fact that in CSS we rely on `nth-child` or `nth-of-type` is a bit fragile if markup is wrong, but we can't prevent people from creating wrong markup, and we can't assume that just "any" markup will work.
To avoid `nth-of-type` issues hopefully soon we could switch to `nth-of-child(X of Selector)` when it's supported by Firefox and Opera.

bartaz:
> 



> > It is simpler now, but I guess to fully benefit from simplification we would need to drop the 12 col grid from row, and redefine it with something like `grid-column-template: 1fr 3fr` for 25/75 (of course keeping the correct width and gap we have in current grid).
> 
> I don't understand this. It is redefining the grid columns for each common pattern by doing exactly that!

Nevermind, diff view got me confused. Indeed it does exactly that.
 

> Yeah in the examples we need the classless wrappers but in reality, those would be rows of their own or cards or buttons or tables and thus wouldn't need a wrapper.

Yes, in some cases there is one element: one heading, one card, one table, but a quick look at ubuntu.com or others shows that it's no the default. You will usually have multiple paragraphs, mixed with images or button, various combinations. Even nested row should not be that common (as our new grid patterns are meant to serve most common layouts).
 
> My issue with the current approach is that it's a weird mix of the children deciding where they sit and the parent placing its children.

Child `col` is not defining its place or its width. It's the parent defining those. The difference is that currently parent row defines the positioning of it's `col` children, not just ANY children. And while it is a subtle change it is important one from point of view of clarity of the HTML code.

 
> But in my opinion, there's no difference between a classless div (or another component) and a `<div class="col">` as neither tells you in which column nor which width it'll be.

Yes, technically there is no difference. But without this class name it looks like useless div, it doesn't look like part of any pattern, because it doesn't have any class name so it appears like it's not target of any CSS. Adding a `col` class name adds this clarity, makes it clear what is the purpose of this element, and is consistent: both internally (every column has a wrapper with `col` class) and "externally" with legacy grid (`.rol` and `.col-X`). The grid pattern consists of row and column children, there are explicit class names for both. 
 
> We could go in the other direction as well and drop all the `row-x-x` in favour of new `col-25` `col-50` `col-75` shorthands that could be used in `row`

Yes, we could, that would be even more consistent with current row and col-X approach, what we would lose is controlling the layout from the parent, especially in responsive sense.

ClementChaumel:
One issue I found while testing is that it doesn't quite align to the 12 column grid since `fr` units ignore the gutters
image

I'll see if I can find a clean fix it when I get the time!

bartaz:
Closing this for now to clean up PR list.

For now it seems that current approach we have is a bit more reliable. We may look into that further once we see new components used in the wild and rework if needed next cycle.

PR #4749, opened 2023-04-27
POC: Add top margin to images

bartaz:
## Done

Testnig out adding margin to images


bartaz:
This is not a good way to approach this issue. We have it reported separately as https://github.com/canonical/vanilla-framework/issues/4904 to be properly defined and solved.

PR #4760, opened 2023-05-04
Sliding animation wip

ClementChaumel:
## Done

Updated @pedoch 's branch to try and match the animation @juanruitina requested

Don't merge, this is just for the demo

## QA

- reduce the viewport size
- click on nesting element to see the animation


juanruitina:
Demo is broken, but could you try this?

* Use this function for the transition: `cubic-bezier(0,0,.58,1)` (seems to be the ease-out [used on iOS](https://github.com/expo/react-apple-easing/blob/master/AppleEasing.js))
* Make sure the top-level darkens when under the opened sub-level as agreed.

PR #4767, opened 2023-05-10
Add sliding nav on mobile

ClementChaumel:
## Done

Added a sliding animation for nested nav elements

Fixes https://warthogs.atlassian.net/browse/WD-2617

## QA

https://vanilla-framework-4767.demos.haus/docs/examples/patterns/navigation/sliding-search
https://vanilla-framework-4767.demos.haus/docs/examples/patterns/navigation/sliding-light
https://vanilla-framework-4767.demos.haus/docs/examples/patterns/navigation/sliding-dark

- check the demo on large and small screens
- try keyboard navigation too

## Screenshots

slidiiningin


bartaz:
@pedoch Good catch with the bugs.

> * On larger screens I can't close the dropdown by clicking the toggle.

To this I'll add that when opening dropdown it should get highlighted as active and arrow should rotate:

dropdowns

Currently setting `is-active` class on dropdown toggle element seems to do this:
https://vanillaframework.io/docs/examples/patterns/navigation/dropdown-dark

juanruitina:
Can we try with just `ease` for the transform? (I noticed there was nothing special about the bezier curve I shared, it was basically `ease-out` 😅)

I also found one bug: when going back to the top level, the text immediately disappears on tap, instead of sliding together with the background:

Screen Recording 2023-05-12 at 10 33 14

ClementChaumel:
> Can we try with just `ease` for the transform? (I noticed there was nothing special about the bezier curve I shared, it was basically `ease-out` sweat_smile)
> 
> I also found one bug: when going back to the top level, the text immediately disappears on tap, instead of sliding together with the background:
> 
> Screen Recording 2023-05-12 at 10 33 14

Done!

bartaz:
There seems to be some issues when resizing the screen with menu open.

First of all the sliding animation triggers making things look bit weird.
resize-animations


But you can also get to a state where you open nested level on mobile, move to desktop view and the dropdown toggle seems active (arrow and background) but dropdown is not visible. Or when you open dropdown on desktop and switch to small screen menu may open on nested level.

I also ended up in some weird state when mobile menu was shifted a bit off screen without any way to fix it:
image

I think it would be safer and more reliable to just reset the state of menus and dropdowns on resize event. We do the same thing with side navigation for similar reasons. So, when screen is resized simply assume all menus are closed (add/remove any necessary class names to achieve this state, reset any relevant state).

bartaz:
@ClementChaumel That's better, thanks. But there is still an issue when you can get to the state where on mobile parent level (0?) thinks child level is open (is shifted), when there is no child level open.

You can follow these steps:

1. open nav on mobile, go to second level
2. make screen bigger, the dropdown will be open (the same one as on mobile)
3. close dropdown on desktop
4. go back to mobile view
5. menu is open on level 0, but shifted, child dropdown is not open (was closed on desktop)

It seems that there is `is-active` class on level 0 that causes it. Maybe it should be cleared when dropdown is closed on desktop as well?

I don't know if mobile and desktop is handled in the same way in terms of JS and adding/removing class names? It seems there is some issue there. We should make sure that any class names (or other attributes that change the style) should be properly handled when toggling dropdowns on / off (both on desktop and mobile).

nav-bug

bartaz:
@juanruitina @lyubomir-popov Do we need this to support the light theme navigation? I guess all designs were featuring only dark one, and currently not everything adjusts well for light theme:

nav-sliding-light

bartaz:
@ClementChaumel From code point of view how does it relate to Ochuko's #4714 PR? Are you using his code? The commits don't show anything like that, so I'm confused a bit what is new code and what is existing…

juanruitina:
I can't recall any use of the navigation with the light theme, but I see it's the default in the Vanilla documentation. From the UX perspective, I would only make the chevrons dark, and perhaps the overlay in the transition lighter (the shadow effect is a bit excessive with the light theme). Visual might have stronger opinions about the overall look and feel, though. @lyubomir-popov

ClementChaumel:
> @ClementChaumel From code point of view how does it relate to Ochuko's #4714 PR? Are you using his code? The commits don't show anything like that, so I'm confused a bit what is new code and what is existing…

This is a completely new reimplementation of the work, I didn't use anything from @pedoch's branch

ClementChaumel:
> I can't recall any use of the navigation with the light theme, but I see it's the default in the Vanilla documentation. From the UX perspective, I would only make the chevrons dark, and perhaps the overlay in the transition lighter (the shadow effect is a bit excessive with the light theme). Visual might have stronger opinions about the overall look and feel, though. @lyubomir-popov

To be clear, should I implement the light version in this PR?

ClementChaumel:
> @ClementChaumel That's better, thanks. But there is still an issue when you can get to the state where on mobile parent level (0?) thinks child level is open (is shifted), when there is no child level open.
> 
> You can follow these steps:
> 
> 1. open nav on mobile, go to second level
> 2. make screen bigger, the dropdown will be open (the same one as on mobile)
> 3. close dropdown on desktop
> 4. go back to mobile view
> 5. menu is open on level 0, but shifted, child dropdown is not open (was closed on desktop)
> 
> It seems that there is `is-active` class on level 0 that causes it. Maybe it should be cleared when dropdown is closed on desktop as well?
> 
> I don't know if mobile and desktop is handled in the same way in terms of JS and adding/removing class names? It seems there is some issue there. We should make sure that any class names (or other attributes that change the style) should be properly handled when toggling dropdowns on / off (both on desktop and mobile).
> 
> nav-bug

Thanks for the step by step! I managed to fix it, PTAL

templates/docs/examples/patterns/navigation/sliding.html
```diff
@@ -0,0 +1,200 @@
...
+    </nav>
+  </div>
+</header>
+
+<style>
+  body { margin: 0; }
+</style>
+
+<script>
+  document.querySelector('.js-menu-button').addEventListener('click', function(e) {
```
bartaz, line 116: Ideally this should be separated to a script that is included here.

Does this only cover mobile animation, or also desktop dropdowns? It would be good if there is only one script that covers all navigation needs.

scss/_patterns_navigation.scss
```diff
@@ -773,11 +939,13 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
 }
 
 @mixin vf-navigation-dark-theme {
+  $opacified-dark-theme--background-active: #373737;
```
bartaz, line 942: Why can't it be transparent colour?

scss/_patterns_navigation.scss
```diff
@@ -651,6 +651,174 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
     }
   }
 
+  @keyframes fade-in {
```
bartaz, line 654: Would be good to namespace it a bit to avoid potential conflicts, something like `vf-nav-fade-it` ?

scss/_patterns_navigation.scss
```diff
@@ -651,6 +651,174 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
...
+    from {
+      opacity: 0;
+    }
+    to {
+      opacity: 0.5;
+    }
+  }
+
+  .is-sliding {
```
bartaz, line 663: Why is this just `.is-sliding`? Is there a specific element it should be added to?
Is it `.p-navigation`? I guess it would be good to explicitly list it to make it more understandable (`.p-navigation.is-sliding`).

This makes me wonder if it should be `.p-navigation--sliding` (that would be more valid BEM name I think…)

templates/docs/examples/patterns/navigation/sliding.html
ClementChaumel: It does everything! I'll abstract it to its file

scss/_patterns_navigation.scss
ClementChaumel: Because the panels slide on top of eachother for the animation and need to be opaque or we'd still see the lvl1 nav through the lvl2

templates/docs/examples/patterns/navigation/sliding.html
```diff
@@ -0,0 +1,118 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Navigation / Sliding{% endblock %}
+
+{% block standalone_css %}patterns_navigation{% endblock %}
+
+{% block content %}
+<header id="navigation" class="p-navigation is-dark p-navigation--sliding">
```
bartaz, line 7: In Vanilla we don't duplicate component and component+modifier class names, so it would need to be `p-navigation--sliding` only.

But to make it work we need to make sure `p-navigation--sliding` implements all the same styles as `p-navigation` does, so it likely means some additional selectors.

ClementChaumel: And we don't want to use the `&--sliding {` syntax, right?

templates/docs/examples/patterns/navigation/_script-sliding.js
```diff
@@ -0,0 +1,81 @@
+document.querySelector('.js-menu-button').addEventListener('click', function (e) {
```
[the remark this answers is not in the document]
ClementChaumel: > Also, please follow the JS convention of other components of having some kind of single entry `setup` function, like `setupSlidingNav()` maybe? JS should be applied when this function is called, not automatically by including the script.

Sure! I don't understand the first comment though

templates/docs/examples/patterns/navigation/sliding.html
bartaz: Yes, but we are not using that part of BEM apparently 🤷 , it's been like that from beginning of Vanilla. We should not change it now for one component.

So, when you have `p-navigation--sliding` it should not need `p-navigation` class, it should already implement everything `p-navigation` needs. Either by SASS extend, or by `p-navigation, p-navigation--sliding` kind of selectors.

I agree, that it would probably be better to strictly follow BEM in this regard, but it's not a point to change it now.

scss/_patterns_navigation.scss
bartaz: What worries me is the fact that it's hardcoded here, while it depends on transparent version of the colour defined elsewhere. No one will remember to find and change it if dark background colour ever changes.

templates/docs/examples/patterns/navigation/sliding.html
ClementChaumel: How is `.is-dark` different then? Shouldn't it be `p-navigation--dark`?

scss/_patterns_navigation.scss
ClementChaumel: Sure but I couldn't find the opaque colour anywhere and it feels extreme to abstract it just for one instance

templates/docs/examples/patterns/navigation/_script-sliding.js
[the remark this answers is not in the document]
ClementChaumel: It'll have to be modified somehow when moved out too but sure!

templates/docs/examples/patterns/navigation/_script-sliding.js
```diff
@@ -0,0 +1,85 @@
+const initNavigationSliding = () => {
+  document.querySelector('.js-menu-button').addEventListener('click', function (e) {
+    if (window.location.hash === '#navigation') {
```
bartaz, line 3: Have a look at nav example with search: https://vanilla-framework-4767.demos.haus/docs/examples/patterns/navigation/search-dark

Try to base this JS on that example.

bartaz:
Menu open on level 0, and after sliding child levels seem to have different backround colours: 

nav-bkg

bartaz:
Something got broken with switching from small screen to large screen. If you have mobile menu open and increase screen size it gets into a weird view:

image

bartaz:
Couple more issues unfortunately:

Third level of navigation in the example with search seems to have a misplaced border. Maybe an issue with markup?

image

https://vanilla-framework-4767.demos.haus/docs/examples/patterns/navigation/sliding-search

---

If you are on second/third level, close the menu, and open it again it stays on the level it was before. I think it should reset back to first level. When you close menu you don't expect it to remember where you left. By the time you open it again you likely want to see menu from the start, and not need to click "back" multiple times.

nav-close-open

---

It seems that mobile menu is larger than the screen. It can be scrolled even if it doesn't need to:

nav-scroll

---

When menu is smaller than the screen some weird scroll issues happen. There is some background leak when you are on nested level, and it seems there are 2 scrollable areas: one within the menu, and another with top bar being scrolled in/out of the view?

nav-scroll-small

scss/_patterns_navigation.scss
```diff
@@ -245,7 +245,7 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
   .p-navigation__tagged-logo {
     display: flex; // to prevent logo link from expanding full width
     margin-right: 0;
-    min-width: 13rem;
+    min-width: 7rem;
```
bartaz, line 248: There was a reason this was set to 13rem. At least on desktop it should be like that.

If we need to reduce it (on smaller screens), we should still keep it at original size on desktop.

templates/docs/examples/patterns/navigation/sliding-dark.html
```diff
@@ -0,0 +1,118 @@
...
+        </a>
+      </div>
+      <ul class="p-navigation__items">
+        <li class="p-navigation__item">
+          <button class="js-menu-button p-navigation__link">Menu</button>
+        </li>
+      </ul>
+    </div>
+    <nav class="p-navigation__nav">
+      <div class="p-navigation__plane" id="main-nav" aria-label="Example main" data-level="0">
```
bartaz, line 25: Is this "__plane" needed? This seems to barely have any CSS on it. Also attributes seem to be moved from `nav` element (id, aria-label - these should be on `nav` I believe), and is `data-level` used for anything?

bartaz: So I guess it's not an HTML you @ClementChaumel created for this CSS, but HTML from #4714 @pedoch implementation? If so, this needs to be cleaned up, because I guess we don't really know what is the difference between this HTML and current navigation HTML (maybe it's just these `__plane` wrappers), but are they even needed in your implementation?

ClementChaumel: Right, I removed it!

scss/_patterns_navigation.scss
ClementChaumel: Reverted

templates/docs/examples/patterns/navigation/sliding-dark.html
bartaz: Doesn't look removed from here ;) I guess you need to update all of examples.

BTW, not sure how much code these examples share (at least dark/light should be practically the same with one class name different), but it would be worth extracting common parts to some partial.

scss/_patterns_navigation.scss
bartaz: @ClementChaumel Could you address that? I think this can be removed here on smaller screens, if not let me know, but we should not assume the width of logo this way.

templates/docs/examples/patterns/navigation/_script-sliding.js
```diff
@@ -0,0 +1,151 @@
+const navigation = document.querySelector('.p-navigation--sliding');
```
bartaz, line 1: Could all these variables be moved inside the `initNavigationSliding` function? So they don't pollute global scope and are initialised when the function is called.

bartaz (changes requested):
Couple small inline comments in the code, but looks good from QA point of view.

scss/_patterns_navigation.scss
```diff
@@ -651,21 +657,214 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
...
+      transform: translateX(-25vw);
+      @media (min-width: $breakpoint-navigation-threshold) {
+        transform: none;
+      }
+      .p-navigation__dropdown {
+        transform: translateX(25vw);
+        @media (min-width: $breakpoint-navigation-threshold) {
+          transform: none;
+        }
+        &[aria-hidden='false'] {
```
bartaz, line 727: This confuses me a bit. Would that work if `aria-hidden` is not on element? 
It seems it would be more natural to have "standard" (visible) styles as default, and `&[aria-hidden='true']` overriding them. Would that work?

scss/_patterns_navigation.scss
```diff
@@ -651,21 +657,214 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
...
+        }
+      }
+    }
+
+    .p-navigation__items:not(:first-child)::before {
+      content: none;
+    }
+  }
+
+  /* stylelint-disable max-nesting-depth */
```
bartaz, line 713: Is this still needed after you moved things up a bit?

scss/_patterns_navigation.scss
```diff
@@ -651,21 +657,214 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
...
+    .p-navigation__link::after {
+      left: 1rem;
+      transform: rotate(90deg);
+    }
+    @media (min-width: $breakpoint-navigation-threshold) {
+      display: none;
+    }
+  }
+
+  /* stylelint-disable max-nesting-depth */
```
bartaz, line 791: Is this still needed?

scss/_patterns_navigation.scss
ClementChaumel: I guess the idea here is that `aria-hidden=true` is basically the default for every item and then occasionally one of them at a time will be `aria-hidden=false` hence why the override is targetting that

juanruitina:
On the accessibility front, I've spotted a couple of issues with tab navigation:

* Focus becomes invisible sometimes. After opening a deeper level, reverse tabbing (Shift + Tab) makes focus invisible, likely because it goes to the list items on the higher level, it should go instead to "Close menu". Similarly, when reaching the end of the opened level, it disappears again and seems to follow with the following higher level's items, instead of jumping to the first item in the page (i guess the logo).
* Something weird: when opening level 3, going "Back" to level 2, and tabbing again to the first item, then suddenly level 3 becomes visible, but links, or at least the "Back" button, don't seem to work:

https://github.com/canonical/vanilla-framework/assets/1409729/ccf22224-c703-44cb-9425-60cf1a66d383

scss/_patterns_navigation.scss
ClementChaumel: Yes! It still complains if I remove it

templates/docs/patterns/navigation/index.md
```diff
@@ -43,6 +43,15 @@ By default, dropdowns are left-aligned with their parent; if you'd prefer the me
 View example of the dropdown pattern
 </a></div>
 
+## Sliding
+
+You can add nested dropdowns to the navigation by nesting one or more `p-navigation__dropdown` in `p-navigation__item--dropdown-toggle` components.
```
bartaz, line 48: Just mention the `p-navigation--sliding` class as well.

And I guess nesting is optional, right? You can have sliding navigation (new style) even if you don't have nested dropdowns, likely without dropdowns as well (obviously not much of a point using it then)?

templates/docs/patterns/navigation/index.md
```diff
@@ -43,6 +43,15 @@ By default, dropdowns are left-aligned with their parent; if you'd prefer the me
...
+They will be rendered as sliding panels that slide in from the right side of the screen on small screens and as dropdowns on larger screens.
+
+<div class="embedded-example"><a href="/docs/examples/patterns/navigation/sliding-search" class="js-example">
```
bartaz, line 51: We don't usually do this, but it would be good to force the narrow view of the example by putting max-width on it.
And there is a setting to make the height of the iframe bigger as well, so dropdowns are visible when opened.

Something like that should work:


```suggestion
<div class="embedded-example" style="max-width: 40rem"><a href="/docs/examples/patterns/navigation/sliding-search" class="js-example" data-height="600">
```

templates/docs/patterns/navigation/index.md
ClementChaumel: Yeah you're right, not much would be sliding then but it's technically possible!

scss/_patterns_navigation.scss
```diff
@@ -651,21 +657,214 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
...
+          transform: none;
+        }
+        &[aria-hidden='false'] {
+          transform: translateX(-75vw);
+          @media (min-width: $breakpoint-navigation-threshold) {
+            transform: none;
+          }
+          &.is-active {
+            transform: translateX(-100vw);
+            @media (min-width: $breakpoint-navigation-threshold) {
```
bartaz, line 729: I don't necessarily like having many little media queries that all practically do the same. But at least they are right next to the thing they are overriding.

Because of how different mobile and desktop navigations are it almost seems like it should be two separate blocks of style.

Anyway, let's leave it for now I think, but there could be some room of improvement once we make this a default or something…

templates/docs/patterns/navigation/index.md
bartaz: And should be, I think at some point we just want this to be default.

scss/_patterns_navigation.scss
ClementChaumel: I don't know I think it makes sense for the component to render differently on different screen sizes. That way we keep the state on resizes and don't have to hack it with JS

bartaz: Sure, I don't mean to remove the media queries at all. I just wonder if they could be "compressed" a bit.

Like here we have multiple ones that do `transform: none` on the same element. 

It almost seems like single:
`.p-navigation--sliding .p-navigation__items { transform: none }` would do similar (but because of nesting and specificity I guess it would need `!important` or something.

On the other hand, it almost seems like it would be a good place to use `max-width` media query, to target small screens only. But there are other issues with that, so…

Anyway, I'm fine with leaving it as is for now, because there is probably no easy and clean way to fix it.

scss/_patterns_navigation.scss
bartaz: The issue with the `[aria-hidden='false']` is that it applies ONLY when aria-hidden is explicitly set to true. If you have no `aria-hidden` on element at all, it means the same (element is not hidden) but the styles won't apply.

Maybe something like `&[aria-hidden='false'], &:not([aria-hidden])` would work? But it shows that this is double negative.
I don't know if it would complicate things to reverse the logic here, but it may make the selector easier, more clear and less reliant on specific JS implementation.

bartaz: Having both: `&[aria-hidden='false'], &:not([aria-hidden])` seems to work in both cases (with aria-hidden false, and without aria-hidden), so I will be fine with that.

ClementChaumel: Let's go with that for now then and see if there's a better way forward

scss/_patterns_navigation.scss
```diff
@@ -244,7 +245,6 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
   .p-navigation__tagged-logo {
     display: flex; // to prevent logo link from expanding full width
     margin-right: 0;
-    min-width: 13rem;
```
bartaz, line 247: @ClementChaumel I didn't mean to remove it completely. We still need it on desktop. But on mobile we don't need max width I think.

So just wraping it in media query should work?

bartaz (commented):
Code is almost good, but there are couple things:

- [x] bring back the min-width of the logo - at least on desktop
- [ ] bump the allowed test specificity to some number to make them pass (next integer that will pass)
- [x] fix accessibility focus bug
- [x] get UX and design +1

juanruitina:
@ClementChaumel Ubuntu logo and "close menu" should still be in the focus order. It does on the top level, not in deeper ones. Can you make that happen?

I also experience the bug @bartaz describes on level 3.

juanruitina:
It's fixed, looks good to me!

Issue #4790, opened 2023-06-14
Change ticks in ticked list


lyubomir-popov:
Ticks should be the grey ones, not green:
This has been flagged by @marcushas during the rebranding audit.

before:
![image](https://github.com/canonical/vanilla-framework/assets/2741678/cfc94b4c-2cda-4acb-a88b-115c62745912)

after: 
![image](https://github.com/canonical/vanilla-framework/assets/2741678/6fb62ec1-755f-4ea7-8448-f91e73661a4b)

bartaz:
Fixed in https://github.com/canonical/vanilla-framework/pull/4792

PR #4792, opened 2023-06-14
Change ticks in lists to grey

ClementChaumel:
## Done

Changes ticks from green to grey

Fixes WD-4590 https://github.com/canonical/vanilla-framework/issues/4790

## QA

- [ticked lists](https://vanilla-framework-4792.demos.haus/docs/examples/patterns/lists/lists-dividers-ticked) should be grey

## Screenshots
The grey circled cross icon doesn't exist so it looks a bit weird here
image


lyubomir-popov:
I think we need ux here as well. Is  everyone ok to change the red one to the matching red outlined icon? 
image
cc @juanruitina @danielmutis please tag any other UX people who m ight have an opinion.

lyubomir-popov:
We discussed this with @marcushas and agreed to make update circle sizes so they are consistent. Here are all 5 icons that need to be updated: 
help
landscape-action
question
video-play
error

juanruitina:
As long as this applies to lists only and the green tick is available for other uses (e.g. statuses), I'm fine with this. We lose the semantic use of green, but the icon meaning is still clear, the colour change makes it less eye-grabbing (which is probably a good thing so it doesn't compete with CTAs for instance), and I see how it works better with the new visual styles. I don't see any issues accessibility wise either.

I think we should also be a bit mindful of when to use these instead of regular bullets or no bullet, but that falls more on the content side of things. Perhaps we can provide some guidance on this on the content guidelines or templates @sophie-32 and I are putting together for brochure pages.

lyubomir-popov:
@juanruitina yes we agreed to keep green and red filled icons for use in applications and elsehwere, but on brochure sites we've received a lot of feedback against those, so it makes sense to have variants I think.

anthonydillon:
What is the fundamental context difference between green and grey ticks and why would they be exclusive to apps vs sites?

lyubomir-popov:
We have received numerous requests from Brand (@marcushas) to stop using lists with green ticks, so in my mind it is a Brand decision.

On the other hand, when we compared the filled vs stroked versions previously, UX felt the stroked versions are not prominent enough in the context of juju and maas tables. 

To me this seems enough justification to provide both options and document when to use which but happy to discuss further if you disagree.

lyubomir-popov:
@ClementChaumel the ticks sit a bit higher now (maybe due to the fact they fill the 16px bounding box a little more than the filld versions?):
image

juanruitina:
I see the changes apply to the corner case I mentioned, statuses: https://vanilla-framework-4792.demos.haus/docs/examples/patterns/lists/lists-status But this might also apply to a list of features where we have both icons, perhaps because we are comparing two different plans of a product.

When putting the red x next to a gray tick, the red  becomes more prominent than green, and that's generally not desirable. In these cases we should use the green tick. How would we be able to override it cleanly with Vanilla?

lyubomir-popov:
@juanruitina have a look at the code, changing the class will give you what you need. At the end of the day, it's a list and you tell it which versions of the icons you need with classnames on the list items:

image

juanruitina:
>  For missing features, the outlined red [x] icon is fine IMO.

Outlined, but red still. If we have both in the same list or somewhat close, red will immediately become more prominent than gray. I think it will be good to have the possibility of setting the checks as green, and in this case only. One close example is this ["What is Kubernetes" page](https://ubuntu.com/kubernetes/what-is-kubernetes), where on a quick scroll "What Kubernetes is not" becomes more eye-grabbing than "What does Kubernetes do?". Arguably, that shouldn't be the case.

Again, I'm happy with setting gray as the default, but it would be desirable to have the option to override it cleanly.

ClementChaumel:
@juanruitina @lyubomir-popov I'm sorry it's unclear what I should do from here. Can you help me?

lyubomir-popov:
@juanruitina how about this for the usecase you pointed to above: 
image

juanruitina:
Thank you @lyubomir-popov for mocking this up. This is the precise case in which I would like to have available something like an override utility so we can keep the green checks. They should be at least as prominent as the red crosses, and that is not the case when they are gray.

When there are no red crosses on sight (which is most of times), gray is fine to me.

Another alternative though is having _both_ the checks and crosses in gray. Of course semantic colour helps, but having as we will a label ("Advantages", "Disadvantages") _and_ an icon, we could arguably do without colour as we would still have two ways of conveying meaning. UX-wise, this is fine. What we would need to assess in this case is to what extent we are OK with the whole section becoming less eye-grabbing; I think this is more of a visual decision. @lyubomir-popov What do you think?

lyubomir-popov:
@juanruitina we have a long-standing practice to flag only errors and warnings, e.g. in apps. I think the logic applies to websites as well, and I think the above does a good job of differentiating between the two. 

Making everything gray, while more democratic, is less nice IMO. But it is far less of an issue than keepign green and grey ticks, and trying to document how to use which, so if you are adamant you want them equally prominent, then let's go with that. 

image

Please @juanruitina can you give a final yes/no so we can unblock this, as it is blocking the feature branch on c.com

juanruitina:
OK, let's go for having both icons as gray. It's not perfect, but it's not wrong; and we are not using them that often anyway. 

(Just as a note, I disagree on that rationale: these are not errors, but pros vs cons. Given we are marketing products in these pages, we should try to not bring more visibility to negatives over positives as a principle.)

juanruitina:
OK, let's go for having both icons as gray. It's not perfect, but it's not wrong; and we are not using them that often anyway. 

(Just as a note, I disagree on that rationale: these are not errors, but pros vs cons. Given we are marketing products in these pages, we should try to not bring more visibility to negatives over positives as a principle.)

lyubomir-popov:
Makes sense @juanruitina, lets go with grey then @ClementChaumel this is now unblocked.

lyubomir-popov:
@ClementChaumel I can't load this in the demos, are you getting  the same? 
image

Will pull though and review that way.

lyubomir-popov:
Pasting a screenshot as the demo didn't work for me:

image

Looks good, +1 from me. @juanruitina are  you ok with this?

lyubomir-popov (approved):
LGTM.

bartaz:
This needs some documentation.

We added new icon, so it should be added to the icons page, as well as mentioned in "what's new" changelog.

bartaz (approved):
Pushed one commit to complete the docs, I think it looks good now.

Issue #4713, opened 2023-03-23
p-text--small-caps lacks sufficient margin bottom


lyubomir-popov:
It looks like it has had u-no-margin--bottom applied to it. It should have the same margin-bottom as the regular paragraph, as it is the same font-size:

![image](https://user-images.githubusercontent.com/2741678/227145295-4711eb93-7b1d-4b80-a9c2-84a3f9acf9bf.png)

bartaz:
Fixed in #4807

PR #4807, opened 2023-06-25
Update muted heading

lyubomir-popov:
## Done

Increase margin-bottom on muted heading

Fixes: https://github.com/canonical/vanilla-framework/pull/4713

Before: 
image

After:
image

### DIFF
--- a/scss/_base_typography-definitions.scss
+++ b/scss/_base_typography-definitions.scss
@@ -197,10 +197,10 @@
     @extend %default-text;
     font-variant-caps: all-small-caps;
     letter-spacing: 0.05em;
-    margin-bottom: map-get($sp-after, default-text) - map-get($nudges, p);
+    margin-bottom: map-get($sp-after, p-small-caps) - map-get($nudges, p);
 
     &.u-no-margin--bottom {
-      @extend %u-no-margin--bottom--small;
+      @extend %u-no-margin--bottom--default-text;
     }
 
     // deprecated: the use of .u-align-text--x-small-to-default on small caps text is deprecated
--- a/scss/_patterns_muted-heading.scss
+++ b/scss/_patterns_muted-heading.scss
@@ -2,5 +2,11 @@
 @mixin vf-p-muted-heading {
   .p-muted-heading {
     @extend %small-caps-text;
+    
+    margin-bottom: #{map-get($sp-after, p) - .5rem - map-get($nudges, p)};
+    
+    &.u-no-margin--bottom {
+      @extend %u-no-margin--bottom--default-text;
+    }
   }
 }
--- a/scss/_patterns_muted-heading.scss
+++ b/scss/_patterns_muted-heading.scss
@@ -2,9 +2,7 @@
 @mixin vf-p-muted-heading {
   .p-muted-heading {
     @extend %small-caps-text;
-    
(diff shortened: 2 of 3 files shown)


bartaz:
@lyubomir-popov you seem to have tons of unrelated commits here - is this supposed to be part of 4.0, or regular release?

lyubomir-popov:
sorry I think I did it on an older version of main, I should probably rebase.

lyubomir-popov:
@bartaz I had pointed this pr at main not 4.0, updated now. It's just one tiny commit.

scss/_patterns_muted-heading.scss
```diff
@@ -2,5 +2,11 @@
 @mixin vf-p-muted-heading {
   .p-muted-heading {
     @extend %small-caps-text;
+    
+    margin-bottom: #{map-get($sp-after, p) - .5rem - map-get($nudges, p)};
```
bartaz, line 6: @lyubomir-popov I think you can use calc here, so SASS can be smart enough to calculate what is possible on build time.

```suggestion
    margin-bottom: calc(map-get($sp-after, p) - .5rem - map-get($nudges, p));
```

But also, why is this so complicated? Weren't nudges supposed to be part of top padding to keep things on baseline grid? Do we use them for bottom margins as well? If so, why aren't they just part of `map-get($sp-after, p)`? If we have specific variable for spacing after paragraph, why do we need to remove nudge from it?

With calculation like this is practically impossible to maintain this. No one in future will know what is the expected margin supposed to be and how to modify it. If we need any calculation here, it needs to be expressed in comment what is it calculating.

When we introduced `p-text--small-caps` it was supposed to be replacing p-muted-heading, now muted heading is using same text style as small caps text, but has different margin. Is this expected? Will designers / devs know which one is which?

lyubomir-popov:
Those are very good points, thanks @bartaz. It's been a while since I contributed,  Iguess I'm a bit rusty. 
I've reworked it now to be exactly as you proposed.

What this is trying to dfo is to get the correct amount of space under a .p-text--small-caps (I hard forgotten about that one, I've now reworked it to target both this and the deprecated p-muted-heading). This is what it should look like: 
image

bartaz (approved):

scss/_base_typography-definitions.scss
```diff
@@ -197,10 +197,10 @@
     @extend %default-text;
     font-variant-caps: all-small-caps;
     letter-spacing: 0.05em;
-    margin-bottom: map-get($sp-after, default-text) - map-get($nudges, p);
+    margin-bottom: map-get($sp-after, p-small-caps) - map-get($nudges, p);
 
     &.u-no-margin--bottom {
-      @extend %u-no-margin--bottom--small;
+      @extend %u-no-margin--bottom--default-text;
```
anthonydillon, line 203: At a glance, I think WTH at this? I understand `--small` might mean no margin bottom and small screen but now we have no margin only on ... Default text?

lyubomir-popov: that’s just fixinga bug. with the small caps in the new variable font, muted heading now uses the same size as p, h5 and h6. to avoid having all those in the nudges and space-after maps, they are grouped under the key default-text. So since the font size of the muted heading changed, we’re updating the calculation to use the appropriate map keys.

PR #4814, opened 2023-06-29
Replace section/block components with container

### LINKED ISSUE #4809: P-section/p-block: improve documentation
The current use cases of p-section and p-block do not cover all cases where they are needed. 
[Here ](https://github.com/canonical/canonical.com/pull/910/files/8113612e4d8e30d3644810c1eca3d714fc76f77a#r1241966721) for example, @mtruj013 rightly pointed out that nesting p-sections is wrong according to the documentation. 

Another related issue - needing a p-section depending on context (volume of content in adjacent elements): https://github.com/canonical/vanilla-framework/issues/4797

I think the intuition is that a 64px margin-bottom is helpful when an element is followed by a large block of content which thus competes for attention with the current element. We need to add a few examples to clearly explain this in the documentation.

bartaz:
## Done

Deprecates section/block components and replaces them with container component with regular/shallow/deep spacing, as this better reflects the usage in designs.

Fixes #4809

## QA

- Open [demo](https://vanilla-framework-4814.demos.haus/docs/patterns/container)
- Review updated documentation and examples:
  - https://vanilla-framework-4814.demos.haus/docs/patterns/container
  
 
## Screenshots
image


ClementChaumel:
Are they meant to be used on the same element as rows or nesting them instead? because on the c.com refresh I've seen a lot of `<div class="row p-section">`

bartaz:
> Are they meant to be used on the same element as rows or nesting them instead? because on the c.com refresh I've seen a lot of `<div class="row p-section">`

I know it sometimes works, but by component definition they shouldn't. This may lead to issues when both define same CSS props.

ClementChaumel:
I know you told me that @anthonydillon wasn't in favour of utility component but this is basically one. All we need it for is the bottom margin
image


Most of the time the feedback I get from design is like "we need a large space between those elements"
Your proposed solution is something like:
```html
<div class="container--large">
  <p>some content</p>
</div>
<p>some other content</p>
```

But I don't find it untuitive when the ask is to put space between those elements.

We could think of a pure spacing component like 
```html
<p>some content</p>
<div class="spacer--large"></div>
<p>some other content</p>
```
It would have the added benefit of keeping a flatter DOM tree which is ultimately better for performances.
Hell, since the container is only applying a margin it is already just that!
```html
<p>some content</p>
<div class="container--large"></div>
<p>some other content</p>
```

Another idea we could explore if we wanted to avoid having some utility html components could be to just have utility css classes like so

```html
<p class="spacing-bottom--large">some content</p>
<p>some other content</p>
```

bartaz:
We discussed it with @lyubomir-popov and on Figma level they are also containers with margin, not just blank spacers. This will hopefully be better reflected in dev view in Figma soon, so it should be clearer for devs when looking at Figma designs to see that part of content is wrapped in a container.

What we need is an easy translation from Figma to code. So if Figma provides container with margin, this should be what is implemented. I know it's not ideal and we are about to explore how it works in practice, but it should give us better consistency between Figma and code for now.

bartaz:
Layout components (grid, strips, containers, etc) don't do anything about their children, but define the layout of the page: where the children should go, what width they have, what spacing is between them.

ClementChaumel:
Strips can define background colours, grids define options for children to position. They all do a bit more than just that. and I find it confusing that it's just called container and not something that hints at it applying bottom margin

What do you think about the utility css class instead?

Again, the feedback from designers has been to add spacing in arbitrary places. I think trying to tie to to a container is the wrong abstraction here

bartaz:
> What do you think about the utility css class instead?

The issue with utilities is that they should be the last resort. If something is not covered by other means. With these spacing needs it doesn't seem like it's gonna be an exception. It's a standard thing that will need to be applied probably to all designs.

Having a container with bottom margin is something we need. The sections of the page are just that (a wrapper for a part of the page, usually with grid inside, and some content about specific topic). The issue is that  with addition of utilities, we will have 2 separate things to do the same - add bottom margin of specific size.

To add on top of that "utilities" are not something that is easy to reflect in Figma. The way it's currently being done in Figma is a wrapper with margin, so it translates well to this kind of component.

Maybe it's mostly an issue of naming? 
I do believe that more "abstract" naming is better than current artificial section/block semantics (that don't translate well to designs). Maybe we can use something else than "container"? We can try to make it more specifically about spacing?

- `p-spacer` `p-spacer--shallow` `p-spacer--deep`
- `p-margin-box` …
- `p-bottom-spacing` … 

I don't know which would make most sense… The last two feel more like names for utils, not components, so I would lean to the spacer (although it gives me a bad memories of spacer.gif 😨 )

lyubomir-popov:
Do we ned to get into a drawn argument on this? We had a decision which is simple and intuitive with `.p-container, `.p-contianer--shallow`, `.p-container--deep`. It implies it is something that groups its contents, it is easy to implement (no need to get into nudge calculations, as would be the case if we do the most semantic thing - provide variations to all headings, and then try to apply the same through utilities (whether they start with "u" or with "p"))...

ClementChaumel:
> Do we need to get into a drawn out argument on this? We had a decision which is simple and intuitive with `.p-container`, `.p-contianer--shallow`, `.p-container--deep`.
> 
> * It draws on people's knowledge of strips, with the only difference being that `.p-container` only adds space at the bottom.

I'd be careful in calling something intuitive when it might be only to you. For example I always found the "shallow / deep" naming to be weird. Why are we talking about depth when we mean height?


> * It implies it is something that groups its contents by visuall distancing them from adjacent elements

In a lot of places sections and block only wrap a single element. So it's not about grouping children it's about adding spacing in some places

> * it is easy to implement (no need to get into nudge calculations, as would be the case if we do the most semantic thing - provide variations to all headings, and then try to apply the same through utilities (whether they start with "u" or with "p")) for non-text elements...

I don't think the ease of implementation should be the first concern when we're talking about a design system that is going to be widely used and maintained for a long time.
Plus a spacer component would be the same exact implementation, the only thing different about it would be the way it's used.


> * and as @bartaz said, it is very easy to annotate in figma so looking at the layers tree gives you a clear outline of the markup

I'm sure Figma can handle and display easily any component and variations we throw at it. Again we shouldn't create anti-patterns because it's easier to draw in Figma.

My main concern still stands, this is a spacing utility and not a container and documenting it as such make things unclear

bartaz:
> I'm sure Figma can handle and display easily any component and variations we throw at it. Again we shouldn't create anti-patterns because it's easier to draw in Figma.

Sorry @ClementChaumel, but I'm not really sure, how we got from renaming `p-section` to `p-container` and `p-block` to `p-container--shallow` to "crating an anti-pattern". You've been in a meeting when we initially talked about this and you didn't have that concerns back then.

It's not just about getting into ideal state. It's about being pragmatic and working with what we have as well. If in Figma we have a wrapper with margin, why not having it in Vanilla as well (whatever we call it). Making it easier in Figma is as important as making it easier in any other place. Design system needs to serve designers as much as developers who build with it.


> My main concern still stands, this is a spacing utility and not a container and documenting it as such make things unclear

And I get this concern. But I don't see any solution that would not have downsides. So we can try to take what we already have (section/blocks) and make them a bit less confusing (which I hope container does). Or we can drop the whole thing and get back to the drawing board. 

But section and blocks are already out there, if we don't change them now, they are going to be released in 4.0. We are not stopping 4.0 release to have spacing discussion.

We still can have them, we can bring this up in Vanilla working group to decide both from design and code perspective how spacing should work - is a function of a container, is it a spacer component placed between elements or something else.

But I'm not sure if blocking this particular PR to have this discussion is valuable.

If we can't get to agree on that, I'm fine with simply updating docs to allow nesting p-section inside each other (because that's what started this whole issue). But this in my opinion is a worse solution than renaming it to container (or whatever we want to call it).

lyubomir-popov:
I second everything @bartaz said and I wish we could put all this energy and passion into c.com right now, where it is very much needed :) We can always reopen this discussion and later.

ClementChaumel:
But the issue here is not renaming sections.


`p-blocks` and `p-sections` were created under the assumption that they were semantic blocks that we would use.
They would describe some sort of hierarchy where sections would contain blocks and would be used in juxtaposition. They would also have rules about what background should be used on them.


Then when came the time to actually use them, @lyubomir-popov kept nesting them and saying things like "let's use the section margin under this title" 
But I think we can all agree that wanting a bit more spacing between a title and a paragraph doesn't warrant declaring that the title is a section on its own.


So the solution has been to strip out most of what the section and blocks were responsible for and only keep its **function** of adding some spacing after the last element it wraps.


From here we noticed that it was weird to have them still as sections and blocks but it is not in my opinion just a naming issue.

What the real issue is, and why it struck us all as weird to keep them with their current name, is that we realised that the components we assumed we wanted (sections and blocks) aren't what we actually needed when building pages (a spacing utility without any semantics attached).

I think the goal of this PR is to actually figure out how things are used "in the wild" and build accordingly.


If you really want to use a wrapper it should wrap 2 elements and declare the gap between them so 
image

Because stripped of all meaning, the relationship of this whitespace is as much to the content at the top of it as it is to the content at the bottom of it.


You say 
> But I'm not sure if blocking this particular PR to have this discussion is valuable.

But I feel like this is our chance to get it right otherwise, down the line you'll say

>But ~section and blocks~ containers are already out there

And it'll be impossible to fix then.

bartaz:
@ClementChaumel While I think I agree in principle, that the container approach isn't perfect, we do need something to work with.

From this discussion it seems that we kind of don't have a clue how we should do the spacing on the brochure sites. Currently proposed section/blocks component don't work as expected and to make use of them we have to break our own documented rules (of not nesting sections).

But without them we are back in the world of using only strips, nesting them, and throwing u-no-padding--top on them.

And we have a 4.0 release on Monday.

If we have no other solution that we can agree on, we release 4.0 with section and blocks and sooner or later people will be using them. We realistically have no time to redefine it, because we are already knees deep in building sites in new style. It will not slow down to wait for our spacing conclusions.

We can, and probably should discuss it via Vanilla working group, but it will take likely weeks to find a solution that works both on design and code level, and may end up being something completely new and potentially breaking change.

But we need something NOW. For 4.0, for canonical.com, for microk4s that comes next.

It's not true that it will be impossible to fix if we release section/blocks or containers now. We have tons of legacy pages anyway that will have to be updated. If we come up with a better solution then we will propose a better solution, deprecate the old, and show teams how to migrate. We already know that 5.0 will be a hugely breaking release, as we plan to remove any legacy stuff from Vanilla, while 4.0 is to support the transition.

So, is there any better proposal that we can make happen on Monday? If not, fine. It's not like we can't change or add stuff later.

ClementChaumel:
> So, is there any better proposal that we can make happen on Monday?

I proposed a few alternatives that would, in my opinion, make more sense from a technical and semantic point of view. I don't feel like you've addressed my points yet

I'd just be careful with thinking
> If not, fine. It's not like we can't change or add stuff later.

because historically we haven't

bartaz:
Superseded by #4823

PR #4823, opened 2023-07-05
Update section component

bartaz:
## Done

Updated section component with shallow and deep version, deprecate block.

## QA

- Open [demo](https://vanilla-framework-4823.demos.haus/docs/patterns/section)
- Review updated section examples and docs


ClementChaumel (approved):
LGTM, is there any plan to revisit the way typography spacing is done so we can also have utilities?

Issue #4818, opened 2023-07-03
P-breadcrumbs: add margin-bottom


lyubomir-popov:
## Pattern to amend

`p-breadcrumbs`

## Visual

Currently, the breadcrumbs have no margin bottom, so the element that comes after sits too close to them:
![image](https://github.com/canonical/vanilla-framework/assets/2741678/8bee61a1-8bfe-464a-9cc6-322f6bd1ede2)

I propose adding .5rem at the bottom, to improve the way it looks:
![image](https://github.com/canonical/vanilla-framework/assets/2741678/0cabc00a-d6d4-4a90-8747-fbda351510cf)

bartaz:
Currently breadcrumbs seem to have 1.5rem margin (indirectly from a `<ol>` list inside of them). Which seems like a bug I guess.

<img width="549" alt="image" src="https://github.com/canonical/vanilla-framework/assets/83575/cf3eb2b7-bbc3-4da6-8f3b-4655e703bc84">


So we want breadcrumbs to ignore default list margins, and only have 0.5rem margin by themselves?

ClementChaumel:
Actually, the issue is a side effect from using `p-text--small-caps` on the `p-breadcrumbs__items` reducing its bottom margin to .5rem instead of 1.5rem. A change that would make more sense here is to use the `p-text--small-caps` text style by default on the breadcrumbs so there's no need to use the two clashing classes

PR #4842, opened 2023-07-24
Update breadcrumb's bottom margin

ClementChaumel:
## Done

Reduce margin-bottom on breadcrumbs to 1rem
Make the breadcrumbs use small caps

Fixes https://github.com/canonical/vanilla-framework/issues/4818

## QA

- Open [demo](https://vanilla-framework-4842.demos.haus/docs/examples/patterns/breadcrumbs)
- See that the bottom margin is 1rem
- See that the breadcrumbs use small caps

### DIFF
--- a/scss/_patterns_breadcrumbs.scss
+++ b/scss/_patterns_breadcrumbs.scss
@@ -10,7 +10,7 @@
   }
 
   .p-breadcrumbs__items {
-    margin-bottom: $spv--large;
+    margin-bottom: 0;
     margin-left: 0;
     padding-left: 0;
   }
@@ -19,7 +19,6 @@
     @extend %small-caps-text;
 
     display: inline-block;
-    margin-bottom: $spv-nudge-compensation;
 
     &:not(:first-of-type) {
       text-indent: $sph--large;


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vanilla-framework",
-  "version": "4.1.0",
+  "version": "4.1.1",
```
bartaz, line 3: We haven't released 4.1 yet, so this is unnecessary

scss/_patterns_breadcrumbs.scss
```diff
@@ -7,11 +7,13 @@
...
     padding: 0;
     width: 100%;
+  }
 
-    &__items {
-      margin-left: 0;
-      padding-left: 0;
-    }
+  .p-breadcrumbs__items {
+    @extend %small-caps-text;
```
bartaz, line 13: I think it would feel more natural for text styles to be moved on text element `p-breadcrumbs__item` instead. Especially that currently `p-breadcrumbs__item` in line 20 extends `%default-text` which conflicts in some styles with `%small-caps-text`, so it would be clearer if `__item` styles only extend `%small-caps-text` and nothing else.

bartaz:
Comparing to canonical.com/careers (which I think Lyubo wants to match) we still have too much of the margin I think:

image

scss/_patterns_breadcrumbs.scss
```diff
@@ -7,15 +7,16 @@
...
+    margin-left: 0;
+    padding-left: 0;
   }
 
   .p-breadcrumbs__item {
-    @extend %default-text;
+    @extend %small-caps-text;
 
     display: inline-block;
     margin-bottom: $spv-nudge-compensation;
```
bartaz, line 22: I think this should be removed, this is some old margin, and we want the margin from small caps text.

scss/_patterns_breadcrumbs.scss
```diff
@@ -7,15 +7,16 @@
...
     padding: 0;
     width: 100%;
+  }
 
-    &__items {
-      margin-left: 0;
-      padding-left: 0;
-    }
+  .p-breadcrumbs__items {
+    margin-bottom: $spv--large;
```
bartaz, line 13: I'm not sure if this is needed. If we have margin from small caps text on the `__item` this probably should not be needed.

ClementChaumel: The margin from small caps is overriden by `    margin-bottom: $spv-nudge-compensation;` which I'm reluctant to touch because it sounds like some baseline grid magic

ClementChaumel: Ah sorry I was confused what your previous comment referred to!

bartaz: Sure, I guess it used to be a margin to align with baseline, but because we are replacing margin completely, changing font text style to something else etc, old styles need to be removed, so we can start with clean state, and then see if they baseline needs adjustments.

But if we use defaults from small caps we should not need anything else (as that should be aligned already).

scss/_patterns_breadcrumbs.scss
```diff
@@ -7,18 +7,18 @@
     margin: 0;
     padding: 0;
     width: 100%;
+  }
 
-    &__items {
-      margin-left: 0;
-      padding-left: 0;
-    }
+  .p-breadcrumbs__items {
```
bartaz, line 12: Thanks for expanding the classname!

bartaz (approved):
Splendid, thanks!

bartaz (approved):

Issue #4748, opened 2023-04-26
Top border visible in split lists


bartaz:
**Describe the bug**

We recently removed top border from lists, this causes issues in split variant of list that displays in 2 columns:

<img width="1126" alt="image" src="https://user-images.githubusercontent.com/83575/234553102-3f0894f0-449e-4064-a808-f36eeaa91799.png">

Top element on second column has top border, because it's not the first element in the list.
I don't know if there is a way in CSS to know if element is first of given column to fix that…

I think this was not released yet (part of 3.14).

bartaz:
Triage: effort high (may not be possible in straight CSS), impact medium/low. To be considered as part of lists in new architecture.

PR #4856, opened 2023-08-01
Fix top border visible in split lists

lemacarl:
## Done

Set border box shadow to bottom in split lists

Fixes [#4748](https://github.com/canonical/vanilla-framework/issues/4748)

## QA

- Open [demo](http://0.0.0.0:8101/docs/patterns/lists#split)
- Review split list border 

## Screenshots
Screenshot from 2023-08-01 17-38-03

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vanilla-framework",
-  "version": "4.1.1",
+  "version": "4.1.2",
   "author": {
     "email": "webteam@canonical.com",
     "name": "Canonical Webteam"
--- a/scss/_patterns_lists.scss
+++ b/scss/_patterns_lists.scss
@@ -111,17 +111,17 @@ $list-step-bullet-margin: $sph--x-large;
 }
 
 @mixin vf-list-item-divided {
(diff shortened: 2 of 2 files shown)


bartaz:
Thanks for trying this one @lemacarl, but I believe this just turns this issue into an opposite one. 

Now, instead of having unnecessary border top in second column, we have unnecessary border bottom in first. We can remove the bottom border from last child, but with just CSS we will not be able to know which element to remove the border from in first column.

lemacarl:
Gotcha @bartaz. Any ideas how this can be done outside css?

bartaz:
> Gotcha @bartaz. Any ideas how this can be done outside CSS?

Not really, that's kind of why this stays in backlog as we haven't figured out a straightforward solution. Anything that would work feels like a workaround (splitting the list into 2 lists in 2 grid columns, or adding some JavaScript to detect which items are in new column, etc.). We don't want to over-engineer the solution, especially that the bug is not really a high priority.

Thanks for trying, but I think we can leave that for now until we have better idea what to do.

lemacarl:
Sure thing @bartaz

PR #4874, opened 2023-09-12
Update publish-on-release.yml

samhotep:
## Done

- Granted Github Actions the write permission to push to `release` branch
- Added `release` ref to the checkout action to set upstream to the release branch automatically

## QA
Real QA can only be run after merging, but here's a demo of the job:
https://github.com/samhotep/codespaces-actions-playground/actions/runs/6162324000/workflow

### DIFF
--- a/.github/workflows/publish-on-release.yml
+++ b/.github/workflows/publish-on-release.yml
@@ -77,4 +77,4 @@ jobs:
           fetch-depth: 0
       - name: Push to release branch
         run: |
-          git push --force
+          git push --force origin release
--- a/.github/workflows/publish-on-release.yml
+++ b/.github/workflows/publish-on-release.yml
@@ -4,6 +4,9 @@ on:
   release:
     types: [published]
 
+permissions:
+  contents: write
+  
 jobs:
   build:
     name: Build Vanilla
@@ -67,10 +70,11 @@ jobs:
 
   publish-website:
     name: Publish to the production website
-    needs: [publish-assets]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: release
       - name: Push to release branch  
         run: |
-          git push --force origin release
+          git push --force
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-on-release.yml
```diff
@@ -67,10 +70,11 @@ jobs:
 
   publish-website:
     name: Publish to the production website
-    needs: [publish-assets]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: release
```
bartaz, line 77: @samhotep As far as I understand it would checkout `release` branch, and push back to it.

What we need is checkout of `main` branch, and force pushing its contents to `release` branch.

bartaz (changes requested):

.github/workflows/publish-on-release.yml
bartaz: I think the issue we have comes from the fact that checkout action only fetches a single commit by default (https://github.com/actions/checkout) so it doesn't have a branch to push.

There are probably 2 ways of working around that:

Fetching whole history with all branches: https://github.com/actions/checkout#Fetch-all-history-for-all-tags-and-branches

Or specifying that we want to force push the commit we are on `git push --force upstream HEAD:release`

samhotep: I see..
So I've been playing around with both options, and the [first one](https://github.com/samhotep/codespaces-actions-playground/actions/runs/6171992084/workflow) seems much easier and cleaner, so maybe we can go with that? Also I couldn't get the [second one](https://github.com/samhotep/codespaces-actions-playground/actions/runs/6171979230/workflow) to work without manually authenticating git :(

bartaz: Thanks for checking that out. It still wont solve the protected branch issue I think.

bartaz:
Surpassed by #4890

Issue #4899, opened 2023-10-23
Enhance Paragraph Justification for Improved Readability


P0Saurabh:
they appear unevenly spaced and do not align properly. This issue aims to address this problem by implementing proper text justification, ensuring that the content appears neat

P0Saurabh:
can you assign me and if possible give hactoberfest tag to it

PR #4900, opened 2023-10-23
Enhance Paragraph Justification for Improved Readability #4899

P0Saurabh:
donec

### DIFF
--- a/templates/contribute.html
+++ b/templates/contribute.html
@@ -98,10 +98,10 @@
     </div>
     <div class="row">
       <div class="col-6">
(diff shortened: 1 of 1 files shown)


P0Saurabh:
What should I do?

bartaz:
Thanks for your contribution, but I'm not sure if merging all the text here into one paragraph and justifying it is an improvement to be honest.

Screenshot 2023-10-23 at 21 17 05

What are you trying to achieve here @P0Saurabh ?

P0Saurabh:
image!
 

Screenshot 2023-10-24 093640

please use a justified form

lyubomir-popov:
Justifying text makes it harder to read, as it alters the distance between letters. The effect gets worse the narrower the column is. If anything, we would add this as an example of what not to do. However, since we don't often get requests like this, I am not sure it is necessary at this point.

Please refer to Robert Bringhurst's "[The elements of typographic style](https://www.goodreads.com/en/book/show/44735)" for a more thorough explanation why justifying text is not a good idea.

P0Saurabh:
thanks

Issue #4911, opened 2023-11-22
bug: dark theme selected dropdown colour turns white


britneywwc:
**Describe the bug**

When an option is selected on the dark theme dropdown, and other options are being hovered, the selected option's colour updates to white, making the selected option unreadable. 

**To Reproduce**

1.  Use Windows/Ubuntu to recreate this bug. Mac will show the default OS dropdown stylings.
2. Go to [dark forms](https://vanillaframework.io/docs/examples/patterns/forms/forms-dark).
3. Select an Ubuntu version from the dropdown
4. Hover over other options.
5. See the error - selected option should turn white.

**Expected behavior**
The selected option should change to a different background colour, making the option readable.

**Screenshots**
![dark-bug](https://github.com/canonical/vanilla-framework/assets/62298176/32eaed4f-cfce-4b58-97c2-6e477986b1bf)

**Desktop (please complete the following information):**

- OS: Windows, Ubuntu Jammy
- Browser Microsoft Edge, Chromium
- Version: 119.0.2151.72

bartaz:
Triage: to be verified now that we introduced CSS variables for theming.

PR #4912, opened 2023-11-22
Fix dark theme dropdown option bug

britneywwc:
## Done

Fixes [#4911](https://github.com/canonical/vanilla-framework/issues/4911), [WD-7399](https://warthogs.atlassian.net/browse/WD-7399)

## QA

- Use Windows/Ubuntu to recreate this bug. Mac will show the default OS dropdown stylings.
- Go to [dark forms](https://0.0.0.0:8108/docs/examples/patterns/forms/forms-dark) on local development
- Select an Ubuntu version from the dropdown
- Hover over other options.
- Selected option should be highlighted in grey, making the option readable.

## Screenshots
Screen Recording 2023-11-22 at 2 56 19 PM

[WD-7399]: https://warthogs.atlassian.net/browse/WD-7399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -17,7 +17,7 @@
     {
       "matchDatasources": ["docker"],
       "matchPackageNames": ["node"],
-      "allowedVersions": "^16 || ^18 || ^20"
+      "versioning": "node"
     }
   ]
 }


scss/_base_forms.scss
```diff
@@ -271,7 +271,7 @@
 
   // stylelint-disable-next-line selector-max-type -- base styles can use type selectors
   option:checked:not(:disabled) {
-    background-color: $color-background-active;
+    background-color: rgba($color-background-option, 0.5);
```
bartaz, line 274: Hey, thanks for suggestion, but I'm not sure if we want to do it as a transparent version of the colour, as it should use the colours directly available in Vanilla.

I'm not sure what is causing the issue (will try to debug it), but maybe it's an issue that background is changing but text color is not? So maybe making sure we set text colour would fix it?

```suggestion
    background-color: $color-background-active;
    color: $color-text;
```

britneywwc: I've tried setting the text colour but it doesn't work for the dark theme case, since the `$color-text` would be white, similar to the existing `background-color`. 

I wasn't sure if the current background color is the intended color. If so, would changing the text color be sufficient in this case? 

Thanks!

bartaz: Yes, text colour is meant to be white in dark theme.

So I guess the issue is with the `$color-background-active`? 
It uses the value of `$colors--dark-theme--background-active` which is transparent white. Maybe that's an issue. Maybe transparency doesn't work (there is no colour behind), so it turns white. 

With your fix also, even though you seem to be using transparent version of `$color-background-option` it looks the same as other options, so probably just using `$color-background-option` instead of `rgba($color-background-option, 0.5);` would do the same…

I guess to make it work we would need the "active" colour that is not transparent.

These colors are also affecting the multiselect version: https://vanillaframework.io/docs/examples/base/forms/select-multiple

So we need to make sure it works there as well, both in light and dark. 

I don't have a good sugestion right now how to fix this to be honest. Let me think about it.

bartaz:
Marking it as blocked, to be revisited when we update forms to new CSS variables theming.

Issue #4769, opened 2023-05-12
u-show utility (display: initial) doesn't work properly with list items.


minkyngkm:
**Describe the bug**

u-show utility (display: initial) doesn't work properly with list items.

Please refer to this issue for details https://github.com/canonical/ubuntu.com/issues/12859

bartaz:
As mentioned in a comment on #4923:

I just realised we can't reliably fix this issue.

Current fix assumes that all list elements need to be reverted back to `list-item`, but it's not true. List items that are part of components may already override the display property to `block` or `inline` or `flex`.

For example any inline list using `u-show` will stop being inline:

<img width="555" alt="image" src="https://github.com/canonical/vanilla-framework/assets/83575/7a329799-2721-4c9c-a873-fd873229869f">

To fix it we would need to create exceptions for all known Vanilla components that style list elements, which is completely unfeasible.

I think the only way around is to currently accept the issue with u-show on lists, and in future deprecate it and only use u-hide.

---

There is no easy way to fix that. There is no reliable way to "revert" the hiding (display: none) and return to previous state, because we don't know what previous state may be.

The only reliable way is to only use u-hide utilities, not u-show.

This needs to be documented, and u-show utility likely to be deprecated and removed in future.

bartaz:
Triage: Effort high, impact high. Too large to tackle in current v4, we should address this in new architecture.

PR #4923, opened 2023-11-27
Fix u-show not working with list-items

aaryanporwal:
## Done

- Added `display: list-item` specifically for `li` on `u-show` class.

Fixes https://github.com/canonical/vanilla-framework/issues/4769

## QA

- [Add QA steps]

## Screenshots

### DIFF
--- a/scss/_utilities_show.scss
+++ b/scss/_utilities_show.scss
@@ -26,4 +26,28 @@
       }
     }
   }
+
+  // stylelint-disable selector-max-type
+  li.u-show {
+    display: list-item !important;
+
+    &--small {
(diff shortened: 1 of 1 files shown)


bartaz:
I just realised we can't reliably fix this issue.

Current fix assumes that all list elements need to be reverted back to `list-item`, but it's not true. List items that are part of components may already override the display property to `block` or `inline` or `flex`.

For example any inline list using `u-show` will stop being inline:

image

To fix it we would need to create exceptions for all known Vanilla components that style list elements, which is completely unfeasible.

I think the only way around is to currently accept the issue with u-show on lists, and in future deprecate it and only use u-hide.

PR #4940, opened 2024-01-09
Reverts postcss-cli update to v11 as it breaks dotrun

bartaz:
This reverts commit 54ba7f2301ead8f36d411bbf0c9d4a28dccdb4b1.

Dotrun still runs on node v16, so we can't update.

## QA

Pull the branch, run `dotrun install`, make sure it not fails on node version incompatibility


jw-can (approved):
runs for me!

PR #4954, opened 2024-01-22
Correct calculation of input padding to take real border width in con…

lyubomir-popov:
…sideration

## Done

Inpuit padding has a correctionm for border, but it uses a hard-coded 1px value. We set borders using a variable, which happens to be 1.5px. This PR corrects that error by referencing the variable instead of the hard-coded 1px value.

## QA

- check buttons and inputs, verify the padding has 1.5px subtracted from the nudge value. This is a sub-pixel difference, but hsould help things adhere better to the baseline grid.
- [Add QA steps]

## Screenshots

### DIFF
--- a/scss/_settings_spacing.scss
+++ b/scss/_settings_spacing.scss
@@ -117,7 +117,7 @@ $sp-after: (
 $spv-nudge: map-get($nudges, p) !default; // top: nudge; bottom: unit - nudge; result: height = exact multiple of base unit
 $spv-nudge-compensation: $sp-unit - $spv-nudge !default;
 $input-margin-bottom: $sp-unit * 4 - $spv-nudge * 2;
-$input-vertical-padding: calc($spv-nudge - 1px);
+$input-vertical-padding: calc($spv-nudge - $input-border-thickness);
 
 // tick element variables
 $form-tick-box-size: 1rem;


bartaz:
The build fails due to undefined variable. Seems that the `$input-border-thickness` that you try to use is defined later in the code.

It would have to be moved up. I guess it would make sense to move it to spacing file. Right now it's defined in `settings_placeholders` (which seems to be a bunch of weird legacy variable definitions).

scss/_settings_spacing.scss
```diff
@@ -117,7 +117,7 @@ $sp-after: (
 $spv-nudge: map-get($nudges, p) !default; // top: nudge; bottom: unit - nudge; result: height = exact multiple of base unit
 $spv-nudge-compensation: $sp-unit - $spv-nudge !default;
 $input-margin-bottom: $sp-unit * 4 - $spv-nudge * 2;
-$input-vertical-padding: calc($spv-nudge - 1px);
+$input-vertical-padding: calc($spv-nudge - $input-border-thickness);
```
bartaz, line 120: Even with this adjustment I'm not sure if the padding is correct. We have border only on the bottom, so it seems we should have calculated it separately for top and bottom padding?

bartaz:
Reimplemented in https://github.com/canonical/vanilla-framework/pull/5389

PR #4986, opened 2024-02-08
Update chips with new theming support

chillkang:
## Done
- update chips with light and dark theme support
- add examples for dark theme version

## Fixes 
- https://warthogs.atlassian.net/browse/WD-7600

## QA
- Open [demo](https://vanilla-framework-4986.demos.haus/docs/examples/patterns/chip/variants-dark) and compare with [the existing version](https://vanillaframework.io/docs/examples/patterns/chip/variants) to check if chips components behave as expected

## Screenshots
Screenshot 2024-02-08 at 6 16 44 PM


templates/docs/examples/patterns/chip/dark.html
```diff
@@ -3,28 +3,29 @@
 
 {% block standalone_css %}patterns_chip{% endblock %}
 
+{% set is_dark = true %}
```
bartaz, line 6: I know that this way would be more consistent with other components we are migrating, but we already have the variants example with both light and dark theme, and I think it's OK to leave it as is, so we can also easier compare if the changes work as expected.

So please revert the variants example back to having both light and dark chips, thanks!

scss/_patterns_chip.scss
```diff
@@ -64,234 +65,204 @@ $chip-border-thickness: 1px;
...
-  @include vf-chip-theme(
-    $chip-type: $chip-type,
-    $is-interactive: $is-interactive,
-    $colors-chip-tinted-backgrounds: $colors-dark-theme--tinted-backgrounds,
-    $colors-chip-tinted-borders: $colors-dark-theme--tinted-borders,
-    $color-chip-value: $colors--dark-theme--text-default,
-    $color-chip-lead: $colors--dark-theme--text-default
-  );
+  // stylelint-disable-next-line selector-max-type
+  body.is-dark .p-chip--information .p-chip__dismiss {
```
bartaz, line 265: The dismiss button doesn't seem to depend on the variant of the chip, only on the theme. It's the same in all the variants (within the same theme), so seems redundant to have it defined for all of them separately: `body.is-dark .p-chip--information .p-chip__dismiss` `body.is-dark .p-chip--caution .p-chip__dismiss`, etc.

Would it work if defined in one place as `body.is-dark .p-chip__dismiss` for all the chips?

bartaz: Also, currently it only work with `is-dark` set on body, but we need it to work with `is-dark` set on chip itself.

So it needs to be something like:

```
body.is-dark .p-chip__dismiss,
.p-chip.is-dark .p-chip__dismiss,
[class*='.p-chip--'].is-dark .p-chip__dismiss { // this [class*=] will match all class names containing `p-chip--` which should be all our chips variants
    @include vf-icon-close(#fff);
}

lyubomir-popov:
can we please use this as the new yellow on the light theme - it is further from red and accessible CC7900

bartaz (approved):
Looks great, thanks for completing this!

Issue #5001, opened 2024-02-16
Top border of table mobile card is different


bartaz:
**Describe the bug**

Top border of mobile view of table mobile cards is thinner (or lighter) than the rest of the card borders (in every card except the first one):

<img width="295" alt="image" src="https://github.com/canonical/vanilla-framework/assets/83575/c7e8c45e-b0c4-4b89-90d0-5a1ab00bc79f">

https://vanillaframework.io/docs/examples/patterns/tables/table-mobile-card

PR #5005, opened 2024-02-23
Fix table mobile card top border

britneywwc:
## Done

- Fixed the issue where top border wasn't consistent with the other border sides

Fixes https://github.com/canonical/vanilla-framework/issues/5001, [WD-9104](https://warthogs.atlassian.net/jira/software/c/projects/WD/boards/801?selectedIssue=WD-9104)

## QA

- Open [demo](https://vanilla-framework-5005.demos.haus/)
- Go to https://vanilla-framework-5005.demos.haus/docs/examples/patterns/tables/table-mobile-card or https://vanilla-framework-5005.demos.haus/docs/examples/patterns/tables/table-mobile-card-dark
- Make sure border sides are all equal size and same colour

## Screenshots

0 0 0 0_8101_docs_examples_patterns_tables_table-mobile-card

0 0 0 0_8101_docs_examples_patterns_tables_table-mobile-card-dark

[WD-9104]: https://warthogs.atlassian.net/browse/WD-9104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scss/_base_tables.scss
+++ b/scss/_base_tables.scss
@@ -71,6 +71,6 @@
   // stylelint-enable selector-max-type
 
   %table-row-border {
-    border-top: 1px solid $colors--theme--border-default;
+    border-top: 1px solid $colors--theme--border-low-contrast;
   }
 }
--- a/scss/_base_tables.scss
+++ b/scss/_base_tables.scss
@@ -71,6 +71,6 @@
   // stylelint-enable selector-max-type
 
   %table-row-border {
-    border-top: 1px solid $colors--theme--border-low-contrast;
+    border-top: 1px solid $colors--theme--border-default;
   }
 }
--- a/scss/_patterns_table-mobile-card.scss
+++ b/scss/_patterns_table-mobile-card.scss
@@ -24,7 +24,8 @@
           width: 100%;
         }
 
-        tr {
+        // tbody selector needed to override the default table styles
+        tbody tr {
           border: 1px solid $colors--theme--border-default;
           display: block;
           margin-bottom: $spv--x-large;
@@ -64,7 +65,7 @@
           }
 
           &:not(:first-child)::after {
-            background-color: $colors--theme--border-default;
+            background-color: $colors--theme--border-low-contrast;
             content: '';
             height: 1px;
             left: 0;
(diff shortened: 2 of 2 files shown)


scss/_base_tables.scss
```diff
@@ -45,7 +45,7 @@
     }
 
     tr {
-      border-bottom: 1px solid $colors--theme--border-default;
+      border-bottom: $input-border-thickness solid $colors--theme--border-default;
```
bartaz, line 48: @lyubomir-popov Is this OK to use `$input-border-thickness` for tables, or is it reserved to other (interactive) elements?

scss/_base_tables.scss
```diff
@@ -71,6 +71,6 @@
   // stylelint-enable selector-max-type
 
   %table-row-border {
-    border-top: 1px solid $colors--theme--border-low-contrast;
+    border-top: $input-border-thickness solid $colors--theme--border-default;
```
bartaz, line 74: @britneywwc Why changing this to `$input-border-thickness`? 

Overall the fix works fine, but I'm worried that changing the width of the border on ALL rows may affect other things.

britneywwc: I tried to match the border thickness with the table-mobile-card's border which inherits from `$border`. 

Would it be better to change it in `p-table--mobile-card` rather than `%table-row-border` so that it uses the same border thickness (1px) as the previous table rows? Just so it won't affect other things and only change within the table-mobile-card component.

bartaz: Yes, I think we need to treat it separate from standard table row border.

In mobile card small screen view we want those to look like card component, so the border around each card (row) should be the same as in card, so as you mentioned it should be `$border`

As in:

```
  %vf-is-bordered {
    border: $border;
  }
```

scss/_base_tables.scss
```diff
@@ -71,6 +71,6 @@
   // stylelint-enable selector-max-type
 
   %table-row-border {
-    border-top: 1px solid $colors--theme--border-low-contrast;
+    border-top: 1px solid $colors--theme--border-default;
```
bartaz, line 74: I think this needs to stay as `-low-contrast` or otherwise all row borders in all columns become move visible than they are now.

Issue #4998, opened 2024-02-15
Stepped lists always show zero count for nested ordered lists


carkod:
**Describe the bug**

When we add an ordered list nested inside of the `p-stepped-list__content` the numbers always start from 0.x, no matter what step we are in. Interestingly, this only happens when we use `p-stepped-list---detailed` or when use `p-list--divided`

**To Reproduce**

1. Go to https://codepen.io/carkod/pen/mdovJWr
2. Scroll to step 4. You can see that the nested ordered list starts from 0.1, instead it should show 4.1

**Expected behavior**

If you change the top parent class from `p-stepped-list--detailed` to `p-stepped-list`, or the nested list from `p-list--divided` to `p-list` it actually works! That's why this is filed as a bug

**Screenshots**
![Screenshot from 2024-02-15 17-54-40](https://github.com/canonical/vanilla-framework/assets/14939793/f65be225-8cd1-4e9d-911a-e8d94bcc3c74)

bartaz:
Thanks. Seems like there is some counter reset happening somewhere. I wonder if it was intended for a reason.

bartaz:
It seems that the CSS counter inside the p-list--divided conflicts with some CSS reset that we do in stepped list (needs a bit more investigation).

See: https://github.com/search?q=repo%3Acanonical%2Fvanilla-framework%20counter-reset&type=code

bartaz:
Fixed in #5011

PR #5011, opened 2024-02-28
Fix nested list counter

codeEmpress1:
## Done
- Fix nested list counter

https://github.com/canonical/vanilla-framework/issues/4998

## QA

- [Add QA steps]

### Before
image

### After
Screenshot from 2024-03-01 11-50-32

### DIFF
--- a/scss/_patterns_lists.scss
+++ b/scss/_patterns_lists.scss
@@ -71,7 +71,7 @@ $list-step-bullet-margin: $sph--x-large;
   }
 
   %vf-list-item-ordered {
-      &::marker {
+    &::marker {
       content: none;
       display: none;
     }
@@ -354,9 +354,8 @@ $list-step-bullet-margin: $sph--x-large;
 }
 
 @mixin vf-p-stepped-list-detailed {
-    .p-stepped-list--detailed {
+  .p-stepped-list--detailed {
     @extend %numbered-step-container;
-    
     margin-left: auto;
 
     @supports (display: grid) {
@@ -427,7 +426,7 @@ $list-step-bullet-margin: $sph--x-large;
 
 @mixin vf-p-list-nested-counter {
   .p-list--nested-counter {
-    @extend %nested-counter;   
+    @extend %nested-counter;
   }
 }
 
--- a/scss/_patterns_lists.scss
+++ b/scss/_patterns_lists.scss
@@ -71,14 +71,16 @@ $list-step-bullet-margin: $sph--x-large;
   }
 
   %vf-list-item-ordered {
-    &::marker {
+      counter-reset: li;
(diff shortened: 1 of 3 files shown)


scss/_patterns_lists.scss
```diff
@@ -71,14 +71,16 @@ $list-step-bullet-margin: $sph--x-large;
   }
 
   %vf-list-item-ordered {
-    &::marker {
+      counter-reset: li;
```
bartaz, line 74: I think the changes you made here are in good direction, but are not necessary in this form.

Overall, I believe the bug comes from the fact that we use 2 different counters in different lists: we use `li` counter and `list-item` counter. The stepped list uses one of them (`li`) and the divided list uses the other (`list-item`).

I believe the difference is not necessary. We can use the same counter in both (so they correctly reset and increment). 

So I think the fix for the issue would be to change any references to counters to use the same one (either `li` or `list-item` - doesn't matter, just the same everywhere).
So any `counter-reset`, `counter-increment` or value of `counter(…)` should use the same (`list-item` for example).

templates/docs/examples/patterns/lists/list-nested-stepped-detailed.html
```diff
@@ -0,0 +1,56 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Lists / Nested Stepped Detailed Ordered List{% endblock %}
```
bartaz, line 2:
```suggestion
{% block title %}Lists / Ordered list nested in stepped list{% endblock %}
```

bartaz (approved):
LTGM, thanks a lot!

bartaz (changes requested):
There is some issue with standard stepped list now, that it increases by 2 on each level:
https://vanilla-framework-5011.demos.haus/docs/patterns/lists#vertical-stepped

image

I haven't checked in detail yet, but maybe `counter-increment` happens twice or something like that.

codeEmpress1:
> There is some issue with standard stepped list now, that it increases by 2 on each level: https://vanilla-framework-5011.demos.haus/docs/patterns/lists#vertical-stepped
> 
> image
> I haven't checked in detail yet, but maybe `counter-increment` happens twice or something like that.

Ohh, Yea. It happens twice for this, I'll look at it

bartaz:
@codeEmpress1 It seems the issue was that we've been using default browser `list-item` counter, which is also increased natively by the browser. So the counter was increased both by `li` element, and by `__title` class.

I moved counter increase to `li` element, so this only happens there and not in the title as well. This seems to help.

Please review/QA if all looks good from your side. Thanks.

codeEmpress1:
LGTM!

PR #5039, opened 2024-03-28
Add equal height row component

mas-who:
## Done

Added the equal height row component which includes the following features:
- Consistent alignment of grid items across all column in a row
- Out of box responsive behaviour for grid items
- Support for use in standard and 25-75 grid layouts
- Insert dividers that can span across all columns in a row (bridges grid gaps) on large screens and breaks up into smaller per column dividers on smaller screen sizes

Fixes: WD-10102

## QA

- Open [demo](https://vanilla-framework-5039.demos.haus/docs/patterns/equal-height-row)
- Inspect the following component examples and make sure behaviour on all screen sizes align with the functionalities outlined above: [Equal height row / Three column row](https://vanilla-framework-5039.demos.haus/docs/examples/patterns/equal-height-row/3-column-row), [Equal height row / 4 item columns](https://vanilla-framework-5039.demos.haus/docs/examples/patterns/equal-height-row/4-items-per-column), [Equal height row / Default](https://vanilla-framework-5039.demos.haus/docs/examples/patterns/equal-height-row/default)
  - [equal height row](https://vanilla-framework-5039.demos.haus/docs/patterns/equal-height-row)

## Screenshots
Screenshot from 2024-04-04 18-40-40
Screenshot from 2024-04-04 18-41-29
Screenshot from 2024-04-04 18-41-56

### DIFF
--- a/releases.yml
+++ b/releases.yml
@@ -4,6 +4,10 @@
       url: /docs/patterns/navigation
       status: Updated
       notes: We've updated the navigation to support new theming.
+    - component: Equal height row
+      url: /docs/patterns/equal-height-row
+      status: New
+      notes: We've introduced new equal height row component (<code>.p-equal-height-row</code>) to better align content across columns.
 - version: 4.9.0
   features:
     - component: Images
--- a/scss/_patterns_equal-height-row.scss
+++ b/scss/_patterns_equal-height-row.scss
@@ -1,0 +1,214 @@
+@import 'settings';
+
+// wrapper mixin that provides relevant class selectors for 1st, 2nd and 3rd dividers
+@mixin position-divider-by-order-in-grid($divider-order, $large-screen: false) {
+  // for large screens, dividers are positioned as pseudo elements at row level
+  @if $large-screen {
+    @if $divider-order == 1 {
+      &.has-1st-divider::before {
+        @content;
+      }
+    }
+
+    @if $divider-order == 2 {
+      &.has-2nd-divider::after {
+        @content;
+      }
+    }
+
+    @if $divider-order == 3 {
+      // when 3rd-divider is present and 1st-divider is not present
+      &.has-3rd-divider:not(.has-1st-divider)::before,
+      // when 3rd-divider is present and 2nd-divider is not present
+      &.has-3rd-divider:not(.has-2nd-divider)::after {
+        @content;
+      }
+
+      // when only 3rd-divider is present
+      &.has-3rd-divider:not(.has-1st-divider):not(.has-2nd-divider)::before {
+        @content;
+      }
+
+      &.has-3rd-divider:not(.has-1st-divider):not(.has-2nd-divider)::after {
+        display: none;
+      }
+    }
+  } @else {
+    // for smaller screens, dividers are positioned as pseudo elements at column level
+    @if $divider-order == 1 {
+      &.has-1st-divider .p-equal-height-row__col::before {
+        @content;
+      }
+    }
+
+    @if $divider-order == 2 {
+      &.has-2nd-divider .p-equal-height-row__col::after {
+        @content;
+      }
+    }
+
+    @if $divider-order == 3 {
+      &.has-3rd-divider:not(.has-1st-divider) .p-equal-height-row__col::before,
+      &.has-3rd-divider:not(.has-2nd-divider) .p-equal-height-row__col::after {
+        @content;
+      }
+
+      &.has-3rd-divider:not(.has-1st-divider):not(.has-2nd-divider) .p-equal-height-row__col::before {
+        @content;
+      }
+
+      &.has-3rd-divider:not(.has-1st-divider):not(.has-2nd-divider) .p-equal-height-row__col::after {
+        display: none;
+      }
+    }
+  }
+}
+
+@mixin divider-styles($large-screen: false) {
+  // For each row or column grid we only have access to two pseudo elements
+  // if 1st-divider (::before) is present, assume 2nd-divider (::after) isn't, then 3rd-divider must be (::after)
+  // if 2nd-divider (::after) is present, assume 1st-divider (::before) isn't, then 3rd-divider must be (::before)
+  @if $large-screen {
+    &.has-1st-divider::before,
+    &.has-2nd-divider::after,
+    &.has-3rd-divider:not(.has-1st-divider)::before,
+    &.has-3rd-divider:not(.has-2nd-divider)::after {
+      @content;
+    }
+  } @else {
+    // For smaller screens, the divider pseudo elements are inserted at the column level
+    &.has-1st-divider .p-equal-height-row__col::before,
+    &.has-2nd-divider .p-equal-height-row__col::after,
+    &.has-3rd-divider:not(.has-1st-divider) .p-equal-height-row__col::before,
+    &.has-3rd-divider:not(.has-2nd-divider) .p-equal-height-row__col::after {
+      @content;
+    }
+  }
+}
+
+@mixin vf-p-equal-height-row {
+  .p-equal-height-row {
+    @extend %vf-row;
+    padding-bottom: $spv--strip-regular * 0.5;
+    position: relative;
+
+    .p-equal-height-row__col {
+      // smaller screens each column will have border top by default
+      border-top-color: $colors--theme--border-default;
+      border-top-style: solid;
+      border-top-width: 1px;
+      display: grid;
+      grid-column: span $grid-columns-small;
+      grid-row: span 4;
+      grid-template-rows: subgrid;
+      margin-top: -1px;
+      padding-bottom: $spv--small;
+      position: relative;
+
+      @media screen and ($breakpoint-small <= width < $breakpoint-large) {
+        grid-column: span $grid-columns-medium;
+        grid-template-columns: subgrid;
+
+        // for medium screen, each column item will take half of the available cols from the parent grid
+        .p-equal-height-row__item {
+          grid-column: span calc($grid-columns-medium / 2);
+        }
+
+        // for medium screen, position the first column item on the left of the grid
+        .p-equal-height-row__item:first-child {
+          grid-row: span 100; // this needs to be sufficiently large so remaining column items won't get stretched
+        }
+      }
+
+      @media screen and (width >= $breakpoint-large) {
+        border: none;
+        grid-column: span calc($grid-columns / 4);
+        padding-bottom: 0;
+      }
+    }
+
+    // divider styles
+    @include divider-styles($large-screen: false) {
+      @extend %vf-pseudo-border;
+    }
+
+    // remove row level dividers for smaller screen sizes
+    @include divider-styles($large-screen: true) {
+      @extend %vf-pseudo-border;
+      display: none;
+    }
+
+    @include position-divider-by-order-in-grid(1) {
+      grid-row: 2;
+    }
+
+    @include position-divider-by-order-in-grid(2) {
+      grid-row: 3;
+    }
+
+    @include position-divider-by-order-in-grid(3) {
+      grid-row: 4;
+    }
+
+    @media screen and ($breakpoint-small <= width < $breakpoint-large) {
+      // We don't need to insert divider below item-1 for medium screen size since item-1 gets positioned on the left
+      @include position-divider-by-order-in-grid(1) {
+        display: none;
+      }
+
+      @include position-divider-by-order-in-grid(2) {
+        grid-column: 4 / 7;
+        grid-row: 2;
+      }
+
+      @include position-divider-by-order-in-grid(3) {
+        grid-column: 4 / 7;
+        grid-row: 3;
+      }
+    }
+
+    @media screen and (width >= $breakpoint-large) {
+      padding-bottom: $spv--strip-regular;
+
+      // remove column level dividers for large screen sizes
+      @include divider-styles($large-screen: false) {
+        display: none;
+      }
+
+      @include divider-styles($large-screen: true) {
+        display: block;
+        grid-column-end: span 12;
+        grid-column-start: 1;
+        margin: auto;
+      }
+
+      @include position-divider-by-order-in-grid(1, $large-screen: true) {
+        grid-row: 2;
+      }
+
+      @include position-divider-by-order-in-grid(2, $large-screen: true) {
+        grid-row: 3;
+      }
+
+      @include position-divider-by-order-in-grid(3, $large-screen: true) {
+        grid-row: 4;
+      }
+    }
+  }
+
+  // support for 25-75 split on large screen size for this pattern
+  .row--25-75-on-large > .col,
+  .row > .col-9 {
+    & .p-equal-height-row {
+      grid-template-columns: repeat($grid-columns-small, minmax(0, 1fr));
+
+      @media screen and ($breakpoint-small <= width < $breakpoint-large) {
+        grid-template-columns: repeat($grid-columns-medium, minmax(0, 1fr));
+      }
+
+      @media screen and (width >= $breakpoint-large) {
+        grid-template-columns: repeat(9, minmax(0, 1fr));
+      }
+    }
+  }
+}
--- a/templates/docs/patterns/equal-height-row.md
+++ b/templates/docs/patterns/equal-height-row.md
@@ -1,0 +1,90 @@
+---
+wrapper_template: '_layouts/docs.html'
+context:
+  title: Equal height row | Components
+---
+
+The equal height row component aims to provide consistent alignment for grid items within a row format. This is achieved using the CSS `subgrid` feature which allows column grids to share the same layout used in the parent row grid.
+
+In addition to enforcing column grid items alignment within each row, responsive behaviour is provided out of the box as shown below:
+
+<table>
+  <thead>
+    <tr>
+      <th style="width: 50ch">Screen size (px)</th>
+      <th>Behaviour</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>Less than <code>$breakpoint-small</code></td>
+      <td>Each column spans across the entire parent grid. Column items are vertically stacked.</td>
+    </tr>
+    <tr>
+      <td><code>$breakpoint-small</code> - <code>$breakpoint-large</code></td>
+      <td>Each column spans across the entire parent grid. The first item within each column is placed on the left of the other column items.</td>
+    </tr>
+    <tr>
+      <td>Greater than <code>$breakpoint-large</code></td>
+      <td>Columns within the row are displayed horizontally. Column items are vertically stacked.</td>
+    </tr>
+  </tbody>
+</table>
+
+This component is an extension of the [grid component](/docs/patterns/grid). There are several usage variations detailed in the examples below.
+
+## Full example
+
+Each column of the component can have up to 4 sub-grid items that will keep equal heights between the columns.
+
+<div class="embedded-example"><a href="/docs/examples/patterns/equal-height-row/4-items-per-column/" class="js-example">
+View example of the four items per column pattern using equal height row component.
+</a></div>
+
+### Cross-column dividers
+
+You can also insert dividers that span across all columns within a row using `.has-1st-divider`, `.has-2nd-divider` and `.has-3rd-divider`.
+
+<div class="p-notification--information">
+  <div class="p-notification__content">
+    <h5 class="p-notification__title">Note:</h5>
+    <p class="p-notification__message">For smaller screen sizes i.e. <code>< $breakpoint-large</code> the divider will appear below relevant items within each column.</p>
+  </div>
+</div>
+
+<div class="p-notification--caution">
+  <div class="p-notification__content">
+    <p class="p-notification__message">You may only have two dividers maximum. For example, if you have <code>.has-1st-divider</code> and <code>.has-2nd-divider</code> set for the row, then <code>.has-3rd-divider</code> will not be shown. This is a limitation due to usage of pseudo elements for visualising dividers that is capable of spanning across grid gaps.</p>
+  </div>
+</div>
+
+## Three column row
+
+You may use the equal height row component nested inside the 25/75 grid split pattern on large screen size (`.row--25-75-on-large`). The row (`.p-equal-height-row`) should be placed within the 75% container with a maximum of three columns.
+
+<div class="embedded-example"><a href="/docs/examples/patterns/equal-height-row/3-column-row/" class="js-example">
+View example of the three column row pattern using equal height row component.
+</a></div>
+
+## Four column row
+
+Used as a four column row grid (`.p-equal-height-row`) spanning the whole width of the default page grid. Each column (`.p-equal-height-row__col`) within the component is a sub-grid and may have up to four items (`.p-equal-height-row__item`).
+
+<div class="embedded-example"><a href="/docs/examples/patterns/equal-height-row/default/" class="js-example">
+View example of the four column row pattern using equal height row component.
+</a></div>
+
+## Import
+
+To import just this component into your project, copy the snippet below and include it in your main Sass file.
+
+```scss
+// import Vanilla and include base mixins
+// this only needs to happen once in a given project
+@import 'vanilla-framework';
+@include vf-base;
+
+@include vf-p-equal-height-row;
+```
+
+For more information see [Customising Vanilla](/docs/customising-vanilla/) in your projects, which includes overrides and importing instructions.
--- a/scss/_patterns_media-container.scss
+++ b/scss/_patterns_media-container.scss
@@ -1,5 +1,10 @@
 @mixin vf-p-media-container {
+  // stylelint-disable selector-max-type -- media container can use selector type
   .p-media-container {
     margin-top: $spv--small;
+    img {
+      display: block;
+    }
   }
+  // stylelint-enable selector-max-type
 }
--- a/scss/_vanilla.scss
+++ b/scss/_vanilla.scss
@@ -4,18 +4,19 @@
 
 @import 'patterns_accordion';
 @import 'patterns_article-pagination';
+@import 'patterns_badge';
 @import 'patterns_breadcrumbs';
 @import 'patterns_buttons';
 @import 'patterns_card';
 @import 'patterns_chip';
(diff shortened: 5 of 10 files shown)


mas-who:
A rough initial setup to get a sense for what needs to be implemented. There are still some discussions to be had about the approach.

templates/docs/examples/patterns/adv-grid/default.html
```diff
@@ -0,0 +1,61 @@
...
+{% block title %}Advanced Grid / Default{% endblock %}
+
+{% block standalone_css %}patterns_adv-grid{% endblock %}
+
+{% block content %}
+
+<div class="adv-grid-demo">
+  <div class="p-adv-grid">
+    <div class="p-adv-grid__title">
+      <p>MORE FROM CANONICAL</p>
```
bartaz, line 11: This should be some kind of heading element (usually h2 or h3 for such component) using `p-muted-heading` component class name. 

Not ALL CAPS in HTML.

templates/docs/examples/patterns/adv-grid/default.html
```diff
@@ -0,0 +1,61 @@
...
+      <hr class='p-adv-grid__row-divider--below-header' />
+      <div class="p-adv-grid__col">
+        <strong class="p-adv-grid__col-header">
+          Some Title
+        </strong>
+        <p class="p-adv-grid__col-content">Use the Ubuntu terminal and run Linux applications on Windows. Use the Ubuntu
+          terminal
+          and run Linux applications on Windows.</p>
+        <footer class="p-adv-grid__col-footer">
+          <hr />
```
bartaz, line 23: Our default for horizontal lines is rule component, so `<hr class="p-rule" />` this makes sure there is no additional unnecessary margin.

lyubomir-popov (changes requested):
looks good, some comments in the code

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,155 @@
...
+@mixin pseudo-border($width) {
+  background: var(--vf-color-border-default);
+  content: '';
+  display: block;
+  height: 1px;
+  position: absolute;
+  width: $width;
+}
+
+@mixin border-top {
```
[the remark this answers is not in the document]
mas-who: aah great! Thanks for the pointer here :+1:

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,155 @@
...
+
+@mixin vf-p-equal-height-row {
+  .p-equal-height-row {
+    display: grid;
+    grid-template-columns: subgrid;
+    position: relative;
+
+    .p-equal-height-col {
+      display: grid;
+      grid-column: span calc($grid-columns / 4);
```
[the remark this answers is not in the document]
mas-who: yes this is for large screen, the media queries below caters for small and medium screen sizes

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,155 @@
...
+      grid-row: span 10;
```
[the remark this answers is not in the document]
mas-who: This can't be auto because `.p-equal-height-col` is a subgrid and its parent grid has `grid-template-rows` set to default auto. This means the subgrid is actually defining the row structure for the parent grid, so we need to explicitly define the row span. Although we can set the span to 4 instead of 10 to cater for maximum 4 grid items, previously this is set to 10 to cater for an arbitrary number of grid items in a column, but we have decided to have maximum 4 items in a column now.

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,155 @@
...
+      .p-equal-height-col .p-equal-height-col-content {
+        grid-column: span calc($grid-columns-medium / 2);
+      }
+
+      .p-equal-height-col .p-equal-height-col-content:first-child {
+        grid-row: span 100;
+      }
+    }
+
+    @media screen and (width < $breakpoint-small) {
```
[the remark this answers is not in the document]
mas-who: got it, this relates to a previous comment as well, will make the update

templates/docs/examples/patterns/equal-height-row/full-screen-dividers.html
```diff
@@ -0,0 +1,64 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Equal height row / Full screen dividers{% endblock %}
+
+{% block standalone_css %}patterns_equal-height-row{% endblock %}
+
+{% block content %}
+
+<div class="row">
+  <div class='p-equal-height-row has-divider-below-item-1'>
```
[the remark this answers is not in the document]
mas-who: If we use "above", then the dividers will have to start at item-2 i.e. `has-divider-above-item-2`, `has-divider-above-item-3`, `has-divider-above-item-4`. Seems a little weird to start at item-2 for me, what do you think?

templates/docs/examples/patterns/equal-height-row/4-item-columns.html
```diff
@@ -0,0 +1,127 @@
...
+
+<div class="row--25-75-on-large">
+  <div class="col">
+    <a href="#">
+      <p class="p-muted-heading">Latest from our blog</p>
+    </a>
+  </div>
+  <div class="col">
+    <div class="p-equal-height-row has-divider-below-item-1 has-divider-below-item-3">
+      <div class="p-equal-height-col">
```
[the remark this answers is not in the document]
bartaz: `p-equal-height-row__col`

Although I'm wondering if it would make sense to "abstract" it away from row/col a bit more and do: `p-equal-height-items` `p-equal-height-items__item`? Hmm… that doesn't look better actually…

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,147 @@
...
+
+      @media screen and (width >= $breakpoint-large) {
+        border: none;
+        grid-column: span calc($grid-columns / 4);
+      }
+    }
+
+    // divider styles
+    // remove row level dividers for small screen
+    &[class*='p-equal-height-row--divider-below-item-']::before,
```
bartaz, line 49: I guess this was renamed together with `__col` child?

I would leave this as `has-divider-…`. When multiple modifiers are possible on a component (like multiple dividers here), we use `is-` or `has-` classes.

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,147 @@
...
+      .p-equal-height-row__col-item:first-child {
+        grid-row: auto;
+      }
+
+      @media screen and ($breakpoint-small <= width < $breakpoint-large) {
+        grid-column: span $grid-columns-medium;
+        grid-template-columns: subgrid;
+        margin-top: -1px;
+
+        .p-equal-height-row__col-item {
```
bartaz, line 32: It's fine to call this `p-equal-height-row__item` - we don't have to explicitly keep the whole structure in the naming.

releases.yml
```diff
@@ -1,3 +1,9 @@
+- version: 4.10.0
+  features:
+    - component: Equal height row
+      url: /docs/patterns/equal-height-row
+      status: New
+      notes: We've introduced new equal height row pattern using the css subgrid feature.
```
bartaz, line 6:
```suggestion
      notes: We've introduced new equal height row component (<code>.p-equal-height-row</code>) to better align content across columns.
```

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,147 @@
...
+    // remove row level dividers for small screen
+    &[class*='p-equal-height-row--divider-below-item-']::before,
+    &[class*='p-equal-height-row--divider-below-item-']::after {
+      @extend %vf-pseudo-border;
+      display: none;
+      width: 100%;
+    }
+
+    // For smaller screens, the divider pseudo elements are inserted at the column level
+    &.p-equal-height-row--divider-below-item-1 .p-equal-height-row__col::before,
```
bartaz, line 57: Would it make sense to use actual border on small screens instead of the before/after pseudo element divider?

So, if we have a class name that adds divider between items 1/2 - the item 2 will have border on top:

```scss
// pseudo-code, just to give idea
&.has-divider-above-item-2 .p-equal-height-row__item:nth-child(2) {
  border-top: 1px solid $colors--theme--border-default;
}
```

scss/_patterns_equal-height-row.scss
mas-who: Yes, I renamed it and removed `has-` as it's getting verbose, I can put it back, but I was wondering if you have any ideas on how to shorten this?

bartaz: `has-divider-1`, etc is less verbose, but also less clear what that does. 

Maybe: `has-item-2-divider`, `has-item-3-divider` that would place a divider above that numbered item? - but again, by making it shorter we are losing a bit of clarity

mas-who: That last one sounds good :slightly_smiling_face:

templates/docs/examples/patterns/equal-height-row/full-screen-dividers.html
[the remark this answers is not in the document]
mas-who: In separate discussion with bartek, we think `has-1st-divider`, `has-2nd-divider` and `has-3rd-divider` might be a good compromise to make the classes less verbose, wdyt?

bartaz:
Please add a screenshot (or screenshots) to PR description, to better show what the component is.

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,190 @@
+@import 'settings';
+
+@mixin divider-position($divider-order, $large-screen: false) {
+  @if $large-screen {
+    @if $divider-order == 1 {
+      &.p-equal-height-row--has-1st-divider::before {
```
bartaz, line 6: This was supposed to be 2 separate class names

`p-equal-height-row` - a main component class name, and `has-1st-divider` as a modifier class name.

In HTML we want this to be:

`<div class="p-equal-height-row has-1st-divider has-3rd-divider" />`

rather than

`<div class="p-equal-height-row p-equal-height-row--has-1st-divider p-equal-height-row--has-3rd-divider" />`.

While the latter is how official BEM naming would be, we don't use it in Vanilla.

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,190 @@
+@import 'settings';
+
+@mixin divider-position($divider-order, $large-screen: false) {
```
bartaz, line 3: Please add come comment explaining what this mixin calculates, it's a bit hard to understand it without context.

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,190 @@
...
+        grid-row: 2;
+      }
+
+      @include divider-position(3) {
+        grid-column: 4 / 7;
+        grid-row: 3;
+      }
+    }
+
+    @media screen and (width >=$breakpoint-large) {
```
bartaz, line 150:
```suggestion
    @media screen and (width >= $breakpoint-large) {
```

scss/_patterns_equal-height-row.scss
bartaz: Is this mixin needed at all?

As far as I can see every time it's used, it's used in different combination of arguments, and every combination of different arguments results in separate conditional if statement.

So for example, instead of 
```
    @include divider-position(1) {
      grid-row: 2;
    }
```
you could just write:

```
     &.p-equal-height-row--has-1st-divider .p-equal-height-row__col::before {
        @content;
      }
``` 

So, while I understand that it does abstract the positioning of divider into one place, it doesn't seem to remove any duplication or anything. It actually seems there is more code now, because of all the nested conditional statements inside this mixin?

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,190 @@
...
+      border-top-style: solid;
+      border-top-width: 1px;
+      display: grid;
+      grid-column: span $grid-columns-small;
+      grid-row: span 4;
+      grid-template-rows: subgrid;
+      margin-top: -1px;
+      position: relative;
+
+      @media screen and ($breakpoint-small <=width < $breakpoint-large) {
```
bartaz, line 85:
```suggestion
      @media screen and ($breakpoint-small <= width < $breakpoint-large) {
```

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,190 @@
...
+          grid-column: span calc($grid-columns-medium / 2);
+        }
+
+        // for medium screen, position the first column item on the left of the grid
+        .p-equal-height-row__item:first-child {
+          grid-row: span 100;
+        }
+      }
+
+      @media screen and (width >=$breakpoint-large) {
```
bartaz, line 100:
```suggestion
      @media screen and (width >= $breakpoint-large) {
```

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,190 @@
...
+    &.p-equal-height-row--has-3rd-divider:not(.p-equal-height-row--has-2nd-divider) .p-equal-height-row__col::after {
+      @content;
+    }
+  }
+}
+
+@mixin vf-p-equal-height-row {
+  .p-equal-height-row {
+    display: grid;
+    grid-template-columns: subgrid;
```
bartaz, line 70: Does this need to be a child of a row? Because I guess that's the only way for the subgrid to work?

I think we discussed this as independent component, with any possible parent (for example we need to be able to wrap it in `p-section`). So the root element of the component can't use subgrid I'm afraid. It needs to extend the `%row`, so that it defines it's own columns (that are aligned with the rest of the grid).

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,190 @@
...
+        grid-template-columns: subgrid;
+
+        // for medium screen, each column item will take half of the available cols from the parent grid
+        .p-equal-height-row__item {
+          grid-column: span calc($grid-columns-medium / 2);
+        }
+
+        // for medium screen, position the first column item on the left of the grid
+        .p-equal-height-row__item:first-child {
+          grid-row: span 100;
```
bartaz, line 96: I guess it doesn't have to be 100? Let's make it just that it needs to be. So 3 I guess (if we have 4 items max in total, then we can have only 3 in the split view).

scss/_patterns_equal-height-row.scss
mas-who: I see, I went with full BEM compliant here, this is way better and much shorter :+1:

scss/_patterns_equal-height-row.scss
mas-who: My aim for moving the code into a `mixin` is more for readability improvements and not necessarily to reduce code. I find having all the divider styles sitting inside `.p-equal-height-row` makes rest of the styling difficult to read. Maybe we can improve the mixin naming to provide better clarity?

scss/_patterns_equal-height-row.scss
mas-who: no it doesn't have to be and extending %vf-row does remove the dependency of having to wrap this component inside of `.row`. I can update this and relevant html + docs sections, although to use it inside a 25-75 grid layout remains the same since it's no different than having a `.row` nested.

scss/_patterns_equal-height-row.scss
mas-who: nice catch, this was set to 100 before we had the discussion to have maximum of 4 items per column. 3 is perfect

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,190 @@
...
+        grid-row: 4;
+      }
+    }
+  }
+
+  // support for 25-75 split on large screen size for this pattern
+  .row--25-75-on-large > .col,
+  .row > .col-9 {
+    & .p-equal-height-row {
+      grid-gap: 0 map-get($grid-gutter-widths, small);
```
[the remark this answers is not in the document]
mas-who: no grid-gap is already defined in `%vf-row`, so this will be removed :+1:

mas-who:
@bartaz as discussed earlier, I've removed unnecessary styling where relevant. Also toyed around having media queries in the divider mixin, it results in having to duplicate the class selectors for different screen sizes so not ideal so I've kept the mixin to be the same. Also did some digging to use `grid-column` to span the dividers, due to the divider pseudo elements are absolutely positioned, they do not respect parent row padding so the only way to cater for the padding (as far as I can tell) is to either define the width for the pseudo element or use left and right attributes to clip the edges, not sure if you have any other suggestions?

scss/_patterns_equal-height-row.scss
```diff
@@ -0,0 +1,194 @@
...
+      // remove column level dividers for large screen sizes
+      @include divider-styles($large-screen: false) {
+        display: none;
+      }
+
+      @include divider-styles($large-screen: true) {
+        display: block;
+        margin: auto;
+        // for a full 12 col grid, there are default padding that must be taken into consideration by the divider
+        width: calc(100% - 2 * map-get($grid-margin-widths, default));
```
bartaz, line 158: Would something like that work instead? To prevent calculating the padding in, and just rely on the grid?

```
grid-column-start: 1;
grid-column-end: span 12;
```

mas-who: I think this would be the same as `grid-column: span 12`, which is what I tested. The issue with this is that the divider will span the correct number of parent grid columns, but the pseudo element is absolutely positioned, so it does not respect the parent grid padding.

bartaz: As far as I can see `grid-column: span 12` relies on auto positioning. 
If you specify `grid-column-start: 1` it seems to set the start position correctly at the start of first column, and the width can be achieved with `grid-column-end: span 12`. 

image

Although this may be relying on some grid CSS edge case usage with position absolute

bartaz:
There is one more assumption made that may break things.

In case of default 4 col example, we currently rely on the fact that it's not nested in any row or grid. 
It's wrong assumption. While it may not be, it doesn't have to be, it can be. Esprcially if it may be wrapped by section component for some padding.

So we need to make sure that cases like:

```
<div class="u-fixed-width">
  <div class="p-equal-height-row has-1st-divider"> …
```  

or the same nested in `row > col-12`.

Right now the divideer width will be wrongly calculated (because padding won't be there if nested):

image

scss/_patterns_equal-height-row.scss
mas-who: Aaah I see, this definitely works, so it seems that if we explicitly define `grid-column-start` and `grid-column-end` then the divider will align against the grid column tracks. Setting `grid-column-end` to `13` also works, but I think `span 12` seems to be a bit clearer

mas-who:
> There is one more assumption made that may break things.
> 
> In case of default 4 col example, we currently rely on the fact that it's not nested in any row or grid. It's wrong assumption. While it may not be, it doesn't have to be, it can be. Esprcially if it may be wrapped by section component for some padding.
> 
> So we need to make sure that cases like:
> 
> ```
> <div class="u-fixed-width">
>   <div class="p-equal-height-row has-1st-divider"> …
> ```
> 
> or the same nested in `row > col-12`.
> 
> Right now the divideer width will be wrongly calculated (because padding won't be there if nested):
> 
> image

Yes agreed, this relates to assuming that width of the divider needs to be calculated with the parent grid padding in mind. The fix based on the other comment changed the divider definition to use `grid-column-start` and `grid-column-end`. This resolved the problem because we no longer have to calculate the width of the divider (also works for `.equal-height-row` nested inside a `col-12`)

Screenshot from 2024-04-08 18-37-48

bartaz:
One last thing would be to update the examples to make them closer to real content.

For default (4 cols) let's use this (hopefully the needed images can be found on assets server):
image


For 3 cols in 25-75 use this:
image


And this could serve as both having 4 items in columns and divider (images like this may not be available, but any placeholder could work for now I think):
image

https://www.figma.com/file/YNYesRrgkpP9bM3JQbFiXz/Responsive-equal-heights-grid?type=design&node-id=0-1&mode=design&t=esDiQnQTIvweaHkG-0

templates/docs/patterns/equal-height-row.md
```diff
@@ -0,0 +1,99 @@
...
+  <div class="p-notification__content">
+    <p class="p-notification__message">You may only have two dividers maximum. For example, if you have <code>.has-1st-divider</code> and <code>.has-2nd-divider</code> set for the row, then <code>.has-3rd-divider</code> will not be shown. This is a limitation due to usage of pseudo elements for visualising dividers that is capable of spanning across grid gaps.</p>
+  </div>
+</div>
+
+<div class="embedded-example"><a href="/docs/examples/patterns/equal-height-row/full-screen-dividers/" class="js-example">
+View example of full screen dividers in equal height row component.
+</a></div>
+
+## 25-75 grid pattern
```
bartaz, line 69: Let's call is "Three column row" and move it up below "Four column row"

templates/docs/patterns/equal-height-row.md
```diff
@@ -0,0 +1,99 @@
...
+    <tr>
+      <td>Greater than <code>$breakpoint-large</code></td>
+      <td>Columns within the row are displayed horizontally. Column items are vertically stacked.</td>
+    </tr>
+  </tbody>
+</table>
+
+The below sections provides examples for using this component to achieve common design patterns.
+
+## Four column row pattern
```
bartaz, line 42: I think the "pattern" word in the titles is not necessary


```suggestion
## Four column row
```

templates/docs/patterns/equal-height-row.md
bartaz:
```suggestion
## Three column row
```

templates/docs/patterns/equal-height-row.md
```diff
@@ -0,0 +1,99 @@
...
+wrapper_template: '_layouts/docs.html'
+context:
+  title: Equal height row | Components
+---
+
+The equal height row component aims to provide consistent alignment for grid items within a row format. This is achieved using the CSS `subgrid` feature which allows column grids to share the same layout used in the parent row grid.
+
+This component is an extension of the [grid component](/docs/patterns/grid). There are two ways to use this component:
+
+1. Used as a four column row grid (`.p-equal-height-row`). Each column (`.p-equal-height-row__col`) within the row is a css grid, and may have up to four grid items (`.p-equal-height-row__item`).
```
bartaz, line 11:
```suggestion
1. Used as a four column row grid (`.p-equal-height-row`) spanning the whole width of the default page grid. Each column (`.p-equal-height-row__col`) within the component is a sub-grid may have up to four items (`.p-equal-height-row__item`).
```

bartaz: Actually, I think this description can be moved inside the "Four column row" section, above the example.

templates/docs/patterns/equal-height-row.md
```diff
@@ -0,0 +1,99 @@
...
+
+2. Nest inside the `25-75` grid pattern on large screen size (`.row--25-75-on-large`). The row (`.p-equal-height-row`) should be placed within the 75% container with a maximum of three columns.
```
bartaz, line 13:
```suggestion
2. Nested inside the 25/75 grid split pattern on large screen size (`.row--25-75-on-large`). The row (`.p-equal-height-row`) should be placed within the 75% container with a maximum of three columns.
```

This also could be moved down inside the "Three column row" as a description above the example.

bartaz (commented):
Some comments on the docs contents.

lyubomir-popov:
assets for thew above 4 images: https://assets.ubuntu.com/manager?tag=equal-heights-component-example-assets

mas-who:
> assets for thew above 4 images: https://assets.ubuntu.com/manager?tag=equal-heights-component-example-assets
> 
> please also replace these with the onese above ^ image

@lyubomir-popov thanks I have replaced the images

bartaz:
@mas-who Could you also add `p-rule--highlight` on top of the titles, as per design? (https://vanillaframework.io/docs/patterns/rule#highlighted)

bartaz:
One last thing (hopefully) as a drive-by. Currently images have unnecessary spacing below them because they are inline.

I think it would be fine to make images block inside `p-media-container`, so could you add `.p-media-container img { display: block; }` to media container component?

image

templates/docs/patterns/equal-height-row.md
```diff
@@ -0,0 +1,88 @@
...
+    <tr>
+      <td>Greater than <code>$breakpoint-large</code></td>
+      <td>Columns within the row are displayed horizontally. Column items are vertically stacked.</td>
+    </tr>
+  </tbody>
+</table>
+
+This component is an extension of the [grid component](/docs/patterns/grid). There are several usage variations detailed in the examples below.
+
+## Four items per column and full-screen dividers
```
bartaz, line 36: To make it feel less like an exception let's simplify this title to "Full example" maybe

```suggestion
## Full example
```

templates/docs/patterns/equal-height-row.md
```diff
@@ -0,0 +1,88 @@
...
+
+Each column can have up to 4 items.
```
bartaz, line 38:
```suggestion
Each column of the component can have up to 4 sub-grid items that will keep equal heights between the columns.
```

templates/docs/patterns/equal-height-row.md
```diff
@@ -0,0 +1,88 @@
...
+
+You can also insert dividers that span across all columns within a row using `.has-1st-divider`, `.has-2nd-divider` and `.has-3rd-divider`.
```
bartaz, line 40: I'd suggest moving the dividers documentation below the "Full example".

Let's make a separate `### Cross-column dividers` heading and keep the dividers documentation and notifications under it, below the embedded example.

bartaz (commented):
Small suggestions for the docs

templates/docs/patterns/equal-height-row.md
mas-who: Yeah that definitely makes more sense, thanks!

lyubomir-popov:
Really nice work @mas-who +1

bartaz (approved):
Awesome work, thanks @mas-who !

PR #5048, opened 2024-04-04
Fix Percy on CircleCI performance issues

bartaz:
## Done

On Thursday 04.04 CircleCI builds for Percy started to timeout.
For some reason we don't fully understand yet, the process of crawling example links out of the website started to be very slow and time consuming which in the end led to timeout.

To remedy the issue this PR replaces the process of crawling the website (to get example URLs) to simply getting list of example files from file system and constructing URLs out of the file names (because they are jinja templates and there is 1:1 relation between the file paths and URLs).

We also increase CircleCI resources (as suggested by CircleCI itself) to use environment with more CPU cores and RAM.

## QA

- Make sure CircleCI Percy job is passing

image

### DIFF
--- a/snapshots.js
+++ b/snapshots.js
@@ -1,15 +1,29 @@
+const fs = require('fs');
+const path = require('path');
+
 const PORT = process.env.PORT || 8101;
 
-module.exports = async () => {
-  const {default: GetSiteUrls} = await import('get-site-urls');
-  let links = await GetSiteUrls(`http://0.0.0.0:${PORT}/`);
-  let urls = [];
+async function getExampleFiles() {
+  //
+  let files = await fs.promises.readdir(path.join(process.cwd(), 'templates/docs/examples'), {recursive: true});
 
-  links = links.found
-    // only snapshot examples, not the whole site
-    .filter((url) => url.includes('/docs/examples/'))
-    // remove standalone examples listing from screenshots
-    .filter((url) => !url.match(/examples\/standalone$/));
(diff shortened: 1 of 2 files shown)


snapshots.js
```diff
@@ -1,15 +1,35 @@
...
 const PORT = process.env.PORT || 8101;
+const DOCS_URL = `http://localhost:${PORT}/docs/examples/`;
 
-module.exports = async () => {
-  const {default: GetSiteUrls} = await import('get-site-urls');
-  let links = await GetSiteUrls(`http://0.0.0.0:${PORT}/`);
-  let urls = [];
+async function getExampleFiles() {
+  // get all example files from templates folder
+  let files = await fs.promises.readdir(path.join(process.cwd(), 'templates/docs/examples'), {recursive: true});
```
mas-who, line 9: I assume we only care about examples for visual regressions?

snapshots.js
```diff
@@ -1,15 +1,35 @@
...
 
-  links = links.found
-    // only snapshot examples, not the whole site
-    .filter((url) => url.includes('/docs/examples/'))
-    // remove standalone examples listing from screenshots
-    .filter((url) => !url.match(/examples\/standalone$/));
+  files = files.concat(standaloneFiles);
+
+  // map the file paths to URLs
+  return files.map((file) => {
```
mas-who, line 24: what happens to `index.html` and `standalone.html` at the root of `/examples`?

mas-who: oh nvm, I see you already filtering them out in `getExampleFiles`

mas-who (approved):

mas-who:
LGTM :+1:

PR #5049, opened 2024-04-04
Remove get-site-urls dependency

bartaz:
## Done

We are not using `get-site-urls` for getting example URLs, so we can remove this dependency.


mas-who (approved):

mas-who:
LGTM :+1:

PR #5053, opened 2024-04-09
Add divider Suru variant

bartaz:
## Done

Adds new divider suru variant

Fixes [WD-8774](https://warthogs.atlassian.net/browse/WD-8774)

## QA

- Open [demo](https://vanilla-framework-5053.demos.haus/docs/examples/patterns/suru/divider)
- Check the example, make sure it works as expected
  - https://vanilla-framework-5053.demos.haus/docs/examples/patterns/suru/divider
  - check the dark and paper versions (light version with white background is not supported)
  - check smaller screen sizes (suru should be hidden)
  - https://vanilla-framework-5053.demos.haus/docs/patterns/suru#divider-suru

## Screenshots

image

[WD-8774]: https://warthogs.atlassian.net/browse/WD-8774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scss/_patterns_suru.scss
+++ b/scss/_patterns_suru.scss
@@ -109,7 +109,7 @@
       background-image: var(--vf-suru-fan-top-right);
       background-position: top right;
       background-repeat: no-repeat;
-      background-size: 512px 385.5px; // based on image dimensions (2048x1542) scaled down by 0.25
+      background-size: 32rem 24.09rem; // based on image dimensions (2048x1542) scaled down by 0.25, coverted to rem
     }
 
     .p-suru--fan-bottom {
@@ -122,7 +122,7 @@
       background-image: var(--vf-suru-pyramid-top-left);
       background-position: top left;
       background-repeat: no-repeat;
-      background-size: 354px 258.75px; // based on image dimensions (1380x1035) scaled down by 0.25
+      background-size: 22.125rem 16.1718rem; // based on image dimensions (1380x1035) scaled down by 0.25, converted to rem
     }
 
     .p-suru--pyramid-right {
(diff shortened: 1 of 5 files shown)


templates/docs/examples/patterns/suru/divider.html
```diff
@@ -0,0 +1,25 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Suru / Divider{% endblock %}
+
+{% block standalone_css %}patterns_suru{% endblock %}
+
+{% set is_paper = True %}
+{% block content %}
+<div class="p-suru--divider">
+    <div class="row--25-75">
```
carkod, line 9: This may even be a question, why are you using `row--25-75` if there is only one column? Shouldn't this not even use grid?

scss/_patterns_suru.scss
```diff
@@ -129,4 +130,32 @@
...
+
+    // only apply suru background images on large screens
+    @media (min-width: $breakpoint-large) {
+      padding-bottom: calc(2 * $spv--strip-deep);
+
+      &::after {
+        background-image: var(--vf-suru-divider);
+        background-position: bottom left;
+        background-repeat: no-repeat;
+        background-size: 1280px 237px; // based on image dimensions (2560x474) scaled down by 0.5
```
carkod, line 149: Don't we usually use `rems` instead of `px`?

carkod (approved):
LGTM, left some doubts, but I suppose there are reasons.

templates/docs/examples/patterns/suru/divider.html
bartaz: Surus are usually used with hero. And hero by default should align with 75% column - not start from the start of the row. And this will happen if you use 25/75 with just one column.

But you are right, in the simplest case, if we don't want any layout, no grid would be needed.

scss/_patterns_suru.scss
bartaz: Yes, but these dimensions are directly related to size of the image (in pixels) - we don't want to scale it in any way. I think… actually, a good question. I'll confirm with Lyubo.

PR #5055, opened 2024-04-15
Fix typography docs

dgtlntv:
## Done

Fixed some of the typography docs to reflect the css more accuratly

## QA

- Look at the typography docs: https://vanilla-framework-5055.demos.haus/docs/base/typography
- Look at vanilla-framework/scss/_base_typography-definitions.scss, vanilla-framework/scss/_settings_font.scss and vanilla-framework/scss/_settings_spacing.scss
- Compare and make sure the typography docs are correct

### DIFF
--- a/templates/docs/base/typography.md
+++ b/templates/docs/base/typography.md
@@ -53,7 +53,7 @@ We use a carefully selected set of weights in our heading hierarchy. As a genera
 
 ## Our type scale
 
-Our type scale consists of 5 font sizes, expressed as rems (root em units). For simplicity, going forward we will refer to sizes in pixels, with the assumption that the base rem unit, which is set by browsers, is at its default of 16 pixels. The following table lists the type sizes and what they are used for:
+Our type scale consists of 8 font sizes, expressed as rems (root em units). For simplicity, going forward we will refer to sizes in pixels, with the assumption that the base rem unit, which is set by browsers, is at its default of 16 pixels. The following table lists the type sizes and what they are used for:
 
 <table style="table-layout: auto">
 <thead>
--- a/templates/docs/base/typography.md
+++ b/templates/docs/base/typography.md
@@ -46,11 +46,10 @@ We use a carefully selected set of weights in our heading hierarchy. As a genera
 | Weight | Use                                                 |
 | ------ | --------------------------------------------------- |
 | 100    | Display headings                                    |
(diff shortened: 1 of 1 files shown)


bartaz:
Thanks @dgtlntv, looks good from my point of view.

@lyubomir-popov could you have a look if proposed changes are good?

templates/docs/base/typography.md
```diff
@@ -103,9 +107,14 @@ Our type scale consists of 5 font sizes, expressed as rems (root em units). For
       <p class="p-heading--1" style="font-size: 2.5rem; line-height: 3rem">H1 heading</p>
     </td>
   </tr>
+  <tr>
+    <td>64px</td>
+    <td>Ad hoc display headings for important bespoke pages on small and medium screens.</td>
```
lyubomir-popov, line 112: why the inline styles?

lyubomir-popov:
seems ok to me. Can I ask for a drive-by - to fix  the column widths of the th's here so these columns in the two tables align:
image

templates/docs/base/typography.md
dgtlntv: In order to make sure that the right size is displayed, independent of the screen you are currently on. But I am now realising, that we need to do it both ways actually. So also need to set inline styles for the large screen version.

bartaz: I guess the question is what do we want to present here: all the available size options, regardless of screen size - even if we never use them on same screen size (for example, both sizes of h2 heading, both sizes of display heading) - I guess 8 options? - this requires examples with inline styles.

Or do we want to show how many options we have on given screen size (5?): so list 1 row for h2 heading, 1 row for display heading, but maybe mention that their size depends on screen size? (and keep examples responsive)

dgtlntv:
> seems ok to me. Can I ask for a drive-by - to fix the column widths of the th's here so these columns in the two tables align: image

 I don't know how to properly do that in Vanilla. Is that something you could take care of @bartaz?

bartaz:
> I don't know how to properly do that in Vanilla. Is that something you could take care of @bartaz?

Resizing columns is one of the most annoying things (especially when it requires to replace markdown table with HTML one…) ;) Consider it done :)

bartaz (approved):
looks good, thanks

PR #5070, opened 2024-04-30
Table layout utilities classes also apply to descendant elements; add table layout example docs

### LINKED ISSUE #2374: Tables probably shouldn't default to fixed width

jmuzina:
## Done

- Updated the `u-table-layout--auto` and `u-table-layout--fixed` utility classes to apply to nested / descended `table` elements as well as elements with those classes directly applied.
- Added example docs for the table layout utility settings. This includes:
  - Default (fixed) table layout (`$table-layout-fixed: true` and no utility class applied directly to a table)
  - Automatic table layout (`$table-layout-fixed: true` and `u-table-layout--auto` applied directly to a table)
  - Automatic table layout, nested (`$table-layout-fixed: true` and `u-table-layout--auto` applied to an ancestor element of a table 

Fixes #2374 , [WD-10925](https://warthogs.atlassian.net/browse/WD-10925) 

## QA

- Open [demo](https://vanilla-framework-5070.demos.haus/docs/settings/table-layout)
- Verify that the example tables in the table layout docs page appear as they do in the attached screenshots.
    - Tables in the table layout utility documentation page should have `table-layout: auto` applied.
  - Docs -> Settings -> Table layout

## Screenshots
image

[WD-10925]: https://warthogs.atlassian.net/browse/WD-10925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jmuzina:
@lyubomir-popov Suggested that this change may be scoped a bit too large. Allowing the css to apply to all tables descended from the utility class element, as Bartek suggested in the Jira card, may be a bit too wide-reaching as this would allow, for example, applying the utility class at the `<body>`, level and making every `<table>` in the application to use `table-layout: auto`. 

Perhaps we should revise this to only apply to direct child elements, like so:

```scss
  @if ($table-layout-fixed) {
    .u-table-layout--auto {
      &,
      & > table {
        table-layout: auto !important;
      }
    }
  } @else {
    .u-table-layout--fixed {
      &,
      & > table {
        table-layout: fixed !important;
      }
    }
  }
```

Will review this with @bartaz upon his return from vacation.

scss/_utilities_layout.scss
```diff
@@ -3,13 +3,21 @@
     @extend %fixed-width-container;
   }
 
+  // stylelint-disable selector-max-type -- table elements can use selector types
   @if ($table-layout-fixed) {
```
bartaz, line 7: While we are updating this, I'd remove the if statement. There is no huge benefit of having only one of those utils in the compiled version of Vanilla (they are tiny and will compress well). And it's just confusing for usage and documentation to have only one of the two available.

So let's just always have both utils available regardless of what is the default.

templates/docs/examples/utilities/table-layout/table-layout-default.html
```diff
@@ -0,0 +1,21 @@
+{% extends "_layouts/examples.html" %}
```
bartaz, line 1: We have enough examples of default tables. I don't think there is a need to this one.

jmuzina: Got it; removed that example from the PR

scss/_utilities_layout.scss
jmuzina: If statement removed from PR and documentation updated accordingly

templates/docs/settings/table-layout.md
```diff
@@ -8,7 +8,17 @@ By default, tables in Vanilla use `table-layout: fixed`.
 
 There are cases where you might want to use `table-layout: auto` - e.g. in automatically generated tables. To do this, you need to set the `$table-layout-fixed` variable to `true`.
 
-Depending on the value of `$table-layout-fixed`, a utility class is generated to allow overriding if necessary:
+Two utility classes are generated to allow overriding if necessary:
 
-- If `$table-layout-fixed` is set to `true` (the default), you can use `u-table-layout--auto` to override it.
-- if `$table-layout-fixed` is set to `auto` it adds a utility called `u-table-layout--fixed`.
+- `u-table-layout--auto` sets a table and its descendant tables to `table-layout: auto`.
```
bartaz, line 13: We don't want people to nest tables, so "its descendant tables" may be a bit misleading, so we should say that this utility sets "the table or any descendant tables of the element it's put on" to auto layout.

scss/standalone/settings_table-layout.scss
```diff
@@ -0,0 +1,4 @@
+@import '../vanilla';
```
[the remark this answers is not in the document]
jmuzina: Removed file & references to it

templates/docs/settings/table-layout.md
jmuzina: Updated documentation language accordingly

bartaz (approved):
LGTM, thanks

PR #5074, opened 2024-05-02
Baseline border fix

dgtlntv:
## Done
On low pixel density devices a 1.5px border would computed as 1px.
If we use $input-border-thickness = 1.5px in the calculation and the user is on a low pixel density screen then the border is rendered as 1px, but we still eg. subtract 1.5px from the padding which will throw off the baseline grid alignment.
So we have to check if the users screen is capable of displaying 0.5px steps at runtime. This is only possible with CSS variables.
This PR uses a CSS variable for input-border-thickness and checks the devices pixelRatio to either give it a 1.5px border or a 1px border.

Additionally the calculation for taking the border width into account was moved from $input-vertical-padding to the components themselves because sometime the border has to be removed twice (button) or only once (input).

## QA
- Open demo
- Look at any input element
- Verify, that the correct border amount has been substracted from the padding - on both high (>1.99dppx) and low pixel density screens.

### Before
Screenshot 2024-05-02 at 10 31 47

### After
Screenshot 2024-05-02 at 10 31 08


dgtlntv:
@bartaz I created a PR for the border calculations based on CSS variables as discussed in the other PR, just so we don't forget about this.

scss/_settings_placeholders.scss
```diff
@@ -2,10 +2,20 @@
 @import 'settings_colors';
 
 // Global placeholder settings
-$input-border-thickness: 1.5px;
```
bartaz, line 5: Thanks for going extra mile and implementing it as a CSS variable.

But for simplicity and consistency with color variables, I think we can keep the `$input-border-thickness` SCSS variable, just make it use the value of CSS variable.

So in places where we use it it won't need to change.

To do that I think you just need to define its value as CSS variable:


```
$input-border-thickness: var(--input-border-thickness);
```

And this should be it. All places that currently use `$input-border-thickness` variable will use it as CSS var automatically without need to change their code to explicitly do that.

scss/_settings_placeholders.scss
```diff
@@ -2,10 +2,20 @@
...
+:root {
+  // On low pixel density devices a 1.5px border would computed as 1px.
+  // If we use $input-border-thickness = 1.5px in the calculation and the user is on a low pixel density screen
+  // then the border is rendered as 1px, but we still eg. subtract 1.5px from the padding which will throw off the baseline grid alignment.
+  // So we have to check if the users screen is capable of displaying 0.5px steps at runtime. This is only possible with CSS variables.
+  --input-border-thickness: 1.5px;
```
bartaz, line 10: Small suggestion, to avoid any future name clashes we namespace our CSS variables with `--vf`, so in this case `--vf-input-border-thickness` (I guess maybe we can have a better name later when we properly define this as a token).

scss/_settings_placeholders.scss
dgtlntv: That makes sense! I'll look into changing that this week. :)

bartaz: Code looks good, but we need to move it elsewhere.

The `settings` part should only define SCSS variables, not produce any CSS. By adding CSS variables here we break this convention.

In `settings_colors.scss` we worked around that by defining mixins with CSS variables, but we use them ("render" to CSS) in `base_themes.scss`. I guess we need some more generic "base variables" file for things like that I guess?… 

I'll try to come up with a proposal, if I'm able I'll push into this branch.

dgtlntv:
@pastelcyborg As Bartek commented on in one of the threads there is this issue with my approach that there should not be CSS in the settings files. I don't really know how I'd go about fixing that so Bartek said he'd be looking into it.

PR #5077, opened 2024-05-02
Navigation "back" button chevron points left instead of right

### LINKED ISSUE #5059: Sliding navigation: "Back" button chevron points to right instead of left
**Describe the bug**

"Back" button chevron is suddenly pointing to the right instead of the left, where it should intuitively point. I think this is a regression introduced by a recent Vanilla update: it's not happening live on ubuntu.com, but it is on [demos](https://ubuntu-com-13522.demos.haus/download/desktop).

**To Reproduce**

1. Open https://vanillaframework.io/docs/examples/patterns/navigation/sliding-dark on a small viewport
2. Open menu, then go to a second level (e.g. "LXD")

**Screenshots**

Screenshot 2024-04-19 at 13 30 54

jmuzina:
## Done

- Fixes a suspected regression in the navigation pattern that causes the "back" button in smaller breakpoints to point right instead of left.

Fixes #5059 

## QA

- Reduce screen size to the medium breakpoint such that the mobile version of the nav appears.
- Tap `Menu`
- Select any sub-menu (such as LXD). Verify that its chevron to open the nested navigation is pointing to the right, indicating drilling into the navigation.
- Open the sub-menu.
- Verify that the back button is visible and its chevron points left.

## Screenshots

Before:
image

After:
image

## Cause
The pseudoelement for the close button needs to have a `transform:rotate(90deg)` applied to point left. However, it was being overridden by the base style for `p-navigation__link::after`, which was moved to further down in the `_patterns_navigation.scss` by [this commit](https://github.com/canonical/vanilla-framework/commit/46b46888caec20d6d5df29923bf9832c1de73385), causing the button to be rotated to the right (`transform: rotate(-90deg)`) instead of the left.

### DIFF
--- a/scss/_patterns_navigation.scss
+++ b/scss/_patterns_navigation.scss
@@ -872,51 +872,55 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
   }
   /* stylelint-enable max-nesting-depth */
 
-  .p-navigation--sliding .p-navigation__item--dropdown-toggle {
-    position: initial;
+  .p-navigation--sliding {
+    // Default positioning for nested dropdown buttons. Overridden by subsequent styles.
+    .p-navigation__item--dropdown-toggle {
+      position: initial;
 
-    &::after {
-      content: none;
-    }
+      &::after {
+        content: none;
+      }
 
-    .p-navigation__link::after {
-      @include vf-icon-chevron-themed;
+      .p-navigation__link::after {
+        @include vf-icon-chevron-themed;
 
-      background-position: center;
-      background-repeat: no-repeat;
-      background-size: contain;
-      content: '';
-      display: block;
-      height: $spv--large;
-      pointer-events: none;
-      position: absolute;
-      right: map-get($grid-margin-widths, small);
-      text-indent: calc(100% + 10rem);
-      top: 1rem;
-      transform: rotate(-90deg);
-      width: map-get($icon-sizes, default);
+        background-position: center;
+        background-repeat: no-repeat;
+        background-size: contain;
+        content: '';
+        display: block;
+        height: $spv--large;
+        pointer-events: none;
+        position: absolute;
+        right: map-get($grid-margin-widths, small);
+        text-indent: calc(100% + 10rem);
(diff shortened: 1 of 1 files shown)


scss/_patterns_navigation.scss
```diff
@@ -919,4 +909,14 @@ $spv-navigation-logo-bottom-position: 0.125rem; // 2px when 1rem is 16px
       }
     }
   }
+
+  .p-navigation--sliding .p-navigation__item--dropdown-close {
```
bartaz, line 913: Is this just a source order changing? 

I'm asking because it's fine if that fixes this bug, but I'm a bit worried that we have issues like this (caused by source order), as this may mean there are other things that are affected by the order here.

jmuzina: Yes, this is a source order issue - caused by the fact that our `.p-navigation__item--dropdown-close` elements themselves are nested within `p-navigation--sliding .p-navigation__item-dropdown-toggle .p-navigation__link` elements. 

Because the "Go Back" button is nested within the dropdown toggle `li` of the parent nav item, and the chevron style is applied to `p-navigation-link` inside of `p-navigation__item-dropdown-toggle`, the source order matters here. If the default style (the button to open a nested navigation) is after the style for the go-back button, then the go-back button's style is overridden, making the chevron point the wrong way.

I believe a source-order independent solution would require either one of the following:

- using the pre-existing `js-back` class in a style selector to exclude `p-navigation__link.js-back` from the style for the right-facing chevron 
- add some other class to the back-nav buttons to exclude them as with the previous option, but this would require users to change their HTML to add another class.

jmuzina: @bartaz I made the requested change per our discussion to encapsulate all of the chevron-related styling into a single block:

image

jmuzina:
Percy shows no changes from baseline: https://percy.io/bb49709b/vanilla-framework/builds/34064034/unchanged/1864481947?browser=firefox&browser_ids=57&category=unchanged&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C800%2C1280

bartaz (approved):

PR #5085, opened 2024-05-07
Release drafter pulls tag number from package.json automatically

jmuzina:
## Done

Release drafter reads the version number from `package.json` and automatically populates tag and release name.

Closes [WD-11112](https://warthogs.atlassian.net/browse/WD-11112)

## QA
Due to the release drafter action being run on pushes to `main`, I tested this in my fork to avoid unnecessarily cluttering our draft releases / `main` commit history.

- View the [release](https://github.com/jmuzina/vanilla-framework/releases/tag/v4.12.0) automatically created by [this workflow run](https://github.com/jmuzina/vanilla-framework/actions/runs/9003871537) (ran by [this action](https://github.com/jmuzina/vanilla-framework/blob/2cc11acf87fb8a851c0c02aeab03097ebb307829/.github/workflows/release-drafter.yml) with [this config](https://github.com/jmuzina/vanilla-framework/blob/2cc11acf87fb8a851c0c02aeab03097ebb307829/.github/release-drafter.yml))
- Verify that the version in my [package.json](https://github.com/jmuzina/vanilla-framework/blob/2cc11acf87fb8a851c0c02aeab03097ebb307829/package.json#L3) matches the version populated into the [release](https://github.com/jmuzina/vanilla-framework/releases/tag/v4.12.0)

[WD-11112]: https://warthogs.atlassian.net/browse/WD-11112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -27,9 +27,8 @@ jobs:
         with:
           # Setting tag and name override the name-template and tag-template.
           # See https://github.com/release-drafter/release-drafter?tab=readme-ov-file#action-inputs for more
-         tag: "v${{ steps.get_version.outputs.version_number }}"
-         name: "v${{ steps.get_version.outputs.version_number }}"
-         version: ${{ steps.get_version.outputs.version_number }}
+         tag: v${{ steps.get_version.outputs.version_number }}
+         name: v${{ steps.get_version.outputs.version_number }}
          commitish: main
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -4,12 +4,33 @@ on:
   push:
     branches:
       - main
+  # temporary to allow testing the release drafter action
+  pull_request:
+    branches:
+      - main
 
 jobs:
   update_release_draft:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          # Only need package.json as it is used by the get_version step to pass tag number into release-drafter
+          sparse-checkout: package.json
+
+      - name: Get version number from package.json
+        id: get_version
+        run: |
+          set -e
+          echo "::set-output name=version_number::$(jq -r '.version' package.json)"
+
       - name: Draft release notes
         uses: release-drafter/release-drafter@v6
+        with:
+          # Setting tag and name override the name-template and tag-template.
+          # See https://github.com/release-drafter/release-drafter?tab=readme-ov-file#action-inputs for more
+         tag: v${{ steps.get_version.outputs.version_number }}
+         name: v${{ steps.get_version.outputs.version_number }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -29,7 +29,7 @@ jobs:
           # See https://github.com/release-drafter/release-drafter?tab=readme-ov-file#action-inputs for more
          tag: "v${{ steps.get_version.outputs.version_number }}"
          name: "v${{ steps.get_version.outputs.version_number }}"
-         version: "v${{ steps.get_version.outputs.version_number }}"
+         version: ${{ steps.get_version.outputs.version_number }}
          commitish: main
         env:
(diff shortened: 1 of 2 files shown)


jmuzina:
Started a PR for the upstream release-drafter to enable templating the tag number from `package.json` into the body of the drafted release; then I noticed by testing that the `$INPUT_VERSION` I was trying to implement is... already supported by release-drafter, but undocumented.

 [Waiting for confirmation](https://github.com/release-drafter/release-drafter/pull/1390#issuecomment-2100693354) on this, but if I'm right, then this will be good to commit to main as-is :)

jmuzina:
Ready for review; we can use `$INPUT_VERSION` in the release body template without making any upstream changes. We should expect this merge to create a draft release of vanilla `v4.11.0`. (no SCSS updates = no version bump) that contains just this update.

.github/workflows/release-drafter.yml
```diff
@@ -9,7 +9,27 @@ jobs:
...
+        with:
+          # Only need package.json as it is used by the get_version step to pass tag number into release-drafter
+          sparse-checkout: package.json
+
+      - name: Get version number from package.json
+        id: get_version
+        shell: bash
+        run: |
+          set -e
+          echo "version_number=$(jq -r '.version' package.json)" >> $GITHUB_OUTPUT
```
bartaz, line 23: Sorry, I'm a bit rusty on GH actions, and they changed how to do things in the past. Is this the recommended way to store some values? Can you point me to some docs?

.github/workflows/release-drafter.yml
```diff
@@ -9,7 +9,27 @@ jobs:
...
+
+      - name: Draft release notes for v${{ steps.get_version.outputs.version_number }}
         uses: release-drafter/release-drafter@v6
+        with:
+          # Setting tag and name override the name-template and tag-template.
+          # See https://github.com/release-drafter/release-drafter?tab=readme-ov-file#action-inputs for more
+         tag: "v${{ steps.get_version.outputs.version_number }}"
+         name: "v${{ steps.get_version.outputs.version_number }}"
+         version: "v${{ steps.get_version.outputs.version_number }}"
```
bartaz, line 32: Not sure if it matters, but is version expected to be a version number (just `1.0.0`), or can it be anything (with `v` in front)?

.github/workflows/release-drafter.yml
jmuzina: @bartaz Yes, this is the "new" way to set an output param for a GHA step. See some docs here: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

You may have seen this in the past with a slightly different syntax that was deprecated recently; maybe this looks more familiar: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

.github/workflows/release-drafter.yml
jmuzina: @bartaz `tag` is filled into the tag field for the release. It is subject to some validation from Github (it has to be a [git ref](https://git-scm.com/docs/git-check-ref-format)). 

`name` is filled into the name of the release. 

`version` is used for the body of the release (where it says "New in Vanilla v{x}"). 

Here's an example image where each field is different:

Screenshot 2024-05-08 at 1 39 52 PM

bartaz: @jmuzina Thanks. So in this case `v` would be duplicated in the body? Because we add it both here into the `version` param, and then again in the template as `v$INPUT_VERSION`. That's why I guess we may want to put just the number of the version into version param. I also makes more sense semantically.

```suggestion
         version: "${{ steps.get_version.outputs.version_number }}"
```

jmuzina: I'm glad you brought this up. It seems like release drafter is doing some sort of logic against `input.tag` to split off the `v` so that `$INPUT_VERSION` template has no `v`. Here's an example of that where you can see that inputting `v4.12.1` as `tag` causes the tag to be resolved to `v4.12.1` and `$INPUT_VERSION` as `4.12.1`.

So, we don't need `input.version` at all in this case; we can just specify `name` for the draft release name and `tag` for the tag to use :)

Screenshot 2024-05-08 at 3 03 30 PM

bartaz (approved):
LGTM, thanks!

Issue #5096, opened 2024-05-22
Remove distinction between horizontal rules with and without margin


lyubomir-popov:
## Component/pattern to amend

Depending on what comes after a line, we use no margin bottom (when followed by text), or .5rem margin-bottom (when followed by images, buttons, etc). This makes things look nicer, but complicates usage - pretty much every pr I review I leave a bunch of comments to apply the p-rule class to remove margins where they are not needed.

In the interest of simplifying the framework, let's drop this distinction and deprecate the p-rule class. This way, every hr will have a .5rem margin-bottom, no exxceptions to document or explain.

Preview:

![image](https://github.com/canonical/vanilla-framework/assets/2741678/7777b37c-e802-44cc-b639-fdd258ac023d)

jmuzina:
Made an [initial draft](https://github.com/jmuzina/vanilla-framework/commit/7510ea509349f8000f1e5fd1c5cca34d59dfd2b2) of a deprecation. Will open a PR after hearing about the above comment.

bartaz:
We need to first clearly decide what do we want to keep.

Sure, default `p-rule` can likely get replaced by clean `<hr />`, but we have muted version, highlighted version, accented version.

Migrating all those on top of "base" `<hr />` seems like a stretch. Having rule component gives it a nice obvious place in the design system, with possibility of optional variants, etc.

So I'm kind of leaning towards keeping `p-rule`, just making sure it looks as `<hr />` does (with bottom margin).

pastelcyborg:
Building on our conversation last week, I still think that, for most (all?) of our components, we could remove the `<hr>` entirely and utilize borders to achieve the desired effect. This would hopefully mean that independent `<hr>` elements are sparingly used.

PR #5097, opened 2024-05-22
Add new image container component and deprecate legacy image variants

jmuzina:
## Done

Added several classes that will help unify the appearance of images in projects that use Vanilla, and deprecated some styles that are no longer part of branding.

- `p-image-container--` classes for setting element aspect ratio.
    - `p-image-container--16-9`
    - `p-image-container--3-2`
    - `p-image-container--2-3`
    - `p-image-container--cinematic`
    - `p-image-container--square`
- `p-image-container .is-highlighted` for distinguishing an image from its surroundings by applying a slight underlay.
- Deprecated image modifiers `p-image--bordered` and `p-image--shadowed`

Closes: #4876 [WD-11412](https://warthogs.atlassian.net/browse/WD-11412)

## QA

- Review updated [Images pattern documentation](https://vanilla-framework-5097.demos.haus/docs/patterns/images)
- Review [updated release notes for 4.12.0](https://vanilla-framework-5097.demos.haus/docs/whats-new)

## Screenshots

Background highlight:
image
Aspect ratio:
image

[WD-11412]: https://warthogs.atlassian.net/browse/WD-11412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/releases.yml
+++ b/releases.yml
@@ -1,3 +1,13 @@
+- version: 4.13.0
+  features:
+    - component: Images
+      url: /docs/patterns/images
+      status: Updated
+      notes: "We've added new classes for controlling image aspect ratio and background."
+    - component: Images
+      url: /docs/patterns/images
+      status: Deprecated
+      notes: "We've deprecated the <code>p-image--bordered</code> and <code>p-image--shadowed</code> classes. Use <code>p-image-container is-highlighted</code> instead."
 - version: 4.12.0
   features:
     - component: Typography
--- a/releases.yml
+++ b/releases.yml
@@ -1,10 +1,10 @@
 - version: 4.13.0
   features:
-    - component: Images
+    - component: Image container
       url: /docs/patterns/images
-      status: Updated
-      notes: "We've added new classes for controlling image aspect ratio and background."
-    - component: Images
+      status: New
+      notes: "We've added a new image container component for controlling image aspect ratio and background."
+    - component: Images / Legacy variants
       url: /docs/patterns/images
       status: Deprecated
       notes: "We've deprecated the <code>p-image--bordered</code> and <code>p-image--shadowed</code> classes. Use <code>p-image-container is-highlighted</code> instead."
--- a/scss/_patterns_image.scss
+++ b/scss/_patterns_image.scss
@@ -1,3 +1,32 @@
+/*
+  @classreference
+  image-container:
+    Image container:
+      .p-image-container:
+        Main element of the image component.
+      "&.is-highlighted":
+        Highlighted variant, to be used to highlight contents.
+      .p-image-container--16-9:
+        Wraps contents in a container with an aspect ratio of 16:9.
+      .p-image-container--3-2:
+        Wraps contents in a container with an aspect ratio of 3:2.
+      .p-image-container--2-3:
+        Wraps contents in a container with an aspect ratio of 2:3.
+      .p-image-container--cinematic:
+        Wraps contents in a container with an aspect ratio of 2.4:1.
+      .p-image-container--square:
+        Wraps contents in a container with an aspect ratio of 1:1.
+    Image:
+      .p-image-container__image:
+        Image element within an image container.
+    Bordered (deprecated):
+      .p-image--bordered:
+        Adds a border around the image.
+    Shadowed (deprecated):
+      .p-image--shadowed:
+        Adds a shadow around the image.
+*/
+
 @import 'settings';
 
 @mixin vf-p-image {
--- a/scss/_patterns_image.scss
+++ b/scss/_patterns_image.scss
@@ -19,12 +19,6 @@
     Image:
       .p-image-container__image:
         Image element within an image container.
-    Bordered (deprecated):
-      .p-image--bordered:
-        Adds a border around the image.
-    Shadowed (deprecated):
-      .p-image--shadowed:
-        Adds a shadow around the image.
 */
 
 @import 'settings';
--- a/scss/_patterns_image.scss
+++ b/scss/_patterns_image.scss
@@ -1,6 +1,37 @@
 @import 'settings';
 
 @mixin vf-p-image {
+  .p-image-container,
+  [class^='p-image-container--'] {
+    align-items: center;
+    display: grid;
+    justify-content: center;
+    text-align: center;
+    .p-image-container__image {
+      object-fit: contain;
+    }
+    &.is-highlighted {
+      background: $colors--theme--background-alt;
+    }
+  }
+
+  .p-image-container--16-9 {
+    aspect-ratio: 16/9;
+  }
+  .p-image-container--3-2 {
+    aspect-ratio: 3/2;
+  }
+  .p-image-container--2-3 {
+    aspect-ratio: 2/3;
+  }
+  .p-image-container--cinematic {
+    aspect-ratio: 2.4/1;
+  }
+  .p-image-container--square {
+    aspect-ratio: 1/1;
+  }
+
+  // Deprecated; will be removed in v5
   .p-image--bordered {
     border: {
       color: $color-mid-light;
@@ -9,6 +40,7 @@
     }
   }
 
+  // Deprecated; will be removed in v5
   .p-image--shadowed {
     box-shadow: 0 1px 5px 1px transparentize($color-mid-light, 0.8);
   }
--- a/templates/docs/examples/patterns/image/container/aspect-ratio/16-9.html
+++ b/templates/docs/examples/patterns/image/container/aspect-ratio/16-9.html
@@ -1,0 +1,10 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Image Container / Aspect Ratio / 16:9{% endblock %}
+
+{% block standalone_css %}patterns_image{% endblock %}
+
+{% block content %}
+<div class="p-image-container--16-9 is-highlighted">
+  <img class="p-image-container__image" src="https://assets.ubuntu.com/v1/9b4c074f-Kernelt%20industries-80-short.png" width="300" alt="">
+</div>
+{% endblock %}
--- a/templates/docs/examples/patterns/image/container/highlighted.html
+++ b/templates/docs/examples/patterns/image/container/highlighted.html
@@ -1,5 +1,5 @@
 {% extends "_layouts/examples.html" %}
-{% block title %}Image container / Highlighted{% endblock %}
+{% block title %}Image Container / Highlighted{% endblock %}
 
 {% block standalone_css %}patterns_image{% endblock %}
 
--- a/templates/docs/examples/patterns/image/container/highlighted.html
+++ b/templates/docs/examples/patterns/image/container/highlighted.html
@@ -1,0 +1,10 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Image container / Highlighted{% endblock %}
+
+{% block standalone_css %}patterns_image{% endblock %}
+
+{% block content %}
+<div class="p-image-container is-highlighted">
+  <img class="p-image-container__image" src="https://assets.ubuntu.com/v1/9b4c074f-Kernelt%20industries-80-short.png" alt="" width="250">
+</div>
+{% endblock %}
--- a/templates/docs/patterns/images.md
+++ b/templates/docs/patterns/images.md
@@ -4,9 +4,45 @@ context:
   title: Images | Components
 ---
 
-Enhance images by adding a variant style with a border or drop shadow.
+Enhance images by distinguishing them from their background or controlling their container's aspect ratio.
+
+## Highlighted image
+
+Insulate an image from its surroundings. This can help to emphasize the image
+and/or prevent its white space from becoming indistinguishable from the surrounding white space.
+
+<div class="embedded-example"><a href="/docs/examples/patterns/image/container/highlighted" class="js-example">
+View example of the image container with a highlighted background
+</a></div>
+
+## Image container with aspect ratio
+
+You can modify the aspect ratio of an image container to neatly wrap an image in a container of the desired aspect ratio.
+This is useful for aligning images or other content with mismatching aspect ratios.
+
+All image containers center the `.p-image-container__image` element inside them by default.
+If you need to change image alignment within the image container, use the [image position utility](/docs/utilities/image-position).
+
+| CSS Class                      | Aspect ratio |
+| ------------------------------ | ------------ |
+| `p-image-container--16-9`      | 16:9         |
+| `p-image-container--3-2`       | 3:2          |
+| `p-image-container--2-3`       | 2:3          |
+| `p-image-container--cinematic` | 2.4:1        |
+| `p-image-container--square`    | 1:1          |
+
+<div class="embedded-example"><a href="/docs/examples/patterns/image/container/aspect-ratio/16-9" class="js-example">
+View example of image container with 16/9 aspect ratio
+</a></div>
 
 ## Image with border
+
+<div class="p-notification--caution">
+  <div class="p-notification__content">
+    <h3 class="p-notification__title">Deprecated</h3>
+    <p class="p-notification__message">Image with border is deprecated. Use `.p-image-container` with `.is-highlighted` modifier instead.</p>
+  </div>
+</div>
 
 A simple key-line around your image.
 
@@ -15,6 +51,13 @@ View example of image with border
 </a></div>
 
 ## Image with drop shadow
+
+<div class="p-notification--caution">
+  <div class="p-notification__content">
+    <h3 class="p-notification__title">Deprecated</h3>
+    <p class="p-notification__message">Image with drop shadow is deprecated. Use `.p-image-container` with `.is-highlighted` modifier instead.</p>
+  </div>
+</div>
 
 Add depth using our drop shadow around your image.
 
--- a/templates/docs/patterns/images.md
+++ b/templates/docs/patterns/images.md
@@ -23,13 +23,7 @@ This is useful for aligning images or other content with mismatching aspect rati
 All image containers center the `.p-image-container__image` element inside them by default.
 If you need to change image alignment within the image container, use the [image position utility](/docs/utilities/image-position).
 
-| CSS Class                      | Aspect ratio |
-| ------------------------------ | ------------ |
-| `p-image-container--16-9`      | 16:9         |
-| `p-image-container--3-2`       | 3:2          |
-| `p-image-container--2-3`       | 2:3          |
-| `p-image-container--cinematic` | 2.4:1        |
-| `p-image-container--square`    | 1:1          |
+See the [class references section](#class-reference) for more information on the available aspect ratio classes.
 
 <div class="embedded-example"><a href="/docs/examples/patterns/image/container/aspect-ratio/16-9" class="js-example">
 View example of image container with 16/9 aspect ratio
@@ -80,6 +74,10 @@ Adds spacing to the top of a media element. This is usually needed when image or
 <div class="embedded-example"><a href="/docs/examples/patterns/image/spacing/" class="js-example">
 View example of media element with spacing
 </a></div>
+
+## Class reference
+
+{{ class_reference("image-container") }}
 
 ## Import
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vanilla-framework",
-  "version": "4.12.2",
+  "version": "4.13.0",
   "author": {
     "email": "webteam@canonical.com",
     "name": "Canonical Webteam"
--- a/templates/docs/examples/patterns/image/container/aspect-ratio/all.html
+++ b/templates/docs/examples/patterns/image/container/aspect-ratio/all.html
@@ -1,0 +1,37 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Image Container / Aspect Ratio / All{% endblock %}
+
+{% block standalone_css %}patterns_image{% endblock %}
+
+{% block content %}
+  <div>
+    <span>16:9</span>
+    <div class="p-image-container--16-9 is-highlighted">
+      <img class="p-image-container__image" src="https://assets.ubuntu.com/v1/9b4c074f-Kernelt%20industries-80-short.png" width="300" alt="">
+    </div>
+  </div>
+  <div>
+    <span>3:2</span>
+    <div class="p-image-container--3-2 is-highlighted">
+      <img class="p-image-container__image" src="https://assets.ubuntu.com/v1/9b4c074f-Kernelt%20industries-80-short.png" width="300" alt="">
+    </div>
+  </div>
+  <div>
+    <span>2:3</span>
+    <div class="p-image-container--2-3 is-highlighted">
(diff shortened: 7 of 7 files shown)


lyubomir-popov:
hi - I think we can drop `u-image-container--padded` as I made an issue to keep hr margins in all cases, so we no longer need to do this. And I think the name of 2-3-4 should be --2.4 (was using 2.34 but the correct one is 2.4 after reading [this ](https://www.panamorph.com/which-home-theater-screen-aspect-ratio-is-best-2-401-2-351-or-169/#:~:text=2.35%3A1%20is%20the%20aspect,movie%20aspect%20ratios%20out%20there))

jmuzina:
As @lyubomir-popov mentioned, `u-image-container--padded` can be removed as we will be deprecating `p-rule`. See #5096

jmuzina:
> I think the name of 2-3-4 should be --2.4 (was using 2.34 but the correct one is 2.4 after reading [this ](https://www.panamorph.com/which-home-theater-screen-aspect-ratio-is-best-2-401-2-351-or-169/#:~:text=2.35%3A1%20is%20the%20aspect,movie%20aspect%20ratios%20out%20there))

Adjusted this to use the class name you described and switched to `aspect-ratio: 1/2.4`.

While it is syntactically possible for us to use `u-aspect-ratio--2.34`, the period in the class name has to be [escaped](https://github.com/jmuzina/vanilla-framework/blob/ff20ec6b7e2d2c40d1a6edbf3c3c71c396ffd4ea/scss/_utilities_aspect-ratio.scss#L11) for the CSS to be valid (`u-aspect-ratio--2\.34`). Then it can be [used as normal](https://github.com/jmuzina/vanilla-framework/blob/ff20ec6b7e2d2c40d1a6edbf3c3c71c396ffd4ea/templates/docs/examples/utilities/aspect-ratio/1-2-4.html#L12).

This causes a [linting failure](https://github.com/canonical/vanilla-framework/actions/runs/9211205543/job/25340035937?pr=5097) due to our [linter configuration](https://github.com/canonical/vanilla-framework/blob/d0197b27f5733789d6d23536eabe68356e045ae1/.stylelintrc.json#L11).

@bartaz @pastelcyborg What do you think is the most semantically clear way to name this class? I don't think we should alter our linting rules for this. Perhaps `u-aspect-ratio--theater` as suggested by the link @lyubomir-popov sent?

bartaz:
I still think we should not introduce them as utilities if they are meant for images only.

Utilities should be a last resort class name that clearly can be applied in ANY context. These from their description seem like are designed to be used with images and images only.

Our current image styling is quite dated (we probably don't want to use the `p-image--bordered` or `p-image--shadowed` anymore https://vanillaframework.io/docs/patterns/images), but maybe it's an opportunity to clean that up and provide a consistent image component with options that we actually need and use.

Creating those image-only class names as utilities is a bad precence in my opition.

@lyubomir-popov What kind of images styles do we need to support?

I see we have images that cover all available space, and illustrations that have some kind of highlight background around them. Any other "background" styles?

The other variable is aspect ratio that is quite well defined.

Is there anything else about images we want to control?

How I see it is that we could possibly have some kind of `p-image` component with any options needed.

So something like:

```<figure class="p-image--aspect-ratio-3-4 is-highighted></figure>```

**Comment by pastelcyborg** [comment]
Thanks for making these changes. I'm wondering if we should change the examples to make it a bit easier to tell what each of them is actually doing; because the image used is so small and there's no background color or border, it's kind of difficult to understand the aspect ratio being applied, since the examples all kind of just look like a big white frame. Even just making the example embed smaller (particularly horizontally) might help here.

File: scss/_patterns_image.scss [hunk 1]
```diff
@@ -1,6 +1,37 @@
 @import 'settings';
 
 @mixin vf-p-image {
+  %vf-image-container--common {
+    img {
+      background-size: contain;
+    }
+  }
+
+  [class^='p-image-container'] {
```
pastelcyborg, line 10: I'd probably make this:
```
.p-image-container,
[class^='p-image-container--'] {
```
Because using the starts-with selector on the entire Block class means it would apply to any Elements also, if we ever were to add any, and that's probably not desirable.

scss/_patterns_image.scss
```diff
@@ -1,6 +1,37 @@
...
+    aspect-ratio: 2/3;
+  }
+  .p-image-container--cinematic {
+    aspect-ratio: 2.4/1;
+  }
+  .p-image-container--square {
+    aspect-ratio: 1/1;
+  }
+
+  .p-image-container--highlighted-background {
```
pastelcyborg, line 30: This also kinda throws a wrench in the starts-with selector above, since it would create a selector containing duplicate styles from the `common` placeholder.

scss/_patterns_image.scss
```diff
@@ -1,6 +1,38 @@
...
+    }
+  }
+
+  .p-image-container,
+  [class^='p-image-container--'] {
+    @extend %vf-image-container--common;
+  }
+
+  [class^='p-image-container'].is-highlighted {
+    background: rgba(0, 0, 0, 0.05);
```
bartaz, line 16: This would need to come from variable, possibly themed one. So we beed a value for both light and dark theme.

@lyubomir-popov What would be the dark version?

lyubomir-popov: pls use the variable called background-alt

scss/_patterns_image.scss
```diff
@@ -1,6 +1,38 @@
...
+      background-size: contain;
+    }
+  }
+
+  .p-image-container,
+  [class^='p-image-container--'] {
+    @extend %vf-image-container--common;
+  }
+
+  [class^='p-image-container'].is-highlighted {
```
bartaz, line 15: Ideally this would be the same selector as the one above in line 10.

I guess for simplicity we can just nest `&.is-highlighted` inside there?

templates/docs/examples/patterns/image/container/aspect-ratio/16-9.html
```diff
@@ -0,0 +1,10 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Image Container / Aspect Ratio / 16/9{% endblock %}
+
+{% block standalone_css %}patterns_image{% endblock %}
+
+{% block content %}
+<div class="p-image-container--16-9 u-align--center u-vertically-center">
```
bartaz, line 7: Utilities shouldn't be required part of a component. They should only be needed if you need to modify the default behaviour that was not specified by the component itself.

If the image is supposed to be horizontally and vertically centered in the container by default, that's what the `p-image-container--` should do itself ideally.

bartaz:
@jmuzina @pastelcyborg What do you think about https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit https://developer.mozilla.org/en-US/docs/Web/CSS/object-position

At the first glance it seems like it could help achieve what we want with just an `img` element. Because as far as I can see we could even set the CSS background on img itself.

templates/docs/examples/patterns/image/container/aspect-ratio/16-9.html
```diff
@@ -0,0 +1,10 @@
...
+  <img src="https://res.cloudinary.com/canonical/image/fetch/f_auto,q_auto,fl_sanitize,c_fill,w_720/https://assets.ubuntu.com/v1/0ecd2ffd-community_wallpaper_noble_little_numbat_boy_by_azskalt.jpeg" alt="">
```
bartaz, line 8: I think it's OK to use direct `https://assets.ubuntu.com/v1/...` links in Vanilla examples. Cloudinary links are usually generated by our image template, so I'd rather not hardcode them into examples.
And it will be also good to avoid using third party service to serve these images as this can impact on Percy snapshots being less stable.

templates/docs/examples/patterns/image/container/aspect-ratio/2-3.html
```diff
@@ -0,0 +1,10 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Image Container / Aspect Ratio / 2/3{% endblock %}
```
bartaz, line 2: I wonder if we need a separate example for every single aspect ratio.

There is probably little value for it, as we likely don't want to embedd them separately in the docs. They are quite self explanatory, so having one of them as an example, with a listing of other available class names should be enough.

Obviously for the sake of reference and testing we should have a single "Image container / All aspect ratios" example that would just showcase all of them all together?

scss/_patterns_image.scss
jmuzina: Moved this (& the centering CSS to remove the alignment utilities from the example) to the `vf-image-container--common` placeholder for simplicity - is that suitable?

templates/docs/examples/patterns/image/container/aspect-ratio/16-9.html
jmuzina: Revised (see [this comment](https://github.com/canonical/vanilla-framework/pull/5097#discussion_r1618861457)) and added note to documentation about centering behavior & how to change image position.

templates/docs/examples/patterns/image/container/aspect-ratio/16-9.html
jmuzina: Revised to use assets API directly

templates/docs/examples/patterns/image/container/aspect-ratio/2-3.html
jmuzina: Revised to use two examples: [16:9](https://vanilla-framework-5097.demos.haus/docs/examples/patterns/image/container/aspect-ratio/16-9?theme=light) (embedded in [documentation](https://vanilla-framework-5097.demos.haus/docs/patterns/images#image-container-with-aspect-ratio) as this is the most frequent use case) and [all](https://vanilla-framework-5097.demos.haus/docs/examples/patterns/image/container/aspect-ratio/all?theme=light)

jmuzina:
> @jmuzina @pastelcyborg What do you think about https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit https://developer.mozilla.org/en-US/docs/Web/CSS/object-position

I've revised to use `object-fit: contain` as it seems to be more appropriate for our use case.

> At the first glance it seems like it could help achieve what we want with just an `img` element. 

Are you suggesting to get rid of the wrapping div with `p-image-container--` classes and switch to `p-image--`? I think we need the div with its own aspect ratio set so the image has a parent container with a set aspect ratio that it can be neatly contained within.

> Because as far as I can see we could even set the CSS background on img itself.

True, we can set the background on img itself. However if we do this and combine usage with the aspect ratio classes, the background is applied only to the image, not the entire container. You'd get a result like this:

image

Instead of the desired:
image

(images from [c.com mongodb page](https://canonical.com/data/mongodb))

bartaz:
> True, we can set the background on img itself. However if we do this and combine usage with the aspect ratio classes, the background is applied only to the image, not the entire container. You'd get a result like this:
> 
> (images from [c.com mongodb page](https://canonical.com/data/mongodb))

Thanks for exploration. Is the issue caused by the fact that our img are inline by default? Seems that setting them to block and 100% width (as they would be a div) seems to solve the issue.
I guess one downside I can see is that it may restrict our ability of setting width/height of actual image (because it would be controlled by object-fit and not width/height of the img element.

jmuzina:
> any idea when is this going to be released?

It could be early next week when @bartaz returns from holiday.

jmuzina:
> One downside I can see is that it may restrict our ability of setting width/height of actual image (because it would be controlled by object-fit and not width/height of the img element.

I think we should avoid doing it this way for this reason. I'd rather not restrict ability to set width & height, which the projects that use Vanilla do quite often.

See c.com: https://github.com/search?q=repo%3Acanonical%2Fcanonical.com%20width%3D&type=code

lyubomir-popov:
to summarise, is the discussion whether to put the background colour on a parent? If so, I think that's the right way to do it. The background fade sets the aspect ratio, and the image is just centered within.

jmuzina:
@lyubomir-popov 
> to summarise, is the discussion whether to put the background colour on a parent?

I believe Bartek is proposing that we implement all of this functionality (aspect ratio & background highlight) on just the image with no parent, at the cost of control over image width/height

pastelcyborg:
I would concur that there should be a container element used here. Removing that and having solely the  itself eliminates the ability of a lot of fine-grained manipulation, particularly to the image's spacing/sizing. It would also make debugging kind of strange, as the  would have a container and metrics totally different from the actual image itself. I have a gut feeling there are other implications to using the  in this way as well, but I'm having trouble articulating exactly what those are without further experimentation.

templates/docs/examples/patterns/image/container/highlighted-background.html
```diff
@@ -0,0 +1,10 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Image container / highlighted background{% endblock %}
```
bartaz, line 2:
```suggestion
{% block title %}Image container / Highlighted{% endblock %}
```

templates/docs/patterns/images.md
```diff
@@ -4,10 +4,46 @@ context:
   title: Images | Components
 ---
 
-Enhance images by adding a variant style with a border or drop shadow.
+Enhance images by distinguishing them from their background or controlling their container's aspect ratio.
+
+## Image container with highlighted background
```
bartaz, line 9: Maybe simply "Highlighted image"?
```suggestion
## Highlighted image container
```

templates/docs/patterns/images.md
```diff
@@ -4,10 +4,46 @@ context:
...
+<div class="embedded-example"><a href="/docs/examples/patterns/image/container/aspect-ratio/16-9" class="js-example">
+View example of image container with 16/9 aspect ratio
+</a></div>
 
 ## Image with border
 
+<div class="p-notification--caution">
+  <div class="p-notification__content">
+    <h3 class="p-notification__title">Deprecated</h3>
+    <p class="p-notification__message">Image with border is deprecated. Use `.p-image-container--highlighted-background` instead.</p>
```
bartaz, line 43:
```suggestion
    <p class="p-notification__message">Image with border is deprecated. Use image container with `.is-highlighted` modifier instead.</p>
```

templates/docs/patterns/images.md
```diff
@@ -16,6 +52,13 @@ View example of image with border
 
 ## Image with drop shadow
 
+<div class="p-notification--caution">
+  <div class="p-notification__content">
+    <h3 class="p-notification__title">Deprecated</h3>
+    <p class="p-notification__message">Image with drop shadow is deprecated. Use `.p-image-container--highlighted-background` instead.</p>
```
bartaz, line 58:
```suggestion
    <p class="p-notification__message">Image with drop shadow is deprecated. Use image container with `.is-highlighted` modifier instead.</p>
```

bartaz (changes requested):
Several places still mention previous `--highlighted-background` instead of new `is-highlighted`. I'm not sure if I found all.

Probably the file `highlighted-background.html` could be renamed to `highlighted.html`, and there is mention in PR description.

bartaz:
@jmuzina @pastelcyborg OK. I agree that settling down on having a wrapping element is a safer approach, let's stick to that.

Now, how about the name of the component itself. Does it have to be "image container"? Or would "image" be enough, `p-image` `p-image--16-9`, etc? Assuming of course that this class name is placed on the wrapping element?

```html
<figure class="p-image--16-9 is-highlighted">
    <img src="…" alt="…" />
</figure>
```

Also, if we have the separate image element inside, and we want to control it we probably should add a class name to it as well? `p-image__img`? Or `p-image-container__image` (in current naming)? What do you think?

jmuzina:
@bartaz 

> Does it have to be "image container"? Or would "image" be enough, `p-image` `p-image--16-9`, etc? Assuming of course that this class name is placed on the wrapping element?

While @pastelcyborg  and I think  "image container" itself is a strange name, it's clearer and safer to use "image container" than to use "image", as the image container's styles are intended for usage on an element that wraps an image. If we use `p-image`, people may get the wrong idea and apply, say, `p-image--16-9` directly to an image, which is unintended usage.

> Also, if we have the separate image element inside, and we want to control it we probably should add a class name to it as well? `p-image__img`? Or `p-image-container__image` (in current naming)? What do you think?

I would normally agree, but I think this sets a different precedent from the media container, which just uses `` tag:
https://github.com/canonical/vanilla-framework/blob/426d15144f7efde7c0dcf0ee52c73b26a1aacdb8/scss/_patterns_media-container.scss#L3-L8

We could make this use `p-image-container__image` as long as we later commit to changing the media container to use an element class name `p-media-container__image` as well.

templates/docs/patterns/images.md
jmuzina: revised

templates/docs/patterns/images.md
jmuzina: revised

templates/docs/patterns/images.md
jmuzina: Revised to highlighted image container

jmuzina:
@bartaz , @pastelcyborg and I met and decided to use `.p-image-container__image` instead of `` tag here. This way the pattern is stricter & also supports other image tags like `<figure>` and `<object>`.

scss/_patterns_image.scss
```diff
@@ -1,6 +1,41 @@
...
+    }
+    align-items: center;
+    display: grid;
+    justify-content: center;
+    text-align: center;
+  }
+
+  .p-image-container,
+  [class^='p-image-container--'] {
+    @extend %vf-image-container--common;
```
bartaz, line 19: Is this placeholder only extended here? Maybe the styles can be defined here directly, as there doesn't seem to be a need for the abstraction yet.

templates/docs/examples/patterns/image/container/highlighted.html
```diff
@@ -0,0 +1,10 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Image container / Highlighted{% endblock %}
+
+{% block standalone_css %}patterns_image{% endblock %}
+
+{% block content %}
+<div class="p-image-container is-highlighted u-align--center">
```
bartaz, line 7: Is the `u-align--center` needed? This should be built into the component ideally.

bartaz (changes requested):
Couple of small comments on the code. It also seems that the branch needs to be rebased due to version changes in package.json

templates/docs/examples/patterns/image/container/highlighted.html
jmuzina: Yes I built this into the common styling and forgot to remove it from the highlighted example as I did with the aspect ratio examples. Thanks; removed!

scss/_patterns_image.scss
jmuzina: Yes, I abstracted it early so it would be easier to extend later; but I've gone ahead and moved its styling to this selector

scss/_patterns_image.scss
```diff
@@ -1,6 +1,41 @@
...
 
 @mixin vf-p-image {
+  %vf-image-container--common {
+    .p-image-container__image {
+      object-fit: contain;
+    }
+    &.is-highlighted {
+      background: $colors--theme--background-alt;
+    }
+    align-items: center;
```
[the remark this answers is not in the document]
jmuzina: Revised

bartaz:
I feel that currently this example is misleading:
image

Because you don't really see the container that has the aspect ratio set, so it's not a good use case for this feature.

I think it makes more sense to use the example with `is-highlighted` background, so that you actually see the container having aspect ratio, with image centered within.
So I guess we could reuse the example from the `is-highlighted`, just add aspect ratio to it.

jmuzina:
> I feel that currently this example is misleading: image
> 
> Because you don't really see the container that has the aspect ratio set, so it's not a good use case for this feature.
> 
> I think it makes more sense to use the example with `is-highlighted` background, so that you actually see the container having aspect ratio, with image centered within. So I guess we could reuse the example from the `is-highlighted`, just add aspect ratio to it.



> I feel that currently this example is misleading: image
> 
> Because you don't really see the container that has the aspect ratio set, so it's not a good use case for this feature.
> 
> I think it makes more sense to use the example with `is-highlighted` background, so that you actually see the container having aspect ratio, with image centered within. So I guess we could reuse the example from the `is-highlighted`, just add aspect ratio to it.

Revised accordingly:
Screenshot 2024-06-13 at 9 10 55 AM

releases.yml
```diff
@@ -1,3 +1,13 @@
+- version: 4.13.0
+  features:
+    - component: Images
```
bartaz, line 3:
```suggestion
    - component: Image container
```

releases.yml
```diff
@@ -1,3 +1,13 @@
...
+      url: /docs/patterns/images
+      status: Updated
```
bartaz, line 5:
```suggestion
      status: New
```

releases.yml
```diff
@@ -1,3 +1,13 @@
...
+      notes: "We've added new classes for controlling image aspect ratio and background."
+    - component: Images
```
bartaz, line 7:
```suggestion
    - component: Images / Legacy variants
```

releases.yml
```diff
@@ -1,3 +1,13 @@
+- version: 4.13.0
+  features:
+    - component: Images
+      url: /docs/patterns/images
+      status: Updated
+      notes: "We've added new classes for controlling image aspect ratio and background."
```
bartaz, line 6:
```suggestion
      notes: "We've added a new image container component for controlling image aspect ratio and background."
```

bartaz (changes requested):
Small tweaks to release notes

templates/docs/examples/patterns/image/container/highlighted.html
```diff
@@ -0,0 +1,10 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Image container / Highlighted{% endblock %}
```
bartaz, line 2: A bit of inconsistent naming:

```suggestion
{% block title %}Image Container / Highlighted{% endblock %}
```

templates/docs/patterns/images.md
```diff
@@ -4,10 +4,46 @@ context:
...
+
+## Image container with aspect ratio
+
+You can modify the aspect ratio of an image container to neatly wrap an image in a container of the desired aspect ratio.
+This is useful for aligning images or other content with mismatching aspect ratios.
+
+All image containers center the `.p-image-container__image` element inside them by default.
+If you need to change image alignment within the image container, use the [image position utility](/docs/utilities/image-position).
+
+| CSS Class                      | Aspect ratio |
```
bartaz, line 26: That reminded me that we actually have a feature to generate list of component class names from the code comment, like this: https://vanillaframework.io/docs/base/code#class-reference

Maybe it would be relatively simple to add something like that to image container? It's basically a yaml at the beginning of the scss file:
https://github.com/canonical/vanilla-framework/blob/main/scss/_patterns_code-snippet.scss#L1-L45

jmuzina: Revised to use class reference for classname documentation

bartaz (approved):
Thanks for all the updates and discussions, all looks good!

PR #5108, opened 2024-06-04
Percy workflow selectivity improvements

jmuzina:
## Done

- Percy tests no longer run against draft PRs.
- Percy tests no longer run against PRs that don't modify examples files or scss files.
- **The above can be circumvented** by applying the `Review: Percy needed` label. All commits to PRs with the Percy review needed label will be tested.
    - Note that name of this label is not stored in code; rather we will use a [Github Repo variable](https://github.com/canonical/vanilla-framework/settings/variables/actions) to define it in one place. It must match the name of a label on our repo that we want to use for requesting Percy testing. I have defined it as "Review: Percy needed".
- Some minor tweaks to the data passing between privileged/nonprivileged workflows to stop unnecessarily compressing the artifact twice. 
- Some preparation work for [WD-11227](https://warthogs.atlassian.net/browse/WD-11227): we are now passing PR numbers into Percy builds, which should enable the Percy integration to create status checks on our PRs once that is set up by IS.
- The "Percy Prepare" workflow was made into a reusable workflow so that it can be called by both the labeling workflow and the on-push workflow without reusing code. 

Closes [WD-11405](https://warthogs.atlassian.net/browse/WD-11405)

## QA

- Review comments & workflows on [testing PR](https://github.com/jmuzina/vanilla-framework/pull/20). 
    - Note that in my fork, the Percy label is "Percy test requested" rather than "Review: Percy needed". Do not worry about this; the name of the label is fetched from a repo variable and in the Canonical repo it will listen for "Review: Percy needed".

[WD-11405]: https://warthogs.atlassian.net/browse/WD-11405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WD-11227]: https://warthogs.atlassian.net/browse/WD-11227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/percy-baseline.yml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - main
+    paths:
```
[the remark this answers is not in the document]
jmuzina: @bartaz Updated so baseline always runs on pushes to main

.github/workflows/pr-percy-prepare-push.yml
```diff
@@ -0,0 +1,25 @@
...
+      - opened
+      - ready_for_review
+      - synchronize
+
+jobs:
+  copy_artifact:
+    name: Copy changed files to GHA artifact
+    # Ignore draft PRS and PRs with the Percy label
+    # If we run tests against PRs with the Percy label, we will run tests twice (test is also ran by the labelling workflow)
+    if: ${{ !github.event.pull_request.draft && !contains(github.event.pull_request.labels.*.name, vars.PERCY_TEST_REQUESTED_LABEL_NAME) }}
```
bartaz, line 20: Would it be possible to have one workflow and construct the conditional in a way that would cover the cases we need?

Right now it seems they are both very similar, and the only difference is in existence of the label? But I guess I'm missing something?

jmuzina: @bartaz 
The reason two are needed is that the `push` workflow is set to only run if the PR changes relevant Percy files, while the label workflow runs on every push to a PR that has the Percy label, regardless of which files were changed. If I have a "paths" selector, on a workflow that needs to run against PRs with the Percy label, the workflow won't run, since the "paths" selector takes precedence.

.github/workflows/pr-percy-prepare-label.yml
```diff
@@ -0,0 +1,22 @@
+# This workflow ensures Percy is executed against PRs with the Percy required label, regardless of which paths were changed
+name: "Percy [labeled] / Prepare build"
```
bartaz, line 2: That's weird, I thought I commented on this already.

I meant for the workflow to be `Percy [labeled]`, and for the job to be `Prepare build`, as now we still have a long and unnecessarily detailed "Copy changed files to GHA artifact" as status name on PR.

```suggestion
name: "Percy [labeled]"
```

.github/workflows/pr-percy-prepare-label.yml
```diff
@@ -0,0 +1,22 @@
...
+      - main
+    types:
+        # workflow runs after a label is added to the PR, or when a commit is added to a PR with the Percy label
+      - labeled
+      - ready_for_review
+      - synchronize
+
+jobs:
+  copy_artifact:
+    name: Copy changed files to GHA artifact
```
bartaz, line 16:
```suggestion
    name: Prepare build
```

These changes would need to be reflected in other places as well.

bartaz (commented):
Small naming adjustments please.

.github/workflows/pr-percy-prepare-label.yml
jmuzina: Ah I misunderstood your original comment and named the entire workflow that; this has been revised!

.github/workflows/pr-percy-prepare-label.yml
jmuzina: Revised

jmuzina:
Making that naming change to the workflows seems to have upset the workflow run trigger... investigating, standby

jmuzina:
> High-level comment: I tend to prefer when workflows/actions are divided into their own directories, as done here:
> 
> https://github.com/pastelcyborg/vanilla-web-scanner/tree/main/.github
> 
> The primary reason for this (atop the structure being a bit clearer) is that it prevents us from having multiple files in the repo called "action.yml," assuming we add other Actions in the future. Every file would then go in one of these two parent directories. Open to other thoughts, though.

@pastelcyborg 
I like this approach. If I understood it correctly, I've made the change to this PR to implement it (moved `percy-snapshot` action into `.github/actions`)

.github/workflows/percy-prepare.yml
```diff
@@ -0,0 +1,48 @@
+name: "Prepare Percy build"
```
bartaz, line 1: Is it a workflow, or an action?

It doesn't seem to be triggered via PR events, but reused in workflows as if it was an action, right?

bartaz: It would be good to have a comment on top what the workflow does, and when it's supposed to be triggered, to have a better understanding of the flow. 

As now it's getting complicated with multiple conditional workflows.

jmuzina: @bartaz 
> Is it a workflow, or an action?
It is a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows). 

> It doesn't seem to be triggered via PR events, but reused in workflows as if it was an action, right?
Yes, it's triggered by the snapshot & label workflows - the workflow runs at job level whereas the action runs at the step level

jmuzina: > It would be good to have a comment on top what the workflow does, and when it's supposed to be triggered, to have a better understanding of the flow.
> 
> As now it's getting complicated with multiple conditional workflows.

Agreed, I'll add a note to the workflow momentarily

bartaz (approved):
LGTM, let's see how that works in practice!

PR #5127, opened 2024-06-11
Add color theme support to Search & Filter component

jmuzina:
## Done
Adds color theme support to Search & Filter component by using new color theme variables

- Color of the search and filter panel headers (it is slightly more contrasted from background now, using default text color rather than `#666`)
- Color of the overflow count - it now uses `$colors--theme--link-default` rather than `$color-information` which is unthemed and does not appear nicely on the dark theme
- Background color of opened search and filter panel (now supports dark theme using themed background color instead of hardcoded to light theme)
- Background color of the search prompt (now uses themed alt background color instead of hardcoded to `#f7f7f7`)
- Background color of the search clear button (now uses themed input background color instead of hardcoded whitesmoke/`f5f5f5`)
- Color of search and filter panel section bottom border (now uses low contrast border instead of `mid-x-light`)
- Enabled color theme switcher on search & filter examples
- Drive-bys:
  - Set chip border thickness to input border thickness
  - Adjusted the position of the chip overflow counter to align to the baseline (it was a bit too high).
  - Removes old blue background color of the aside navigation in the JAAS application layout example

Fixes [WD-11868](https://warthogs.atlassian.net/browse/WD-11868)

## QA

- Open all [search and filter examples](https://vanilla-framework-5127.demos.haus/docs/examples/standalone) and verify that the changes above appear as expected in all color themes.

## Screenshots
Screenshot 2024-08-07 at 10 03 12 AM
Screenshot 2024-08-07 at 10 07 19 AM

[WD-11868]: https://warthogs.atlassian.net/browse/WD-11868?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scss/_base_forms-range.scss
+++ b/scss/_base_forms-range.scss
@@ -2,7 +2,7 @@
   $thumb-shadow: 0 0 $bar-thickness 1px rgba(0, 0, 0, 0.2);
   $thumb-size: 1rem;
   $thumb-radius: 50%;
-  $thumb-border: $border-high-contrast;
+  $thumb-border: $input-border-thickness solid $colors--theme--border-high-contrast;
   $track-height: $bar-thickness;
   $track-radius: $bar-thickness;
 
--- a/scss/_base_forms-range.scss
+++ b/scss/_base_forms-range.scss
@@ -2,7 +2,7 @@
   $thumb-shadow: 0 0 $bar-thickness 1px rgba(0, 0, 0, 0.2);
   $thumb-size: 1rem;
   $thumb-radius: 50%;
-  $thumb-border: $input-border-thickness solid $colors--theme--border-high-contrast;
+  $thumb-border: $border-high-contrast;
   $track-height: $bar-thickness;
   $track-radius: $bar-thickness;
 
--- a/scss/_patterns_search-and-filter.scss
+++ b/scss/_patterns_search-and-filter.scss
@@ -56,7 +56,7 @@
         cursor: pointer;
         position: absolute;
         right: 0.5rem;
-        top: $focus-outline-offset;
+        top: 0.3rem;
       }
 
       &[aria-expanded='true'] {
--- a/scss/_patterns_search-and-filter.scss
+++ b/scss/_patterns_search-and-filter.scss
@@ -168,7 +168,7 @@
     }
 
     .p-filter-panel-section__heading {
-      color: $colors--theme--text-default;
+      color: $colors--theme--text-muted;
       font-size: 1rem;
       margin-bottom: 0;
     }
--- a/scss/_patterns_search-and-filter.scss
+++ b/scss/_patterns_search-and-filter.scss
@@ -22,7 +22,7 @@
   $input-height: calc(map-get($line-heights, default-text) + $input-vertical-padding + $spv-nudge - $input-border-thickness);
 
   .p-search-and-filter {
-    border-bottom: $input-border-thickness solid $colors--theme--border-high-contrast;
+    border-bottom: $border-high-contrast;
     position: relative;
 
     .p-search-and-filter__search-container {
--- a/scss/_patterns_search-and-filter.scss
+++ b/scss/_patterns_search-and-filter.scss
@@ -22,7 +22,7 @@
   $input-height: calc(map-get($line-heights, default-text) + $input-vertical-padding + $spv-nudge - $input-border-thickness);
 
   .p-search-and-filter {
-    border-bottom: $border-high-contrast;
+    border-bottom: $input-border-thickness solid $colors--theme--border-high-contrast;
     position: relative;
 
     .p-search-and-filter__search-container {
--- a/scss/_patterns_search-and-filter.scss
+++ b/scss/_patterns_search-and-filter.scss
@@ -123,8 +123,6 @@
     }
 
     .p-search-and-filter__input {
-      // Background color of the search and filter input must be opaque to cover the chip overflow amount
-      background-color: $colors--theme--background-inputs--opaque;
       border: 0;
       box-shadow: none;
       flex-grow: 1;
--- a/scss/_patterns_search-and-filter.scss
+++ b/scss/_patterns_search-and-filter.scss
@@ -52,7 +52,7 @@
       }
 
       .p-search-and-filter__selected-count {
-        color: $colors--theme--border-information;
+        color: $colors--theme--link-default;
         cursor: pointer;
         position: absolute;
         right: 0.5rem;
--- a/scss/_patterns_search-and-filter.scss
+++ b/scss/_patterns_search-and-filter.scss
@@ -168,8 +168,8 @@
     }
 
     .p-filter-panel-section__heading {
+      @extend %vf-heading-5;
       color: $colors--theme--text-muted;
-      font-size: 1rem;
       margin-bottom: 0;
     }
 
--- a/scss/_patterns_search-and-filter.scss
+++ b/scss/_patterns_search-and-filter.scss
@@ -170,7 +170,6 @@
     .p-filter-panel-section__heading {
       @extend %vf-heading-5;
       color: $colors--theme--text-muted;
-      margin-bottom: 0;
     }
 
     .p-filter-panel-section__chips {
--- a/scss/_patterns_search-and-filter.scss
+++ b/scss/_patterns_search-and-filter.scss
@@ -185,7 +185,7 @@
 
     .p-filter-panel-section__counter {
       bottom: 0;
-      color: $colors--theme--border-information;
+      color: $colors--theme--link-default;
       cursor: pointer;
       position: absolute;
       right: 0.25rem;
--- a/scss/_patterns_search-and-filter.scss
+++ b/scss/_patterns_search-and-filter.scss
@@ -52,7 +52,7 @@
       }
 
       .p-search-and-filter__selected-count {
-        color: $color-information;
+        color: $colors--theme--border-information;
         cursor: pointer;
         position: absolute;
         right: 0.5rem;
@@ -185,7 +185,7 @@
 
     .p-filter-panel-section__counter {
       bottom: 0;
-      color: $color-information;
+      color: $colors--theme--border-information;
       cursor: pointer;
       position: absolute;
       right: 0.25rem;
--- a/scss/_patterns_search-and-filter.scss
+++ b/scss/_patterns_search-and-filter.scss
@@ -123,6 +123,8 @@
     }
 
     .p-search-and-filter__input {
+      // Background color of the search and filter input must be opaque to cover the chip overflow amount
+      background-color: $colors--theme--background-inputs--opaque;
       border: 0;
       box-shadow: none;
       flex-grow: 1;
--- a/scss/_patterns_search-and-filter.scss
+++ b/scss/_patterns_search-and-filter.scss
@@ -72,7 +72,7 @@
     .p-search-and-filter__panel {
       @include vf-transition(opacity, fast);
 
-      background-color: $colors--theme--background-default;
+      background-color: $colors--theme--background-inputs;
       box-shadow: $box-shadow;
       opacity: 1;
       padding: $input-vertical-padding $sph--large 0;
@@ -87,7 +87,7 @@
     }
 
     .p-search-and-filter__search-prompt {
-      background: $colors--theme--background-alt;
+      background: $colors--theme--background-inputs;
       cursor: pointer;
       margin-left: -1rem;
       margin-right: -1rem;
--- a/scss/_patterns_search-and-filter.scss
+++ b/scss/_patterns_search-and-filter.scss
@@ -22,12 +22,12 @@
   $input-height: calc(map-get($line-heights, default-text) + $input-vertical-padding + $spv-nudge - $input-border-thickness);
 
   .p-search-and-filter {
-    border-bottom: $input-border-thickness solid $colors--light-theme--border-high-contrast;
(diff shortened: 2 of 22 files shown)


scss/_base_forms-range.scss
```diff
@@ -2,7 +2,7 @@
   $thumb-shadow: 0 0 $bar-thickness 1px rgba(0, 0, 0, 0.2);
   $thumb-size: 1rem;
   $thumb-radius: 50%;
-  $thumb-border: $input-border-thickness solid $colors--theme--border-high-contrast;
+  $thumb-border: $border-high-contrast;
```
pastelcyborg, line 5: I would probably recommend a different name for this variable - "high contrast" traditionally means a display mode in which colors are completely inverted for users with severe vision issues.

pastelcyborg (commented):
- I notice there's a "+2" in-field helper on Dark and Paper themes that's not visible on Light theme. Is this possibly missing styles?
- Some crazy stuff occurs if you Tab while within the search field - a "Search" button is focused, but it's about half cut off on the right side of the input, and the "Search and filter" placeholder becomes unevenly vertically aligned
- In addition to the previous bullet, hitting Tab while focused within the default component shifts focus to the Search button, also hiding the chips pane; as it is, I don't believe a user would be able to use this component in an accessible manner due to this
- It's possible to get the search input/chips into a state where input doesn't turn into chips and will - seemingly randomly - remove field input; this is most obvious on the With Search Prompt example if you alternate between clicking the chip, typing into the input, clicking the Search for... link, and refreshing the page

I recognize most of the above aren't related to theming directly, but just wanted to capture them during this review.

scss/_base_forms-range.scss
bartaz: We are going to be revisiting colour variables when working on design tokens, so this can be addressed then.

bartaz:
> I recognize most of the above aren't related to theming directly, but just wanted to capture them during this review.

@pastelcyborg This is a weird component in Vanilla sense. It was [originally built in React](https://canonical.github.io/react-components/?path=/docs/search-and-filter--docs) and it's styling has been upstreamed to Vanilla, but there is way too much functionality to it to replicate it in full in plain JS here.

So all we did is create some static examples of different states the component can be to make sure all the styling is in place, but none of the examples is fully funcional (outside of React). So there certainly can be issues if you try to interact with the component.

If you are interested you could try similar tests in [React examples](https://canonical.github.io/react-components/?path=/docs/search-and-filter--docs) and if there are any bugs report them in React components directly. At this point in Vanilla CSS we are only focusing on styling of it.

I don't know if there is a easy better way to approach such component. I guess ideally we should be able to just embed the React component, as this component has practically no use as pure HTML/CSS.

pastelcyborg:
Good to know, thanks @bartaz. Optimally I think the first 2 of my bullets should still be fixed here, as they're related to visual presentation/theming and not any JS functionality.

jmuzina:
I've attempted to fix the chip overflow problem by introducing an opaque version of the inputs background for each color theme and setting the background of `p-search-and-filter__input` to it. This appears to solve the issue, but please [review](https://vanilla-framework-5127.demos.haus/docs/examples/standalone/patterns/search-and-filter/chip-overflow?theme=light) and let me know what you think @bartaz @lyubomir-popov @pastelcyborg 

As for @pastelcyborg 's second bullet point: 
> Some crazy stuff occurs if you Tab while within the search field - a "Search" button is focused, but it's about half cut off on the right side of the input, and the "Search and filter" placeholder becomes unevenly vertically aligned

Maybe we should make a new issue for this and solve it separately, since it isn't related to the theming.

bartaz:
What was the `+2` showing up behind the input? I woudn't expect anything to be behind it if it's covered - maybe that's something we can fix in the component itself rather than by introducing new colours?

But it is another situation when we have issues caused by transparent backgrounds. I think we need a clear direction on this. And I believe that if we address it, it should be the other way around - backgrounds should be opaque by default, and only if transparent version is needed for any reason there would be a specific transparent version and it would be clear that it is transparent (in all themes).

jmuzina:
@bartaz 
> What was the `+2` showing up behind the input? 

It was the `p-search-and-filter__selected-count` from the chip overflow example. In the example, it is always in the DOM and set to visible, but covered by `p-search-and-filter__input` when it is visible. However this only works on light theme, as dark & paper input background colors have transparency that makes the chip overflow count visible.

> backgrounds should be opaque by default, and only if transparent version is needed for any reason there would be a specific transparent version and it would be clear that it is transparent (in all themes).

I agree; I suppose we may need to adjust the values used for the inputs backgrounds on dark & paper. I wonder if there are any apps/sites that are relying on that transparency... I hope not.

jmuzina:
@lyubomir-popov When you have some time, please look through the conversation in this PR and let me know your thoughts on best direction forward for this

lyubomir-popov:
Color of the search and filter panel headers (it is slightly more contrasted from background now, using default text color rather than #666) - _I think these should use the muted text colour_
Color of the overflow count - _just the normal link colour should work fine_
Background color of opened search and filter panel (now supports dark theme using themed background color instead of hardcoded to light theme) - _can we use the input background?_ 
Background color of the search prompt _can we use the input background?_ 
Background color of the search clear button (now uses themed input backgroundcolor instead of hardcoded whitesmoke/f5f5f5)
_themed input sounds ccorrect, same as the two above._ 
Color of search and filter panel section bottom border - _any borders on non-interactive elements should either use the default or low-contrast border styling_

As a drive-by, added the global variable $border-high-contrast and replaced all usages of $input-border-thickness solid $colors--theme--border-high-contrast with $border-high-contrast. _don't we already have that at the theme level?_

jmuzina:
Need to adjust so the overflow count is on the baseline

scss/_base_forms-range.scss
jmuzina: Removed this change.

scss/_patterns_search-and-filter.scss
```diff
@@ -72,7 +72,7 @@
     .p-search-and-filter__panel {
       @include vf-transition(opacity, fast);
 
-      background-color: $color-x-light;
+      background-color: $colors--theme--background-default;
```
jmuzina, line 75: Lyubo wants to use input background color

scss/_patterns_search-and-filter.scss
```diff
@@ -87,7 +87,7 @@
     }
 
     .p-search-and-filter__search-prompt {
-      background: $color-light;
+      background: $colors--theme--background-alt;
```
jmuzina, line 90: Use input background color

jmuzina (commented):
Inline-commenting where to switch to input background color for my own sake

scss/_patterns_search-and-filter.scss
jmuzina: Updated to use input background color (flattened)

scss/_patterns_search-and-filter.scss
jmuzina: Updated to use input background color (flattened)

jmuzina:
Updated to use flattened inputs background color on paper & dark per design suggestion. This partially fixes the transparency problem, but there is still an issue on hover state due to the dark & paper themes using a partially transparent hover color:

Screenshot 2024-08-06 at 5 26 39 PM

That will need to be updated to a flattened color as well.

jmuzina:
@pastelcyborg and @bartaz I've updated this per suggestions from Lyubo - have a look if you like, but we should wait to merge this until @lyubomir-popov is back from his leave and can review it

jmuzina:
For now just review the colors updates; I still need to fix the baseline grid alignment of the chip overflow as a drive-by that Lyubo requested

jmuzina:
Baseline grid alignment adjusted!

jmuzina:
@lyubomir-popov Ready for another design review, the one thing I'm not sure about is the opened search panel background color - it being the same color as the search prompt is a bit strange

scss/_settings_colors.scss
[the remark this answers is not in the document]
jmuzina: It'd be nice to use the same `adjust-color` calculations that light theme uses so we can take advantage of `(input|active|hover)-background-opacity-amount` for consistency's sake.

I'll have to look through our variables to see if there are any available that we can use to get these same variable values using an expression instead of hex values

jmuzina:
I think the color vals I used here are slightly off, as there's a lot of change in unrelated dark/paper examples that use inputs. I'll adjust to avoid that change.

jmuzina:
Needed to add new overrides to the JAAS application layout example's colors, as the current example is relying on background color transparency to achieve its look.

scss/_settings_colors.scss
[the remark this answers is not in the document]
jmuzina: @pastelcyborg Just to confirm from our [earlier discussion](https://chat.canonical.com/canonical/pl/j1kssqtqcidybfp1o794xdr4dc) - this isn't something that we can trivially implement, correct?

jmuzina:
Had to also override the background color of the search field in the MAAS docs layout example, as it was relying on transparency as well.

lyubomir-popov:
managed to break [this one](https://vanilla-framework-5127.demos.haus/docs/examples/standalone/patterns/search-and-filter/default?theme=light) - pressing tab to move the focus away leads to the text being misaligned:
image

lyubomir-popov:
driveby: can we make chip borders 1.5px (using $input-border-thickness)? They differ from all other interactive elements that have borders:
image

jmuzina:
> managed to break [this one](https://vanilla-framework-5127.demos.haus/docs/examples/standalone/patterns/search-and-filter/default?theme=light) - pressing tab to move the focus away leads to the text being misaligned:
> 
> image
> 
> 
> 
> also, the focus moves to a previously invisible element, making it visible 
> 
> image
> 

I think this is related to [5289](https://github.com/canonical/vanilla-framework/issues/5289), not the theming - but I'll have another look in the morning!

jmuzina:
> driveby: can we make chip borders 1.5px (using $input-border-thickness)? They differ from all other interactive elements that have borders

@lyubomir-popov  Sure thing, this has been added to the PR

lyubomir-popov:
Apart from the glitches that we'll tackle separately, looks good!

jmuzina:
> just noticed these use h3 - can we please switch to h5 (not necessarily in the markup, could be with p-heading--5 if there's a strong reason why these should be h3s semantically)

@lyubomir-popov Updated to use `h5` text. This doesn't change the font size (it was previously manually set to `1rem` anyway), but it does slightly alter the line height and other spacing. I set this in the scss (the heading class now extends the h5 behavior) and also updated the markup for clarity.

jmuzina:
@pastelcyborg does the [commit](https://github.com/canonical/vanilla-framework/pull/5127/commits/b37a9caf81f469437ef86e972cd8af85b2fa4dce) I added per @lyubomir-popov feedback after your code +1 look good?

Also, this one will need a pretty hefty Percy review, lots of things changing around slightly due to input background colors being tweaked.

lyubomir-popov:
Well spotted @pastelcyborg ! The margin is reset to 0, throwing things off the baseline grid. I don't think that's necessary - here it is with the default h5 margin-bottom:
image

To me that looks better.

jmuzina:
@lyubomir-popov I've updated to remove the collapsed bottom margin from the headings, in favor of using the default h5 margin bottom of `1.125rem`

templates/docs/examples/layouts/application/JAAS.html
```diff
@@ -45,6 +45,14 @@
   .is-jaas-background {
     background: #003b4e !important;
   }
+
+  .is-jaas-background .p-side-navigation__link[aria-current="page"] {
```
bartaz, line 49: This is such an outdated example (in terms of colours), maybe it would be better to get rid of custom background and make it plain dark?

templates/docs/examples/patterns/search-and-filter/chip-overflow.html
```diff
@@ -56,7 +55,7 @@
   </div>
   <div class="p-search-and-filter__panel" aria-hidden="true">
     <div class="p-filter-panel-section">
-      <h3 class="p-filter-panel-section__heading">Cloud</h3>
+      <h5 class="p-filter-panel-section__heading">Cloud</h5>
```
bartaz, line 58: Would be good to make sure this change is reflected in the react component (not sure if the heading level is hardcoded there).
Or even better, if it's a styling change, it should happen in CSS `p-filter-panel-section__heading` rather than via HTML.

bartaz (approved):
Some small suggestions/questions in comments, but looks good to me.

templates/docs/examples/patterns/search-and-filter/chip-overflow.html
jmuzina: This is done in HTML and CSS [here](https://github.com/canonical/vanilla-framework/blob/f0e4bd851b7c6bb98a6401a06ecad38b4345293f/scss/_patterns_search-and-filter.scss#L170-L173)

I wanted to change it in both places so that the markup is still semantically clear. However, using h3 with the updated styling has no visual differences from just using h5.

That being said, the markup change should be added to React Components for completeness.

jmuzina:
@bartaz I've updated this to remove the custom color from the JAAS aside navigation in the application layout example, as requested.

templates/docs/examples/layouts/application/JAAS.html
jmuzina: Removed the JAAS custom background, it is now plain dark

PR #5137, opened 2024-06-13
Update Pull Requests README with additional information and steps

pastelcyborg:
## Done

- Update README with additional information and steps to improve clarity on PR submission process

Fixes [WD-12174](https://warthogs.atlassian.net/browse/WD-12174)

[WD-12174]: https://warthogs.atlassian.net/browse/WD-12174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/guides/pull-requests.md
+++ b/guides/pull-requests.md
@@ -93,23 +93,10 @@ large to comfortably `Squash and merge`, then default to `Rebase and merge`.
 ## Documentation site status labels
 
 To help draw a user's attention to new features while navigating the docs, we
-use status labels in the sidebar navigation, and within the example's
+use status labels in the sidebar navigation and within the example's
 documentation.
 
-Any affected examples in your PR should have an appropriate label - `new`,
-`updated`, or `deprecated` - added to the
-[sidebar navigation](/templates/_layouts/docs.html). This can be done by passing
-a third argument to any instance of `side_nav_item()` in that file, e.g.:
-
-```
-{{ side_nav_item("/docs/patterns/contextual-menu", "Contextual menu", 'updated) }}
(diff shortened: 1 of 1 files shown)


guides/pull-requests.md
```diff
@@ -4,52 +4,109 @@
...
+## Documentation site status labels
+
+To help draw a user's attention to new features while navigating the docs, we
+use status labels in the sidebar navigation, and within the example's
+documentation.
+
+Any affected examples in your PR should have an appropriate label - `new`,
+`updated`, or `deprecated` - added to the
+[sidebar navigation](/templates/_layouts/docs.html). This can be done by passing
+a third argument to any instance of `side_nav_item()` in that file, e.g.:
```
jmuzina, line 102: I believe this has actually been refactored into using a macro. 

https://github.com/search?q=repo%3Acanonical%2Fvanilla-framework%20side_nav_item(&type=code

We no longer manually call `side_nav_item`; it look like this was updated in #4681 but was never added to the README. So we can probably remove this bit; we pass in the change status from `releases.yaml` instead.

bartaz: Yes, when you change the status of existing navigation item it's taken from `releases.yaml`. 
Only when you add new page you have to add it to `side-navigation.yaml`. So no need to manually mess with templates anymore.

It was changed a while ago, but seems we forgot to update this doc.

bartaz (approved):
Huge improvement, thanks for looking into this!

PR #5174, opened 2024-06-19
Percy examples combination - Add combined template name to combined example body / meta head

jmuzina:
While splitting the combined examples and working on the utilities combination, I noticed that it could become somewhat difficult to keep track of which example you are looking at when looking at bits of a combined example. If there were a problem with a combined example in a Percy snapshot, it would take a bit of work to determine which example the problem came from. 

This PR updates the examples template to do the following for combined examples:

- Embed the name of the current template in the meta head
- Show the name of the current template in the example body 

## Screenshot
Here's an example of how this will look, with the Suru combined example.
image

### DIFF
--- a/templates/_layouts/examples.html
+++ b/templates/_layouts/examples.html
@@ -27,11 +27,16 @@
     {% endif %}
 
     <script defer src="{{ versioned_static('js/example-tools.js') }}"></script>
+    {% else %}
+      <meta name="combined_template_name" content="{{ self._TemplateReference__context.name }}">
     {% endif %}
   
   </head>
 
   <body {% if is_dark %}class="is-dark"{% elif is_paper %}class="is-paper"{% endif %}>
+    {% if is_combined %}
+    <small>{{ self._TemplateReference__context.name }}</small>
+    {% endif %}
     {% block content %}{% endblock %}
   </body>
 </html>


templates/_layouts/examples.html
```diff
@@ -27,11 +27,16 @@
     {% endif %}
 
     <script defer src="{{ versioned_static('js/example-tools.js') }}"></script>
+    {% else %}
+      <meta name="combined_template_name" content="{{ self._TemplateReference__context.name }}">
```
bartaz, line 31: What's the purpose of this? Are we gonna use this info from meta in any way?

bartaz:
> I generally like this, but it's worth noting that if/when browsers do change their text rendering methodology, the text on the page may cause visual regression test failures. Hopefully Percy is smart enough to ignore such minor changes.

If browsers change their text rendering methodology **all** of our visual tests will fail anyway, right?

templates/_layouts/examples.html
```diff
@@ -27,11 +27,16 @@
...
     {% endif %}
   
   </head>
 
   <body {% if is_dark %}class="is-dark"{% elif is_paper %}class="is-paper"{% endif %}>
+    {% if is_combined %}
+    <small>{{ self._TemplateReference__context.name }}</small>
```
bartaz, line 38: That's very much needed indeed. It would be annoying not to know which example to look at.

Could we make this into a link? So that when visiting combined example in the browser we could click to get to single one directly?

templates/_layouts/examples.html
jmuzina: It was sort of an artifact of testing to make sure that `TemplateReference` was working properly, but no intention of really using it at this point. I've removed it

templates/_layouts/examples.html
jmuzina: @bartaz 
Changed it to a link:
image

jmuzina:
@bartaz I rolled the changes we discussed to only embed the block content into this PR. Here's an example of how this looks:

[card-combined.txt](https://github.com/user-attachments/files/15917707/card-combined.txt)
[card-default.txt](https://github.com/user-attachments/files/15917708/card-default.txt)

jmuzina:
Contents of #5176 have been merged into this PR for review simplicity's sake (PRs are being combined)

bartaz (approved):

PR #5175, opened 2024-06-19
Percy examples combination - `utilities/`

jmuzina:
WIP


bartaz:
I'm not sure if it makes sense. I mean sure - that would save us a bit of snapshots, but could be annoying to review in Percy, as they are all so different. Every utility has completely different purpose.

Combining examples per component makes sense from semantic point of view. This seems like going a bit too far for the sake of optimisation.

PR #5176, opened 2024-06-19
Percy examples combination - Jinja vertical spacing option

jmuzina:
Some combined examples (such as search & filter, context menu, navigation) have overlay/non-contained contents that can overlap other combined examples if we're not careful. 

In my PRs so far I have been handling this case manually by creating multple divs with `u-sv3` after examples that need extra space below them. However, this came to be tedious so I have come up with a (hopefully) more streamlined approach for this.

This does the following:

- Creates a new `combined.scss` that the examples template can import to load vertical spacing utility styling. This prevents having to create separate `-combined.scss` style files for combined examples that need to include the vertical spacing utility.
- Allows passing `spacing_below` into an example template. If `spacing_below` is defined and the example is a combined example, then a div with `margin-bottom: <spacing_below>rem` will be added below the block content to make space for whatever content is needed.

## Example 
Example usage, with [Search & Filter](https://github.com/jmuzina/vanilla-framework/tree/examples-combination-spacing-patterns-search-and-filter)
### Markup
```html
{% extends "_layouts/examples.html" %}
{% block title %}Search and filter / Combined{% endblock %}

{% block standalone_css %}patterns_search-and-filter{% endblock %}

{% block content %}
{% with is_combined = true %}
<section>{% include 'docs/examples/patterns/search-and-filter/chip-overflow.html' %}</section>
<section>{% include 'docs/examples/patterns/search-and-filter/default.html' %}</section>
<section>{% include 'docs/examples/patterns/search-and-filter/expanded.html' %}</section>
<section>
  {% with spacing_below = 17 %}
    {% include 'docs/examples/patterns/search-and-filter/with-chips.html' %}
  {% endwith %}
</section>
<section>{% include 'docs/examples/patterns/search-and-filter/with-search-prompt.html' %}</section>
{% endwith %}
{% endblock %}
```
### Screenshot
image

### DIFF
--- a/scss/standalone/combined.scss
+++ b/scss/standalone/combined.scss
@@ -1,3 +1,0 @@
-@import '../settings';
-@import '../utilities_vertical-spacing';
-@include vf-u-vertical-spacing;
--- a/scss/standalone/combined.scss
+++ b/scss/standalone/combined.scss
@@ -1,0 +1,3 @@
+@import '../settings';
+@import '../utilities_vertical-spacing';
+@include vf-u-vertical-spacing;
--- a/templates/_layouts/examples.html
+++ b/templates/_layouts/examples.html
@@ -7,31 +7,40 @@
     <title>{% block title %}{% endblock %}{% if self.title() %} | {% endif %}Examples | Vanilla framework documentation</title>
 
     {% if not is_combined %}
-    {% if is_standalone %}
-    <link rel="stylesheet" type="text/css" href="{{ versioned_static('build/css/standalone/' + (self.standalone_css() if self.standalone_css is defined else 'base') + '.css') }}" />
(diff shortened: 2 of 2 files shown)


bartaz:
Each individual example in the combined examples is wrapped in it's own `<section>` - would it be enough to set some `min-height` on this section wrapper (when we know it's needed)? Seems a bit less involved then generating dozens of spacer divs.

Or, if we really want to automate this, it would be better for the individual example itself to define how tall it needs to be, rather than combined example to define how much space to put.

jmuzina:
@bartaz Revised to use rems for spacing, as discussed in our 1:1

jmuzina:
Rolling this into #5174 for review simplicity.

PR #5177, opened 2024-06-19
Percy example combinations - `patterns/search-and-filter`

jmuzina:
Please review #5176 first as this depends on the vertical spacing introduced by #5176.

## Search prompt script change

Introduces a change to the `_search-prompt-script` to make sure it is only called against search & filter components that actually have search prompt elements. 

### Why?
Without this check, the search & filter script is also called against search & filter components that aren't meant to have the search prompt script run against it, as more search & filter components are now in the DOM.  This causes errors at the `clear.addEventListener` and `prompt.addEventListener` steps as those elements are not found.

### What's the change?
Majority of the change is tabbing/formatting. The bit that is actually changed is here:
```js
// Add click handler for clicks on elements with aria-controls
[].slice
  .call(document.querySelectorAll('.p-search-and-filter'))
  .filter(function (pattern) { // filter step was added to check if the component has a search prompt
    return pattern.querySelector('.p-search-and-filter__search-prompt');
  })
  .forEach(function (pattern) {
// setup search prompt controls....
```

## QA

- Review [combined search & filter example](https://vanilla-framework-5177.demos.haus/docs/examples/patterns/search-and-filter/combined?theme=light)
- Verify there are no console errors in the combined example & the search & filter JS functionality works.

### DIFF
--- a/templates/docs/examples/patterns/search-and-filter/combined.html
+++ b/templates/docs/examples/patterns/search-and-filter/combined.html
@@ -7,10 +7,9 @@
 {% with is_combined = true %}
 <section>{% include 'docs/examples/patterns/search-and-filter/chip-overflow.html' %}</section>
 <section>{% include 'docs/examples/patterns/search-and-filter/default.html' %}</section>
-<section>{% include 'docs/examples/patterns/search-and-filter/expanded.html' %}</section>
 <section>
-  {% with spacing_below = 15 %}
-    {% include 'docs/examples/patterns/search-and-filter/with-chips.html' %}
+  {% with spacing_below = 23 %}
+  {% include 'docs/examples/patterns/search-and-filter/expanded.html' %}
   {% endwith %}
 </section>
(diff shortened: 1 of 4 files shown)


templates/docs/examples/patterns/search-and-filter/combined.html
```diff
@@ -0,0 +1,17 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Search and filter / Combined{% endblock %}
+
+{% block standalone_css %}patterns_search-and-filter{% endblock %}
+
+{% block content %}
+{% with is_combined = true %}
+<section>{% include 'docs/examples/patterns/search-and-filter/chip-overflow.html' %}</section>
+<section>{% include 'docs/examples/patterns/search-and-filter/default.html' %}</section>
```
bartaz, line 9: Nit picking a bit, but could we have some order to the combined examples, at least having the default to be the first?

jmuzina: @bartaz revised!

PR #5184, opened 2024-06-24
fix percy snapshots widths test failing for responsive combined examples

jmuzina:
Fixes the percy snapshots test [failing](https://github.com/canonical/vanilla-framework/actions/runs/9647707795/job/26607152387) on combined examples that embed responsive examples.

Adds logic to check for the responsive breakpoint being snapshotted against combined examples only if they are configured to be responsive.

## QA
- Verify the PR tests pass


tests/snapshots.test.js
```diff
@@ -3,17 +3,33 @@ const snapshotsTest = require('../snapshots');
 
 const PORT = process.env.PORT || 8101;
 
+/**
+ * Combined examples that embed responsive examples.
+ * @type {Map<string, boolean>} Mapping of example urls to whether they embed responsive examples.
+ */
+const RESPONSIVE_COMBINED_EXAMPLES = new Map([
```
bartaz, line 10: Is there a particular reason for this being a Map instead of simple array?

jmuzina: This is a habit of mine when I have a specific value that I'm searching for, since it's faster, but I suppose in the case of a two-value array it wouldn't really matter

bartaz (approved):

bartaz (changes requested):
OH wait, why are tests failing now?

PR #5189, opened 2024-06-25
Hero Pattern: Build Hero 50-50 HOC

jmuzina:
## Done

- Implements 50/50 examples of the Hero HOC per [Design](https://www.figma.com/design/Hll6ZRvHyTTGBZZSoRd8ei/Higher-order-components-(HOCs)?node-id=694-14112&m=dev)
- Changed the CTA block example to use an anchor link element instead of button.
- Moves Hero from "Components" to "Patterns" in side navigation
- Implements Hero with macros, similar to #5219 

This **does not build documentation** for the hero pattern's new layouts; it merely implements the new layouts. Documentation will be completed as part of a separate PR to avoid blowing up scope of the initial implementation.

Fixes [WD-13474](https://warthogs.atlassian.net/browse/WD-13474)

## QA
- Review all changed hero examples
  - [50-50](https://vanilla-framework-5189.demos.haus/docs/examples/patterns/hero/hero-50-50)
  - [50-50 with full cover image](https://vanilla-framework-5189.demos.haus/docs/examples/patterns/hero/hero-50-50-full-cover-image.html)
  - [50-50 split on medium](https://vanilla-framework-5189.demos.haus/docs/examples/patterns/hero/hero-50-50-split-on-medium)
- Review [combined hero example](https://vanilla-framework-5189.demos.haus/docs/examples/patterns/hero/combined?theme=paper)
- Review [4.15.0 release notes](https://vanilla-framework-5189.demos.haus/docs/whats-new)
- Review [CTA block example](https://vanilla-framework-5189.demos.haus/docs/examples/patterns/cta-block/default?theme=light) and verify it still appears as expected, but the "Learn more" button is a link.

## Screenshots
image
image

[WD-12214]: https://warthogs.atlassian.net/browse/WD-12214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WD-13474]: https://warthogs.atlassian.net/browse/WD-13474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jmuzina:
I've temporarily made the CTA block `display: inline-block` using an inline style to prevent it from wrapping the link's text rather than wrapping the entire link, like this:

Before:
image

After: 
image

1. Is this functionality needed?
2. If so, is there some class in Vanilla as-is that I can use to achieve this rather than using inline styles?

Here's an example: https://vanilla-framework-5189.demos.haus/docs/examples/patterns/hero/hoc-50-50

@bartaz @pastelcyborg @lyubomir-popov

bartaz:
While we did call those "higher order components" at the time of planning, and it's still mentioned in some places, we want to call those "patterns" in the design system. So let's avoid the "hoc" in documentation, file names, etc.

bartaz:
We already have a bunch of existing examples for hero pattern, so I hoped this will be just making sure that we cover what is needed, and possibly don't need to add a bunch of completely new ones.

image

templates/docs/patterns/hero.md
```diff
@@ -304,3 +304,41 @@ For more complicated layouts utilise both columns of the 25/75% split and a nest
 Suru backgrounds are a common way to provide visual separation between hero and rest of the page.
 
 Check out the [Suru component documentation](/docs/patterns/suru) for more information on how to use it and relevant examples.
+
+## Higher Order Component
```
bartaz, line 308: "HOC" is not any specific version of hero.

Hero is a "hoc" (or pattern, as we should call it). So the whole hero documentation page is about it being a pattern. We should just make sure how existing examples for hero relate to new pattern designs, modify them if needed, only add new ones if really necessary (we already have quite a few of them).

lyubomir-popov:
> * Is this functionality needed?

yes that's what it should do, wrap as entire block.

jmuzina:
Blocked until team decision about CTA block component (#5195)

lyubomir-popov:
what does "is-shallow" do? 
image

jmuzina:
@lyubomir-popov Addressing your feedback:

- [x] `.is-shallow`: this was unnecessary and not actually doing anything - think I stole it from `p-strip`. It's been removed
- [x] Image aspect ratio classes: Added!
- [x] Paragraphs should be wrapped in `.p-section--shallow`: Updated!
- [ ] H1/H2 should be closer together
- [ ] Spell out "paragraph"
- [ ] Bottom padding should be the padding for `p-section`, not shallow.
- [ ] Stretch image to full height
- [ ] Macros: Will have to discuss this with @pastelcyborg as macros are still in-flight and he's been primarily working on those.

scss/_patterns_section.scss
```diff
@@ -7,6 +7,11 @@
     @media (min-width: $breakpoint-large) {
       padding-bottom: $spv--strip-regular;
     }
+
+    // Signpost is used with an H1 to its right, so we need to align top of the image to the top of the H1
+    .p-section__signpost-image {
```
bartaz, line 12: If there is any CSS needed for this it would not be part of section component.

"Signpost" are purely part of hero, and don't have much use outside of it as far as I understand.
We do have examples with signpost images already in Vanilla without any additional styling, so not sure why it's needed here.
https://vanillaframework.io/docs/examples/patterns/hero/hero-signpost

templates/docs/examples/patterns/hero/hero-50-50-stacked-on-medium.html
```diff
@@ -0,0 +1,35 @@
...
+          hand side.</p>
+      </div>
+      <div class="p-cta-block">
+        <a href="#" class="p-button--positive">Learn more</a>
+        <a href="#">Contact us ›</a>
+      </div>
+    </div>
+    <div class="col">
+      <div class="p-section--shallow">
+        <div class="p-image-container--16-9">
```
bartaz, line 26: I think all images need `is-cover` (if they are photograph), otherwise they don't respect aspect ratio and end up having weird spacing on sides.

```suggestion
        <div class="p-image-container--16-9 is-cover">
```

templates/docs/examples/patterns/hero/hero-50-50-vertical-image.html
```diff
@@ -0,0 +1,45 @@
...
+        </p>
+      </div>
+      <div class="p-cta-block">
+        <a href="#" class="p-button--positive">Learn more</a>
+        <a href="#">Contact us ›</a>
+      </div>
+    </div>
+    <div class="col">
+      <div class="p-section--shallow">
+        <div class="p-image-container--2-3">
```
bartaz, line 36:
```suggestion
        <div class="p-image-container--2-3 is-cover">
```

templates/docs/examples/patterns/hero/hero-50-50.html
```diff
@@ -0,0 +1,35 @@
...
+          hand side.</p>
+      </div>
+      <div class="p-cta-block">
+        <a href="#" class="p-button--positive">Learn more</a>
+        <a href="#">Contact us ›</a>
+      </div>
+    </div>
+    <div class="col">
+      <div class="p-section--shallow">
+        <div class="p-image-container--3-2">
```
bartaz, line 26:
```suggestion
        <div class="p-image-container--3-2 is-cover">
```

templates/docs/examples/patterns/hero/hero-75-25.html
```diff
@@ -0,0 +1,35 @@
...
+          hand side.</p>
+      </div>
+      <div class="p-cta-block">
+        <a href="#" class="p-button--positive">Learn more</a>
+        <a href="#">Contact us ›</a>
+      </div>
+    </div>
+    <div class="col-3">
+      <div class="p-section--shallow">
+        <div class="p-image-container--2-3">
```
bartaz, line 26:
```suggestion
        <div class="p-image-container--2-3 is-cover">
```

templates/docs/examples/patterns/hero/hero-fallback.html
```diff
@@ -0,0 +1,49 @@
...
+        </p>
+      </div>
+      <div class="p-cta-block">
+        <a href="#" class="p-button--positive">Learn more</a>
+        <a href="#">Contact us ›</a>
+      </div>
+    </div>
+    <div class="col">
+      <div class="p-section--shallow">
+        <div class="p-image-container--3-2">
```
bartaz, line 40:
```suggestion
        <div class="p-image-container--3-2 is-cover">
```

bartaz:
I feel like we have too many options and examples here, it's getting hard to understand what are the differences between those.

If the main first distinction is going to be a layout, so let's maybe focus on 50/50 only first.

What I'm missing from Figma design is a better understanding of what are the different variations and how they behave responsively.

My understanding currently is:

* 50/50
     - text content on left, image on right (optional)
     - text content on top (h1 on left, rest on the right), image on the bottom (optional)
* 75/25
     - text content on left, image on right (optional)
* 25/75
     - (optional) signpost image on left, text content on right, (optional) image on bottom
     
Anything else seems to me a content difference (aspect ratio of the image, some optional content being used or not). There are no other different layout options, or am I missing something?

scss/_patterns_section.scss
jmuzina: The separate styling comes from the padding shown here in Figma:
image

With the padding:
image

Without:
image

Existing Signpost logo example does not have this padding so the logo and the heading are not aligned. 

@lyubomir-popov Is this new padding correct?

bartaz: Whatever this padding is, it should not be part of hero pattern. 

If we need something to align images, it needs to be in image component. Some images will have whitespace on top (like many of our logos have), so setting any arbitrary top padding may depend on the image in question.

jmuzina:
@lyubomir-popov Can you please re-org the Hero Figma design in a similar way to the split list? In SU today we were having a hard time differentiating the different variants & breakpoints. 

Also, can you let us know which variants should "stack" on medium (use `-on-large`) or not?

jmuzina:
Per SU today, this PR is being split to only tackle the 50/50 variants & add macro support.

lyubomir-popov:
> @lyubomir-popov Can you please re-org the Hero Figma design in a similar way to the split list? In SU today we were having a hard time differentiating the different variants & breakpoints.

yes here https://www.figma.com/design/Hll6ZRvHyTTGBZZSoRd8ei/Higher-order-components-(HOCs)?node-id=1584-5346&t=Rs4MS0K3bl2zhzIg-1

they all have the exact same medium/small, I haven't added a split version on medium for simplicity. It's already complicated enough.

templates/_macros/cta/cta-block.jinja
```diff
@@ -0,0 +1,22 @@
+# Params
+# primary_text: primary CTA button text, shown in a primary-colored button
+# primary_url: primary CTA button URL
+# secondary_text: secondary CTA button text, shown as a link
+# secondary_url: secondary CTA button URL
+{% macro cta_block(
+  primary_text = "Learn more",
```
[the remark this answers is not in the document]
jmuzina: I don't believe so, but now I'm recalling the preference to use caller "slots" for this sort of thing and use params for flags. I'll try that out for consistency's sake - though I was running into issues calling the CTA block macro from within the hero macro earlier:
```
macro 'cta_block' was invoked with two values for the special caller argument. This is most likely a bug.
```
I'll give it another shot

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,109 @@
...
+  cta_secondary_text = "",
+  cta_secondary_url = "#"
+)-%}
+  {% set has_subtitle = subtitle_text|trim|length > 0 %}
+  {% set description_content = caller('description') %}
+  {% set has_description = description_content|trim|length > 0 %}
+  {% set has_specific_image_aspect_ratio = image_aspect_ratio|trim|length > 0 %}
+  {% set has_cta = show_cta and cta_primary_text|trim|length > 0 %}
+
+  {% macro hero_title_block() -%}
```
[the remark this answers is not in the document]
jmuzina: My thinking was that there is some content that has to live at different places in the markup depending on variants (flags) - so rather than define them and their surrounding containers etc multiple times it would be nice to simply define them after all of the macro-scoped variables etc, so the "internal" macros have access to them without needing to worry about passing them around as parameters.

This makes things for the calling portion of the macro quite convenient, and makes the main body of the macro very readable at a high level:

```html
<div class="col">
  {% if image_full_cover %}
    {{ hero_title_block() }}
  {% else %}
    <div class="p-section--shallow">
      {{ hero_title_block() }}
      {{ hero_subtitle_block() }}
    </div>
    {{ hero_description_block() }}
    {{ hero_cta_block() }}
  {% endif %}
</div>
```

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,109 @@
...
+  {% macro hero_description_block() -%}
+    {% if has_description %}
+      <div class="p-section--shallow">
+        {{ description_content }}
+      </div>
+    {% endif %}
+  {% endmacro %}
+
+  {% macro hero_image() -%}
+    <div class="p-image-container{% if has_specific_image_aspect_ratio %}--{{ image_aspect_ratio|trim }}{% endif %} {% for breakpoint in image_hidden_breakpoints %}u-hide--{{ breakpoint }} {% endfor %} is-cover">
```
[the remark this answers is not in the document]
jmuzina: Yes, I saw your image container macro - just didn't use it yet as I wasn't sure if it was still in-flight as it was only in the experimentation PR as far as I could tell.

If it doesn't complicate the scope of this PR too much, I can cherry-pick your image container macro & documentation into this PR.

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,109 @@
...
+{% macro hero(
+  title_text="",
+  subtitle_text="",
+  is_full_width_on_tablet = true,
+  image_src = "",
+  image_alt = "",
+  image_aspect_ratio = "",
+  image_full_cover = false,
+  image_hidden_breakpoints = [],
+  show_cta = false,
```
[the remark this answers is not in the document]
jmuzina: Yes - checking for truthiness and positive length of `cta_primary_text` will be sufficient

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,109 @@
...
+  {% endmacro %}
+
+  {% macro hero_image() -%}
+    <div class="p-image-container{% if has_specific_image_aspect_ratio %}--{{ image_aspect_ratio|trim }}{% endif %} {% for breakpoint in image_hidden_breakpoints %}u-hide--{{ breakpoint }} {% endfor %} is-cover">
+      <img class="p-image-container__image" src="{{ image_src }}" alt="{{ image_alt }}">
+    </div>
+  {% endmacro %}
+
+  <section class="p-section--hero">
+    <div class="row--50-50{% if is_full_width_on_tablet %}-on-large{% endif %}">
```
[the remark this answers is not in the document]
jmuzina: Do you mean something like this?
```html
<div class="{% if is_full_width_on_tablet %}row--50-50-on-large{% else %}row--50-50{% endif %}">
```

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,109 @@
...
+
+{% macro hero(
+  title_text="",
+  subtitle_text="",
+  is_full_width_on_tablet = true,
+  image_src = "",
+  image_alt = "",
+  image_aspect_ratio = "",
+  image_full_cover = false,
+  image_hidden_breakpoints = [],
```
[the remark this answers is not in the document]
jmuzina: I suppose we could accept device names instead of breakpoint names, and then have some mapping in the macro, like this

| Device identifier | Breakpoint identifier |
|-------------------|-----------------------|
| mobile            | small                 |
| tablet            | medium                |
| desktop           | large                 |


Then, input such as this could be used to hide the image on "mobile" size (small breakpoint):

```html
{% call(slot) hero(
  image_hidden_on_sizes=["mobile"]
) %}
```

Thoughts?

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,109 @@
...
+  cta_secondary_url = "#"
+)-%}
+  {% set has_subtitle = subtitle_text|trim|length > 0 %}
+  {% set description_content = caller('description') %}
+  {% set has_description = description_content|trim|length > 0 %}
+  {% set has_specific_image_aspect_ratio = image_aspect_ratio|trim|length > 0 %}
+  {% set has_cta = show_cta and cta_primary_text|trim|length > 0 %}
+
+  {% macro hero_title_block() -%}
+    <h1 class="{% if image_full_cover %}u-no-margin--bottom{% endif %}">
```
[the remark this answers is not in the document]
jmuzina: As far as I'm aware there is no need for the title to contain arbitrary HTML. Though if we did need it to, we could  use `title_text | safe`?

This one is also a bit complicated as I need to set `u-no-margin-bottom` on the header in full-cover mode (to prevent the h1-h2 being so far from eachother as [Lyubo noticed](https://github.com/canonical/vanilla-framework/pull/5189#issuecomment-2233829522)); so I need control over the element tag itself rather than letting the user pass it in.

templates/_macros/hero.jinja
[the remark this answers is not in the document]
jmuzina: Added your image container macro into this PR

templates/_macros/hero.jinja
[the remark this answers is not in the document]
jmuzina: This has been added to the PR, let me know your thoughts

templates/_macros/cta/cta-block.jinja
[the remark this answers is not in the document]
jmuzina: Fixed - I just hadn't properly implemented `caller`, as you said - thanks :)
I also switched the CTA block macro to use caller for passing in the button/link content, rather than caller. I kept the href as a parameter since that seems to make more sense.

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,115 @@
...
+# image_aspect_ratio: image aspect ratio (16/9, 3/2, 2/3, cinematic, square)
+# image_full_cover: whether the image should span the full width of the bottom of the hero section
+# image_hidden_on_devices: list of devices where the image should be hidden, e.g. ["desktop", "tablet"] to only show on small screens
+# cta_primary_url: primary CTA button URL
+# cta_secondary_url: secondary CTA button URL
+# Slots
+# description: slot for additional paragraph-style content below the title and subtitle
+
+{% macro hero(
+  title_text="",
```
bartaz, line 20: Most of the content will have to be passed to slots as it would require HTML.

For images also, we probably just need to allow a slot for image, and let devs use image container.
Especially that they will want to use image template module to serve image, not pass src to img directly.

As far as I understand the only params we want is flags that define the layout/variant of the component. Any kind of content should be HTML passed to slots.

templates/_macros/hero.jinja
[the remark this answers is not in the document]
bartaz: Hmm… that's a tricky one.

We have to allow HTML in titles (in terms of inline styling `<code>` or things like that).

But, in case of hero at least, we don't need a flexibility of changing level of headings. They are always gonna be h1 (mandatory title), and h2-sized paragraph (subtitle). We don't have to (and we don't want to) give devs tools to change those, so they can just pass text. Which in 95% cases will be just a string.

So… kind of starts to make sense to make them params instead of slots?
Of course we can also make them slots for content **inside** h1/h2, but that will be inconsistent with other slots where you are providing the "wrapping" elements, like paragraphs etc.

Let's discuss it.

templates/docs/examples/patterns/image/container/highlighted.html
```diff
@@ -1,10 +1,10 @@
...
+
 {% block title %}Image Container / Highlighted{% endblock %}
 
 {% block standalone_css %}patterns_image{% endblock %}
 
 {% block content %}
-<div class="p-image-container is-highlighted">
-  <img class="p-image-container__image" src="https://assets.ubuntu.com/v1/9b4c074f-Kernelt%20industries-80-short.png" alt="" width="250">
-</div>
+  {{ image_container('https://assets.ubuntu.com/v1/9b4c074f-Kernelt%20industries-80-short.png', width='250', alt='Sample image', is_highlighted=true) }}
```
bartaz, line 9: Please don't create macros for components yet. We don't need them at this point, and can be problematic in some cases where devs are used to do things with more freedom.

For the image case, devs are almost always using image module (https://github.com/canonical/canonicalwebteam.image-template) to create the image element. We can't create a macro that doesn't support that.

So if we want to provide one it would have to take this into account, and it's way outside the scope of this PR.

jmuzina: Altered to pass in image content slot rather than image properties, and no longer using image container macro

templates/_macros/hero.jinja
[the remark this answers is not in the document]
jmuzina: Hidden image breakpoints have been removed from the PR entirely per Bartek's suggestion; users now have responsibility to pass in image content with whatever classes they need

templates/_macros/cta/cta-block.jinja
```diff
@@ -0,0 +1,23 @@
+# Params
+# primary_url: primary CTA button URL
+# secondary_url: secondary CTA button URL
+# Slots
+# primary: primary CTA button text
+# secondary: secondary CTA button text
+{% macro cta_block(
```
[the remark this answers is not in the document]
jmuzina: CTA now uses HTML with no macros

templates/_macros/hero.jinja
jmuzina: Image content passed in by slot now. 

Title passing to be discussed as part of [this conversation](https://github.com/canonical/vanilla-framework/pull/5189#discussion_r1683908905)

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
+# Params
+# header_text: Hero header text (required)
+# subheader_text: Hero subheader text
```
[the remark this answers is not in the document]
jmuzina: Resolved

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
+# Params
+# header_text: Hero header text (required)
```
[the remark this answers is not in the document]
jmuzina: Resolved

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
...
+# description: paragraph-style content below the header and subheader
+# cta_primary: contents of the primary CTA button
+# cta_secondary: contents of the secondary CTA button
+# image: slot for image content (required)
+
+{% macro hero(
+  header_text="",
+  subheader_text="",
+  layout="fallback",
+  layout_responsive_on_tablet = false,
```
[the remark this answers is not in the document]
jmuzina: Updated to `is_`

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
...
+  layout_responsive_on_desktop = true,
+  image_full_cover = false,
+  cta_primary_url = "#",
+  cta_secondary_url = "#"
+)-%}
```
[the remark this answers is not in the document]
jmuzina: Adopted, thanks!

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
+# Params
+# header_text: Hero header text (required)
+# subheader_text: Hero subheader text
+# layout: layout of hero section. Options are '50/50', '75/25', '25/75', 'fallback'
+# layout_responsive_on_tablet: whether the layout should be applied to the tablet size specifically. Applies "-on-medium" to row class in 50/50 and 25/75 layouts
```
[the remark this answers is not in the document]
jmuzina: I struggled with how to name these parameters. They are appending `-on-medium` and `-on-large`, respectively, to the row class. 

`layout_responsive_on_tablet` makes a layout (50-50 or 25-75) apply only on medium, and stack on small and large.

`layout_responsive_on_desktop` makes the layout apply only on large, and stack on small and medium.

I may rename these to `is_split_on_tablet` and `is_split_on_desktop` as these are easier to understand. 

They have no effect on `75/25` and `fallback` currently, as those PRs will be created separately to keep scope of this PR smaller.

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
...
+  subheader_text="",
+  layout="fallback",
+  layout_responsive_on_tablet = false,
+  layout_responsive_on_desktop = true,
+  image_full_cover = false,
+  cta_primary_url = "#",
+  cta_secondary_url = "#"
+)-%}
+  {% set has_subheader = subheader_text|trim|length > 0 %}
+  {% set description_content = caller('description') %}
```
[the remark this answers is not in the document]
jmuzina: I think I prefer `_content`, as `_element` suggests you can only pass in one element to the slot, which is not true

templates/docs/examples/patterns/hero/hero-50-50-full-cover-image.html
```diff
@@ -0,0 +1,35 @@
...
+{% from "_macros/hero.jinja" import hero %}
+
+{% block header %}Hero / 50/50 / Full cover image{% endblock %}
+{% block standalone_css %}patterns_hero{% endblock %}
+
+{% set is_paper = true %}
+{% block content %}
+
+{% call(slot) hero(
+  header_text="<code>H1</code> - ideally one line, up to two",
```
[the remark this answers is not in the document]
jmuzina: Good call, adjusted

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
...
+  {% set image_full_cover = image_full_cover and layout == '50-50' %}
+
+  {# Construct grid classes from params input #}
+  {% set row_classes = [] %}
+  {% set col_classes = ['col', 'col'] %}
+  {% if layout in ['50-50', '25-75'] %}
+    {# Responsive breakpoint classes were requested #}
+    {% if layout_responsive_on_tablet or layout_responsive_on_desktop %}
+      {% if layout_responsive_on_tablet %}
+        {% set _ = row_classes.append('row--' + layout + '-on-medium') %}
```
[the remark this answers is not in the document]
jmuzina: This is a way to append to a list without adding "None" to the rendered markup.

I've replaced it with `row_classes.append() or ""`, which gets rid of the placeholder LHS (but has to use an `or ""` to [write an empty string](https://stackoverflow.com/questions/71850614/jinja2-prints-none-when-appending-to-a-loop) rather than `None`

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
+# Params
+# header_text: Hero header text (required)
+# subheader_text: Hero subheader text
+# layout: layout of hero section. Options are '50/50', '75/25', '25/75', 'fallback'
```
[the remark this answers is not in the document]
bartaz: We attempted things like that in the past without much success.

If we try to come up with some "universal" name that will suit everyone (like "variant") it's usually very vague and ends up not meaning anything.

I would focus on making things be as clear as possible on a single component level (keeping some naming conventions).

If something works out as a boolean flag (meaning, there are only 2 possible options), let's use a flag.

If there are more possible values, we need another name.

In this case I think I'm ok with layout. That's exactly what it is: you define the layout of the pattern: 50/50, 25/75, 75/25, or "fallback".

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
...
+# cta_primary: contents of the primary CTA button
+# cta_secondary: contents of the secondary CTA button
+# image: slot for image content (required)
+
+{% macro hero(
+  title_text='',
+  subtitle_text='',
+  layout="fallback",
+  is_split_on_tablet = false,
+  is_split_on_desktop = true,
```
bartaz, line 21: Do we need that flag? I suppose it's the "layout" that defines the default desktop "split", so not sure if we need options to disable it?

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
...
+# description: paragraph-style content below the title and subtitle
+# cta_primary: contents of the primary CTA button
+# cta_secondary: contents of the secondary CTA button
+# image: slot for image content (required)
+
+{% macro hero(
+  title_text='',
+  subtitle_text='',
+  layout="fallback",
+  is_split_on_tablet = false,
```
bartaz, line 20: I guess that's where we need a convention, to avoid using `is_split` in some components and `is_full_width` in others.

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
...
+  is_split_on_desktop = true,
+  is_image_full_cover = false,
```
bartaz, line 22: "Full cover" sounds like it turns on the "cover" feature on the image itself, while it's an option for layout.

I'd suggest `is_image_full_width` or `has_full_width_image`

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
...
+  cta_primary_url = '#',
```
bartaz, line 23: Sooner or later someone will need to change the style of the CTA buttons.
I think we need to make it a slot `cta_block` slot, where they just put whatever links/buttons they want.

templates/_macros/hero.jinja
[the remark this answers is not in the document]
bartaz: Is that a question about name of internal variable, or name of the slot?

Yes, we should have some general conventions for slot naming (for it to be consistent), but for internal vars it doesn't matter that much as long as it's clear.

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
...
+
+  {% macro _hero_description_block() -%}
+    {% if has_description %}
+      <div class="p-section--shallow">
+        {{ description_content }}
+      </div>
+    {% endif %}
+  {% endmacro %}
+
+  <section class="p-section--hero">
```
bartaz, line 99: I know that potentially it would be a little bit of duplication, but maybe it would be clearer to have the `{% if is_image_full_cover %}` conditional on top level? To avoid checking for it 3 times inside? Currently it's quite hard to understand what the actual structure will end up being.

Suggestion:

```html
<section class="p-section--hero">
  <div class="{{ row_classes | join(' ') }}">
    {% if is_image_full_cover %}
        <!-- full width image structure -->
    {% else %}
        <!-- other  -->
    {% endif %}
  </div>
</section>
```

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
...
+      {% if is_split_on_desktop %}
+        {{ row_classes.append('row--' + layout + '-on-large') or "" }}
+      {% endif %}
+    {% else %}
+      {{ row_classes.append('row--' + layout) or "" }}
+    {% endif %}
+  {% elif layout  == '75-25' %}
+    {# there is no .row--75-25 class, so we use unmodified row and col-9 and col-3 to create 75-25 layout #}
+    {{ row_classes.append('row') or "" }}
+    {% set col_classes = ['col-9', 'col-3'] %}
```
bartaz, line 62: If the lack of `row--75-25` is making this more complicated, lets add it.

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
...
+# is_split_on_tablet: whether the layout is split on tablet. Applies "-on-medium" to row class in 50/50 and 25/75 layouts. If false, the layout is stacked on tablet.
+# is_split_on_desktop: whether the layout is split on desktop. Applies "-on-large" to row class in 50/50 and 25/75 layouts. If false, the layout is stacked on desktop, provided is_split_on_tablet is true.
+# is_image_full_cover: whether the image should span full width of the hero. Only supported on 50/50 layout
+# cta_primary_url: primary CTA button URL
+# cta_secondary_url: secondary CTA button URL
+# Slots
+# description: paragraph-style content below the title and subtitle
+# cta_primary: contents of the primary CTA button
+# cta_secondary: contents of the secondary CTA button
+# image: slot for image content (required)
```
bartaz, line 14: Only h1 title is required. All the other content is optional.

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,127 @@
+# Params
+# title_text: Hero title text (required)
+# subtitle_text: Hero subtitle text
+# layout: layout of hero section. Options are '50/50', '75/25', '25/75', 'fallback'
+# is_split_on_tablet: whether the layout is split on tablet. Applies "-on-medium" to row class in 50/50 and 25/75 layouts. If false, the layout is stacked on tablet.
+# is_split_on_desktop: whether the layout is split on desktop. Applies "-on-large" to row class in 50/50 and 25/75 layouts. If false, the layout is stacked on desktop, provided is_split_on_tablet is true.
+# is_image_full_cover: whether the image should span full width of the hero. Only supported on 50/50 layout
```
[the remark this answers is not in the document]
jmuzina: Full cover is also available on 25-75 signpost, but it is [mandatory](https://www.figma.com/design/Hll6ZRvHyTTGBZZSoRd8ei/Higher-order-components-(HOCs)?node-id=693-12631&m=dev) there for heros that have images

So, in practice, this flag is only used in 50-50 layout; you're right that it is a bit odd to have the flag in this case. I came to the same conclusion that it's better to handle the variation in this macro with a flag than create a new macro or different layout option

templates/_macros/hero.jinja
jmuzina: This flag can be removed, I added it to support `-on-large` and `-on-medium` but it is not really needed.

templates/_macros/hero.jinja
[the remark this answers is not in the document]
jmuzina: My understanding is this is a discussion of the naming of the variable that holds slot content

templates/_macros/hero.jinja
jmuzina: changing to something like this
```html
<div class="p-cta-block">
    {{ contents }}
</div>
```

templates/_macros/hero.jinja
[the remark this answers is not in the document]
jmuzina: Desktop param removed and image full width param moved into its own layout per review with @bartaz

templates/_macros/hero.jinja
jmuzina: Working on this today as #5234

jmuzina: Removed the 75-25 code from this PR entirely, it'll be re-introduced in 75-25 PR

templates/_macros/hero.jinja
jmuzina: Good call, made changes to param docs and also made change to make sure the image slot is optional

templates/_macros/hero.jinja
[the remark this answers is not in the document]
jmuzina: This was moved to a `50/50-full-width-image` layout for now. `layout='25-75'` and `layout='50/50-full-width-image'` will both set `is_image_full_width=true`, but it will not be exposed to the caller directly; it will be set based on `layout` choice.

templates/_macros/hero.jinja
jmuzina: Per our conversation this morning, `is_full_width` is no longer an exposed param, though it would be technically possible to use a layout that sets full width with `is_split_on_medium`. We just don't have any examples that will do this.

templates/_macros/hero.jinja
jmuzina: Implemented

templates/_macros/hero.jinja
```diff
@@ -0,0 +1,111 @@
...
+# subtitle: Hero subtitle text
+# layout: layout of hero section. Options are '50/50', '50/50-full-width-image', '75/25', '25/75', 'fallback'
+# is_split_on_medium: whether the layout is split on tablet. Applies "-on-medium" to row class in 50/50 and 25/75 layouts. If false, the layout is stacked on tablet.
+# Slots
+# description: paragraph-style content below the title and subtitle
+# cta: call-to-action block below the description
+# image: slot for image content
+
+{% macro hero(
+  title='',
```
bartaz, line 12: Can we enforce passing title, by not setting the default value (not sure if that works this way in macros)?

bartaz (approved):
LGTM, good enough to merge to feature branch and build on top of it.

templates/_macros/hero.jinja
jmuzina: I'll try it out for the 25-75 PR!

PR #5191, opened 2024-06-26
Fix Percy tests being queued against PRs with Percy label when the ne…

jmuzina:
Fixes Percy tests being triggered against PRs that have the Percy label on label events when the label that was just applied is not the Percy needed label.

The logic has been refined as such:

- If the trigger was a PR synchronization (push to a live PR), Percy is run if the PR has the Percy needed label.
- If the trigger was a label being added to the PR, Percy is run if the new label is the Percy needed label.

## QA

- Review [QA PR](https://github.com/jmuzina/vanilla-framework/pull/26); its comments demonstrate all use cases covered by this change
- Verify the actions YAML code for the if conditional in the QA PR is identical to the proposed code for this PR, with the exception of the echo job that is used for QA output.

### DIFF
--- a/.github/workflows/pr-percy-prepare-label.yml
+++ b/.github/workflows/pr-percy-prepare-label.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   copy_artifact:
     name: Copy changed files to GHA artifact
-    if: "contains(toJson(github.event.pull_request.labels.*.name), 'Review: Percy needed')"
+    if: "(github.event.action == 'synchronize' && contains(toJson(github.event.pull_request.labels.*.name), 'Review: Percy needed')) || (github.event.action == 'labeled' && github.event.label.name == 'Review: Percy needed')"
     uses: ./.github/workflows/percy-prepare.yml
     with:
       pr_number: ${{ github.event.number }}


jmuzina:
@bartaz @pastelcyborg Be careful applying labels to PRs that already have the Percy label on them until this is merged - they'll run Percy tests multiple times. I've been manually cancelling runs against workflows that were already tested; we should merge this asap

.github/workflows/pr-percy-prepare-label.yml
```diff
@@ -13,7 +13,7 @@ on:
 jobs:
   copy_artifact:
     name: Copy changed files to GHA artifact
-    if: "contains(toJson(github.event.pull_request.labels.*.name), 'Review: Percy needed')"
+    if: "(github.event.action == 'synchronize' && contains(toJson(github.event.pull_request.labels.*.name), 'Review: Percy needed')) || (github.event.action == 'labeled' && github.event.label.name == 'Review: Percy needed')"
```
bartaz, line 16: Is there an "else"? Not in scope of this PR, but I wonder if there is a way for logs to tell us "skipping because label XYZ, etc", to have better understing why something runs and why it's skipped.

bartaz (approved):

.github/workflows/pr-percy-prepare-label.yml
jmuzina: @bartaz We could add an else - the only consideration is that because this is the only job in the workflow and it is skipped if this case is not met, that shows the entire workflow as skipped, which may be helpful:

image

If we add a job that *does* run in the opposite case, then this workflow will always run, and appear as a green check in PRs & the workflow list.

PR #5214, opened 2024-07-11
Add form legend required symbol

britneywwc:
## Done

- Add `is-required` class to fieldset legend

## QA

- Open https://vanilla-framework-5214.demos.haus/docs/examples/patterns/forms/fieldset-required?theme=light
- See that the asterisk shows for fieldset legend

## Screenshots

### DIFF
--- a/scss/_base_forms.scss
+++ b/scss/_base_forms.scss
@@ -228,7 +228,7 @@
 
   legend {
     &.is-required::before {
-      content: '*' !important;
+      content: '*';
       margin-right: $spv--x-small;
     }
   }
--- a/scss/_base_forms.scss
+++ b/scss/_base_forms.scss
@@ -226,4 +226,11 @@
     padding: calc($spv--large - $input-border-thickness);
   }
   // stylelint-enable selector-max-type
+
+  legend {
+    &.is-required::before {
+      content: "* " !important;
+      position: relative !important;
+    }
+  }
 }
--- a/scss/_base_forms.scss
+++ b/scss/_base_forms.scss
@@ -228,8 +228,7 @@
 
   legend {
     &.is-required::before {
-      content: '*';
-      margin-right: $spv--x-small;
+      content: '* ';
     }
   }
   // stylelint-enable selector-max-type
--- a/scss/_base_forms.scss
+++ b/scss/_base_forms.scss
@@ -228,8 +228,8 @@
 
   legend {
     &.is-required::before {
-      content: '* ' !important;
-      position: relative !important;
+      content: '*' !important;
+      margin-right: $spv--x-small;
     }
   }
   // stylelint-enable selector-max-type
--- a/scss/_base_forms.scss
+++ b/scss/_base_forms.scss
@@ -225,12 +225,12 @@
     margin-right: 0;
     padding: calc($spv--large - $input-border-thickness);
   }
-  // stylelint-enable selector-max-type
 
   legend {
     &.is-required::before {
-      content: "* " !important;
+      content: '* ' !important;
       position: relative !important;
     }
   }
+  // stylelint-enable selector-max-type
 }
--- a/scss/_base_reset.scss
+++ b/scss/_base_reset.scss
@@ -243,7 +243,7 @@
     white-space: normal; /* 1 */
     
     &.is-required::before {
-      content: '* ' !important;
+      content: "* " !important;
       position: relative !important;
     }
   }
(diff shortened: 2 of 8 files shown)


bartaz:
Thanks @britneywwc, could you share some context to it, where do we show required on fieldset level instead of input level?

britneywwc:
Hi @bartaz. Yeah sure, when checkbox/radio inputs are grouped under a fieldset and we wish to set the legend for that fieldset group to show that it's required. We set required on the input level as well but the intention behind this is to show the required label (*) on the corresponding legend.

Example use case here: https://github.com/canonical/ubuntu.com/blob/56e02eb43914c928fc20e553f75eb70ae56e4d74/templates/kernel/real-time/contact-us.html#L312

Please let me know it makes sense to add it to the legend field, it's ongoing work so I figured this could be upstreamed if useful.

bartaz:
In such case I'd suggest making an example that reflects this use case (have a group of checkboxes inside the fieldset, rather than text input).

scss/_base_reset.scss
```diff
@@ -241,6 +241,11 @@
     max-width: 100%; /* 1 */
     padding: 0; /* 3 */
     white-space: normal; /* 1 */
+    
+    &.is-required::before {
```
bartaz, line 245: The top of the file says `// DO NOT edit this file except to upgrade Normalize in it's entirety` :)

This is a CSS that comes form external reset library. It's meant to provide a clean state. No class names or specific styling should be defined here.

Please move the legend styling to `_base_form.scss`.

bartaz:
There seem to be some linting/prettier issues that fail the tests. You can run `dotrun test` locally to see them as well.

For the linter failing that we don't allow element selectors, just disable linting for that (we are doing that for all base styles, because we expect them to be on elements).

templates/docs/examples/patterns/forms/fieldset-required.html
jmuzina, line 1: Please add this new example to `templates/docs/examples/patterns/forms/combined.html` - you can just add a new `<section>` that includes this example. This will make sure it shows up in our Percy tests

jmuzina (changes requested):
Added a note about adding this to combined forms example to ensure this gets tested by Percy.

Also, I think this should set `package.json` package version to `4.15.1` as `4.15.0` has already been released and no PR has updated `main` to `4.15.1` yet.

templates/docs/examples/patterns/forms/fieldset-required.html
```diff
@@ -0,0 +1,24 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Forms / Fieldset required{% endblock %}
+
+{% block standalone_css %}patterns_forms{% endblock %}
+
+{% block content %}
+<form>
+  <fieldset>
+    <legend class="is-required" for="list-input-1">Required input</legend>
```
jmuzina, line 9: I don't think the `for` attribute is supported for the `<legend>` element ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend))

templates/docs/examples/patterns/forms/fieldset-required.html
```diff
@@ -0,0 +1,24 @@
...
+    <label class="p-radio">
+      <input type="radio" class="p-radio__input" name="radioPattern" aria-labelledby="radioExample1">
```
jmuzina, line 11: These three `span`s all have the same `id`, and the `inputs` have the same `aria-labelledby` - can we adjust so the `id` fields are unique and `aria-labelledby` references the ID of the adjacent label?

jmuzina (changes requested):

britneywwc:
Thanks for all the suggestions, I've addressed them in the recent pushes. Please have another look!

jmuzina:
Thanks @britneywwc ! Another thing - can we get an entry in releases.yml announcing the new class variant?

scss/_base_forms.scss
```diff
@@ -230,5 +230,12 @@
     margin-right: 0;
     padding: calc($spv--large - $input-border-thickness);
   }
+
+  legend {
+    &.is-required::before {
+      content: '* ' !important;
```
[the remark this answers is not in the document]
jmuzina: @pastelcyborg  This makes this component inconsistent with `label.is-required` from `_base_forms.scss` - should we also adjust the spacing to use `margin-right` there? ([src](https://github.com/canonical/vanilla-framework/blob/1525b35b87e14f5f85cfbc5d4eb892f9f8e38bff/scss/_base_forms.scss#L111-L114))

I'm inclined to say yes.

jmuzina (approved):

PR #5217, opened 2024-07-15
Add new theme variable for Positive Button focus state

### LINKED ISSUE #5194: Positive button's focus in invisible on dark theme
**Describe the bug**

image

https://vanillaframework.io/docs/examples/patterns/buttons/dark

Focus color for positive button is very dark and invisible in dark theme.

pastelcyborg:
## Done

- Add new theme color variable for positive button focus state, which combines the existing black color on light mode with the light blue color used for focus on other elements on dark mode

Fixes #5194 

## QA

- Open [Positive Button demo](https://vanilla-framework-5217.demos.haus/docs/examples/patterns/buttons/positive?theme=dark)
- Switch through themes and tab through buttons to witness focus state color
- Verify focus state color appears as expected on all themes

image

image

### DIFF
--- a/scss/_patterns_buttons.scss
+++ b/scss/_patterns_buttons.scss
@@ -80,7 +80,7 @@
       $button-disabled-border-color: $colors--theme--button-positive-default,
       $button-text-color: $colors--theme--button-positive-text
     );
-    @include vf-focus($color-focus-positive);
+    @include vf-focus($colors--theme--positive-focus);
   }
 }
 
--- a/scss/_settings_colors.scss
+++ b/scss/_settings_colors.scss
@@ -39,9 +39,9 @@ $color-link-visited-dark: #a679d2 !default;
 $color-focus-dark: #9cf !default;
 
 // Focus modifications to meet AA 3:1 contrast ratio against
-// button background for positive/negative buttons
+// button background for positive buttons
 $color-focus-positive: #003008 !default;
-$color-focus-negative: #3b0006 !default;
+$color-focus-positive-dark: #9cf !default;
 
 // Button background color changes
 $input-background-opacity-amount: 0.04;
@@ -93,6 +93,7 @@ $states: (
 // --link-default    - default link color
 // --link-visited    - visited link color
 // --focus           - focus highlight/outline color
+// --positive-focus  - focus highlight/outline color for positive buttons
 //
 // Background colors:
 // --background           - default background color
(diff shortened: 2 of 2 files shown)


lyubomir-popov:
That actually fixes an accessibility failure (adjacent colour contrast). The blue just about passes the non-text contrast requiremenet of 3.0 
image

scss/_patterns_buttons.scss
```diff
@@ -80,7 +80,7 @@
       $button-disabled-border-color: $colors--theme--button-positive-default,
       $button-text-color: $colors--theme--button-positive-text
     );
-    @include vf-focus($color-focus-positive);
+    @include vf-focus($colors--theme--positive-focus);
```
bartaz, line 83: Would be great if we don't have to introduce a new theme colour just for the sake of one variant in single component. Seems like a bit overkill.

We don't have any good means of components to "add" colours to themes (themes are defined centrally), but I wonder if we could utilise this somehow: 
https://github.com/canonical/vanilla-framework/blob/4c15e4009ba37baa3c6b7ef4a6b3f386d766a461/scss/_base_themes.scss#L47-L52

It uses a CSS variable with theme flag (light/dark) to switch between 2 values.

jmuzina: > An alternative might be adding another helper that just returns a value based on theme (that could then be passed to functions like `vf-focus`), but I think that's a bit out of scope for this PR.

Agree, this is the best way to do this but we needn't implement it here. 

Maybe we should implement this with a new variable for now, and clean things up later?

jmuzina (approved):

scss/_patterns_buttons.scss
bartaz: It's fine for now, but I kind of don't like the precedent of us fixing this by adding new theme variable for a single variant of a single component. We probably wouldn't want to add a new design token for that, but that's kind of what we are doing.

bartaz: > Unfortunately the use of `vf-focus` for focus states kind of precludes using helpers like this

@pastelcyborg Of course it would not with current `vf-focus`. This focus placeholder is an old piece of code that was clearly not created with theming in mind. At first when migrating to new theming I assumed we will still use the same focus color regardless of the theme, so this was not addressed in any way.

But it looks like we need to address it. I created https://warthogs.atlassian.net/browse/WD-13363 to cover this.

PR #5219, opened 2024-07-16
WIP - Additional macro experimentation (DO NOT MERGE)

pastelcyborg:
## Done

- New macro examples for Image Container, Hero Section, List Divided, Tabs, and Application Layout

This PR is for experimentation and proofs of concept - DO NOT MERGE.

### DIFF
--- a/templates/_macros/split-list.jinja
+++ b/templates/_macros/split-list.jinja
@@ -1,90 +1,91 @@
 # Params
-# title: split list title text
-# description: split list description text
 # is_description_full_width_on_desktop: whether description has its own row on
 #   desktop (vs. 50/50 split)
 # is_list_full_width_on_tablet: whether list title/description each have their
 #   own row on tablet (vs. 50/50 split)
-# list_items: array of content objects for list
-#   [{ 'title': string, 'description': string }]
 {% macro split_list(
-  title='',
-  description='',
   is_description_full_width_on_desktop=true,
-  is_list_full_width_on_tablet=true,
-  list_items=[{}]) -%}
-{% set has_description = description != '' %}
+  is_list_full_width_on_tablet=true) -%}
+  {% set title_content = caller('title') %}
+  {% set description_content = caller('description') %}
+  {% set has_description = description_content|trim|length > 0 %}
 
-<div class="u-fixed-width">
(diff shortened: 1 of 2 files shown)


templates/docs/examples/patterns/tabs/content.html
```diff
@@ -1,54 +1,33 @@
...
-        id="charmhub"
-        tabindex="-1"
-      >Charmhub</button>
-    </div>
-  </div>
-
-  <div tabindex="0" role="tabpanel" id="olm-tab" aria-labelledby="olm">
-    <p>A system to help you move from configuration management to application management across your hybrid cloud estate - through sharable, reusable, tiny applications called Charmed Operators.</p>
-  </div>
+{% call(content) tabs(buttons=['Charmed Operator Lifecycle Manager', 'Charmed Operator SDK', 'Charmhub'], aria_label='Juju technology') %}
```
bartaz, line 9: OK, nice. This is promising.
I guess this is the closest we can get to "slots" in macro.

But the notation worries me a bit. The purpose of this work is to make it easier for (potentially junior) web devs, so that they don't have to maintain complex HTML structures.
Now we are asking them to maintain complex macro calls using a notation that is likely even more alien to them than HTML.

bartaz: I guess one suggestion I would have is to rename `content` to `slot` (the naming used in web components). So this may turn to be a bit more "familiar".

```
{% call(slot) some_component_macro() %}
  {% if slot = 'header' %}…{% endif %}
  {% if slot = 'footer' %}…{% endif %}
{% endcall %}
```

or, if we can pass flags instead of strings from macro with `caller({ 'header': True })`.

```
{% call(slot) some_component_macro() %}
  {% if slot.header %}…{% endif %}
  {% if slot.footer %}…{% endif %}
{% endcall %}
```

bartaz:
If we are focusing our efforts on split list here in this PR, let's maybe remove unnecessary code? Or close/park this PR, and create one from scratch that would focus on one thing, as it's hard to know what too look at here.

jmuzina:
Can we close this PR to clean up the PRs list @pastelcyborg ?

PR #5233, opened 2024-07-21
Remove bots from the contributors list

### LINKED ISSUE #5098: Filter bots from contributors list
Currently there are several bot accounts that appear in our contributors list. Three of the contributors listed on our contributors page are bots, and [renovate-bot](https://github.com/renovate-bot) appears in the top 10 contributors. 

To better recognize human contributors, we should filter out bots from the contributors list. 

We use [this endpoint](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repository-contributors) to fetch contributors. The current list of contributors can be seen [here](https://api.github.com/repos/canonical/vanilla-framework/contributors). Each user has a `type` attribute that can be `User` or `Bot`. We can filter out users with `type=='Bot'`. The exception is [renovate-bot](https://github.com/renovate-bot) which, while being a bot, is not treated like a bot by Github as it has `type='User'`.

Satishpokala124:
## Done

Removes bots from the contributors list

Fixes #5098 

## QA

- Open [contributors page](insert-demo-url)
- Check if bots are removed from the contributors list.

## Screenshots

image

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -178,15 +178,27 @@ def _get_contributors():
     return contributors
 
 
-def _filter_contributors(contributors):
-    # Distinguish team_members from contributors
-
-    member_usernames = [member["login"] for member in TEAM_MEMBERS]
+def _filter_team_members_from_contributors(contributors):
+    member_usernames = {member["login"] for member in TEAM_MEMBERS}
     return [
         contributor
         for contributor in contributors
         if contributor["login"] not in member_usernames
     ]
+
+
+def _filter_bots_from_contributors(contributors):
+    return [
+        contributor
+        for contributor in contributors
+        if contributor["type"].lower() != "bot"
+    ]
+
+
+def _filter_contributors(contributors):
+    return _filter_bots_from_contributors(
+        _filter_team_members_from_contributors(contributors)
+    )
 
 
 # Global context settings
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -187,11 +187,21 @@ def _filter_team_members_from_contributors(contributors):
     ]
 
 
+def _is_not_renovate_bot(contributor):
+    # Using "login" to check for renovate-bot as
+    # GitHub treats it as a user, i.e., type: "User"
+    
+    return contributor["login"] != "renovate-bot"
+
+
 def _filter_bots_from_contributors(contributors):
     return [
         contributor
         for contributor in contributors
-        if contributor["type"].lower() != "bot"
+        if (
+            contributor["type"].lower() != "bot" 
+            and _is_not_renovate_bot(contributor)
+        )
     ]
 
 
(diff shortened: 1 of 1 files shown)


jmuzina:
Hi @Satishpokala124, thanks for the contribution! 

Could you update this to also filter out [renovate-bot](https://github.com/renovate-bot) please? It is currently the bot with the most contributions so I'd like to make sure that it is also filtered out by this change.

Satishpokala124:
Hi @jmuzina, the [renovate-bot](https://github.com/renovate-bot) is removed using its login. Let me know if that's ok.

webapp/app.py
```diff
@@ -178,17 +178,39 @@ def _get_contributors():
...
         for contributor in contributors
         if contributor["login"] not in member_usernames
     ]
 
 
+def _is_not_renovate_bot(contributor):
+    # Using "login" to check for renovate-bot as
+    # GitHub treats it as a user, i.e., type: "User"
+    
+    return contributor["login"] != "renovate-bot"
```
jmuzina, line 194: I'd rather not make a function specifically for the `renovate-bot`, and as far as I know this bot is the only special case we need to worry about currently. So for now, can you un-extract this renovate-specific function?

jmuzina: `login` can be changed, so this filtering is somewhat [brittle](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-a-user-using-their-id) in the case that renovate decides to change its username. Can we use `id` of `renovate-bot` (found [here](https://api.github.com/users/renovate-bot)) instead?

webapp/app.py
```diff
@@ -178,17 +178,39 @@ def _get_contributors():
...
+
+
+def _filter_bots_from_contributors(contributors):
+    return [
+        contributor
+        for contributor in contributors
+        if (
+            contributor["type"].lower() != "bot" 
+            and _is_not_renovate_bot(contributor)
```
jmuzina, line 203: We only have one case of this currently, so I don't think we need to create a whole extra function for special-case bots yet; how about something like this?

```suggestion
            and contributor["id"] != 25180681 # renovate-bot
```

webapp/app.py
```diff
@@ -178,17 +178,39 @@ def _get_contributors():
     return contributors
 
 
-def _filter_contributors(contributors):
-    # Distinguish team_members from contributors
```
jmuzina, line 182: Can we keep this comment about distinguishing team_members in the code please?

jmuzina (changes requested):
Left a few comments, but this is almost ready to go! :)

webapp/app.py
Satishpokala124: Got it! Using `id` instead of `login`.

webapp/app.py
Satishpokala124: Sure. I have added it back.

jmuzina (approved):
LGTM, thanks for the contribution!

@bartaz this is looking good & ready to merge

bartaz:
Thanks @Satishpokala124 for the fix and @jmuzina for review!

PR #5240, opened 2024-07-23
Bugfix: Allows for long strings to be wrapped within the notification container.

Kxiru:
## Done

- Added ```word-wrap: break-word``` to p-notification__message classes.

## QA

- Open [demo](https://vanilla-framework-5240.demos.haus/docs/examples/patterns/notifications/notifications?theme=light)
- [Add QA steps]
/docs/patterns/notification

## Screenshots

Default implementation wraps strings with spaces.
image

Default implementation does not wrap long strings with no spaces.
image

Additional scss attribute wraps long strings with no spaces:
image

### DIFF
--- a/scss/_patterns_notifications.scss
+++ b/scss/_patterns_notifications.scss
@@ -142,6 +142,7 @@ $notification-text-margin-bottom: $spv--large - $spv-nudge;
       margin: 0;
       max-width: unset;
       padding: 0;
+      word-wrap: break-word;
     }
 
     .p-notification__close {


jmuzina:
Good find! This may happen in multiple places throughout vanilla and may be a wider-reaching issue; I was able to reproduce this in [Modal](https://vanillaframework.io/docs/examples/patterns/modal/default?theme=light) also (albeit it has `overflow: auto` to scroll the long string horizontally)

[Screencast from 07-23-2024 01:19:41 PM.webm](https://github.com/user-attachments/assets/563a604b-c24f-47a7-a21d-566e639b04a1)

scss/_patterns_notifications.scss
```diff
@@ -119,6 +119,7 @@ $notification-text-margin-bottom: $spv--large - $spv-nudge;
       .p-notification__title,
       .p-notification__message {
         display: inline;
+        word-wrap: break-word;
```
bartaz, line 122: This one is likely not needed.
This is specifically for inline version, but it will inherit any styling from the default one. So if it's specified for general `p-notification__message` (as below), it will also work here.

bartaz:
> This may require a broader change to [base_typography](https://github.com/canonical/vanilla-framework/blob/90d793aaeca47e5701c1b0019bc82bce5984db2b/scss/_base_typography.scss#L11) - thoughts @bartaz @lyubomir-popov @pastelcyborg ?

While it is likely an issue with any component that has borders, I'm not sure if we want to handle this by default, or on a use case basis.

In most cases things like that are not realistic, they don't happen in natural language and are only needed for specific arbitrary strings (like tokens - which is the use case in this case as far as I understand), or in very narrow spaces (table columns).

Kxiru:
@bartaz Thanks for pointing that out. I was wondering whether or not it was going to wrap. I should have tested it then and there. In any case. Looks good!
image

I have taken that attribute out of the inline styling. Is this good to go, then? :).

Kxiru:
> Good find! This may happen in multiple places throughout Vanilla and may be a wider-reaching issue; I was able to reproduce this in [Modal](https://vanillaframework.io/docs/examples/patterns/modal/default?theme=light) also (albeit it has `overflow: auto` to scroll the long string horizontally)
> 
>  Screencast.from.07-23-2024.01.19.41.PM.webm 
> Also occurs on [pull quote](https://vanillaframework.io/docs/examples/patterns/pull-quotes/default?theme=light) image
> 
> And [card](https://vanillaframework.io/docs/examples/patterns/card/default?theme=light): image
> 
> This may require a broader change to [base_typography](https://github.com/canonical/vanilla-framework/blob/90d793aaeca47e5701c1b0019bc82bce5984db2b/scss/_base_typography.scss#L11) - thoughts @bartaz @lyubomir-popov @pastelcyborg ?

Thanks for the review! Looks like a thorough investigation, too :).

jmuzina:
There's a small CSS change here, so this should be accompanied by a [patch version number bump](https://github.com/canonical/vanilla-framework/blob/main/guides/pull-requests.md#updating-the-vanilla-version-number) in `package.json`, I think.

jmuzina:
> There's a small CSS change here, so this should be accompanied by a [patch version number bump](https://github.com/canonical/vanilla-framework/blob/main/guides/pull-requests.md#updating-the-vanilla-version-number) in `package.json`, I think.

Disregard this, we already have a release in progress :)

bartaz:
Looking into it I realised we actually have a solution for it.

Assuming your use case is a token @Kxiru, if you wrap it in `<code>` it should correclty break, because code has a word break build in:

image

Would that be an OK solution for you problem?

Kxiru:
> Looking into it I realised we actually have a solution for it.
> 
> Assuming your use case is a token @Kxiru, if you wrap it in `<code>` it should correclty break, because code has a word break build in:
> 
> image
> Would that be an OK solution for you problem?

We did think to do that but useNotify does not allow for take jsx.

bartaz:
> We did think to do that but useNotify does not allow for take jsx.

OK. Sorry to kind of going into circles with that, but assuming that we have a solution for it on Vanilla side (use `<code>` to wrap on any arbitrary strings that may need wrapping), we should allow using those in React notifications.

Obviously, allowing JSX opens up a risk of devs abusing this, but I'm not sure if we can easily limit that to just "inline" things like code.

Anyway, I guess we are back at the point to discuss a proposal of allowing JSX (not just strings) in the NotificationProvider.

So I will close this PR (sorry!). Feel free to dicuss the NotificationProvider changes in React Guild channel, or directly create a PR for that and get it discussed there.

Kxiru:
> > We did think to do that but useNotify does not allow for take jsx.
> 
> OK. Sorry to kind of going into circles with that, but assuming that we have a solution for it on Vanilla side (use `<code>` to wrap on any arbitrary strings that may need wrapping), we should allow using those in React notifications.
> 
> Obviously, allowing JSX opens up a risk of devs abusing this, but I'm not sure if we can easily limit that to just "inline" things like code.
> 
> Anyway, I guess we are back at the point to discuss a proposal of allowing JSX (not just strings) in the NotificationProvider.
> 
> So I will close this PR (sorry!). Feel free to dicuss the NotificationProvider changes in React Guild channel, or directly create a PR for that and get it discussed there.

No worries, Bartek! At this stage our current solution works for us and if anything changes or we have any broader suggestions, we will let you know!

PR #5267, opened 2024-07-31
Add color, font, and opacity Design Tokens from Core Components Figma file

pastelcyborg:
## Done

- Added new color, font, and opacity Design Tokens from [Core Components Figma file](https://www.figma.com/design/kwRMtSCI8m3lQipDeusEoI/%F0%9F%92%A0-Vanilla---Core-component-library-(Copy)?node-id=1-8&t=qShynCPuMuYCEmg8-1)

Note that any build tasks related to token compilation or updating Vanilla to use tokens will occur in separate PRs. This PR is intended to be submitted first to validate our approach can be used in Figma - I have already confirmed these tokens can be properly compiled down to usable Sass variables for use in Vanilla.

Fixes [WD-14858](https://warthogs.atlassian.net/browse/WD-14858)

## QA

- Open [Core Components Figma file](https://www.figma.com/design/kwRMtSCI8m3lQipDeusEoI/%F0%9F%92%A0-Vanilla---Core-component-library-(Copy)?node-id=1-8&t=qShynCPuMuYCEmg8-1)
- Compare local variables in color, font, and opacity categories in Figma file to values in token JSON files
- Ensure there is a matching token for each local variable in Figma file (caveat: `dimensions` and `baseline spacing` categories are not represented here and will be included in a separate PR)

[WD-14858]: https://warthogs.atlassian.net/browse/WD-14858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sd.config.json
```diff
@@ -0,0 +1,16 @@
+{
+  "source": ["tokens/**/*.json"],
+  "platforms": {
+    "scss": {
+      "transformGroup": "scss",
+      "prefix": "vf",
+      "buildPath": "build/scss/",
+      "files": [
+        {
+          "destination": "_variables.scss",
```
bartaz, line 10: Let's call them `_tokens.scss` maybe? To make it clearer when we import that the file contains?

scss/_settings_colors.scss
```diff
@@ -55,23 +60,34 @@ $active-background-opacity-percentage: $active-background-opacity-amount * 100%;
 $hover-background-opacity-percentage: $hover-background-opacity-amount * 100%;
 
 // NON-SEMANTIC COLOURS
-$color-label-validated: #006b75;
-$color-code-background: rgba($color-x-dark, 0.03);
-$color-code-background-dark: rgba($color-x-light, 0.3);
-$color-code-heading-background: rgba($color-x-dark, 0.08);
+// TODO: This is unused
+$color-label-validated: $vf-color-non-semantic-label-validated;
```
bartaz, line 64: I'm not sure at what point and why these were called "non-semantic", this may be quite a legacy header.

I'd rather not upstream it to tokens naming. Make it part of "base" or whatever. Creating a separate group doesn't seem to have much value I think.

scss/_settings_colors.scss
```diff
@@ -108,7 +124,7 @@ $states: (
 // --border-low-contrast  - low contrast version of border color, to be used when border needs more visibility (separators)
 
 // Light theme
-$colors--light-theme--text-default: #000 !default;
+$colors--light-theme--text-default: $vf-color-font-default !default;
```
bartaz, line 127: Why "font" and not "text"?

scss/_settings_colors.scss
```diff
@@ -1,47 +1,52 @@
...
+$color-x-light: $vf-color-x-light !default;
+$color-light: $vf-color-light !default;
+$color-mid-x-light: $vf-color-mid-x-light !default;
+$color-mid-light: $vf-color-mid-light !default;
+$color-mid: $vf-color-mid !default;
+$color-mid-dark: $vf-color-mid-dark !default;
+$color-dark: $vf-color-dark !default;
+$color-x-dark: $vf-color-x-dark !default;
+// TODO: This is unused
+$color-input-shadow: $vf-color-input-shadow !default;
```
bartaz, line 16: To remove a variable we would need to make sure no one else is using it as well, so it's a bit of a breaking change territory.

But good to identify them. We can either just comment on them that they are deprecated, or maybe move them all to a separate `_settings_deprecated.scss` or something,  where we keep all unused values to be removed in next major release.

scss/_settings_colors.scss
```diff
@@ -55,23 +60,34 @@ $active-background-opacity-percentage: $active-background-opacity-amount * 100%;
...
-
-$color-caution-background: #f8efe8;
-$color-caution-background--hover: #f5e7dd;
-$color-caution-background--focus: #f3e2d6;
-
-$color-negative-background: #f8e9e8;
-$color-negative-background--hover: #f5dfdd;
-$color-negative-background--focus: #f3d8d6;
+// TODO: This is unused
+$color-positive-background: $vf-color-background-positive;
```
bartaz, line 72: If this is unused, don't upstream it to tokens.

jmuzina:
> Please note that Percy is currently failing due to the new variables file not existing in its workspace - didn't want to bother updating the Github Actions until we decide where exactly all this should live.

Which file(s) would need to be copied to the Percy runner workspace?

scss/_settings_colors.scss
bartaz: Looking at the sheet from @dgtlntv, they also used `text` for text colour variables: https://docs.google.com/spreadsheets/d/1AY6BLFU-Cnj48ggHUnNVlWRxwWZ6Z-66s7jxJC1HLSU/edit?gid=1738212977#gid=1738212977

package.json
```diff
@@ -70,11 +71,13 @@
...
+    "style-dictionary": "4.0.1",
     "stylelint": "16.8.1",
     "stylelint-config-recommended-scss": "14.1.0",
     "stylelint-order": "6.0.4",
     "stylelint-prettier": "5.0.2",
     "svgo": "3.3.2",
     "yaml": "2.5.0"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
bartaz, line 82: Where is this coming from? Haven't seen it before.

scss/_settings_placeholders.scss
```diff
@@ -1,5 +1,6 @@
 @import 'settings_spacing';
 @import 'settings_colors';
+@import 'settings_colors_deprecated';
```
bartaz, line 3: Is it needed here, if they are deprecated?

scss/_settings.scss
```diff
@@ -12,6 +12,7 @@
 @import 'settings_assets';
 @import 'settings_breakpoints';
 @import 'settings_colors';
+@import 'settings_colors_deprecated';
```
bartaz, line 15: I'd suggest importing deprecated colors in `settings_colors.scss`, so that anything that imports colours gets all of them automatically.

tokens/color/link.json
```diff
@@ -0,0 +1,20 @@
+{
+  "color": {
+    "link": {
+      " ": {"value": "#06c", "type": "color", "themeable": true},
```
jmuzina, line 4: What does the empty string for key mean here and elsewhere? Is it the default i.e., `$color-link` with no modifiers?

jmuzina:
As far as I know, this should be ready to be rebased and tested with Percy

jmuzina:
Percy is completely identical to the baseline, which is a great sign - just need to check the docs as you suggested to be sure those are styled properly as well.

jmuzina:
Since SD builds are passing now, you can remove the error ignore that we added to `package.json` in #5291 

https://github.com/canonical/vanilla-framework/blob/510f79dcac1b426f145f1324c214dc2f9f03993e/package.json#L32

This can be:
```json
 "build-tokens": "style-dictionary build --verbose --config ./sd.config.json", 
```

bartaz:
@shivjoshi1996 Shall we wait out a bit with this one until the tokens discussion we have next week with designers, what do you think?

bartaz:
@pastelcyborg 
There seems to be a conflict caused by vf-focus merged previously.

Is there anythnig else this PR is waiting for? I would suggest getting it to mergable state (as is, with current values of tokens). Next step (in following PR) would be to introduce the primitive tokens defined by Max and Juan. WDYT?

package.json
```diff
@@ -29,7 +29,7 @@
   "scripts": {
     "start": "yarn build && yarn serve",
     "build": "yarn build-scss && yarn build-js && yarn build-class-references",
-    "build-tokens": "((style-dictionary build --verbose --config ./sd.config.json) || true)",
+    "build-tokens": "node ./sd.config.mjs",
```
jmuzina, line 32: You would also have to change this in `main` before this can be built on Percy, using the `|| true` trick in the first PR, then reverting it to this in this PR.

So, first PR is:
```
build-tokens": "((node ./sd.config.mjs) || true)",
```

Then this PR can remain as-is:
```
"build-tokens": "node ./sd.config.mjs",
```

.github/workflows/pr-percy-snapshots.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Remove SCM directories that will be replaced by artifact files
-        run: rm -rf templates/docs/examples/ templates/_macros/ tokens/ sd.config.json scss/
+        run: rm -rf templates/docs/examples/ templates/_macros/ tokens/ sd.config.mjs scss/
```
jmuzina, line 29: These Percy file changes will have to be merged to `main` separately before this can be tested

jmuzina (commented):
Your Percy-related changes will need to come in a separate PR that is merged to `main` ahead of this PR, like we did with #5291 

The more we do this, the less I'm a fan of it and am looking forward to doing it better in the new arch :\

sd.config.mjs
```diff
@@ -0,0 +1,130 @@
...
+  source: [`tokens/color/light/*.json`],
+  platforms: {
+    scss: {
+      transformGroup: 'vanilla/scss',
+      buildPath: 'build/scss/',
+      files: [
+        {
+          destination: 'tokens.light.scss',
+          format: 'scss/variables',
+          filter: (token) => token.path[1] === 'light',
```
jmuzina, line 98: Not sure what exactly is passed into `token` param, but is `token.path.length >= 2` a safe assumption here?

sd.config.mjs
```diff
@@ -0,0 +1,130 @@
...
+          destination: 'tokens.dark.scss',
+          format: 'scss/variables',
+          filter: (token) => token.path[1] === 'dark',
+        },
+      ],
+    },
+  },
+});
+
+await darkSd.buildAllPlatforms();
```
jmuzina, line 127: Can `darkSd` and `lightSd` be built in parallel, or are there execution order reasons that they need to be done sequentially? I'd initially expect the flow to be:
```js
await baseSd.buildAllPlatforms();
await Promise.all([lightSd.buildAllPlatforms(), darkSd.buildAllPlatforms()]);
```

jmuzina (approved):

bartaz:
Not sure how much it matters (as these are just the values for now), but I'm a bit confused by the structure.

We discussed that the theme cannot be part of tokens, but we do have separate folders per theme. Or do I miss some context here?

Also some of the values in the per-theme files seem to be exactly the same, either the same exact value (like "canonical-logo" one, or referencing the same other token. Could/should they be "extracted" to be outside of the theme?

jmuzina (approved):

PR #5304, opened 2024-08-15
WIP: Experiment with converting Tiered List Macro to use variant param (DO NOT MERGE)

pastelcyborg:
The intent of this PR is to explore changing our Macro API to use a unified `variant` param for layout permutations, rather than multiple, inconsistent string/boolean params. There are a few goals for doing this:
- Provide unified, predictable param across all our Macros
- Allow for much more flexible param that will give us the option of easily adding new variants as design specs change
- Allow for easier documentation on docs site, explaining the param and its effects in a clear manner

## Done

- Converted tiered list macro to use new `variant` param instead of multiple boolean params
- Added new `variant` documentation table to docs page

Fixes [WD-14207](https://warthogs.atlassian.net/browse/WD-14207)

## QA

- Open [demo](https://vanilla-framework-5304.demos.haus/docs/patterns/tiered-list)
- Read docs for Macro and new `variant` param

[WD-14207]: https://warthogs.atlassian.net/browse/WD-14207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/_macros/tiered-list.jinja
+++ b/templates/_macros/tiered-list.jinja
@@ -9,9 +9,7 @@
 # list_item_title_[1-25]: title element of each child list item
 # list_item_description_[1-25]: description element of each child list item
 # cta: CTA block element
-{% macro tiered_list(
-  is_description_full_width_on_desktop=true,
-  is_list_full_width_on_tablet=true) -%}
+{% macro tiered_list(variant=1) -%}
   {%- set title_content = caller('title') -%}
   {%- set description_content = caller('description') -%}
   {%- set has_description = description_content|trim|length > 0 -%}
@@ -22,7 +20,7 @@
     <hr class="u-hide--small">
 
     {% if has_description == true -%}
-      {%- if is_description_full_width_on_desktop == true -%}
+      {%- if variant in [2, 4] -%}
         <div class="u-fixed-width">
           <div class="p-section--shallow">
             {{ title_content }}
@@ -63,7 +61,7 @@
         {%- set has_description_content = list_item_description_content|trim|length > 0 -%}
 
         {%- if has_title_content and has_description_content -%}
-          {%- if is_list_full_width_on_tablet == true %}
+          {%- if variant in [1, 4] %}
             <div class="row">
               <div class="col-9 col-start-large-4">
                 <hr class="is-muted">
--- a/templates/_macros/tiered-list.jinja
+++ b/templates/_macros/tiered-list.jinja
@@ -1,8 +1,6 @@
 # Params
-# is_description_full_width_on_desktop: whether description has its own row on
-#   desktop vs. 50/50 split between title/description
-# is_list_full_width_on_tablet: whether list title/description each have their
-#   own row on tablet vs. 50/50 split between title/description
+# variant: integer representing layout variant; see docs for options:
+# https://vanillaframework.io/docs/patterns/tiered-list
 # Slots
 # title: top-level title element
 # description: top-level description element
--- a/templates/docs/examples/patterns/tiered-list/50-50-with-description.html
+++ b/templates/docs/examples/patterns/tiered-list/50-50-with-description.html
@@ -6,7 +6,7 @@
 {% block standalone_css %}patterns_all{% endblock %}
 
 {% block content %}
-{%- call(slot) tiered_list(is_description_full_width_on_desktop=false, is_list_full_width_on_tablet=false) -%}
+{%- call(slot) tiered_list(variant=3) -%}
   {%- if slot == 'title' -%}
     <h2>H2 - up to two lines; ideally one.</h2>
   {%- endif -%}
--- a/templates/docs/examples/patterns/tiered-list/full-width-with-description.html
+++ b/templates/docs/examples/patterns/tiered-list/full-width-with-description.html
@@ -6,7 +6,7 @@
 {% block standalone_css %}patterns_all{% endblock %}
 
 {% block content %}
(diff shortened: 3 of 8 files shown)


templates/docs/patterns/tiered-list/index.md
```diff
@@ -27,6 +27,27 @@ The tiered list pattern is composed of the following elements:
...
+| --------- | ---------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------- | ------------- | -------------- |
+| `1`       | [50/50 on desktop with description](#5050-on-desktop-with-description)                                                                               | Yes                           | Full-width    | 50/50          |
+| `2`       | [50/50 on tablet without description](#5050-on-tablet-without-description), <br>[50/50 on tablet with description](#5050-on-tablet-with-description) | No                            | 50/50         | Full-width     |
+| `3`       | [50/50 with description](#5050-with-description)                                                                                                     | Yes                           | 50/50         | 50/50          |
+| `4`       | [Full-width without description](#full-width-without-description), <br>[Full-width with description](#full-width-with-description)                   | No                            | Full-width    | Full-width     |
```
jmuzina, line 49: While this table is perfectly clear, I'm still not 100% comfortable with using integers to represent layouts for readability reasons.

templates/docs/examples/patterns/tiered-list/full-width-with-description.html
```diff
@@ -6,7 +6,7 @@
 {% block standalone_css %}patterns_all{% endblock %}
 
 {% block content %}
-{%- call(slot) tiered_list(is_description_full_width_on_desktop=true, is_list_full_width_on_tablet=true) -%}
+{%- call(slot) tiered_list(variant=4) -%}
   {%- if slot == 'title' -%}
```
jmuzina, line 10: My readability concern is that the user can't tell anything about what `variant=4` means just by looking at this macro call

templates/_macros/tiered-list.jinja
```diff
@@ -22,7 +18,7 @@
     <hr class="u-hide--small">
 
     {% if has_description == true -%}
-      {%- if is_description_full_width_on_desktop == true -%}
+      {%- if variant in [2, 4] -%}
```
jmuzina, line 21: Relying on integers also makes our DX here a little worse, as the developer has to check the documentation table to translate these integers to their meaning. However, we already do this for proper CSS class usage, so it's not necessarily a deal-breaker.

templates/_macros/tiered-list.jinja
```diff
@@ -1,17 +1,13 @@
...
 # Slots
 # title: top-level title element
 # description: top-level description element
 # list_item_title_[1-25]: title element of each child list item
 # list_item_description_[1-25]: description element of each child list item
 # cta: CTA block element
-{% macro tiered_list(
-  is_description_full_width_on_desktop=true,
-  is_list_full_width_on_tablet=true) -%}
+{% macro tiered_list(variant=1) -%}
```
jmuzina, line 10: What happens if the user passes in an invalid `variant`? Is it a valid default layout?

jmuzina (commented):
While this does make adding new variants a lot easier, I'm not 100% sold on this integer-based approach due to its effects on our DX & UX. I know we'll end up having a wider team discussion about this, so just leaving some thoughts in the comments below.

I wonder if there is some compromise we could come to that makes our parameter juggling a bit more manageable while also not harming DX or UX. Perhaps we could keep this macro with one layout parameter (either a string or an integer), then add macros that are more user-focused and more semantically meaningful? For example:

```html
{% macro tiered_list_full_width_desktop() %}
  {% call tiered_list(2) %}
    // slots added here
  {% endcall %}
{% endmacro %}
```

templates/docs/examples/patterns/tiered-list/50-50-with-description.html
```diff
@@ -6,7 +6,7 @@
 {% block standalone_css %}patterns_all{% endblock %}
 
 {% block content %}
-{%- call(slot) tiered_list(is_description_full_width_on_desktop=false, is_list_full_width_on_tablet=false) -%}
+{%- call(slot) tiered_list(variant=3) -%}
```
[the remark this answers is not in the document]
bartaz: While "simpler" it is completely obscured. What does the "3" mean? How would anyone know what it means and what value to choose? Would designers and developers understand them in the same way? What if we need to add another slightly different version of existing one? Throwing in just another number (like `variant=5`) if that's a new modification for responsiveness of `variant=1` seems weird and confusing.

With `is_description_full_width_on_desktop=false, is_list_full_width_on_tablet=false` I at least know more or less what options I'm setting. With `variant=3` I know nothing.

Developers are likely used to using words for naming Vanilla options, because they have to do it in CSS using class names. This way, the name of a variant is at least somewhat descriptive. This also allows us to have separate flags if there are independent options to change, rather than hiding everything behind a number.

I guess I wouldn't be so opposed to that if instead of numbers we would use some more descriptive names, like "full-width" "full-width-on-large" "full-width-on-medium" … but, with macros we can't really enforce the values with any enum, so it becomes a free string (which is quite error prone). The named parameters at least give us a bit of control over what is being passed to the macro.

Overall, I wonder why do we want to change this? Has current approach turned out to be so complex people don't know how to use it? Is the proposed one actually better in this regard?

bartaz:
Thanks @pastelcyborg but we are not gonna do it!

PR #5330, opened 2024-08-28
Add macro install instructions to docs pages

pastelcyborg:
## Done

- Add macro install instructions to Building with Vanilla page
- Add macro import instructions to pattern pages

Fixes [WD-14189](https://warthogs.atlassian.net/browse/WD-14189)

## QA

- Open the following pages:
  - [Building with Vanilla](https://vanilla-framework-5330.demos.haus/docs/building-vanilla#jinja-macros)
  - [Hero](https://vanilla-framework-5330.demos.haus/docs/patterns/hero#import)
  - [Tiered list](https://vanilla-framework-5330.demos.haus/docs/patterns/tiered-list#import)
- Confirm documentation is logical and functions as expected

### Check if PR is ready for release

If this PR contains Vanilla SCSS code changes, it should contain the following changes to make sure it's ready for the release:

- [X] PR should have one of the following labels to automatically categorise it in release notes:
  - `Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.
- [X] Vanilla version in `package.json` should be updated relative to the [most recent release](https://github.com/canonical/vanilla-framework/releases/latest), following semver convention:
  - if CSS class names are not changed it can be bugfix release (x.x.**X**)
  - if CSS class names are changed/added/removed it should be minor version (x.**X**.0)
  - see the [wiki for more details](https://github.com/canonical/vanilla-framework/wiki/Release-process#pre-release-tasks)
- [X] Any changes to component class names (new patterns, variants, removed or added features) should be listed on the [what's new page](https://github.com/canonical/vanilla-framework/blob/main/releases.yml).

[WD-14189]: https://warthogs.atlassian.net/browse/WD-14189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5329, opened 2024-08-28
Add macros to npm package; rename macros to include `vf_` namespace

pastelcyborg:
## Done

- Add `_macros/` directory to npm package for publishing to npm
- Rename macros to include `vf_` namespace in both filename and macro name

Fixes #5322 
Fixes [WD-14402](https://warthogs.atlassian.net/browse/WD-14402)

## QA

- Pull down this branch
- Run `npm publish --dry-run`
- See that macro files are now included with published files

### Check if PR is ready for release

If this PR contains Vanilla SCSS code changes, it should contain the following changes to make sure it's ready for the release:

- [X] PR should have one of the following labels to automatically categorise it in release notes:
  - `Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.
- [X] Vanilla version in `package.json` should be updated relative to the [most recent release](https://github.com/canonical/vanilla-framework/releases/latest), following semver convention:
  - if CSS class names are not changed it can be bugfix release (x.x.**X**)
  - if CSS class names are changed/added/removed it should be minor version (x.**X**.0)
  - see the [wiki for more details](https://github.com/canonical/vanilla-framework/wiki/Release-process#pre-release-tasks)
- [X] Any changes to component class names (new patterns, variants, removed or added features) should be listed on the [what's new page](https://github.com/canonical/vanilla-framework/blob/main/releases.yml).

[WD-14402]: https://warthogs.atlassian.net/browse/WD-14402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5325, opened 2024-08-26
Document hero macro

jmuzina:
## Done

- Adds documentation of the hero macro to the hero page.
- Adds some markup comments to the major blocks in the hero macro to make each element of the pattern more distinct in the HTML code snippets.
- Minor hero macro API changes per naming convention discussions

Note: this PR does not show the hero macro call. #5321 introduces the ability to view the Jinja macro call in the docs page.

Fixes [WD-14522](https://warthogs.atlassian.net/browse/WD-14522)

## QA

- Open [hero docs](https://vanilla-framework-5325.demos.haus/docs/patterns/hero)
- Review the macro documentation and verify you have a clear idea of how to use the macro.

### Check if PR is ready for release

If this PR contains Vanilla SCSS code changes, it should contain the following changes to make sure it's ready for the release:

- [x] PR should have one of the following labels to automatically categorise it in release notes:
  - `Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.
- [ ] Vanilla version in `package.json` should be updated relative to the [most recent release](https://github.com/canonical/vanilla-framework/releases/latest), following semver convention:
  - if CSS class names are not changed it can be bugfix release (x.x.**X**)
  - if CSS class names are changed/added/removed it should be minor version (x.**X**.0)
  - see the [wiki for more details](https://github.com/canonical/vanilla-framework/wiki/Release-process#pre-release-tasks)
- [ ] Any changes to component class names (new patterns, variants, removed or added features) should be listed on the [what's new page](https://github.com/canonical/vanilla-framework/blob/main/releases.yml).

## Screenshots
Screenshot 2024-08-29 at 2 24 00 PM

[WD-14522]: https://warthogs.atlassian.net/browse/WD-14522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5342, opened 2024-09-11
feat: Publish & Document Jinja Macros

jmuzina:
## Done

Feature branch PR for publishing & documenting Jinja macros

- [x] #5325 
- [x] #5329 
- [x] #5330 
- [x] #5321 
- [x] #5351 

## QA

- See QA instructions on the above PRs if you want to re-QA the feature PR.

[WD-14523]: https://warthogs.atlassian.net/browse/WD-14523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bartaz:
Looks good overall.

One downside is that the start of the page is very table-heavy, full of props you need to parse (before you even see what the pattern actually looks like).

image

I'd personally suggest moving the "Jinja Macro" section to the bottom of the page. I know we want devs to have access to this spec, but to me it seems like having the examples first would give a better overview of the component. They are gonna copy the snippet in 90% of cases anyway. Then, if they want to have a look at the reference, they can find it on the page.

Thoughts?

jmuzina:
> One downside is that the start of the page is very table-heavy, full of props you need to parse (before you even see what the pattern actually looks like). I'd personally suggest moving the "Jinja Macro" section to the bottom of the page. I know we want devs to have access to this spec, but to me it seems like having the examples first would give a better overview of the component. They are gonna copy the snippet in 90% of cases anyway. Then, if they want to have a look at the reference, they can find it on the page. Thoughts?

@bartaz Good point. I've moved it to just above the "import" section as I feel it might be a bit disjointed to read examples -> import instructions -> Jinja API docs

We could also move the Jinja API docs to their own tab / sub-page, any thoughts on putting it at bottom of page vs on a separate sub-page? @bartaz @pastelcyborg @advl

apollo13:
Hi, this looks interesting. The slot checking in the macro calling is a bit ugly imo -- the following jinja issue has an alternative that is nicer I think: https://github.com/pallets/jinja/issues/482#issuecomment-431331117  

This would kinda look like this:
```
{% macro vf_hero(
  title_text,
  subtitle_text='',
  layout="fallback",
  is_split_on_medium=false
) -%}
  {% set has_subtitle = subtitle|trim|length > 0 %}
  {% set has_subtitle = subtitle_text|trim|length > 0 %}
  {% set description_content = caller(description=True) if 'description' in caller.arguments else '' %}
  {% set cta_content = caller(cta=True) if 'cta' in caller.arguments else '' %}
```

(see the last lines) and then used in call like this:
```
{% call(description, cta) vf_hero(…) %}
{% if description %}
…
{% else if cta %}
…
{% endif %}
{% endif %}
```

Personally I think this is more readable.

bartaz (dismissed):
LGTM

jmuzina:
> Hi, this looks interesting. The slot checking in the macro calling is a bit ugly imo -- the following jinja issue has an alternative that is nicer I think: [pallets/jinja#482 (comment)](https://github.com/pallets/jinja/issues/482#issuecomment-431331117)

Hi @apollo13 , thanks for your feedback! We've discussed it as a team and decided to stick with the current approach, at least for now - we've established this style as a team standard for this opening round of macros, but will keep it in mind for the future. It's nice to have some feedback from someone who is so familiar with these!

apollo13:
Hi Julie, ok fine with me. The good thing is that if you decide to change it in the future you can support both syntaxes in parallel during a deprecation period. So it is not an end-of-world type decision.

PR #5343, opened 2024-09-11
Fix top margin of validation messages

### LINKED ISSUE #5262: Form validation and help messages don't work with dense elements
via @pierre-H https://github.com/canonical/vanilla-framework/pull/5101

When an element is dense, the help text or the validation text has a negative top margin, which should be disabled.

## Screenshots

Fix : 
image

image

jmuzina:
## Done

Removes negative top margin from form validation error messages & help text

Fixes #5262 , [WD-13179](https://warthogs.atlassian.net/browse/WD-13719)

### Open question
As implemented, `.p-form-validation__select-wrapper.is-dense` does not apply dense styling to the input element inside it. However, .p-form-validation__select-wrapper.is-dense`  is necessary because `.is-dense` must be at the level of the component that precedes `.p-form-validation__message` in order to properly remove the negative top margin.

So, we end  up with two `.is-dense` classes:
- One on `.p-form-validation__select-wrapper` to remove the negative top margin from the following validation error message
- One on ``.p-form-validation__select-wrapper .`.p-form-validation__input` in order to actually apply the dense styling to the input

This could be resolved by making `.is-dense` apply to child elements as well, but I'm not sure if that's desired.

## QA

- Open [form validation dense example](url)
  - Verify top margin removed from the validation error messages
- Open [form help text dense example](url)
  - Verify the top margin is removed from the help messages

## Screenshots
Form validation
image

Form help text
image


jmuzina:
@bartaz @pastelcyborg @advl Any thoughts on the open question in the PR description?

bartaz:
> @bartaz @pastelcyborg @advl Any thoughts on the open question in the PR description?

Form styles are probably one of the oldest in Vanilla, and likely some of component and class naming guidelines were not even fully in place when they were implemented, so it's always a bit of painful and inconsistent to maintain and update those.

This, as far as I understand, only affects select inputs, and only them (for reason I don't know) require this validation wrapper. So not to spend too much time on this, and start refactoring of any sort (we can leave that for new arch) I think it's OK to move `is-dense` to the wrapper, assuming the old way is still going to work, because you only need this wrapper when validation is in place.

So yes, `is-dense` applied to the wrapper will affect the nested select input automatically, but you can also use `is-dense` on select itself.

jmuzina:
> assuming the old way is still going to work, because you only need this wrapper when validation is in place. So yes, `is-dense` applied to the wrapper will affect the nested select input automatically, but you can also use `is-dense` on select itself.

`.is-dense` is defined as a valid flag class on any element that extends ` %vf-input-elements` .

The following selectors can be used with `.is-dense`:
https://github.com/canonical/vanilla-framework/blob/b90c2c651d5d1068d055fdde2e8a72d4f71a991a/scss/_base_forms.scss#L128-L144

So, `.is-dense` on the wrapper does not automatically apply the dense styling to its child select element, as the wrapper is just a `<div>`. A SCSS change has to be made to allow this behavior.

Screenshots from staging:
`.is-dense` on just the wrapper (has no effect)
image
`.is-dense` on the select element inside the wrapper (applies dense styling, but does not affect top margin of the validation message):
image

From this demo:
`.is-dense` on just the wrapper (applies dense styling to the select element and removes the negative top margin from the following validation error message):
image

jmuzina:
> I'm having trouble following this comment thread, but the TLDR for me is that I concur with @bartaz - applying `is-dense` to any parent element, including the `select` in this case, should automatically apply to any child elements. That seems to be what's happening in the code in this PR right now, but the comments seem to conflict with this?

@pastelcyborg 

Yes, the PR is changing things so `.is-dense` applied to the parent select wrapper applies the dense styling to the input as well. Just clarifying that a change is needed to support that, since `main` doesn't support that currently.

bartaz (dismissed):
LGTM, thanks!

bartaz:
As we are adding new way of using `is-dense` with selects and validation, probably worth mentioning it as an update in the changelog and use the `Updated` label in side nav and docs.

jmuzina:
> As we are adding new way of using `is-dense` with selects and validation, probably worth mentioning it as an update in the changelog and use the `Updated` label in side nav and docs.

@bartaz Updated. I tried adding links to the new docs sections in the changelog, but noticed an issue with using hashlinks to docs sections with status labels in their headers. It has been filed as #5393

PR #5347, opened 2024-09-12
Add support for chip links

### LINKED ISSUE #5317: Add support for chip links
**Describe the bug**

[Chip](https://vanillaframework.io/docs/patterns/chip) styles don't work great when `p-chip` is applied to an `a` element instead of a button. I'd expect the same hover styles than the button: that means slightly darker background and no underline.

**To Reproduce**

1. Go to https://vanillaframework.io/docs/patterns/chip
2. Turn any chip into an `a` element
3. Hover

**Expected behavior**

No styling difference between links and buttons

**Screenshots**

Screenshot 2024-08-21 at 17 07 47

jmuzina:
## Done

- Applies button styling to link `p-chip` elements
- Drive-by: indentation fixes for the chip variants example, and removal of the set of chips inside a dark strip (we are now capturing examples in all themes anyway, so there is no need for this)

Fixes #5317 , [WD-14348](https://warthogs.atlassian.net/browse/WD-14348)

## QA

- [Add QA steps]

## Screenshots

image


juanruitina:
LGTM!

scss/_patterns_chip.scss
```diff
@@ -73,6 +73,19 @@
 
 // Default notification styling
 @mixin vf-chips-default {
+  // stylelint-disable selector-max-type -- allow button selector for interactive chips
+
+  // Status-color chips all use the button pattern mixin to set their colors, so they already have most button styling.
+  // We need to add the default button styling to the link default chip as well.
+  a.p-chip {
+    @include vf-button-pattern;
```
bartaz, line 81: Would it make sense to make `.p-chip` include button styles by default, regardless of the type of element?

lyubomir-popov: I think we discussed but never implemented chips that are not clickable, and do not have an affordance that satisfies 3:1 contrast (meaning the border). So from that perspective, I'd say we shoudln't.

lyubomir-popov: possible caveat - if wcag has anything on chips, and clickability is mandatory in their spec, that would change things.

juanruitina (dismissed):

templates/docs/examples/patterns/chip/variants.html
```diff
@@ -10,9 +10,7 @@
 
 <span class="p-chip">
   <span class="p-chip__value">21.10</span>
-  <button class="p-chip__dismiss">
-    Dismiss
-  </button>
+  <button class="p-chip__dismiss"> Dismiss </button>
```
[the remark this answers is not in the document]
jmuzina: Fixed

scss/_patterns_chip.scss
jmuzina: @bartaz Good catch, I've updated this to unify link chip and button chip styling

scss/_patterns_chip.scss
```diff
@@ -71,8 +71,38 @@
...
+  }
+}
+
 // Default notification styling
 @mixin vf-chips-default {
+  // stylelint-disable selector-max-type -- allow button selector for interactive chips
+
+  // Status-color chips all use the button pattern mixin to set their colors, so they already have most button styling.
+  // We need to add the default button styling to the link default chip as well.
+  a.p-chip {
```
bartaz, line 100: As this is more specific than `.p-chip` would it make sense to move it below? Somewhere alongside the `button.p-chip` (this seems very similar).

jmuzina: Done

templates/docs/examples/patterns/chip/link.html
```diff
@@ -0,0 +1,111 @@
...
+  </a>
+
+  <a href="#" class="p-chip" aria-pressed="true">
+    <span class="p-chip__value">Selected</span>
+  </a>
+
+  <a href="#" class="p-chip">
+    <span class="p-chip__lead">Owner</span>
+    <span class="p-chip__value">Bob</span>
+    <button class="p-chip__dismiss">Dismiss</button>
```
bartaz, line 23: I don't think we should allow this. Having nested interactive/clickable elements leads to confusing UX and accessibility issues.

With button chips we don't do that. If chip is dismissable, it's a span with a button inside it. It should not be interactive as a whole in such case.
I don't think that link chips are any different in this regards.

I guess nothing in the code/styling needs to change, but we should remove examples that use buttons inside links

bartaz (changes requested):
Please remove examples of link chips with button inside of them.

templates/docs/examples/patterns/chip/link.html
jmuzina: Fixed

bartaz (approved):
Looks good, thanks!

PR #5352, opened 2024-09-17
Feat: Implement grid with 4 columns on medium

jmuzina:
## Done

Implements the 4-column grid on medium, with `.row--4-cols`.

`.row--4-cols` is a grid row that shares most of the behavior of `.row`, but has 4 columns on medium screens, instead of 6.

Fixes [WD-14972](https://warthogs.atlassian.net/browse/WD-14972)

## QA

- Review all 4-column-medium examples. Be sure to verify that they look correct on all breakpoints.
  - [4 columns on medium](https://vanilla-framework-5352.demos.haus/docs/examples/patterns/grid/4-columns-medium-responsive?theme=light)
  - [4 columns on medium, mixed responsive](https://vanilla-framework-5352.demos.haus/docs/examples/patterns/grid/4-columns-medium-mixed-responsive?theme=light)
  - [4 columns on medium, with offsets](https://vanilla-framework-5352.demos.haus/docs/examples/patterns/grid/4-columns-medium-offsets-responsive?theme=light)
  - [4 columns on medium, with reordering](https://vanilla-framework-5352.demos.haus/docs/examples/patterns/grid/4-columns-medium-reordering-responsive?theme=light)
  - [4 columns on medium, with nesting](https://vanilla-framework-5352.demos.haus/docs/examples/patterns/grid/4-columns-medium-nested-responsive?theme=light)
  - [Comparison of 4-columns on medium and 6-columns on medium](https://vanilla-framework-5352.demos.haus/docs/examples/patterns/grid/4-columns-medium-6-columns-medium-comparison-responsive?theme=light)
- Review [4 col medium grid docs](https://vanilla-framework-5352.demos.haus/docs/patterns/grid#4-column-grid)
- Review [4.17.0 release notes](https://vanilla-framework-5352.demos.haus/docs/whats-new)

## Screenshots
image

[WD-14972]: https://warthogs.atlassian.net/browse/WD-14972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scss/_base_grid-definitions.scss
+++ b/scss/_base_grid-definitions.scss
@@ -5,6 +5,31 @@
   padding-right: 0;
 }
 
+// Apply grid properties for a given breakpoint
+@mixin vf-b-grid-template-impl($breakpoint, $columns) {
+  grid-gap: 0 map-get($grid-gutter-widths, $breakpoint);
+  grid-template-columns: repeat($columns, minmax(0, 1fr));
+  // Expose the number of columns available for child elements as a CSS var, so that columns can span full grid width without increasing selector specificity\
+  --vf-grid-columns: #{$columns};
+
+  & > * {
+    grid-column-end: span $columns;
+  }
+}
+
+// Apply grid properties for a given breakpoint. If a min_width is provided, wrap the properties in a media query
+@mixin vf-b-grid-template($breakpoint, $columns, $min_width: None) {
+  @supports (display: grid) {
+    @if $min-width == None {
+      @include vf-b-grid-template-impl($breakpoint, $columns);
+    } @else {
+      @media (min-width: $min_width) {
+        @include vf-b-grid-template-impl($breakpoint, $columns);
+      }
+    }
+  }
+}
+
 @mixin vf-b-grid-definitions {
   %display-block {
     // make columns explicitly display:block; the use of a placeholder is to ensure the rule appears only once in the compiled css
@@ -29,45 +54,50 @@
     }
   }
 
-  %vf-row {
+  %vf-row-small {
+    @include vf-b-grid-template(small, $grid-columns-small);
+  }
+
+  %vf-row-medium-4 {
+    @include vf-b-grid-template(default, $grid-columns-medium-4, $threshold-4-small-4-med-col);
+  }
+
+  %vf-row-medium-6 {
+    @include vf-b-grid-template(default, $grid-columns-medium, $threshold-4-6-col);
+  }
+
+  %vf-row-large {
+    @include vf-b-grid-template(default, $grid-columns, $threshold-6-12-col);
+  }
+
+  %vf-row-base {
     @extend %fixed-width-container;
 
     @supports (display: grid) {
       display: grid;
-      grid-gap: 0 map-get($grid-gutter-widths, small);
-      grid-template-columns: repeat($grid-columns-small, minmax(0, 1fr));
       grid-template-rows: auto;
       margin-left: auto;
       margin-right: auto;
       max-width: $grid-max-width;
 
-      & > * {
-        grid-column-end: span $grid-columns-small;
-      }
-
       [class*='#{$grid-column-prefix}'] {
         grid-column-start: auto;
       }
-
-      // set static gutter width per breakpoint
-      @media (min-width: $threshold-4-6-col) {
-        grid-gap: 0 map-get($grid-gutter-widths, default);
-        grid-template-columns: repeat($grid-columns-medium, minmax(0, 1fr));
-
-        & > * {
-          grid-column-end: span $grid-columns-medium;
-        }
-      }
-
-      @media (min-width: $threshold-6-12-col) {
-        grid-gap: 0 map-get($grid-gutter-widths, default);
-        grid-template-columns: repeat($grid-columns, minmax(0, 1fr));
-
-        & > * {
-          grid-column-end: span $grid-columns;
-        }
-      }
     }
+  }
+
+  %vf-row {
+    @extend %vf-row-base;
+    @extend %vf-row-small;
+    @extend %vf-row-medium-6;
+    @extend %vf-row-large;
+  }
+
+  %vf-row-medium-4 {
+    @extend %vf-row-base;
+    @extend %vf-row-small;
+    @extend %vf-row-medium-4;
+    @extend %vf-row-large;
   }
 
   %vf-grid-container-padding {
--- a/scss/_patterns_grid.scss
+++ b/scss/_patterns_grid.scss
@@ -3,20 +3,25 @@
 // CSS grid implementation of columns for all screens sizes
 @mixin vf-grid-column($col) {
   @supports (display: grid) {
-    grid-column-end: span #{$col};
+    // Get the smaller of the parent column span count or the grid row's maximum column count
+    // This prevents columns from using more columns than their grid row allows
+    --vf-grid-colspan: min(#{$col}, var(--vf-grid-columns));
+
+    grid-column-end: span var(--vf-grid-colspan);
 
     //nesting
     @if $col > 1 {
-      & .row {
-        grid-template-columns: repeat($col, minmax(0, 1fr));
+      & .#{$grid-row-class},
+      & .#{$grid-row-4-cols-class} {
+        grid-template-columns: repeat(var(--vf-grid-colspan), minmax(0, 1fr));
       }
     }
   }
 }
 
-@mixin vf-grid-column-reordering($label, $col-count, $reset: false) {
+@mixin vf-grid-column-reordering($label, $col-count, $reset: false, $row-class: $grid-row-class) {
   @for $i from 1 through $col-count {
-    .row [class*='#{$grid-column-prefix}'].#{$grid-column-prefix}start-#{$label}-#{$i} {
+    .#{$row-class} [class*='#{$grid-column-prefix}'].#{$grid-column-prefix}start-#{$label}-#{$i} {
       @if $reset {
         grid-column-start: initial;
       } @else {
@@ -34,6 +39,17 @@
   }
 }
 
+// span full width on some breakpoint (to match block level element behaviour)
+@mixin vf-grid-span-full-grid($min-width: None) {
+  @if $min-width == None {
+    grid-column: auto / span var(--vf-grid-columns);
+  } @else {
+    @media (min-width: $min-width) {
+      grid-column: auto / span var(--vf-grid-columns);
+    }
+  }
+}
+
 @mixin vf-p-grid {
   // FIXME: this should be removed from framework SCSS
   // (see https://github.com/canonical/vanilla-framework/issues/3199)
@@ -43,18 +59,29 @@
     margin-bottom: $spv--small;
   }
 
-  .row {
+  .#{$grid-row-class} {
     @extend %vf-row;
+  }
+
+  .#{$grid-row-4-cols-class} {
+    @extend %vf-row-medium-4;
   }
 
   // mobile grid
 
   // by default medium and large screen col classes should span full width on mobile (to match block level element behaviour)
   %span-full-grid-on-mobile {
-    grid-column: auto / span $grid-columns-small;
+    @include vf-grid-span-full-grid;
   }
 
   @for $i from 1 through $grid-columns-medium {
+    .#{$grid-medium-col-prefix}#{$i} {
+      @extend %span-full-grid-on-mobile;
+      @extend %display-block;
+    }
+  }
+
+  @for $i from 1 through $grid-columns-medium-4 {
     .#{$grid-medium-col-prefix}#{$i} {
       @extend %span-full-grid-on-mobile;
       @extend %display-block;
@@ -79,12 +106,9 @@
   }
 
   // tablet grid
-
   // on medium/tablet screens, small and large grid class names span full width (to match block level element behaviour)
   %span-full-grid-on-tablet {
-    @media (min-width: $threshold-4-6-col) {
-      grid-column: auto / span $grid-columns-medium;
-    }
+    @include vf-grid-span-full-grid($threshold-4-small-4-med-col);
   }
 
   @for $i from 1 through $grid-columns-small {
@@ -100,8 +124,20 @@
   }
 
   // col-medium-X classes define grid for medium screens
+  // col-medium-X classes for grids with 6 columns on medium
   @media (min-width: $threshold-4-6-col) {
     @for $i from $grid-columns-medium through 1 {
+      .#{$grid-medium-col-prefix}#{$i} {
+        @include vf-grid-column($i);
+
+        width: 100%;
+      }
+    }
+  }
+
+  // col-medium-X classes for grids with 4 columns on medium
+  @media (min-width: $threshold-4-small-4-med-col) {
+    @for $i from $grid-columns-medium-4 through 1 {
       .#{$grid-medium-col-prefix}#{$i} {
         @include vf-grid-column($i);
 
@@ -114,9 +150,7 @@
 
   // on desktop screens small and medium grid class names span full width (to match block level element behaviour)
   %span-full-grid-on-desktop {
-    @media (min-width: $threshold-6-12-col) {
-      grid-column: auto / span $grid-columns;
-    }
+    @include vf-grid-span-full-grid($threshold-4-med-12-col);
   }
 
   @for $i from 1 through $grid-columns-small {
@@ -131,8 +165,14 @@
     }
   }
 
+  @for $i from 1 through $grid-columns-medium-4 {
+    .#{$grid-medium-col-prefix}#{$i} {
+      @extend %span-full-grid-on-desktop;
+    }
+  }
+
   // col-X class names define grid on large/desktop screens
-  @media (min-width: $threshold-6-12-col) {
+  @media (min-width: $threshold-4-med-12-col) {
     @for $i from $grid-columns through 1 {
       // set col-* to span respective number of columns on desktop
       .#{$grid-large-col-prefix}#{$i} {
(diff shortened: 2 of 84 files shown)


scss/_patterns_grid.scss
```diff
@@ -87,6 +102,12 @@
...
 
+  %span-full-grid-on-tablet-new {
+    @media (min-width: $threshold-4-small-4-med-col) {
+      grid-column: auto / span $grid-columns-medium-new;
+    }
+  }
+
   @for $i from 1 through $grid-columns-small {
     .#{$grid-small-col-prefix}#{$i} {
       @extend %span-full-grid-on-tablet;
```
jmuzina, line 113: May need to extend the span full 4-col-tablet-grid placeholder here as well, for consistency

lyubomir-popov:
looks great to me. questions around naming:
- what is the shortest name we can come up with? `.vf-row` maybe?
- once inside a grid-row, do we need separate col classnames? maybe the original ones can be reused as the context is set by the row

scss/_patterns_grid.scss
```diff
@@ -87,6 +102,12 @@
     }
   }
 
+  %span-full-grid-on-tablet-new {
+    @media (min-width: $threshold-4-small-4-med-col) {
+      grid-column: auto / span $grid-columns-medium-new;
```
jmuzina, line 107: @bartaz @pastelcyborg @lyubomir-popov  This behavior here (and in another handful of "spanning placeholders") makes it hard to use `.col` instead of `.grid-col` - the `.col` class needs to know how many columns each breakpoint has available, so that the full grid can be spanned.

I suppose we could modify this so that `.row` children that extend the spanning placeholders span 4/6/12, and `.grid-row` children span 4/4/12. 

We could also keep the existing `.grid-col` classes. 

Any thoughts?

scss/_patterns_grid.scss
jmuzina: Fixed

scss/_patterns_grid.scss
lyubomir-popov: not sure I understand - this ensures elements with no explicitly defined col-* classes occupy the entire available space. as long as the variable  on line 107 specifies the correct number of columns available, an arbitrary div placed there should fill the available space IIRC. So how does this get in the way of using say `.col-medium-3`?

jmuzina: Here's a larger code snippet that shows placeholder definition and usage, from current `main`

https://github.com/canonical/vanilla-framework/blob/b90c2c651d5d1068d055fdde2e8a72d4f71a991a/scss/_patterns_grid.scss#L84-L94

This is setting `col-small-x` to span full grid width on the medium breakpoint, if no medium grid classes are used. So the placeholder must know how many columns to span, depending on the grid.

Now that I think about it more, we could maybe use `grid-column: 1 / -1` to automatically span all columns, without needing to know how many columns to span. I'll try that out

jmuzina: `1 / -1` and `auto / -1` ended up not working as negative grid positions are [only supported in explicit grids](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column#integer_custom-ident). 

To work around this, I first tried to use parent-child selectors, like this:
```scss
  %span-full-grid-on-tablet {
    @media (min-width: $threshold-4-small-4-med-col) {
      [class^='#{$grid-row-class}'] > & {
        grid-column: auto / span $grid-columns-medium;
      }
      [class^='#{$grid-row-4-cols-class}'] > & {
        grid-column: auto / span $grid-columns-medium-new;
      }
    }
  }
```

However, this increase in selector specificity necessitates changes to `vf-grid-column`:
```scss
@mixin vf-grid-column($col) {
  @supports (display: grid) {
    [class^='#{$grid-row-class}'] > & {
      grid-column-end: span #{$col};

      //nesting
      @if $col > 1 {
        & .#{$grid-row-class},
        & .#{$grid-row-4-cols-class} {
          grid-template-columns: repeat($col, minmax(0, 1fr));
        }
      }
    }
  }
}
```
All these increases in specificity added up, causing tests to fail due to large stylesheet size (failing the Parker tests). 

I've implemented another workaround: setting a CSS attribute `--vf-grid-columns` that spanning placeholders can then use to span full grid width without increasing selector specificity or increasing stylesheet size too much:

```scss
@mixin vf-b-grid-template-impl($breakpoint, $columns) {
  grid-gap: 0 map-get($grid-gutter-widths, $breakpoint);
  grid-template-columns: repeat($columns, minmax(0, 1fr));
  // Expose the number of columns available for child elements as a CSS var, so that columns can span full grid width without increasing selector specificity\
  --vf-grid-columns: #{$columns};
  // .....
}
```
```scss
  %span-full-grid-on-tablet {
    @media (min-width: $threshold-4-small-4-med-col) {
      grid-column: auto / span var(--vf-grid-columns);
    }
  }
```

jmuzina:
Implementation for sharing the `col-` class across both grids took a bit longer than I thought it would - will clean this up more tomorrow to hopefully get it out of draft stage

scss/_settings_grid.scss
```diff
@@ -5,15 +5,29 @@
...
 $grid-columns: 12 !default;
+
 $threshold-4-6-col: $breakpoint-small !default;
+$threshold-4-small-4-med-col: $breakpoint-small !default;
 $threshold-6-12-col: $breakpoint-large !default;
+$threshold-4-med-12-col: $breakpoint-large !default;
 
 $grid-column-prefix: 'col-' !default;
+$grid-row-class: 'row' !default;
+$grid-row-4-cols-class: '#{$grid-row-class}--4-cols' !default;
```
bartaz, line 18: If we don't have a better semantic name for it, I'd suggest calling it `row--medium-4-cols` or `row--4-cols-medium` to at least make it clearer in the name that 4 cols change only affects medium screen size.

scss/_settings_grid.scss
```diff
@@ -5,15 +5,29 @@
 $grid-max-width: 80rem !default;
 $grid-columns-small: 4 !default;
 $grid-columns-medium: 6 !default;
+$grid-columns-medium-new: 4 !default;
```
bartaz, line 8: Please rename to denote it's 4 cols, like `grid-columns-medium-4` or something, I know it's weird, because it technically just aligns the variable name with it's value, but in the end that's what we are doing.

And `new` doesn't say anything, and will get "old" pretty quick.

scss/_settings_grid.scss
```diff
@@ -5,15 +5,29 @@
...
 $grid-column-prefix: 'col-' !default;
+$grid-row-class: 'row' !default;
+$grid-row-4-cols-class: '#{$grid-row-class}--4-cols' !default;
 $grid-small-col-prefix: '#{$grid-column-prefix}small-' !default;
 $grid-medium-col-prefix: '#{$grid-column-prefix}medium-' !default;
 $grid-large-col-prefix: #{$grid-column-prefix} !default;
 
+// Common grid patterns
+$col-50-medium: calc($grid-columns-medium / 2);
+$col-50-medium-new: calc($grid-columns-medium-new / 2);
```
bartaz, line 25: Not sure if we need to complicate this. We know exactly how many columns is there.

```suggestion
$col-50-medium-4-cols: calc(4 / 2);
```

Maybe even directly say 
```suggestion
$col-50-medium-4-cols: 2;
```

templates/docs/patterns/grid/index.md
```diff
@@ -53,6 +53,16 @@ Read also: [Breakpoints](/docs/settings/breakpoint-settings)
     View example of the default grid
 </a></div>
 
+## 4-column grid
```
bartaz, line 56: Would make sense to mention "medium" screen size somewhere I think.

scss/_settings_grid.scss
```diff
@@ -5,15 +5,29 @@
...
+
+$col-25: calc($grid-columns / 4);
+$col-50: calc($grid-columns / 2);
+$col-75: calc(($grid-columns / 4) * 3);
```
bartaz, line 29: This exposes a lot of variables to global SCSS scope.

If they are needed in a single file/mixin, would be good to keep it there.
If they need to be global, worth namespacing them a bit, at least with `$grid-` prefix.

bartaz (commented):
Couple naming suggestions in inline comments. But the overall approach is solid I think, thanks!

scss/_settings_grid.scss
jmuzina: Renamed

scss/_settings_grid.scss
jmuzina: Removed this variable as it is not needed currently.

scss/_settings_grid.scss
jmuzina: Moved these back to the grid pattern's scope

templates/docs/patterns/grid/index.md
jmuzina: Updated!

scss/_settings_grid.scss
jmuzina: Updated to `row--4-cols-medium`

scss/_base_grid-definitions.scss
```diff
@@ -5,6 +5,31 @@
   padding-right: 0;
 }
 
+// Apply grid properties for a given breakpoint
+@mixin vf-b-grid-template-impl($breakpoint, $columns) {
```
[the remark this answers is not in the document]
jmuzina: I intend it to mean "implementation" - I am using `vf-b-grid-template` as an abstraction over the implementation because it has a conditionally-applied media query, and the styles applied are many lines long, so an abstraction saves us from some repetition:

I.E., we get this:

```scss
// Apply grid properties for a given breakpoint
@mixin vf-b-grid-template-impl($breakpoint, $columns) {
  grid-gap: 0 map-get($grid-gutter-widths, $breakpoint);
  grid-template-columns: repeat($columns, minmax(0, 1fr));
  // Expose the number of columns available for child elements as a CSS var, so that columns can span full grid width without increasing selector specificity\
  --vf-grid-columns: #{$columns};

  & > * {
    grid-column-end: span$columns;
  }
}

// Apply grid properties for a given breakpoint. If a min_width is provided, wrap the properties in a media query
@mixin vf-b-grid-template($breakpoint, $columns, $min_width: null) {
  @supports (display: grid) {
    @if $min_width {
      @media (min-width: $min_width) {
        @include vf-b-grid-template-impl($breakpoint, $columns);
      }
    } @else {
      @include vf-b-grid-template-impl($breakpoint, $columns);
    }
  }
}
```
Instead of this:
```scss
// Apply grid properties for a given breakpoint. If a min_width is provided, wrap the properties in a media query
@mixin vf-b-grid-template($breakpoint, $columns, $min_width: null) {
  @supports (display: grid) {
    @if $min_width {
      @media (min-width: $min_width) {
         grid-gap: 0 map-get($grid-gutter-widths, $breakpoint);
         grid-template-columns: repeat($columns, minmax(0, 1fr));
          --vf-grid-columns: #{$columns};
        
          & > * {
            grid-column-end: span $columns;
          }
      }
    } @else {
         grid-gap: 0 map-get($grid-gutter-widths, $breakpoint);
         grid-template-columns: repeat($columns, minmax(0, 1fr));
          --vf-grid-columns: #{$columns};
        
          & > * {
            grid-column-end: span $columns;
          }
    }
  }
}
```

templates/docs/patterns/grid/index.md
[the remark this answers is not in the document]
jmuzina: Good point, updated the table to show 6 vs 4 columns depending on class

Screenshot 2024-09-27 at 4 12 50 PM

scss/_base_grid-definitions.scss
```diff
@@ -5,6 +5,31 @@
...
+
+  & > * {
+    grid-column-end: span$columns;
+  }
+}
+
+// Apply grid properties for a given breakpoint. If a min_width is provided, wrap the properties in a media query
+@mixin vf-b-grid-template($breakpoint, $columns, $min_width: null) {
+  @supports (display: grid) {
+    @if $min_width {
```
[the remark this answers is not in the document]
jmuzina: Good call, switched to using `$min-width == None`

scss/_base_grid-definitions.scss
```diff
@@ -5,6 +5,31 @@
...
 
+// Apply grid properties for a given breakpoint
+@mixin vf-b-grid-template-impl($breakpoint, $columns) {
+  grid-gap: 0 map-get($grid-gutter-widths, $breakpoint);
+  grid-template-columns: repeat($columns, minmax(0, 1fr));
+  // Expose the number of columns available for child elements as a CSS var, so that columns can span full grid width without increasing selector specificity\
+  --vf-grid-columns: #{$columns};
+
+  & > * {
+    grid-column-end: span$columns;
```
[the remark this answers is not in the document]
jmuzina: Fixed

jmuzina:
@pastelcyborg 

>     1. If it weren't for the code sample for the new 4-column variant, I would find the documentation/guidance on grids now quite confusing. I'm wondering if there's some way to more clearly indicate that there are essentially 2 top-level grids now, and that the 4-column variant is newer and recommended? It almost makes me think we should consider adding a deprecation notice to the old grid...

I have updated the structure table to show the difference in the two different grid classes, maybe this helps some:
Screenshot 2024-09-27 at 4 12 50 PM

In terms of stating that the 4-column variant is newer and recommended, I will end up using your "New" label from #5353 for sure. Also, we won't really be "deprecating" `.row` per-se in v5 - my understanding is that we will switch `.row` over to have 4 columns on medium (basically, `.row--4-cols` is a preview of the new `.row`). We just can't switch over the functionality in a minor version as it's breaking. 

Maybe we could add some language there along the lines of "4 columns on medium will become the default behavior of `.row` in Vanilla v5. Please use `.row--4-cols` to build layouts with multiples of 4 columns on all breakpoints to avoid unexpected layout disruptions when v5 released."?


>  2. There's only one example using the new 4-column variant - it might be nice to include some more examples in the docs themselves.

Not sure if I understand you fully here - I have made 6 examples that use the 4-column variant. See the PR description.

Do you mean that we should embed more of them in the docs page?

bartaz:
As discussed in standup, let's document it as a new default.

This means:
- update the grid documentation accordingly, mark `.row` as legacy/deprecated and not recommended to be used on new pages
- update Vanilla examples where relevant to feature new `row-4-cols-medium` by default
- update Vanilla website to use new row where relevant

(I'm not sure if we need to do it all in one PR, but I guess would be good to make them part of the same release)

jmuzina:
@bartaz I've updated the PR as requested, have a look and let me know what you think.

bartaz:
Now that we made the new `row--4-col-medium` a de-facto new default I kinda dislike how verbose the name is. I can't imagine how annoying it will be to use it all the time instead of `row`. We need to simplify it I think.

But I can't see a very good alternative:

- `p-row`
- `grid-row`
- `row--new`, `row--next`
- `row--4`, `row--4c`, `row--4-col`
- …

templates/docs/patterns/grid/index.md
```diff
@@ -25,10 +25,13 @@ Vanilla has a responsive grid with the following columns and gutters:
...
-      <td>6</td>
-      <td>2.0rem</td>
-      <td>1.5rem</td>
+      <td rowspan="2"><code>$breakpoint-small</code> - <code>$breakpoint-large</code></td>
+      <td>6 (using <code>.row</code>)</td>
+      <td rowspan="2">2.0rem</td>
+      <td rowspan="2">1.5rem</td>
+    </tr>
+    <tr>
+      <td>4 (using <code>.row--4-cols-medium</code> - see <a href="#4-column-grid-on-medium-screens">docs</a>).</td>
```
bartaz, line 34: If that's the new default we should list it first. And maybe add `deprecated` red status label to the cell listing `.row`

jmuzina:
This afternoon I did some exploration of updating the common grid patterns to use 4 cols on medium, as discussed in review today. 

Attempting to change the behavior of the common grid patterns (grid shorthands) to use 4 columns on medium is proving to be much more difficult than I thought it would be. Updating the base case (no nested grid) is trivial, but handling grid nesting inside these shorthands is difficult to do as a shorthand needs to be able to handle the existing `.row` (which sets itself to use 6 cols on medium) and the new `.p-row`. 

Essentially this creates a case where `.row` sometimes has up to 6 cols on medium (if it is not nested), and sometimes has up to 4 cols on medium (when it is nested inside a shorthand grid). This creates a *lot* of extra complication and is very brittle.

I suspect it may be easier (if messier) to introduce separate `.p-` prefixed shorthand classes (and expect people to update their nested rows inside these shorthands to also use `.p-row`). 

I have updated this PR to change the class of the 4 col grid to `p-row` as discussed, and updated the documentation to more explicitly state that the old `.row` is to be deprecated.

@bartaz @pastelcyborg

bartaz:
> Essentially this creates a case where `.row` sometimes has up to 6 cols on medium (if it is not nested), and sometimes has up to 4 cols on medium (when it is nested inside a shorthand grid). This creates a _lot_ of extra complication and is very brittle.

I'm getting to the point where I feel we should not do it in current codebase.

If this issue would be to go through out new triaging process it would not land in v4 due to high effort and compatibility issues.

Having these 2 grids working side by side would be extremely confusing and problematic for devs.

I see 2 options:
- update existing grid `row` class to be 4 columns and deal with any breaking changes (find them and fix them if needed)
- don't do it in current codebase and leave it for new architecture

bartaz:
To summarise recent decisions made:

- we don't want to postpone the new grid to new architecture
- we want to separate the implementation of the new grid and the old one, for clearer separation
  - this means having a copy of `_patterns_grid.scss` file for current and new grid
  - the current/old grid will use hardcoded values for column counts (4/6/12)
  - new grid will use values from grid settings that we will set to 4/4/8 (notice both medium and large screen grid changes!)
  - we need to take care of `%vf-row` and likely duplicate it as well
- new grid needs to use new class names for both row and columns (`grid-row`, `grid-col-` - that's the initial suggestion, Josh suggested using something more semantic if possible)
- we currently don't need to implement shorthands (like `--50-50` etc) for the new grid
- existing components that use grid don't need to be updated right away, we can tackle them once new grid is out
- from design guidelines perspective
  - it's OK to have sections with different grids on the page (one section using new grid, other section using old `row` or `row--50-50`
  - it's NOT OK to mix grids within one section, especially nesting one in another
    - we may actively make it break or have some debug red colour borders if devs try to do it

PR #5392, opened 2024-10-21
Fix slider margin

### LINKED ISSUE #5391: Sliders off baseline grid
slider height not accounted for in margi-declaraitons, resulting in the overall slider misalignign with the baseline grid:

image

to fix, subtract half the thickness from the margins top and bottom.

mcslayer:
## Done
Based on the comment of @lyubomir-popov, it seems we need to reduce the top and bottom margins by half the height of the track. 
I'm not sure if the calculations needed to be enclosed in parentheses, but I did so for better code readability.

Fixes #5391

## QA
- Please review the [updated component Slider(Range input)](https://vanilla-framework-5392.demos.haus/docs/examples/base/forms/range)

### DIFF
--- a/scss/_base_forms-range.scss
+++ b/scss/_base_forms-range.scss
@@ -15,7 +15,7 @@
     // stylelint-enable property-no-vendor-prefix
     background-color: $color-transparent;
     border-radius: $track-radius;
-    margin: $sp-small 0;
+    margin: ($sp-small - $track-height * 0.5) 0;
     padding: 0;
     width: 100%;
 
--- a/scss/_base_forms-range.scss
+++ b/scss/_base_forms-range.scss
@@ -15,8 +15,10 @@
     // stylelint-enable property-no-vendor-prefix
     background-color: $color-transparent;
     border-radius: $track-radius;
+    height: $track-height;
     margin: ($spv--small - $track-height * 0.5) 0;
     padding: 0;
+    vertical-align: bottom;
     width: 100%;
 
     // Chrome
--- a/scss/_base_forms-range.scss
+++ b/scss/_base_forms-range.scss
@@ -15,7 +15,7 @@
     // stylelint-enable property-no-vendor-prefix
     background-color: $color-transparent;
     border-radius: $track-radius;
-    margin: ($sp-small - $track-height * 0.5) 0;
+    margin: ($spv--small - $track-height * 0.5) 0;
     padding: 0;
     width: 100%;
 


jmuzina:
I think it looks good, @lyubomir-popov can you give a design review [here](https://vanilla-framework-5392.demos.haus/docs/examples/base/forms/range?theme=light)?

bartaz:
There still seems to be a bit of baseline drift if text in paragraphs is added between sliders. You can see that the middle paragraph moves onto the baseline a bit, and the one on the bottom even more.

The baseline alignment is quite tricky, especially in such old components like slider, so no worries if you can't figure that one out @mcslayer. We can have a look ourselves as well.

scss/_base_forms-range.scss
```diff
@@ -15,7 +15,7 @@
     // stylelint-enable property-no-vendor-prefix
     background-color: $color-transparent;
     border-radius: $track-radius;
-    margin: $sp-small 0;
+    margin: ($sp-small - $track-height * 0.5) 0;
```
bartaz, line 18: This is using a legacy spacing variable. Should be replaced with new dedicated to vertical spacing `$spv--small` now that we update it. Thanks!

bartaz:
The math seems to be correct. 3px height + 2* 6.5px margins add up to 16px which should be exactly right.

image

Not sure where the drift comes from. Maybe rounding and/or subpixes issues. Because all of those px values are calculated from fractions of rem values, so maybe rounding kicks in and moves things by a pixel somewhere.

mcslayer:
Yes, I’ve double-checked the calculation, and it is correct. Interestingly, if I adjust the margin to `6.6px 0 6.4px`, it looks perfect!

scss/_base_forms-range.scss
mcslayer: It's fixed now!

mcslayer:
We can apply `vertical-align: bottom` to `input[type=range]`, though I am unsure if this solution fully meets your requirements. If this approach is acceptable, I can proceed with pushing my changes to this PR.

Screenshot 2024-10-22 at 15 52 01

jmuzina:
I think there is something Chromium-specific (or possible Chrome on Mac-specific) happening here. I can't replicate the issue on Firefox, but I can on chrome-based browers.

Firefox:
Screenshot 2024-10-22 at 11 11 10 AM

Brave (chrome-based):
Screenshot 2024-10-22 at 11 12 06 AM

bartaz:
That looks promising. I think you can push the code so that we can see it in the demo server. Thanks for exploring!

mcslayer:
Unfortunately, I found a problem with Firefox where the slider height was calculated incorrectly because Firefox includes the thumb in the input height by default. I figured out how to fix it, and I've applied the fix. I've attached screenshots showing Safari, Chrome, and Firefox (before and after the fix) for comparison.

Before:
Screenshot 2024-10-22 at 19 09 40

After:
Screenshot 2024-10-22 at 19 10 21

jmuzina (commented):
Demo updated to include latest changes

Looks good to me - tested on firefox, chrome, and safari and there are no longer spacing differences between them and the margin is fixed on all of them. 

@bartaz Have a look and let me know if I've missed something 👀

bartaz (approved):
Thanks for chasing that one down @mcslayer

PR #5398, opened 2024-10-23
Remove padding-bottom in cta-block

### LINKED ISSUE #5397: CTA Block: Excessive bottom padding
While building https://github.com/canonical/canonical.com/pull/1406, I noticed a visual quirk with the [CTA block](https://vanillaframework.io/docs/patterns/cta/block): we have given it `padding-bottom: $spv--x-large` here:

https://github.com/canonical/vanilla-framework/blob/78211ca79d33f316815a660be4ff91089fef40d3/scss/_patterns_cta.scss#L19

This causes some extra spacing beneath the CTA block that is not present in CTA blocks that don't use our `.p-cta-block`.

Default:
Screenshot 2024-10-23 at 3 19 22 PM

Expected (no bottom padding):
Screenshot 2024-10-23 at 3 19 37 PM

mcslayer:
## Done

Remove padding-bottom in cta-block, based on issue #5397

Fixes #5397

## QA

- Open `CTA Block` page

### DIFF
--- a/scss/_patterns_cta.scss
+++ b/scss/_patterns_cta.scss
@@ -16,7 +16,6 @@
     border-top: 1px solid $colors--theme--border-low-contrast;
     display: flex;
     flex-wrap: wrap;
-    padding-bottom: $spv--x-large;
     padding-top: $spv--small;
 
     &.is-borderless {


bartaz:
Thanks @mcslayer for another contribution. We want to first consult with designer if this is an intended change. There is a bit of inconsistency between designs in Figma and our component in this regard.

jmuzina:
I think I found where that bottom padding came from: quote wrapper has a link-only CTA that depends on padding-bottom to space it away from content beneath it

Before/after:
Screenshot 2024-10-24 at 10 50 08 AM

For our higher order components like quote block, we do not write any Sass; we just consume other styles. So, a good way of fixing this is probably to wrap the cta block in the quote wrapper in a `.p-section--shallow`

jmuzina:
@lyubomir-popov @dgtlntv @mattea-turic Can you have a look at the CTA block component on its [docs](https://staging.vanillaframework.io/docs/patterns/cta-block), [usage in hero](https://staging.vanillaframework.io/docs/patterns/hero), [usage in quote wrapper](https://staging.vanillaframework.io/docs/patterns/quote-wrapper), and [usage in tiered list](https://staging.vanillaframework.io/docs/patterns/tiered-list) and let us know what the correct bottom spacing is for those cases?

lyubomir-popov:
> and it can be fixed by wrapping in a shallow section there instead of relying on padding-bottom from the CTA itself.

we have our examplke wrong: we should place anchor elements (as they are inline) directly in divs. We wrap them in paragraphs or headings, which applies the needed padding-top/margin-bottom to align the text to the baseline grid and add whatever space is needed after.

jmuzina:
> maybe we ca nhavecan have a class that applies the paragraph stylignstyling to simple anchors, to avoid the nesting - what do you think @bartaz?

With this already released, we should continue supporting .p-cta-block to avoid a breaking change - maybe the easiest way to do this is do remove the bottom spacing on the .p-cta-block class itself, and make .p-cta-block extend paragraph styling like so:

```scss
.p-cta-block {
  @extend %paragraph;
  // ......
}
```

Which would extend this behavior:
https://github.com/canonical/vanilla-framework/blob/ce8551bdec28187663d3de35f199db5829509cb6/scss/_base_typography-definitions.scss#L153-L165

lyubomir-popov:
> ```css-scss
> .p-cta-block {
>   @extend %paragraph;
>   // ......
> }
> ```

this definitely seems confusing. If it extends a paragraph, then we shouldn't have a `button` inside.

@bartaz I guess what I am proposing is .p-text-link class that makes inline anchors into text that aligns with the baseline grid (which is something we expect of all text we display, regardless of how a developer coded it).

bartaz:
On one hand I understand what the intention is there with `p-text-link`, but how would devs know when to use it?

We would need to require devs to add this new class name to links if they are not in paragraph, but not add it, when link in inside of the other text elements. It is gonna be very easily forgotten. 

Not sure what is the good way to get around that.

@jmuzina Would it help at all, if for the case of Quote component we don't use CTA block component at all (if it's never meant to contain a button anyway, just a link)?

jmuzina:
> @jmuzina Would it help at all, if for the case of Quote component we don't use CTA block component at all (if it's never meant to contain a button anyway, just a link)?

@bartaz 

We could wrap the quote wrapper's CTA contents with a `<p>` and preface the `<p>` with `hr.p-rule--muted.is-fixed-width` for roughly the same look. There is a slight change in top margin since we're now using paragraph, but this actually puts the CTA text on the baseline (it currently is not).

Before:
Screenshot 2024-10-30 at 3 23 31 PM

After:
Screenshot 2024-10-30 at 3 23 29 PM

PR #5523, opened 2025-05-19
DON'T MERGE (yet): grouped table example

jmuzina:
A version of the documentation table from [docs.u.om](https://docs.ubuntu.com/) that uses borders between groups, rather than group shading.

Also, the group headers have been set to use `rowspan` and `scope="rowgroup"` for compliance with https://www.w3.org/WAI/tutorials/tables/irregular/ .

This shouldn't be considered fully design-compliant yet, pending further review by design - it merely aims to get the current table looking better in the short term while a more permanent approach is investigated. 

View the live example [here](https://vanilla-framework-5523.demos.haus/docs/examples/patterns/tables/table-grouped?theme=light).

### DIFF
--- a/templates/docs/examples/patterns/tables/table-grouped.html
+++ b/templates/docs/examples/patterns/tables/table-grouped.html
@@ -1,0 +1,157 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Table / Grouped{% endblock %}
+
+{% block standalone_css %}base{% endblock %}
+
+{#- https://www.w3.org/WAI/tutorials/tables/irregular/ -#}
+
+{% block content %}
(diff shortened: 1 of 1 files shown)


jmuzina:
CC @lyubomir-popov - I built this quickly and sent it to @nbellol as a short-term "more legal" version of the docs index table. I'm sure there are improvements we could make over more time - if you have any, post them here. 

I'll be out of office until the 27th of May - just getting this out there so they have something that is better than what they currently have.

jmuzina:
Closing as there hasn't been a demonstrated need for using this sort of table outside of docs.u.com, as far as I'm aware. docs.ubuntu.com has been updated to use this table, so it can reside in their site tier.

PR #5050, opened 2024-04-05
Fix typo in release notes

bartaz:
## Done

Fixes typo in changelog data that prevented link to appear on what's new page

## QA

- Open [demo](https://vanilla-framework-5050.demos.haus/docs/whats-new)
- Make sure that the first item in changelog links correctly to Navigation page

image

PR #5531, opened 2025-05-30
Added new TemplateFinder that removes tagcontent from header ids

### LINKED ISSUE #5393: Status labels in docs headings break section links
**Describe the bug**

Our new system for adding status labels in docs pages causes links to docs sections to break. It seems like the status label's text is included in the search for a matching page section, causing navigation to those sections to break.

Consider this heading:

```md
### Dense help text {{ status('updated') }}
```

This should allow you to navigate to that heading with `#dense-help-text`. However, you can only navigate to the requested heading by removing the status label. 

**To Reproduce**

1. Checkout the repo
2. Find any docs heading that has a {{ status() }} label. 
3. Attempt to navigate to it directly by appending the url encoding of the heading to your URL. I.E [colour-coding](https://vanillaframework.io/docs/patterns/badge#colour-coding) for a "Colour coding" section.
4. See that the requested heading is not scrolled to on page load.  You should also see an error in console:
```
Uncaught TypeError: title is null
    fixScroll http://0.0.0.0:8101/static/js/example.js?v=0134365:485
    resizeInterval http://0.0.0.0:8101/static/js/example.js?v=0134365:369
    setInterval handler*renderIframe http://0.0.0.0:8101/static/js/example.js?v=0134365:365
    renderExample http://0.0.0.0:8101/static/js/example.js?v=0134365:291
    async* http://0.0.0.0:8101/static/js/example.js?v=0134365:81
    <anonymous> http://0.0.0.0:8101/static/js/example.js?v=0134365:80
    EventListener.handleEvent* http://0.0.0.0:8101/static/js/example.js?v=0134365:77
    <anonymous> http://0.0.0.0:8101/static/js/example.js?v=0134365:512
example.js:485:7
```
5. Remove the status label from the markdown title. Retry navigating to the section.
6. See that the page scrolls to the requested section on load.

**Expected behavior**

Navigating to a documentation section by using a section name in a hash causes the browser to scroll to that section on load, regardless of status label.

JuanFernandoCastaneda:
## Done

Added new TemplateFinder that removes tag content from header ids. That way, a header named "X" with a tag "New", won't have id "x-new", but "x" alone. 

Fixes #5393

A less intrusive (but less powerful) solution will be pull requested as well.

## QA

- Go to `http://.../docs/patterns/hero#5050-with-no-image`
- Check that the redirection occurs correctly, even if the page is not refreshed.
- Check that the header id is now `5050-with-no-image` instead of `5050-with-no-image-new`

In case this solution is acceptable, it shouldn't be merged yet, because links like the one in line 16 of `templates/docs/patterns/hero/index.md` `(#5050-with-no-image-new)` would break.

Does not contain Vanilla SCSS or macro code changes

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -34,7 +34,9 @@ with open("releases.yml") as releases_file:
 with open("side-navigation.yaml") as side_navigation_file:
     # maps values of `side_navigation_file.subheadings.ordering` to their implementations
     supported_orderings = {
-        "alphabetical": lambda subheadings, by_attribute: sorted(subheadings, key=lambda subheading: subheading[by_attribute])
+        "alphabetical": lambda subheadings, by_attribute: sorted(
+            subheadings, key=lambda subheading: subheading[by_attribute]
+        )
     }
 
     SIDE_NAVIGATION = yaml.load(
@@ -54,11 +56,15 @@ with open("side-navigation.yaml") as side_navigation_file:
         subheadings_ordering_fn = None
         try:
             subheadings_ordering_identifier = heading["ordering"]
-            subheadings_ordering_fn = supported_orderings[subheadings_ordering_identifier]
+            subheadings_ordering_fn = supported_orderings[
+                subheadings_ordering_identifier
+            ]
         except KeyError:
             return heading
 
-        heading["subheadings"] = subheadings_ordering_fn(heading["subheadings"], by_attribute)
+        heading["subheadings"] = subheadings_ordering_fn(
+            heading["subheadings"], by_attribute
+        )
 
         return heading
 
@@ -80,7 +86,7 @@ TEAM_MEMBERS = [
     {"login": "anthonydillon", "role": "Engineering Director"},
     {"login": "advl", "role": "Engineering Manager"},
     {"login": "lyubomir-popov", "role": "Lead Visual Designer"},
-    {"login": "jmuzina", "role": "Web Engineer"}
+    {"login": "jmuzina", "role": "Web Engineer"},
 ]
 
 
@@ -92,9 +98,7 @@ def _get_title(title):
 
 def _get_examples():
     # get all example files (but ignore partials that start with _)
-    example_files = glob.glob(
-        "templates/docs/examples/*/**/[!_]*.html", recursive=True
-    )
+    example_files = glob.glob("templates/docs/examples/*/**/[!_]*.html", recursive=True)
     examples = {}
 
     for filepath in sorted(example_files):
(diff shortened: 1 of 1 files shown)


jmuzina:
Thanks for the contribution @JuanFernandoCastaneda !
Starting up a demo [here](https://vanilla-framework-5531.demos.haus/). I'll give you an initial review sometime today.

> In case this solution is acceptable, it shouldn't be merged yet, because links like the one in line 16 of templates/docs/patterns/hero/index.md (#5050-with-no-image-new) would break.

I haven't read through the code yet, but can you elaborate a bit on this? How do you see this being addressed?

templates/static/js/example.js
```diff
@@ -481,7 +481,7 @@
   function fixScroll() {
     const titleId = window.location.hash;
     if (titleId) {
-      const title = document.querySelector(titleId);
+      const title = document.querySelector(`[id="${titleId.substring(1)}"]`);
```
jmuzina, line 484: Can you add a comment here explaining the purpose for the substring operation?

Also, I'm not sure if this is necessary - your Python change seems to have successfully changed the ID to exclude the status label.

Current [prod](https://vanillaframework.io/docs/base/forms#dense-help-text):
Screenshot 2025-05-30 at 20 00 07

This [PR](https://vanilla-framework-5531.demos.haus/docs/base/forms#dense-help-text):
Screenshot 2025-05-30 at 20 00 10

JuanFernandoCastaneda: Yes. I found another error related to the autoscroll, but didn't seem big enough to make a different pull req (if it is needed, then I will). 

If you have the query as `titleId` only, instead of the substring one, errors like this appear on the ids that start with a number. 

[fixScroll.webm](https://github.com/user-attachments/assets/5ed6b5ac-3409-4bf9-a164-b2a7d256ff56)

The second and onward times you visit the site, the autoscroll goes crazy, and shows this error 

image

webapp/app.py
```diff
@@ -279,7 +280,19 @@ def utility_processor():
     return {"class_reference": class_reference, "image": image_template, "status": status_label}
 
 
-template_finder_view = TemplateFinder.as_view("template_finder")
+# Overrides the representation of the headers that have spans inside them to not add the label content to the id of the header. Enforces a restriction in which two headers with the same text but different tags should not exist. 
+class TemplateFinderStatusTags(TemplateFinder):
+    def dispatch_request(self, *args, **kwargs):
+        inherited_result = super().dispatch_request(*args, **kwargs) 
+        labeled_headers = [result[0] for result in re.findall(r"(<(h\d+).*?>.*?(<span.*?>.*</span>).*?</\2>)", inherited_result)]
```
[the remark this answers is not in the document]
JuanFernandoCastaneda: Fixed it 🫡

jmuzina (changes requested):
Thanks for this. However, with the current phase of maintenance Vanilla is in (gradually shifting over to [DS-25](https://github.com/canonical/ds25), I think the solution here and in #5532 is a bit complex and if anything is demonstrating that our architecture itself is contributing to problems that are making us build band-aid solutions like this.

There is one thing I'd like to keep in this PR however - you noticed a new bug with fixScroll() for sections that start with numbers. Can you drop the rest of the changes from this and just keep the fixScroll() update and we can merge that?

PR #5538, opened 2025-06-04
Add logo section negative margin right

### LINKED ISSUE #5522: Add negative margin: Logo section
## Component/pattern to amend

Logo section

## Visual

Image

## Context

Can ewe add a negative margin-right (`-2rem`) to `p-logo-section__items`? This way, the margin-right on whichever logo happens to be last at the current screen width (they wrap), will go into that negative margin, allowing more logos to fit onto one line. (Thanks to @kim-isaac) for bringing this up.

jmuzina:
## Done

Fixes #5522 
Fixes [WD-22109](https://warthogs.atlassian.net/browse/WD-22109)

## QA

- Open logo section examples and verify the spacing is correct
- Review 4.24.1 release notes

- [] Any changes to component class names (new patterns, variants, removed or added features) or macros should be listed on the [what's new page](https://github.com/canonical/vanilla-framework/blob/main/releases.yml).

## Screenshots

Screenshot 2025-06-04 at 09 28 35

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vanilla-framework",
-  "version": "4.24.0",
+  "version": "4.24.1",
   "author": {
     "email": "webteam@canonical.com",
     "name": "Canonical Webteam"
--- a/scss/_patterns_logo-section.scss
+++ b/scss/_patterns_logo-section.scss
@@ -46,6 +46,7 @@
       // compensate for the negative margin on the logos
       padding-bottom: $logo-section-offset-small;
       padding-top: $logo-section-offset-small;
+      margin-right: #{-1 * $logo-section-item-gap};
 
       @media (min-width: $breakpoint-small) {
         padding-bottom: $logo-section-offset;


terameta:
image

On https://percy.io/bb49709b/web/vanilla-framework/builds/40802237/changed/2191994322?browser=chrome&browser_ids=64&group_snapshots_by=similar_diff&subcategories=unreviewed%2Cchanges_requested&utm_source=github_status_private&viewLayout=side-by-side&viewMode=new&width=375&widths=375%2C800%2C1280

As far as I can see, first row has less logo than before.

In all the other examples on Percy, I verified the PR is successful except this one.

jmuzina:
@lyubomir-popov @terameta is referring to the [line breaks example](https://vanilla-framework-5538.demos.haus/docs/examples/patterns/logo-section/logo-section-line-breaks?theme=light), which now displays more logos in the second row than the first row at the Percy test width of 375px. As the page shrinks, a logo is redistributed up to the first row. Would you consider this to be a visual regression?

https://github.com/user-attachments/assets/e4d98f1e-c646-4d69-aff2-dc920d3c8acf

lyubomir-popov:
> As the page shrinks, a logo is redistributed up to the first row. Would you consider this to be a visual regression?

why tdoes it happen? do the logos scale ata different ratethan the screen width? I wouldn't consider it a regression, but if there's a simple way to avoid it, would be nice. If it's just a result of the natural wrapping, then I'de leave it.

jmuzina:
@lyubomir-popov It seems to be an artifact of browser-specific content wrapping behavior. This doesn't occur on Firefox, but it does occur on Chrome.

I was able to make it consistent across browsers by making the items list a flexbox rather than a sequence of `inline-block` items, but this would be breaking for our "with line breaks" variant support as it would create undue gap after line breaks.

jmuzina:
With @lyubomir-popov : we don't need to worry about the inconsistent wrapping behavior mentioned [here](https://github.com/canonical/vanilla-framework/pull/5538#issuecomment-2940216073). One possible way to reduce this is to set `font-size: 0` on the items list to eliminate space taken up by whitespace characters, but this should be tested to avoid breaking cases where we do place text inside the logo sections.

jmuzina:
> With @lyubomir-popov : we don't need to worry about the inconsistent wrapping behavior mentioned [here](https://github.com/canonical/vanilla-framework/pull/5538#issuecomment-2940216073). One possible way to reduce this is to set `font-size: 0` on the items list to eliminate space taken up by whitespace characters, but this should be tested to avoid breaking cases where we do place text inside the logo sections.

As far as I can tell, we don't add any text inside `p-logo-section__items`. In the past, we've added text in `p-logo-section__title`, and sometimes directly in `p-logo-section` ([ref](https://vanillaframework.io/docs/patterns/logo-section#fallback-for-misaligned-logos)), but never in `p-logo-section__items`. 

The one risk here is image alt text. What if a logo doesn't load? The image's alt text would be invisible if font size is set to 0, and page content would be lost, which could be especially problematic considering we often use these to indicate partners.


### HP logo failing to load, with and without font-size: 0
Screenshot from 2025-06-11 07-41-28
Screenshot from 2025-06-11 07-41-36

We could keep this if we added font size overrides for images within logo section items blocks. 

The resulting whitespace difference:

[Screencast from 2025-06-11 08-20-08.webm](https://github.com/user-attachments/assets/89e481bd-e665-4844-aa93-8fdaac6b7d5a)

lyubomir-popov:
@jmuzina what about the logo section with links? if it resets the font size to whatever is needed, then maybe we can use the 0 font size hack

jmuzina:
> @jmuzina what about the logo section with links? if it resets the font size to whatever is needed, then maybe we can use the 0 font size hack

@lyubomir-popov Can you elaborate on what you mean by this? It sounds like it may be a separate issue for the linked logo section pattern.

I found a few cases where there is actually text inside the logo section items, for which this would constitute a breaking change unless font size is reset at the `logo-section__item` level. 

- [c.com/data/opensearch/support](https://canonical.com/data/opensearch/support) has a logo section "Come for OpenSearch support, get the full stack" that has text inside the logo section items block
- same for [c.com/data/mongodb/support](https://canonical.com/data/mongodb/support)
- [u.com/security/fedramp](https://ubuntu.com/security/fedramp)

Revising to reset the font size at the `logo-section__item` level, rather than at the image level.

bartaz:
The only thing that I'd maybe only suggest to verify the proposed solution, is to try it out on the production websites.
You could either point a local environment, or a branch of a product (like canonical.com) to this branch as dependency, or hack it in as a local SCSS override.

This way you could make sure nothing will break when applied. We did have situations in the past (with logo section), where seemingly non-breaking change in Vanilla appeared breaking in production because of some unusual use of the component.

jmuzina:
Closing this due to high complexity of verifying the safety of these changes and limited capacity. If a contributor would like to investigate and propose a solution, I would welcome this.

PR #5569, opened 2025-07-08
fix(Badge): Truncate badges with long text, make the smallest badge a perfect circle

### LINKED ISSUE #5568: [Bug]: Badge max-width and line-height not applied
**Describe the bug**
On the badge component the max-width and line-height are not applied correctly. Seemingly because in an inline context max-width and line-height are not respected. Since we are applying the badge on a span, which is inline, max-width and line-height are not correctly applied.

To fix `display: inline-block;` needs to be applied

**Screenshots**
Without inline-block:
Image

With inline-block:
Image

jmuzina:
## Done

- Set `display: inline-block` on the badge component, to ensure its `max-width` and `line-height` are properly enforced
- Set a minimum width on badges so that badges with the smallest size possible are perfect 16x16 circles
  - Set badge text to center-align (otherwise very short badge text would not be horizontally centered)

Fixes [WD-23540](https://warthogs.atlassian.net/browse/WD-23540)
Fixes #5568 

## QA

- Open [badge](https://vanilla-framework-5569.demos.haus/docs/examples/patterns/badge/default?theme=light)
- Verify that the "9" badge is a perfect 16x16 circle
- Use "Inspect element" to change any badge to have very long text, longer than 4 characters.
- Verify that the badge with the long text does not show more than 4 full characters, and the rest of the characters are cut off (not displayed)

## Screenshots

image

[WD-23540]: https://warthogs.atlassian.net/browse/WD-23540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scss/_patterns_badge.scss
+++ b/scss/_patterns_badge.scss
@@ -7,6 +7,7 @@
     border-radius: 1rem;
     box-sizing: content-box;
     color: $colors--theme--background-default; // inverse the theme by using the baackground color
+    display: inline-block;
     margin-bottom: 0;
     max-width: 4ch;
     overflow: hidden;
--- a/scss/_patterns_badge.scss
+++ b/scss/_patterns_badge.scss
@@ -10,8 +10,10 @@
     display: inline-block;
     margin-bottom: 0;
     max-width: 4ch;
+    min-width: $sph--small;
     overflow: hidden;
     padding: 0 $spv--x-small;
+    text-align: center;
   }
 
   .p-badge,
--- a/scss/_patterns_badge.scss
+++ b/scss/_patterns_badge.scss
@@ -10,7 +10,10 @@
     display: inline-block;
     margin-bottom: 0;
     max-width: 4ch;
-    min-width: $sph--small;
+    // the minimum content width should be the total desired width (the line-height)
+    // minus the space already taken up by the left and right padding
+    // Makes the shortest badge a perfect circle and keeps it related to line-height (of %x-small-text) and padding (below)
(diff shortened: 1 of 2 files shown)


jmuzina:
@dgtlntv Do you think this would constitute a breaking change, since badges that contain text over 4 characters long will now have some content hidden?

dgtlntv:
I dont know if it might break some implementations. But generally I think I would personally not regard it as a breaking change because the documentation was quite clear that you should not use more than 4 characters. And the code did have max-width set, it was just not properly applied because of this bug. So I would personally regard it as a patch of that bug rather than a breaking change.

scss/_patterns_badge.scss
```diff
@@ -6,11 +6,14 @@
     background-color: $colors--theme--text-default; // inverse the theme by using the text color
     border-radius: 1rem;
     box-sizing: content-box;
-    color: $colors--theme--background-default; // inverse the theme by using the baackground color
+    color: $colors--theme--background-default; // inverse the theme by using the background color
+    display: inline-block;
     margin-bottom: 0;
     max-width: 4ch;
+    min-width: $sph--small;
```
dgtlntv: I wonder if the min-width should be the same variable as is used on the line height. As the height of the badge is determined by the line-height of the text. So if the line-height of the text were to change it would not be a circle anymore.

jmuzina:
> Apart from the thought regarding the variable usage on the min-width LGTM. Thanks for the quick fix! :)

@dgtlntv adjusted!

PR #5598, opened 2025-07-24
wip: Implement Basic Section pattern

jmuzina:
## Done

Implements [Basic section](https://www.figma.com/design/1brHCsyeTE6FU7hfJd6ao2/%F0%9F%9A%A7---WIP--Sites---Core-component-library?node-id=47-3043&m=dev) pattern.

## QA

- [Add QA steps]

## Screenshots
Screen Shot 2025-07-24 at 17 51 51

### DIFF
--- a/templates/_macros/vf_basic-section.jinja
+++ b/templates/_macros/vf_basic-section.jinja
@@ -1,0 +1,304 @@
+{% from "_macros/shared/vf_dedent.jinja" import vf_dedent %}
+{% from "_macros/vf_linked-logo-section.jinja" import vf_linked_logo_section %}
+
+# description_config
+# - type: "text" | "html" (default is "text")
+# - content: The content of the description. Could be plain text (if type is "text") or HTML (if type is "html").
+{% macro _basic_section_description(description_config={}) %}
+  {%- set type = (description_config["type"] or "text") | trim -%}
+  {%- if type not in ["text", "html"] -%}
+    {%- set type = "text" -%}
+  {%- endif -%}
+
+  {% if type == "text" %}
+    <p>{{- description_config["content"] -}}</p>
+  {% elif type == "html" %}
+    {{- description_config["content"] | safe -}}
+  {% endif %}
+{% endmacro %}
+
+# image_config
+# - aspect_ratio: "16-9" | "3-2" | "" (default is "")
+# - attrs: A dictionary of attributes to apply to the image
+{% macro _basic_section_image(image_config={}) %}
+  {% set aspect_ratio = image_config["aspect_ratio"] | trim %}
+  {% if aspect_ratio not in ["16-9", "3-2"] %}
+    {% set aspect_ratio = "" %}
+  {% endif %}
+
+  <div class="p-image-container{% if aspect_ratio | length > 0 %}--{{- aspect_ratio -}}{% endif %} is-highlighted">
+    <img class="p-image-container__image{{- image_config["attrs"]["class"] -}}"
+      {% for attr, value in image_config["attrs"].items() %}
+        {% if attr != "class" %}
+          {{ attr }}="{{ value }}"
+        {% endif %}
+      {% endfor %}
+    />
+  </div>
+{% endmacro %}
+
+# video_config
+# - attrs: A dictionary of attributes to apply to the video iframe
+{% macro _basic_section_video(video_config={}) %}
+  <div class="u-embedded-media">
+    <iframe class="u-embedded-media__element {{- video_config["attrs"]["class"] -}}"
+      {% for attr, value in video_config["attrs"].items() %}
+        {% if attr != "class" %}
+          {{ attr }}="{{ value }}"
+        {% endif %}
+      {% endfor %}
+    ></iframe>
+  </div>
+{% endmacro %}
+
+# list_config
+# - list_item_type: "bullet" | "tick" | "cross" | "number" | "" (default is "")
+# - list_items: A list of items to be displayed in the section.
+# - - content: The HTML content of the list item.
+# - - sublist: A nested list configuration
+{% macro _basic_section_list(list_config={}) %}
+  {% if list_config["list_items"] | length > 0 %}
+    {#- Note: namespace() requires jinja2 2.10 or later -#}
+    {% set ns = namespace(list_tag="ul") %}
+
+    {#- If any of the list items is numbered, the entire list is numbered -#}
+    {% for list_item in list_config["list_items"] %}
+      {% if list_item["list_item_type"]|trim|lower == "number" %}
+        {% set ns.list_tag = "ol" %}
+      {% endif %}
+    {% endfor %}
+
+    <{{ ns.list_tag }} class="p-list--divided">
+    {% for list_item in list_config["list_items"] %}
+      {% set list_item_type=list_item["list_item_type"] | trim | lower %}
+      {#- If the list is ordered, ignore list item type. Prevents from drawing both a tick and a number, for example. -#}
+      {% if ns.list_tag == "ol" %}
+        {% set list_item_type = 'number' %}
+      {% elif list_item_type|length > 0 and list_item_type not in ['bullet', 'tick', 'number', 'cross'] %}
+        {% set list_item_type = '' %}
+      {% endif %}
+
+      {% if list_item_type == "bullet" %}
+        {% set list_item_style_class = "has-bullet" %}
+      {% elif list_item_type == "tick" %}
+        {% set list_item_style_class = "is-ticked" %}
+      {% elif list_item_type == "cross" %}
+        {% set list_item_style_class = "is-crossed" %}
+      {% endif %}
+
+      <li class="p-list__item {{ list_item_style_class }}">
+        {{- list_item["content"] | safe -}}
+        {% if "sublist" in list_item %}
+          {{- _basic_section_list(list_config=list_item["sublist"]) -}}
+        {% endif %}
+      </li>
+    {% endfor %}
+    </{{ ns.list_tag }}>
+  {% endif %}
+{% endmacro %}
+
+# code_block_config
+# - content: The HTML content to be displayed in the code block. Will be wrapped in a <pre> tag.
+{% macro _basic_section_code_block(code_block_config={}) %}
+  <pre>{{- vf_dedent(code_block_config["content"]) | safe -}}</pre>
+{% endmacro %}
+
+{#- TODO does this need to handle the specific column count case? -#}
+# logo_block_config
+# - logos: A list of logos to be displayed in the section.
+# - - attrs: A dictionary of attributes to apply to the logo
+# - - is_fixed_width: Boolean to indicate if the logo should be wrapped in a fixed-width container (default is true).
+# - - include_line_break_after: Boolean to indicate if a line break should be included after the logo (default is false). This will be hidden on small screens.
+#       See https://vanillaframework.io/docs/patterns/logo-section#line-breaks
+{% macro _basic_section_logo_block(logo_block_config={}) %}
+  {% if is_fixed_width %}<div class="u-fixed-width">{% endif %}
+    <div class="p-logo-section">
+      {%- for logo in logo_block_config["logos"] -%}
+        <div class="p-logo-section__item">
+          <img class="p-logo-section__logo {{- logo["attrs"]["class"] -}}"
+            {% for attr, value in logo["attrs"].items() %}
+              {% if attr != "class" %}
+                {{ attr }}="{{ value }}"
+              {% endif %}
+            {% endfor %}
+          />
+          {%- if logo.get("include_line_break_after", false) %}
+            <br class="u-hide--small">
+          {% endif %}
+        </div>
+      {%- endfor -%}
+    </div>
+  {% if is_fixed_width %}</div>{% endif %}
+{% endmacro %}
+
+# linked_logo_block_config
+#   - title_text (string) (required): The text to be displayed as the heading
+#   - links (array) (required) : Array of links, each including 'href', 'label', 'text', and 'image_attrs'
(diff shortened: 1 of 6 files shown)


kim-isaac:
Hi @jmuzina , thanks so much for organizing and building out this complex pattern!
I’ve listed a few comments below:

- [x] Overall, the default rule at the top of each section is missing. Could you please add that?
- [x] Label: Please remove the top & bottom padding.
- [x] Would it be okay to add an extra note at the 'pattern level' docs?: Deep padding is applied to the final section at the bottom of the page to create sufficient spacing from the page footer. Shallow padding is used when a related full-width element (e.g., an image or table) is placed directly below the section.
- [x] Logo section: Currently, two separate logo sections are used across different lines. Would it be possible to consolidate them into a single component? Also, the spacing around the Lenovo logo on the second row seems off.
image

also, I'll add @eliman11 for the UX review!

eliman11:
Thank you @jmuzina and @kim-isaac this looks great! Sorry this is super minor and annoying, but could we change all the headings to use sentence case so that we're consistent with the other patterns (although I also just noticed "Quote Wrapper" is in title case) - so Basic section, Basic section / Mixed content, Basic section / Linked title, etc

Also could we have an h3 HTML tag for the label instead of an h4 to avoid skipping levles in the heading hierarchy? Everything else looks good to me!
Screenshot 2025-07-30 at 14 29 19

jmuzina:
Thanks for your reviews @eliman11 and @kim-isaac - I've adjusted per your feedback. Have another look and let me know what you think!

kim-isaac:
Thanks @jmuzina ! I have two comments below:

- [ ] It seems like there's still bottom padding left on the label, even when I set top and bottom padding to none. Is this something that can't be changed within the pattern?
- [ ] I had a chat with Elaine, and a title like "default padding" at the pattern level might be misinterpreted as a built-in part of the pattern. Since you’ve already included a link to the documentation that I asked to add, I believe the codepen sample could be omitted. What are your thoughts?

Besides that all looks great! Thanks again.

jmuzina:
@kim-isaac 
> It seems like there's still bottom padding left on the label, even when I set top and bottom padding to none. Is this something that can't be changed within the pattern?

This is bottom margin, not padding:

Screenshot 2025-08-01 at 10 14 08

If you like, we can add `u-no-margin--bottom` to reduce (but not remove) the bottom margin from `10px` to `2px`. The 2px is kept to maintain baseline alignment.

Screenshot 2025-08-01 at 10 15 21


Removing the dev tools overlay so you can see what that would look like:

Split:
Screenshot 2025-08-01 at 10 17 20

Stacked:
Screenshot 2025-08-01 at 10 17 50

Would that work for you?

> a title like "default padding" at the pattern level might be misinterpreted as a built-in part of the pattern. Since you’ve already included a link to the documentation that I asked to add, I believe the codepen sample could be omitted. What are your thoughts?

Yes, that's workable, and won't have negative effects for visual regression coverage because there are other examples that use the "default padding". I'll push that in a moment

kim-isaac:
Hi @jmuzina 

> If you like, we can add u-no-margin--bottom to reduce (but not remove) the bottom margin from 10px to 2px. The 2px is kept to maintain baseline alignment.

Yes, that would be great!

eliman11:
Thanks Julie! Looks good :) Also very minor but can we also have the "Basic Section" heading on the page itself in sentence case? Will +1 for UX though, thanks both!

kim-isaac:
Hi @jmuzina , I've checked that the bottom margin on the label has been reduced - thank you!
What I meant by “pattern level” was the entire Codepen sample. (Similarly, there's a chance that “deep padding” and “shallow padding” could also be misread as part of the pattern) Would it be possible to remove those? Just a suggestion though - feel free to keep them if you think they’re still helpful!

jmuzina:
@eliman11 
>  Can we also have the "Basic Section" heading on the page itself in sentence case?

Done, thanks!

@kim-isaac 
> there's a chance that “deep padding” and “shallow padding” could also be misread as part of the pattern

Good point, I removed those bits of text that were outside of the patterns and instead am relying on the text inside the examples themselves to demonstrate which padding is being used.

> What I meant by “pattern level” was the entire Codepen sample. 

I think it's important we keep the example embed in some form - it helps developers to quickly get started using the pattern with a given configuration (in this case different padding settings) by copy-pasting. I've reworded the documentation here to make these things clear : 

1. Section padding is being applied by the pattern itself (the user doesn't have to apply p-section themselves)
2. To request a different padding type, the `padding` parameter can be used
3. Guidance on when to use which padding level is available in the section docs (with a link)

With these changes and the removal of the misleading out-of-pattern titles in the example, do you think the page is okay now?

kim-isaac:
Hi @jmuzina , thanks, all looks good to me!
I'll add +1 as well :)

templates/_macros/vf_linked-logo-section.jinja
```diff
@@ -39,36 +66,44 @@
...
     {%- if layout == "50-50" -%}
       <div class="p-section--shallow">
         <div class="grid-row">
+          {% if has_title %}
           <div class="grid-col-4">
             <h2>{{ title_text }}</h2>
           </div>
-          <div class="grid-col-4 grid-col-medium-4">
+          {% endif %}
+          <div class="grid-col-4 grid-col-start-large-5 grid-col-medium-4 grid-col-start-medium-5">
```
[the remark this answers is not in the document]
jmuzina: copilot doesn't understand that grid-col-start-x classes apply a `grid-column-start` which ensures the col is placed at the same starting point regardless of whether the preceding col was rendered or not. This is the intended approach.

templates/_macros/vf_linked-logo-section.jinja
```diff
@@ -1,13 +1,19 @@
 {#
   Params
-  - title_text (string) (required): The text to be displayed as the heading
+  - title_text (string): The text to be displayed as the heading
   - layout: layout of linked logo section. Options are '50/50', '25/75', 'full-width'
   - links (array) (required) : Array of links, each including 'href', 'label', 'text', and 'image_html'
+  - - if present, `image_attrs` will be used over `image_html` as it provides a more flexible way to pass image attributes.
```
muhammad-ali-pk:
```suggestion
  - if present, `image_attrs` will be used over `image_html` as it provides a more flexible way to pass image attributes.
```

templates/_macros/vf_basic-section.jinja
```diff
@@ -0,0 +1,340 @@
...
+# code_block_config
+# - content: The HTML content to be displayed in the code block. Will be wrapped in a <pre> tag.
+# - is_code_snippet: Boolean to indicate if the content is a code snippet (default is false). If true, the content will be wrapped in a <code> tag, within the <pre> tag.
+{% macro _basic_section_code_block(code_block_config={}) %}
+  {%- set is_code_snippet = code_block_config.get("is_code_snippet", False) -%}
+  <pre>
+    {%- if is_code_snippet -%}
+    <code>
+    {%- endif -%}
+    {{- vf_dedent(code_block_config.get("content", "")) | safe -}}
```
muhammad-ali-pk, line 123: Why is a custom indendentation calculating macro needed here? Wouldn't the whitespace be preserved automatically for code blocks wrapped in pre tag?

muhammad-ali-pk (approved):
@jmuzina This is so well done. Love the implementation!

templates/_macros/vf_basic-section.jinja
jmuzina: In [this example](https://github.com/canonical/vanilla-framework/blob/4d6cb55cd529ff666f186f8e033ead942e07eb4c/templates/docs/examples/patterns/basic-section/code-variants.html#L16-L31), I noticed that without the dedent, the indentation in the source code is rendered into the macro.

With the dedent:

image

Without it:

image

For reference, here's how the block is generated:

https://github.com/canonical/vanilla-framework/blob/4d6cb55cd529ff666f186f8e033ead942e07eb4c/templates/_macros/vf_basic-section.jinja#L117-L128

Do you think there's a way I could do this without specifically handling the indentation? I might've missed something

jmuzina:
Had a conversation with @mattea-turic about the horizontal rules used in #5607 and I have a few questions for the group of both @mattea-turic and @kim-isaac :

I've been told that the pricing block should use a muted rule when it's a subsection that is preceded by other subsections, no rule when it's the only or first subsection, and a highlighted rule when it's a top-level section.

In this PR, the linked logo section is used inside the basic section, **always with no top rule**. All of the blocks in the basic section's right side never have rules separating them. I just want to be sure that these two sets of guidance don't conflict - is the rule behavior in this PR (never using a top rule for the linked logo used as a block within basic section) correct?

mattea-turic:
@jmuzina @kim-isaac 
> I just want to be sure that these two sets of guidance don't conflict - is the rule behavior in this PR (never using a top rule for the linked logo used as a block within basic section) correct?

Just so I understand, do you mean we're not currently using muted rules above _only_ the linked logo section as part of the basic section (against other subsections), or any elements we use on the right side are currently not (and shouldn't be) using rules above – so essentially no separators?

I'm asking this as we've had conversations about our overuse of rules, but I think it's still important to add a distinction between subsections, depending on the content. For example, using no rules reduces clarity (example [here](https://www.figma.com/design/1brHCsyeTE6FU7hfJd6ao2/%F0%9F%9A%A7---WIP--Sites---Core-component-library?node-id=2111-13060&t=cdRX8m9seKQs63TC-1)), compared to when we _are_ using them (example [here](https://www.figma.com/design/1brHCsyeTE6FU7hfJd6ao2/%F0%9F%9A%A7---WIP--Sites---Core-component-library?node-id=2138-9224&t=cdRX8m9seKQs63TC-1)). 

So all this to say, is this a discussion on how we're handling our use of muted rules, in general?

jmuzina:
> do you mean we're not currently using muted rules above _only_ the linked logo section as part of the basic section (against other subsections), or any elements we use on the right side are currently not (and shouldn't be) using rules above – so essentially no separators?

Currently none of the subsection blocks in the basic section have top borders, see [example](https://vanilla-framework-5598.demos.haus/docs/examples/patterns/basic-section/mixed-content?theme=light). This includes the linked logo block.

> is this a discussion on how we're handling our use of muted rules, in general?

I think it is, yes - a concrete set of guidelines for when to use which sort of rules would be helpful. Per our initial conversations @kim-isaac I had the understanding that the top rule on the linked logo block was a very simple choice - default rule if it's a section, no rule if it's a block embedded somewhere else (like the basic section). Now, with examples such as [pricing block](https://github.com/canonical/vanilla-framework/pull/5613), it seems this is less clear, and there are possibilities for any of the rule types to be used.

> it's still important to add a distinction between subsections, depending on the content

The [design spec](https://www.figma.com/design/1brHCsyeTE6FU7hfJd6ao2/%F0%9F%9A%A7---WIP--Sites---Core-component-library?node-id=41-9326&m=dev) for the pattern did not specify top item separators/rules, do you @kim-isaac @mattea-turic think it should?

----

In summary:

1. There is some common set of parameters that every pattern should support in order to allow for it to be embedded in other patterns - so far this seems like `top_rule_variant` and `padding`. These may have different defaults, depending on the pattern, but they all need to be individually customizable, instead of using a single `scope` param like this PR currently does.
2. @mattea-turic and @kim-isaac may need to discuss the use of rules to separate the basic section subsections. Possibly the outcome of this could influence guidelines for other patterns. 

@mattea-turic @kim-isaac let me know if the summary is accurate

kim-isaac:
Hi @jmuzina , @mattea-turic and I discussed this and came to this conclusion:
The content in the basic section is intentionally designed without dividers/rules to minimize visual noise and because its purpose is to cover a single topic. When covering different topics, we plan to use separate patterns such as the horizontal/vertical tabbed section or stepped list. For that reason, there are no rules within the basic section. 

And here’s my thought:
“Block” elements(e.g., linked logo block) embedded within a pattern (e.g., basic section) should not have rules by default. I believe defining them at the parent pattern level is more suitable for covering a wider range of cases.

The muted rule I mentioned refers to when separate sections/patterns are connected.
_Example: the first section uses a regular rule + shallow padding, and the second (last) connected section uses a muted rule + regular padding._

I agree that we need to define the parameters for rules and padding! It might be good to set up a meeting for this. Thanks for raising this Julie.

jmuzina:
> “Block” elements(e.g., linked logo block) embedded within a pattern (e.g., basic section) should not have rules by default. I believe defining them at the parent pattern level is more suitable for covering a wider range of cases. The muted rule I mentioned refers to when separate sections/patterns are connected. _Example: the first section uses a regular rule + shallow padding, and the second (last) connected section uses a muted rule + regular padding.

OK, this makes sense. This means that embeddable patterns should support removing their top rules or using different top rule variants, up to the designer's discretion. Usages of embedded patterns will, by default, have no top rule, but this can be adjusted at the parent pattern level if needed by Design on a pattern-by-pattern basis.

So, I'll need to update this PR to support more granular control of linked logo section top rule & padding, to align with Pricing Block - then we should be good to go.

jmuzina:
@muhammad-ali-pk I've updated this PR per [this comment](https://github.com/canonical/vanilla-framework/pull/5598#issuecomment-3176050157) to make the approach to the top rule and padding more consistent with your work on the pricing block - can you have a final look please?

releases.yml
```diff
@@ -1,3 +1,9 @@
+- version: 4.30.0
+  features:
```
muhammad-ali-pk, line 2: Can you please update the release notes to include the changes for the linked logo section?

muhammad-ali-pk (approved):

mattea-turic:
@jmuzina 
> So, I'll need to update this PR to support more granular control of linked logo section top rule & padding, to align with Pricing Block - then we should be good to go.

This sounds good, thanks so much!

releases.yml
jmuzina: Done

muhammad-ali-pk (approved):

PR #5624, opened 2025-08-25
WD-26007 - Add more sliding animations to the navigation

### LINKED ISSUE #5620: [Component amendment]: p-navigation--sliding
## p-navigation--sliding

When trying to migrate snapcraft.io to use the "sliding" global-nav we noticed that just changing the class to _p-navigation--sliding_ wasn't giving as result a functioning sliding navigation. There are styles missing in order for the animations to be displayed properly.

## Visual

[snapcraft.io](https://snapcraft.io/)

## Context

This amendment will allow us to migrate all the projects that currently use _p-navigation_ to be updated to use _p-navigation--sliding_ without having to add any custom styles repeatedly in each project in order for the change to work.

alvaromateo:
## Done

Add styles to make the navigation of the menu in mobile view slide in and out (from the right side).
Add styles to make the item's dropdown in desktop view slide down and up.
Add necessary HTML and JS to the examples for the new changes to work.

Fixes #5620 / [WD-26007](https://warthogs.atlassian.net/browse/WD-26007)

## QA

- Open the following URLs:
  - [Docs](https://vanilla-framework-5624.demos.haus/docs/patterns/navigation)
  - [Example: Sliding](https://vanilla-framework-5624.demos.haus/docs/examples/patterns/navigation/sliding)
  - [Example: Sliding dark](https://vanilla-framework-5624.demos.haus/docs/examples/patterns/navigation/sliding-dark)
  - [Example: Sliding with search](https://vanilla-framework-5624.demos.haus/docs/examples/patterns/navigation/sliding-search)
  - [Example: Sliding full width](https://vanilla-framework-5624.demos.haus/docs/examples/patterns/navigation/dropdown-full-width)
- Check that the menu in mobile view slides in and out from the right of the screen.
- Check that the dropdown items in desktop view slide down and up.

The other navigation types should not be affected by these changes.
- Open the following URLs:
  - [Example: Reduced](https://vanilla-framework-5624.demos.haus/docs/examples/patterns/navigation/reduced)
  - [Example: Dropdown](https://vanilla-framework-5624.demos.haus/docs/examples/patterns/navigation/dropdown)
- Check that the navigation still works as before.

## Check if PR is ready for release

[WD-26007]: https://warthogs.atlassian.net/browse/WD-26007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scss/_patterns_navigation.scss
+++ b/scss/_patterns_navigation.scss
@@ -190,6 +190,20 @@ $navigation-height: calc(map-get($settings-text-p, line-height) + 2 * $spv--larg
     }
   }
 
+  .p-navigation--sliding {
+    @property --right {
+      inherits: false;
+      initial-value: 0%;
+      syntax: '<percentage>';
+    }
+
+    @property --up {
+      inherits: false;
+      initial-value: 0%;
+      syntax: '<percentage>';
+    }
+  }
+
   // navigation row
   .p-navigation__row,
   .p-navigation__row--full-width {
@@ -1061,6 +1075,86 @@ $navigation-height: calc(map-get($settings-text-p, line-height) + 2 * $spv--larg
       }
       @media (min-width: $breakpoint-navigation-threshold) {
         display: none;
+      }
+    }
+  }
+
+  @media (width < $breakpoint-navigation-threshold) {
+    .p-navigation--sliding {
+      .p-navigation__nav {
+        display: block;
+        margin-top: 3.5rem;
+        min-width: 100%;
+        position: absolute;
+        transform: translateX(var(--right));
+        // same function as for the global-nav dropdown
+        transition: transform 0.333s cubic-bezier(0.215, 0.61, 0.355, 1);
+        // starts on the right outside of the screen
+        --right: 110%;
+      }
+
+      &.has-menu-open {
+        overflow-x: hidden;
+      }
+
+      &.has-menu-open .p-navigation__nav {
+        --right: 0%;
+      }
+
+      &.menu-closing .p-navigation__nav {
+        --right: 110%;
+      }
+
+      .p-navigation__items--section {
+        margin: 0;
+        padding: 0;
+      }
+    }
+  }
+
+  @media (min-width: $breakpoint-navigation-threshold) {
+    .p-navigation--sliding {
+      &.has-menu-open {
+        box-shadow: none;
+      }
+
+      .p-navigation__items {
+        justify-content: space-between;
+
+        &--section {
+          display: flex;
+          margin: 0;
+          padding: 0;
+        }
+      }
+
+      .p-navigation__item--dropdown-toggle {
+        position: relative;
+      }
+
+      .p-navigation__dropdown--container {
+        // a bit extra % to be able to show the shadow of the box if present
+        clip-path: rect(0 105% 105% 0);
+        display: block;
+        min-width: 100%;
+        position: absolute;
+      }
+
+      .p-navigation__dropdown--container > .p-navigation__dropdown {
+        display: block;
+        position: static;
+        transform: translateY(calc(-1 * var(--up)));
+        // same function as for the global-nav dropdown
+        transition: transform 0.333s cubic-bezier(0.215, 0.61, 0.355, 1);
+
+        &[aria-hidden='true'] {
+          // extra 10% to not let the shadow show underneath the navbar
+          --up: 110%;
+        }
+
+        &[aria-hidden='false'] {
+          --up: 0%;
+        }
       }
     }
   }
--- a/scss/_patterns_navigation.scss
+++ b/scss/_patterns_navigation.scss
@@ -1103,32 +1103,21 @@ $navigation-height: calc(map-get($settings-text-p, line-height) + 2 * $spv--larg
       &.menu-closing .p-navigation__nav {
         --right: 110%;
       }
-
-      .p-navigation__items--section {
-        margin: 0;
-        padding: 0;
-      }
     }
   }
 
   @media (min-width: $breakpoint-navigation-threshold) {
     .p-navigation--sliding {
-      &.has-menu-open {
+      &.has-menu-open.desktop-no-shadow {
         box-shadow: none;
-      }
-
-      .p-navigation__items {
-        justify-content: space-between;
-
-        &--section {
-          display: flex;
-          margin: 0;
-          padding: 0;
-        }
       }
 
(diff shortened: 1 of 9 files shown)


scss/_patterns_navigation.scss
```diff
@@ -190,6 +190,20 @@ $navigation-height: calc(map-get($settings-text-p, line-height) + 2 * $spv--larg
     }
   }
 
+  .p-navigation--sliding {
+    @property --right {
```
bartaz: Is there a reason to define those using full `@property` notation, instead of just defining their initial value:

```scss
--right: 0%;
--up: 0%;
```
?

Also, even though they are local variables inside navigation, it could be a good practice to prefix them in similar way other Vanilla CSS variables are prefixed, something like `--vf-nav-sliding-right`.

templates/docs/examples/patterns/navigation/sliding-dark.html
```diff
@@ -36,70 +36,76 @@
...
-              <a href="#" class="p-navigation__dropdown-item">Getting started - Command line</a>
-            </li>
-            <li>
-              <a href="#" class="p-navigation__dropdown-item">Getting started - OpenStack</a>
-            </li>
-            <li>
-              <a href="#" class="p-navigation__dropdown-item">Getting started - OpenNebula</a>
-            </li>
-          </ul>
+          <div class="p-navigation__dropdown--container">
```
bartaz, line 39: It looks like the change requires markup change (adding the additional div)?

If that's the case this is a likely breaking change.
We should potentially keep a legacy example of previous version of the markup (without the div) and make sure that it still works.

We should not expect people to update their markup when they update Vanilla. So any markup changes should ideally be optional (if they want a new feature, full functionality), but the component should still work without it.

If it's not possible, we would have to reconsider how to apply this fix.

scss/_patterns_navigation.scss
```diff
@@ -1064,4 +1078,71 @@ $navigation-height: calc(map-get($settings-text-p, line-height) + 2 * $spv--larg
...
+
+      &.menu-closing .p-navigation__nav {
+        --right: 110%;
+      }
+    }
+  }
+
+  @media (min-width: $breakpoint-navigation-threshold) {
+    .p-navigation--sliding {
+      &.has-menu-open.desktop-no-shadow {
```
bartaz: By convention in Vanilla such status class names start with `is-` or `has-`. It may be a bit weirder with a negative, but `has-no-shadow-on-large` could work?

But, is this option needed? I guess it was a design decision to have overlay. It would be best to consistently use it or not. Otherwise it would be unclear when would you use navigation with or without overlay on desktop.

templates/docs/examples/patterns/navigation/_script-sliding.js
```diff
@@ -5,21 +5,29 @@ const initNavigationSliding = () => {
   const toggles = document.querySelectorAll('.p-navigation__nav .p-navigation__link[aria-controls]:not(.js-back-button)');
   const searchButtons = document.querySelectorAll('.js-search-button');
   const menuButton = document.querySelector('.js-menu-button');
-  const dropdownNavLists = document.querySelectorAll('.js-dropdown-nav-list');
-  const topNavList = [...dropdownNavLists].filter((list) => !list.parentNode.closest('.js-dropdown-nav-list'))[0];
+  const dropdownNavLists = document.querySelectorAll('.p-navigation__dropdown');
```
bartaz: We are not very consistent with this, but we were trying to avoid using styling CSS class names in JS.

So in this case having `js-dropdown-nav-list` was meant to allow having a specific pointers for JS, that are not connected to CSS styles. 
Ideally we want to avoid using `p-` class names in JS, and either use custom `js-` class names, or other ways to select items (via DOM, parents, data attributes, aria attributes, etc).

templates/docs/examples/patterns/navigation/_script-sliding.js
```diff
@@ -5,21 +5,29 @@ const initNavigationSliding = () => {
...
+    const closeMenuHandler = () => {
+      if (secondaryNavigation) {
+        secondaryNavigation.classList.remove('has-menu-open');
+      }
+      navigation.classList.remove('has-menu-open');
+      navigation.classList.remove('menu-closing');
+      resetToggles();
+    };
+
+    // the time is aproximately the time of the sliding animation
```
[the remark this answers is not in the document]
alvaromateo: Fixed

templates/docs/examples/patterns/navigation/_script-sliding.js
```diff
@@ -5,21 +5,29 @@ const initNavigationSliding = () => {
...
+    setTimeout(closeMenuHandler, ANIMATION_SNAP_DURATION);
```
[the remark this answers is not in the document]
alvaromateo: It is defined at line 2.

scss/_patterns_navigation.scss
alvaromateo: Removed full notation and updated names to follow Vanilla CSS conventions.

scss/_patterns_navigation.scss
alvaromateo: I've removed the class. I added it cause I thought it was looking nicer without the overlay when in desktop view (obscuring all the page just for a little dropdown seemed overkill to me). But you're right to bring it up, cause I'm no designer and this decision is not on my hands.

templates/docs/examples/patterns/navigation/_script-sliding.js
alvaromateo: Reverted back to `js-dropdown-nav-list`

templates/docs/examples/patterns/navigation/sliding-dark.html
alvaromateo: This breaking change is the reason why at the beginning I added the styles in snapcraft.io instead of vanilla-framework.
I will research if there's a way to make this change optional without breaking the sites that don't apply it.

bartaz:
We discussed the changes with @alvaromateo and decided to park it for now.
We will review and revisit these custom changes on the snapcraft.io side, and see if they are necessary and worth upstreaming.

alvaromateo:
As agreed with @bartaz, I'm closing the PR because it's possible that it will cause some regressions.
We will discuss further if the changes are wanted/needed.

PR #5626, opened 2025-08-26
[DE-210] Publish contribution process

juanruitina:
## Done

- Create new level in the side navigation under Docs for contributions
- Add 3 pages for design contribution process: "Contribute to design", "Share an idea", "Submit a design proposal"
- Move content currently in /contribute to /docs/contribute, rename as "Contribute to code"
- Redirect /contribute to /docs/contribute
- Show contributors in the docs template on the /docs/contribute page, below the content as agreed. Small style tweaks for responsiveness.

## Pending/Known issues

- [x] Review changed code in app.py for injecting contributor info in /docs/contribute. @jmuzina please review that bit, I generated it with Copilot.
- [x] Fix sidebar: `aria-current="page"` seems to be incorrectly added to pages that have the last segment of the current url in their url. For instance, when in /docs/contribute/design, /docs/contribute/design-proposal is styled as if active.
- [ ] Ideally lists should have `max-width: 40em` as paragraphs do for optimal readability

## To be fixed later

- Add links to refreshed GitHub forms
- Content in /docs/contribute should perhaps be reworked a bit so it also provides a path for designers

## QA

Open [demo](https://vanilla-framework-5626.demos.haus/docs/contribute)

DE-210


.spelling
```diff
@@ -94,3 +94,13 @@ alt
...
 A11y
 standalone
+Figma
+FigJam
+Miro
+Penpot
+Charmhub
+dotrun
+behavior
+behaviors
```
jmuzina, line 104: Our mdspell spellchecker was throwing errors on your usage of "behavior" rather than "behaviour" as this project is still using EN-UK. We need to convert it to EN-US to follow suit with the rest of the company but this hasn't been done yet, so I've added it as a spelling exception. No action needed from you, just letting you know

templates/docs/contribute/design-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+|                                      | Variants                                     | Different implementations of the component Examples like blog cards versus Charmhub cards                                                                                                                                                                     |
+|                                      | Properties                                   | Configurable properties like size and color Show/hide options Editable text fields                                                                                                                                                                            |
+| When should we use the component?    | When to use (best practices)                 | Context and problem-solving scenarios Alternative component options Comparative advantages UI examples showing correct implementation Recommended practices or do’s and dont’s Usage limitations (such as primary button frequency) Implementation guidelines |
```
jmuzina: "do's and dont's" is currently causing a [spellcheck error](https://github.com/canonical/vanilla-framework/actions/runs/17241928642/job/48921252408?pr=5626).



We can add a spelling exception to allow this, but we have a choice ahead of us, maybe you should check with someone from Content.

[This article](https://www.vocabulary.com/articles/wc/dos-and-donts-or-dos-and-donts/) says that some style guides recommend "dos and don'ts", others recommend "do's and don'ts.", and others recommend "do's and don't's". We need to check the style guide to see what the recommendation across the company is.

juanruitina: Checked with Content, "dos and don'ts" it is. I fixed it in the md file. Would you mind adding the exception? (Dunno how apostrophes are handled...)

jmuzina:
Opening for more formal review and merging following [meeting about the interim publishing of the contribution process and its future in Pragma](https://docs.google.com/document/d/1HHhTD80-nya5CkrWOzWitDsehDdILz0mkPPHzIySahs/edit?tab=t.eu1fiayu8uv1#heading=h.q63oe87dtriw). I'll have a look shortly :)

webapp/app.py
```diff
@@ -320,26 +320,24 @@ def standalone_example(example_path):
...
+            "team_members": list(_get_team_members(contributors)),
+            "contributors": _filter_contributors(contributors),
+        }
+    return {}
+
+@app.after_request
+def set_contribute_cache_control(response):
+    if flask.request.path.rstrip("/") == "/docs/contribute":
+        response.cache_control.max_age = 86400
+        response.cache_control.public = True
```
jmuzina: Seems like a sensible change, copilot did an OK job here.

Previously, `/contribute` rendered its own template `contribute.html`, so it made sense for it to have its own route.

Now, however, `/contribute` reroutes to `/docs/contribute`, which inherits the `templates/_layouts/docs` template and renders Markdown from the appropriate contribution document. So, we no longer need an `@app.route()` for it, as the routing is handled by the `TemplateFinder`. All this code is doing is removing the route for contribute and injecting contributor context and response caching when the route matches the contribute page. 

I'm making a small readability change here to keep the naming of `all_contributors` to contain all contributors prior to filtering - I think it's cleaner for the `contributors` context var in the template to have the same name as the var in the context function

jmuzina:
>  Ideally lists should have `max-width: 40em` as paragraphs do for optimal readability

I'd suggest proposing this as a separate contribution/issue outside the scope of this PR @juanruitina

templates/docs/contribute/design-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+| :----------------------------------- | :------------------------------------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| What are the parts of the component? | Parts, or ‘anatomy’                          | List of component parts and their names Illustrations to show structure                                                                                                                                                                                      |
+|                                      | Variants                                     | Different implementations of the component Examples like blog cards versus Charmhub cards                                                                                                                                                                    |
```
jmuzina: Your "for example" column has list-style items in a single paragraph-style block, which means the text wrapping points are dependent on device size. This can cause your points to break at unexpected points:

Image

I'd recommend either:
- Adding explicit line breaks between the items
- (if "legal"): using [unordered lists](https://vanillaframework.io/docs/base/typography#unordered-list) inside the table cell.

templates/docs/contribute/design-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+| When should we use the component?    | When to use (best practices)                 | Context and problem-solving scenarios Alternative component options Comparative advantages UI examples showing correct implementation Recommended practices or dos and don’ts Usage limitations (such as primary button frequency) Implementation guidelines |
+|                                      | When not to use                              | Scenarios to avoid Team agreements on component usage                                                                                                                                                                                                        |
+| How is your contribution accessible? | Show how you meet accessibility requirements | Follow this [accessibility checklist](https://discourse.canonical.com/t/quick-accessibility-audit-for-designers/340)                                                                                                                                         |
```
jmuzina: I think this Discourse instance is private. I need to log in to access it. Maybe the article should be moved to the Ubuntu Discourse, blog, canonical.design, or directly inside this document?

templates/docs/contribute/design-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+- Gather concerns and validate them through research to help build consensus. User research can help if there’s disagreement on the solution or if conversations get too hypothetical.
+- Document your decisions and keep a list of sources to back them up, especially if your proposal follows common patterns or industry standards. You may not need additional research.
+
+Make sure to keep track of your conclusions and add them to the relevant sections of the design contributions form.
+
+## Suggesting changes to Figma libraries
+
+If your proposal is a component, pattern, or page template, you may need to create Figma components.
+
+1. Check our Figma libraries for [instructions on adding a component](https://www.figma.com/community/file/1435297834108003391). You can use [our Figma template](https://www.figma.com/files/team/989818926921821605/project/354426501/%F0%9F%AA%9A-Design-Systems-Component-Contributions?fuid=802129372533337888) to check what we expect.
```
jmuzina: I'm not able to view the second link ("our Figma template") when logged into my personal Figma account rather than my Canonical account. Maybe the permissions need to be checked.

templates/docs/contribute/design-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+**Here’s how to prepare a fully fledged design proposal that we can review and approve to add to our Design System.**
+
+We hold these proposals to the same standard as we do for design work at Canonical, but we are here to walk you through the process. If you just want to make a suggestion for our Design System, you can [submit an idea](/docs/contribute/idea) instead.
+
+Generally speaking, we accept proposals that:
+
+- Are flexible enough to handle many different use cases
+- Consider how other designers might use the design in the future
+- Match the styles from the design foundations of the Design System (font scale, colors, spacing)
+- Consider how [accessibility](https://vanillaframework.io/accessibility) will be taken into account
```
jmuzina: Given our conversation this morning with @advl and others, it may be prudent to add markdown comments to Vanilla-specific references throughout these documents so that we know what to replace when we need to adapt or migrate these documents to the new design system

Something like this would be nice, here and elsewhere where there are Vanilla-specific references

```md
- Match the styles from the design foundations of the Design System (font scale, colors, spacing)Add a comment on line R15Add diff commentMarkdown input: edit mode selected.WritePreviewAdd a suggestionHeadingBoldItalicQuoteCodeLinkUnordered listNumbered listTask listMentionReferenceSaved repliesAdd FilesPaste, drop, or click to add filesCancelAdd review commentReturn to code
<!-- TODO revisit/replace Vanilla-specific references when ready to add contribution process for the new design system -->
- Consider how [accessibility](https://vanillaframework.io/accessibility) will be taken into account
```

templates/docs/contribute/design-proposal.md
```diff
@@ -0,0 +1,102 @@
...
+Not all of these steps are mandatory, but a proposal can only be approved if it has a clear rationale.
+
+### 1. Discovery
+
+Before submitting an idea, look through the [Vanilla documentation](https://vanillaframework.io/docs) or [the backlog on GitHub](https://github.com/canonical/vanilla-framework/issues) in case we’ve addressed it already.
+
+Once you confirm there’s a need for it, we recommend that you:
+
+- Check articles or guidelines from reputable, external sources to inform your approach.
+- Benchmark solutions from other design systems, particularly those from leading organisations. We frequently look at systems like [Carbon](https://carbondesignsystem.com/) (IBM) and [Material Design](https://material.io/) (Google) for inspiration.
```
jmuzina: "organisations" is British English - normally this is fine in Vanilla as we haven't undertaken the effort to convert all of our pages to American English. However, you are using American English in this same document with "behaviors" and "color", and even this same word as "organization"  ("looking at how other organizations solve similar problems").

I think it's OK if the site has a mix of AE and BE, due to low capacity for us to convert the entire site to AE - but pages should be internally consistent and stick to one or the other.

```suggestion
- Benchmark solutions from other design systems, particularly those from leading organizations. We frequently look at systems like [Carbon](https://carbondesignsystem.com/) (IBM) and [Material Design](https://material.io/) (Google) for inspiration.
```

templates/docs/contribute/index.md
```diff
@@ -0,0 +1,36 @@
...
+wrapper_template: '_layouts/docs.html'
+context:
+  title: Contribute to code
+---
+
+When [submitting a new issue](https://github.com/canonical/vanilla-framework/issues/new/choose), please check that it hasn't already been raised by someone else. We've provided a template for new issues which will help you structure your issue to ensure it can be picked up and actioned easily.
+
+Please provide as much information possible detailing what you're currently experiencing and what you'd expect to experience.
+
+To work on an issue, please fork this repo and create a new branch on your local fork. When you're happy and would like to propose that changeset to be merged back upstream, open a pull request to merge from your local origin/main to upstream/main.
```
jmuzina: I think we should use code elements for code identifiers like variables, branch names, CLI commands, etc - it flags them as pieces of technical text rather than regular text

```suggestion
To work on an issue, please fork this repo and create a new branch on your local fork. When you're happy and would like to propose that changeset to be merged back upstream, open a pull request to merge from your local `origin/main` to `upstream/main`.
```

templates/_layouts/_contribute.html
```diff
@@ -0,0 +1,43 @@
+<h2 class="u-sv1">Meet the team</h2>
+
+<div class="grid-row">
+    {% for team_member in team_members %}
+    <div class="grid-col-2 grid-col-medium-2 grid-col-small-4 p-media-object">
```
jmuzina: This grid, and the contributors grid below, get a little cramped at the bottom end of the large breakpoint - see this screenshot taken at `1036px` width:

Image

Maybe we should use [`grid-row--50-50`](https://vanillaframework.io/docs/patterns/grid#common-patterns-new) instead, so these are 50/50 on medium and large, and 100 on small?

https://github.com/user-attachments/assets/50c69d70-7c56-4b86-ae3c-042e92a5f5c8

I'll leave this up to you or maybe @lyubomir-popov though, I'm no designer :)

jmuzina (changes requested):
Some small tweak suggestions and notes about internal resource access settings!

jmuzina:
Hey @eliman11 and @lyubomir-popov , can you provide UX and visual reviews, please?

templates/_layouts/_contribute.html
juanruitina: Yes, I felt two columns left them too spaced out in bigger screens, but my solution isn't optimal either. @lyubomir-popov your opinion will be greatly appreciated here,  you're more familiar with the options we got.

templates/docs/contribute/design-proposal.md
juanruitina: Replaced to a link to /accessibility, thank you for flagging it up!

templates/docs/contribute/design-proposal.md
juanruitina: Removed the link, it was not that necessary (the previous link to the community Figma file is good enough)

templates/docs/contribute/design-proposal.md
juanruitina: I added the TODO comment to all links pointing anywhere in Vanilla.

templates/docs/contribute/design-proposal.md
juanruitina: Thanks a lot for flagging this up. I turned the table into HTML to use better markup (including `rowspan`) and to indeed use unordered lists inside the table cells.

templates/_layouts/_contribute.html
lyubomir-popov: how about a 2x3 or square picture above the text? this way, we can keep 4 columns, and text will align weith text above. We could keep the avatars in circles if needed too.

redirects.yaml
```diff
@@ -10,6 +10,7 @@
 /docs/examples/patterns/x-small-capitalised: /docs/examples/patterns/text-small-caps
 /docs/base/separators: /docs/patterns/rule
 /docs/base/paper: /docs/base/themes
+/contribute: /docs/contribute
```
jmuzina, line 13: It might be confusing, from a UX point of view, that the "contribute" navigation link is not active-highlighted while we are on the contribute docs, since they are now classified under the `/docs` domain and `/contribute` is just a redirect to `/docs/contribute`. 

Screenshot 2025-09-05 at 10 18 25


I do think it's important for a contribution link to appear somewhere on the main page, and perhaps on all pages (nav or footer). Just not sure what should happen with this navigation link. What do you think?

One option could be to update the contribute nav link to point to `/docs/contribute`, and consider it the active route on these contribution pages - even though these pages are inside the `/docs` area.

templates/docs/contribute/design-proposal.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Suggesting changes to Figma libraries
+
+If your proposal is a component, pattern, or page template, you may need to create Figma components.
+
+1. Check our Figma libraries for [instructions on adding a component](https://www.figma.com/community/file/1435297834108003391)
+2. Create a share link for your file and set permissions to “anyone can view”
+3. Add the link to the GitHub issue
+
+The Figma library maintainers at Canonical will then review your work and add it into the shared library
```
jmuzina: Nit: missing punctuation
```suggestion
The Figma library maintainers at Canonical will then review your work and add it into the shared library.
```

templates/_layouts/_contribute.html
juanruitina: Just tweaked it, looks pretty neat. I am also used muted text instead of small for the number of contributors, looks cleaner IMO. I think avatars should remain circles as they come from GitHub (and they are circled there).

image

redirects.yaml
juanruitina: I think it's fine: the active button helps the user know where they are, but one is sufficient. Having "Docs" as active in the top nav and "Contribute to code" in the sidebar exactly communicates where the user is in the hierarchy, so I'd argue there's no need to have "Contribute" highlighted as well in the top nav. I think its lack of state is rather inconspicuous, I don't think it will confusing. @eliman11 would you agree?

eliman11: I agree - it's not ideal but if we go with the alternative of having the Contribute tab in the horiziontal nav in it's active state I feel like it would be equally confusing for it to stay active if users go through other pages using the side nav. Would suggest keeping as is as well!

eliman11:
Sorry for the delay! 
- [x] V minor but "Licences" should be spelled "Licenses" with AE spelling

- [ ] The tables in "Share an idea" and "Submit a design proposal" truncate on smaller screens with no way of viewing the rest of the content. Should this table instead allow for horizontal scrolling or switch into cards?

Screenshot 2025-09-09 at 12 22 43

Screenshot 2025-09-09 at 12 17 54

- [ ] The table of contents on the right get pushed to the bottom of the page on medium screens - I'm not sure if this makes sense for it to appear at the bottom. Should it instead be hidden for medium breakpoints? More of a question than anything though - feel free to ignore the suggestion.

Screenshot 2025-09-09 at 12 18 38

juanruitina:
Thank you for the review, Elaine. It seems that, for tables, neither horizontal scroll or switching into cards work well out of the box in Vanilla. 

For horizontal scroll, I think Vanilla lacks the styles? In the [CVEs pages](https://ubuntu.com/security/CVE-2025-6558) in ubuntu.com there's some custom code that could be reused:

```scss
@media screen and (width < $breakpoint-x-large) {
    .u-cve-table-horizontal-scroll {
      overflow-x: hidden;
      .u-cve-table-horizontal-scroll__table {
        display: block;
        overflow-x: auto;
        white-space: nowrap;
      }
    }
  }
```

If switching into cards, content doesn't wrap (`white-space: nowrap;`); I couldn't find an easy override with utilities. I'd argue that wrapping should be the default in cards as it is for [regular tables](https://vanillaframework.io/docs/base/tables#formatting): I find hard to justify hiding content for visual reasons. In any case, cards wouldn't work for the table in /contribute/design-proposal because there are joined cells.

An alternative would be to turn this content into something other than a table. Will discuss options with Julie.

As for the placement of the table of contents on mobile, that seems addressed in the [newer Documentation layout](https://vanillaframework.io/docs/layouts/documentation). @jmuzina shall we open a separate ticket for doing that update?

juanruitina:
After discussing with Julie, I've turned tables into headings and lists so we don't become blocked by current limitations in Vanilla tables. @eliman11 Can you re-review the previously tabulated content and +1 if OK? The table of contents showing up at the end is an issue across all documentation, because it's not using the new documentation template; it's not specific to these new pages, so should be handled separately.

There's a minor visual issue with the bottom margins in list items with nested items that seems to have to do with Markdown parsing (text within lists is wrapped in `p` tags). Julie will explore, but shouldn't be a blocker either.

eliman11:
That makes sense, thank you for the thorough explanation @juanruitina! Adding +1!

lyubomir-popov (approved):
LGTM

jmuzina:
> There's a minor visual issue with the bottom margins in list items with nested items that seems to have to do with Markdown parsing (text within lists is wrapped in p tags). Julie will explore, but shouldn't be a blocker either.

Having a look at this today - will keep you updated on what kind of solution I find

jmuzina:
I think I've fixed that list spacing bug @juanruitina - there were some extra line breaks in the code that were causing the list items that followed to be wrapped in a `<p>`. Did I fix all of the problem spots?

juanruitina:
LGTM! I think this is ready to merge now :)

jmuzina (approved):

PR #5641, opened 2025-09-15
For modal height use dvh to make it respect open url bars in mobile browsers

edlerd:
## Done

- For modal height use dvh to make it respect open url bars in mobile browsers
- Related issue reported in https://github.com/canonical/lxd-ui/issues/1496 with respect to a Network ACL configuration modal. The buttons in the footer are cut, as the height of the modal (100vh) does not take into account the reduced available height due to the open url bar. The change to 100dvh respects that and resolves the issue.


jmuzina (approved):
lgtm, thanks for this!

PR #5643, opened 2025-09-17
fix(badge): reset text-indent to prevent inheritance (#5629)

### LINKED ISSUE #5629: [Bug]: Badges are broken when used in filters
**Describe the bug**

When using badges in side `p-checkbox` (for filters) they don't display correctly anymore.
There is no text and the badge itself is misaligned with text.

**To Reproduce**

Example Codepen: 
https://codepen.io/bartoszopka/pen/raOZgRK

**Expected behavior**

Badges should display correctly regardless of their placement.

**Screenshots**

Before:

Image

After (latest Vanilla):

Image

Safvan-tsy:
## Done

Badges inside p-checkbox inherited a negative text-indent, causing the text to disappear. 
This change resets text-indent to 0 for p-badge to prevent inheritance.

Fixes #5629 

## QA

- [Add QA steps]

### DIFF
--- a/scss/_patterns_badge.scss
+++ b/scss/_patterns_badge.scss
@@ -17,6 +17,7 @@
     overflow: hidden;
     padding: 0 $spv--x-small;
     text-align: center;
+    text-indent: 0;
   }
 
   .p-badge,


jmuzina (commented):
Created a version of  @bartaz 's codepen from the underlying issue that we can use for QA here, since there are no badges inside forms/labels currently on the Vanilla docs sites.

https://codepen.io/jmuzina/pen/myPjjme

The badge indeed appears to render as expected when inside a checkbox/label - and there are [no visual regressions](https://percy.io/bb49709b/web/vanilla-framework/builds/45143627?utm_campaign=bb49709b&utm_content=vanilla-framework&utm_source=github_status_private).

@bartaz I want to check with you here, as in your [underlying issue](https://github.com/canonical/vanilla-framework/issues/5629#issuecomment-3237164325) you reported a workaround where you selectively reset `text-indent`, whereas this PR resets it in all cases. I didn't find any visual regressions with this approach, do you think there are situations where this approach would cause issues? 

Per the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-indent), indentation is applied relative to the "containing block-level element's content box" - which in this case is the badge itself - so I am led to believe that this approach will work. Let me know if we're missing something.

bartaz (approved):
Yes, I believe `text-indent` is fully safe to be applied this way.
My scoped approach was mostly regarding the vertical align (which was needed due to how we use badge ourselves), so it's OK for it to be omitted here I think.

PR #5648, opened 2025-09-18
feat(utilities): add top, bottom, text-top, and text-bottom to vertical-align utilities

### LINKED ISSUE #5644: [New utility]: Vertical-align utilities
Vanilla currently provides a utility class for vertical-align ([u-vertical-align--middle](https://vanillaframework.io/docs/utilities/align#vertical-alignment)), but lacks equivalents for other [values](https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align) like top and bottom.
Adding these would make alignment utilities more flexible and ensure consistent handling of inline elements like badges, icons, and form controls across different contexts.

## Visual

Image

## Context

These utilities would be useful in:
- Badges next to labels (aligning text and count consistently).
- Icons inline with text, where alignment sometimes needs to be top or bottom instead of middle.
- Form components (checkboxes, radio buttons, etc.) where inline alignment flexibility is required.

## State

Does this pattern have state? 
- No, these are static utility classes.

## Progressive enhancement

A pattern should be designed small screen first. How does this pattern scale up?
- Mobile-first design remains unaffected since vertical-align is a CSS inline property and degrades gracefully.

How will this pattern degrade on less competent browsers?
- On less capable browsers, inline elements will fall back to default alignment (baseline).

Safvan-tsy:
## Done

- Added new utility classes:
  - `u-vertical-align--top`
  - `u-vertical-align--bottom`
  - `u-vertical-align--text-top`
  - `u-vertical-align--text-bottom`
- Updated /docs/utilities/align#vertical-alignment to showcase all available vertical-align options.

Fixes  #5644 

## QA

- [Add QA steps]

## Screenshots

Screenshot from 2025-09-18 16-20-45

### DIFF
--- a/templates/docs/examples/utilities/align/vertical-align.html
+++ b/templates/docs/examples/utilities/align/vertical-align.html
@@ -2,10 +2,18 @@
 {% block title %}Align / Vertical align{% endblock %}
 
 {% block content %}
+<div style="overflow-x: auto; white-space: nowrap;">
 <h2>
+  <span style="padding: 4px 8px; border: 1px solid #aaa9a9; display: inline-block;">
   Vertical align
+  </span>
+  <span class="p-chip--information is-readonly is-inline u-vertical-align--top">top</span>
+  <span class="p-chip--information is-readonly is-inline u-vertical-align--text-top">text-top</span>
   <span class="p-chip--information is-readonly is-inline u-vertical-align--middle">middle</span>
+  <span class="p-chip--information is-readonly is-inline u-vertical-align--bottom">bottom</span>
+  <span class="p-chip--information is-readonly is-inline u-vertical-align--text-bottom">text-bottom</span>
 </h2>
+</div>
 {% endblock %}
 
 
--- a/templates/docs/utilities/align.md
+++ b/templates/docs/utilities/align.md
@@ -22,13 +22,20 @@ If you only need text to be aligned, you can use the `u-align-text` utilities.
 View example of the text align utility
 </a></div>
 
-## Vertical alignment
+## Vertical alignment {{ status('updated') }}
 
-If you need to vertically align inline elements to the middle of their
-container, you can use the `u-vertical-align--middle` utility.
+If you need to vertically align elements, you can use the vertical alignment utilities. These utilities work with inline and inline-block elements.
+
+The available vertical alignment utilities are:
+
+- `u-vertical-align--top` - Aligns the top of the element with the top of the tallest element on the line
+- `u-vertical-align--middle` - Aligns the middle of the element with the middle of the parent element
+- `u-vertical-align--bottom` - Aligns the bottom of the element with the bottom of the line
+- `u-vertical-align--text-top` - Aligns the top of the element with the top of the parent element's font
+- `u-vertical-align--text-bottom` - Aligns the bottom of the element with the bottom of the parent element's font
 
 <div class="embedded-example"><a href="/docs/examples/utilities/align/vertical-align" class="js-example">
-View example of the vertical align middle utility
+View example of the vertical alignment utilities
 </a></div>
 
 ## Import
--- a/releases.yml
+++ b/releases.yml
@@ -12,6 +12,10 @@
       url: /docs/patterns/logo-block
       status: Updated
       notes: We have renamed the formerly named "Logo section" component to "Logo block"
+    - component: Utilities / Align
+      url: /docs/utilities/align#vertical-alignment
+      status: Updated
+      notes: We've introduced new vertical alignment utility classes including <code>u-vertical-align--top</code>, <code>u-vertical-align--bottom</code>, <code>u-vertical-align--text-top</code>, and <code>u-vertical-align--text-bottom</code> for better control over alignment of an element.
 - version: 4.33.0
   features:
     - component: Newsletter signup pattern
--- a/scss/_utilities_content-align.scss
+++ b/scss/_utilities_content-align.scss
@@ -7,6 +7,10 @@
   @include vf-u-align--right;
   @include vf-u-align--bottom;
   @include vf-u-vertical-align--middle;
+  @include vf-u-vertical-align--top;
+  @include vf-u-vertical-align--bottom;
+  @include vf-u-vertical-align--text-top;
+  @include vf-u-vertical-align--text-bottom;
   @include vf-u-align--text;
 }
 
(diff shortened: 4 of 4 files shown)


jmuzina:
Thanks for the work @Safvan-tsy !

Starting demo at https://vanilla-framework-5648.demos.haus/

templates/docs/examples/utilities/align/vertical-align.html
```diff
@@ -2,10 +2,18 @@
 {% block title %}Align / Vertical align{% endblock %}
 
 {% block content %}
+<div style="overflow-x: auto; white-space: nowrap;">
```
jmuzina, line 5: Is this new outer div necessary? It seems you're using it to ensure the example renders on one line, but I wouldn't worry about this - if the elements are wrapping, they would never be vertically aligned anyway, so it's likely not a problem we need to solve.

We generally try to avoid using inline style in our examples if we can, so that the classes used in the examples are representative of the behavior a user would see if they used the classes in their own project.

I was able to remove the outer div and the example still appears fine:

Image

templates/docs/examples/utilities/align/vertical-align.html
```diff
@@ -2,10 +2,18 @@
...
 <h2>
+  <span style="padding: 4px 8px; border: 1px solid #aaa9a9; display: inline-block;">
```
jmuzina: What's your rationale/priority for these inline styles; to create a bordered, box-model reference point for the vertical alignment of the subsequent `span`s?

Maybe we could use [`u-visualise-font-metrics`](http://127.0.0.1:8101/docs/utilities/font-metrics) class on the `h2`, I think it is more purpose-built for this. 

Here's an example, @lyubomir-popov what do you think, does this send the right message in docs?

Image

templates/docs/utilities/align.md
```diff
@@ -22,13 +22,20 @@ If you only need text to be aligned, you can use the `u-align-text` utilities.
...
+- `u-vertical-align--top` - Aligns the top of the element with the top of the tallest element on the line
+- `u-vertical-align--middle` - Aligns the middle of the element with the middle of the parent element
+- `u-vertical-align--bottom` - Aligns the bottom of the element with the bottom of the line
+- `u-vertical-align--text-top` - Aligns the top of the element with the top of the parent element's font
+- `u-vertical-align--text-bottom` - Aligns the bottom of the element with the bottom of the parent element's font
 
 <div class="embedded-example"><a href="/docs/examples/utilities/align/vertical-align" class="js-example">
-View example of the vertical align middle utility
+View example of the vertical alignment utilities
 </a></div>
```
jmuzina, line 39: It might be slightly confusing from an API/UX point of view for us to introduce `u-vertical-align--text-top` and `u-vertical-align--text-bottom`, considering there is also `u-align-text--{left|center|right}`. We are mixing a bit the domains of vertical alignment and text alignment.

Image

@eliman11 Do you have any thoughts here on the best naming / documentation structuring strategy here?

templates/docs/utilities/align.md
```diff
@@ -22,13 +22,20 @@ If you only need text to be aligned, you can use the `u-align-text` utilities.
 View example of the text align utility
 </a></div>
 
-## Vertical alignment
+## Vertical alignment {{ status('updated') }}
 
-If you need to vertically align inline elements to the middle of their
-container, you can use the `u-vertical-align--middle` utility.
+If you need to vertically align elements, you can use the vertical alignment utilities. These utilities work with inline and inline-block elements.
```
jmuzina:
```suggestion
If you need to vertically align elements, you can use the vertical alignment utilities. These utilities work with `inline` and `inline-block` elements.
```

Nit: these are being used as code keywords, I'd recommend using code ticks here to make them stand out.

templates/docs/examples/utilities/align/vertical-align.html
Safvan-tsy: That was mainly to ensure the chips stay on a single line on smaller screens (otherwise, once they wrap, the visual difference in vertical alignment isn’t obvious anymore).

without :

https://github.com/user-attachments/assets/eefad3b3-0cb3-47c7-93cc-05d8e642cb66


with that wrapper div:

https://github.com/user-attachments/assets/ef2dfb5b-eb3e-4710-9ea3-d7bdefc156d7


> We generally try to avoid using inline style in our examples if we can, so that the classes used in the examples are representative of the behavior a user would see if they used the classes in their own project.

Since there aren’t existing utilities for overflow-x or white-space in vanilla, I used inline styles here.

* Wdyt ? should i drop that div entirely and let it wrap on mobile or keep it is as it is now?

templates/docs/utilities/align.md
Safvan-tsy: Addressed this in f3c104f

templates/docs/examples/utilities/align/vertical-align.html
Safvan-tsy: Yes, that’s correct, the bordered span was mainly meant as a reference box to make the vertical alignment of the chips clearer.

Thanks for pointing out `u-visualise-font-metrics` , I wasn’t aware of that utility. That does seem like a more appropriate choice. 

* updated in [d26f264](https://github.com/canonical/vanilla-framework/pull/5648/commits/d26f2645f2590352dc624447a043c3a1b0c14e21)

templates/docs/utilities/align.md
Safvan-tsy: I see the concern about mixing vertical alignment of elements with text alignment.

Currently, the docs describe:

> u-vertical-align--text-top - Aligns the top of the element with the top of the parent element's font
> u-vertical-align--text-bottom - Aligns the bottom of the element with the bottom of the parent element's font

I think this explanation is fairly clear, but I’m happy to improve it.

Would you prefer:
* Rename the utilities to something like `u-vertical-align--font-top` / `u-vertical-align--font-bottom` to make it explicitly text/font-specific, or
* Remove them if they’re rarely needed or feel redundant?

I’m happy to follow whichever approach makes this clearer

jmuzina:
- I'm checking with a designer (@dgtlntv ) to make sure the example visuals are appropriate for what we want to publish on the site
- I'm pinging a UX designer (@eliman11 ) for opinions on the utility naming

templates/docs/examples/utilities/align/vertical-align.html
jmuzina: Checked with @dgtlntv and we are okay with leaving the div with inline styles as we need the text to not wrap in order to demonstrate vertical alignment

Safvan-tsy:
Just a gentle follow up - still here and happy to adjust anything needed when someone has time 😊

jmuzina:
> Just a gentle follow up - still here and happy to adjust anything needed when someone has time 😊

Hey @Safvan-tsy - I've talked this over with our design team and they think some more conversation within the [underlying issue](https://github.com/canonical/vanilla-framework/issues/5644) is warranted. They'd like to understand the use-case a bit more. For now I'll be closing the PR, it may be re-opened depending on the outcome of that conversation.

Thanks for your patience.

PR #5659, opened 2025-10-07
wip: blog pattern

jmuzina:
## Done

Builds the [blog pattern](https://www.figma.com/design/1brHCsyeTE6FU7hfJd6ao2/%F0%9F%9A%A7---WIP--Sites---Core-component-library?node-id=3168-31983&m=dev)

## QA

- [Add QA steps]

### DIFF
--- a/templates/_macros/vf_basic-section.jinja
+++ b/templates/_macros/vf_basic-section.jinja
@@ -1,4 +1,5 @@
 {% from "_macros/shared/vf_dedent.jinja" import vf_dedent %}
+{% from "_macros/shared/vf_section_top_rule.jinja" import vf_section_top_rule %}
 {% from "_macros/vf_linked-logo-section.jinja" import vf_linked_logo_section %}
 
 # description_config
@@ -291,33 +292,6 @@
   </h2>
 {%- endmacro -%}
 
-{%- macro basic_section_top_rule(top_rule_variant="default") -%}
-  {%- set top_rule_variant = top_rule_variant | trim | lower -%}
-  {%- if top_rule_variant not in ["default", "muted", "highlighted", "none"] -%}
-    {%- set top_rule_variant = "default" -%}
-  {%- endif -%}
-
-  {%- if top_rule_variant != "none" -%}
-    {%- set top_rule_classes = "p-rule" -%}
-    {%- if top_rule_variant != "default" -%}
-      {#-
-        p-rule--highlighted doesn't exist (use p-rule--highlight instead), but p-rule--muted does.
-        We keep the external API here consistent ("-ed" suffix) for simplicity but need to handle this internally.
-      -#}
-      {%- if top_rule_variant == "highlighted" -%}
-        {%- set top_rule_classes = "p-rule--highlight" -%}
-      {%- else -%}
-        {#- Other cases: just append the top_rule_variant to the p-rule class. -#}
-        {%- set top_rule_classes = top_rule_classes + "--" + top_rule_variant -%}
-      {%- endif -%}
-    {%- endif -%}
-  {%- endif -%}
-
-  {%- if top_rule_variant != "none" -%}
-    <hr class="{{- top_rule_classes -}}"/>
-  {%- endif -%}
-{%- endmacro -%}
-
 # Params
 # title: A dictionary with "text" and optional "link_attrs" (a dictionary of link attributes for the title).
 # subtitle: A dictionary with "text" (required) and optional "heading_level" (default is 4).
@@ -358,7 +332,7 @@
 
 <section class="{{ padding_classes }}">
   <div class="grid-row--50-50{%- if not is_split_on_medium -%}-on-large{%- endif -%}">
-    {{ basic_section_top_rule(top_rule_variant) }}
+    {{ vf_section_top_rule(top_rule_variant) }}
     <div class="grid-col">
       {%- if has_label -%}
         <h3 class="p-muted-heading u-no-padding--top u-no-margin--bottom">{{- label_text -}}</h3>
--- a/templates/_macros/vf_blog.jinja
+++ b/templates/_macros/vf_blog.jinja
@@ -69,25 +69,23 @@
 # Params
 # title: A dictionary with "text" and optional "link_attrs" (a dictionary of link attributes for the title).
 # articles: A list of article dictionaries, each with "title", "image", "description", "author", and "date".
+# num_templates: An integer for the number of article templates to render if no articles are provided (for dynamic loading scenarios).
 {% macro vf_blog(
     title={},
-    articles=[]
+    articles=[],
+    num_templates=0
 ) -%}
-    {#
-        TODO strong assumption - if no articles are present, we are in "dynamic" mode and assume user will use JS to populate the articles themselves
-        Saves us from adding a "mode" param
-    #}
-
     {#- todo add title styling more properly -#}
     {{ title }}
 
-    {%- for article in articles -%}
-
-        
-
-    {%- endfor -%}
-
-
-
+    {%- if num_templates <= 0 -%}
+        {%- for article in articles -%}
+            {{- _blog_article(article) -}}
+        {%- endfor -%}
+    {%- else -%}
+        {%- for i in range(num_templates) -%}
+            {{- _blog_article() -}}
+        {%- endfor -%}
+    {%- endif -%}
 
 {% endmacro %}
--- a/templates/_macros/vf_blog.jinja
+++ b/templates/_macros/vf_blog.jinja
@@ -1,91 +1,111 @@
+{% from "_macros/shared/vf_section_top_rule.jinja" import vf_section_top_rule %}
+{% from "_macros/shared/vf_blog_article_block.jinja" import vf_blog_article_block %}
+
 # Params
+# - col_start: An integer override for the column start (for grid-col-start-*)
+# - truncation_config:
 # - article_config: A dictionary with the article configuration.
     # - "title": A dictionary with "text" and optional "link_attrs" (a dictionary of link attributes for the title).
     # - "image": A dictionary with "attrs" dict containing "src", "alt", and other image attributes. Always wrapped in 16:9 cover image container by the pattern and has `p-image-container__image` added. Fallback image used if no image passed, which can be overriden by "fallback_attrs" attr within this object.
     # - "description": A dictionary with "text", and "truncation_length" (an integer for max characters before truncation).
     # - "author": A dictionary with "text" and optional "link_attrs" (a dictionary of link attributes for the author).
     # - "date": A string with the publication date.
-{%- macro _blog_article(article_config={}) -%}
-    {%- set image = article_config.get("image", {}) -%}
-    {%- set title = article_config.get("title", {}) -%}
-    {%- set description = article_config.get("description", {}) -%}
-    {%- set description_truncation_length = description.get("truncation_length", 300) -%}
-    {%- set author = article_config.get("author", {}) -%}
-    {%- set has_author = author.get("text") | trim | length > 0 -%}
-    {%- set date = article_config.get("date", {}) -%}
-    {%- set has_date = date | trim | length > 0 -%}
-
-    <div class="p-section--shallow">
-        <div class="p-image-container--16-9 is-cover">
-        {#- todo fallback image -#}
-            <img class="p-image-container__image {{ image.get("attrs", {}).get("class", "") -}}"
-                {%- for attr, value in image.get("attrs", {}).items() -%}
-                    {% if attr != "class" %}
-                        {{ attr }}="{{ value }}"
-                    {%- endif -%}
-                {%- endfor -%}
-            />
-        </div>
-    </div>
-    <h5>
-        <a
-            {%- for attr, value in title.get("link_attrs", {}).items() -%}
-                {{ attr }}="{{ value }}"
-            {%- endfor -%}
-        >
-            {{ title.get("text", "Article Title")[:description_truncation_length] }}
-        </a>
-    </h5>
-    <p>
-        {{ description.get("text", "Article description goes here.") }}
-    </p>
-    {%- if has_author or has_date -%}
-        <small>
-            {%- if has_author -%}
-                {% if author.get("link_attrs") -%}
-                    <a
-                        {%- for attr, value in author.get("link_attrs", {}).items() -%}
-                            {{ attr }}="{{ value }}"
-                        {%- endfor -%}
-                    >
-                {%- endif -%}
-                {{ author.get("text") }}
-                {%- if author.get("link_attrs") -%}
-                    </a>
-                {%- endif -%}
-            {%- endif -%}
-            {%- if has_author and has_date -%}
-                &middot;
-            {%- endif -%}
-            {%- if has_date -%}
-                <time datetime="{{ date }}">{{ date }}</time>
-            {%- endif -%}
-        </small>
-    {%- endif -%}
-
-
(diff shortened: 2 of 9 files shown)


kim-isaac:
Thanks @jmuzina ! It looks great, I have some comments below:

[Docs related]

- [x] Since dynamic content is the most common use case, could we place it at the top?

> [For static content, blog descriptions are automatically truncated after 180 characters.](https://vanilla-framework-5659.demos.haus/docs/patterns/blog#dynamic-content:~:text=For%20static%20content%2C%20blog%20descriptions%20are%20automatically%20truncated%20after%20180%20characters.)

- [x] While static content is recommended to stay within 300 characters, there shouldn’t be any restriction or truncation within the pattern itself.

[Pattern related]
image

- [x] Would it be possible to lock the image ratio is always 16:9? (If different ratio is used, it could be center-aligned and filled.)
- [x] It looks like the block item spacing has been applied to all elements (image, title, description, and metadata). Could the title, description, and metadata all follow the paragraph spacing(6px top and 18px bottom)?
image

- [x] It seems the image height is fixed and not maintaining the correct ratio. Could we keep the height at 16:9?


+Also, even though it’s not used in this specific pattern, could we have an example of the resource block item for reference?

Thanks again and let me know if you have any questions!

I'll invite @eliman11 to the UX review as well :)

templates/_macros/vf_basic-section.jinja
```diff
@@ -291,33 +292,6 @@
   </h2>
 {%- endmacro -%}
 
-{%- macro basic_section_top_rule(top_rule_variant="default") -%}
```
muhammad-ali-pk, line 294: I believe this macro was being used in some of the other patterns as well, (e.g., `divided section`). Can you please update all those references to use the shared `vf_section_top_rule`?

templates/_macros/shared/vf_article_block.jinja
```diff
@@ -0,0 +1,187 @@
...
+        <div class="p-article-block__item">
+            {{- _article_title(title=title) -}}
+        </div>
+
+        <div class="p-article-block__item">
+            {{- _article_description(
+                description=description,
+                truncation=truncation_config.get("description", {})
+            ) -}}
+        </div>
```
muhammad-ali-pk: If no description is provided (e.g., a resources variant without descriptions), this code renders empty divs with some spacings, contributing to overall spacing of the block. Can you please add a conditional to only render this block if a description was provided?

image

templates/_macros/shared/vf_article_block.jinja
```diff
@@ -0,0 +1,187 @@
...
+
+        <div class="p-article-block__item">
+            {{- _article_metadata(authors=authors, date=date, template_mode=template_mode) -}}
+        </div>
```
muhammad-ali-pk: If no metadata is provided (e.g., a resources variant without metadata), this code renders empty divs with some spacings, contributing to overall spacing of the block. Can you please add a conditional to only render this block if metadata was provided?

image

jmuzina:
> It seems the image height is fixed and not maintaining the correct ratio. Could we keep the height at 16:9?

@kim-isaac Yes, thanks - this is due to the [latest news module](https://github.com/canonical/latest-news) not being able to add our image container image class, and the images coming from our Blog system not being a unified 16:9. I've raised an [issue](https://github.com/canonical/latest-news/issues/54) on latest-news that I can submit a contribution for, and added a temporary workaround [here](https://vanilla-framework-5659.demos.haus/docs/examples/patterns/blog/templates?theme=light) to show what the results will be.

Update: raised a PR to allow adding classes like image container classes to the article images, to address this behavior upstream. https://github.com/canonical/latest-news/pull/55

eliman11:
Thanks @jmuzina and @kim-isaac! Some minor comments - 
- [x] Doesn't affect the previews as we're using the muted heading style anyway, but could we change "Latest from our Blog" to use sentence case (i.e. "Latest from our blog")
- [x] Since these blog sections are usually standalone sections with the same hierarchy as other H2 headings on the page, could we use the muted heading style but use the H2 label instead of the H4? More so for users using screen readers to ensure we aren't skipping heading hierarchies
- [x] Similarly the article headings should then take on H3 labels instead of H5s
- [ ] Maybe more of a question for @kim-isaac but should we allow for some articles within the same block to show metadata, and some articles without metadata? Or even if it is possible with the pattern, would we want to showcase this as the first example?
Screenshot 2025-10-23 at 17 54 26

jmuzina:
Thanks @eliman11 I addressed all your points except for the question for @kim-isaac about mixed metadata presence, can you have another look?

templates/_macros/shared/vf_article_block.jinja
jmuzina: The presence of these divs, even when empty, is intentional, as all of the divs need to be present to enable the subgrid alignment to work properly - we rely on the blocks having the same number of items, so that if a description isn't present in one block, the next block's metadata still vertically aligns with the previous blocks. 

Further, because the articles may be loaded asynchronously (in dynamic contents mode), we can't know at request-time whether the description/metadata/title etc is empty. The markup/styling needs to work with empty slots for title/desc/metadata etc. So, we at minimum must always render the `p-article-block__item` and the empty elements for blog fields, and then apply spacing to those nodes only if they have content.

However, you're right that this means these items should not contribute to spacing in any case. I've updated the styling so `p-article-block__item` doesn't add any margin/padding; instead, its children are responsible for spacing. 

I've also left this comment to make this clearer:

>  p-article-block relies on subgrid to align items vertically across the grid.
        This means that we must always create 4 p-article-block__items, even if some are empty.
        This allows the subgrid to maintain alignment, even when some blocks are missing some items.
        Note that this means `p-article-block__item` elements should never define padding or margin, as this would cause empty items to take up space.

templates/_macros/shared/vf_article_block.jinja
jmuzina: See https://github.com/canonical/vanilla-framework/pull/5659#discussion_r2457425023

templates/_macros/vf_basic-section.jinja
jmuzina: done in https://github.com/canonical/vanilla-framework/pull/5659/commits/eb79c9a453820c544136b6629829830d2b94ef6e, thanks!

jmuzina:
@kim-isaac I've addressed your comments, please have another look and let me know what you think!

kim-isaac:
Thanks @jmuzina ! Here are a few more comments:

- [ ] Regarding @eliman11 's question, I agree that the metadata should either appear consistently or not appear at all within a single section.
- [x] In the document, please use the default title style (not the link style) as this would be the standard setting.
image

- [x] It seems both p-article-block__description article-excerpt and the nested p have 6px top padding, causing overlap - could we remove one of them?
- [x] Could we add 8px bottom spacing between stacked blocks in medium and small dimensions to match the current equal-heights row?

eliman11:
LGTM thanks @jmuzina and @kim-isaac for clarifying :)

muhammad-ali-pk:
@jmuzina When providing `link_attrs` to `article-titles`, the attributes do not appear to be forwarded to `anchor` element itself:

image


Also, if no `link_attrs` are provided, the title is still a link:

image

jmuzina:
### @kim-isaac 's comments

> I agree that the metadata should either appear consistently or not appear at all within a single section.

Yes, I would agree. To correctly align the grid and to properly support dynamic contents loading, I have to create a slot for metadata in all cases. I am handling this by only applying spacing to the item if content is passed.

I have updated the static contents examples to only show consistent presence/lack of metadata. 

I do have a static contents example that shows what might happen in various contexts where pieces of data are populated or not populated. This is to help ensure the spacing is consistent in cases where the blog API might return data in a shape we don't expect, such as one article having metadata and another not having any. 

This isn't exposed in public docs, but helps us for visual testing purposes.

image

> In the document, please use the default title style (not the link style) as this would be the standard setting.

Understood, done.

> It seems both p-article-block__description article-excerpt and the nested p have 6px top padding, causing overlap - could we remove one of them?

Done!

> Could we add 8px bottom spacing between stacked blocks in medium and small dimensions to match the current equal-heights row?

Done! Note that you may see this in inspection tools as `26px` or `1.625rem`. This is the sum of bottom margin for the last item (paragraph bottom margin of `1.125rem`/`18px`) and the desired stacked item gap (`0.5rem`/`8px`) and is handled this way for ease of implementation regarding handling the possible presence/lack of contents from the blog API.

### @muhammad-ali-pk 's comments

> When providing link_attrs to article-titles, the attributes do not appear to be forwarded to anchor element itself

Good catch, this is because I was expecting the attribute to be received as "link", rather than "link_attrs". I've updated the article block macro accordingly, it should work with `link_attrs` now.

> If no link_attrs are provided, the title is still a link

Good catch, this was indeed a bug. I've fixed it by updating the article block macro.

templates/_macros/shared/vf_article_block.jinja
```diff
@@ -0,0 +1,183 @@
...
+  {%- endif -%}
+    <h3
+      class="p-article-block__title article-title {{ title.get("class", "") }}"
+      {% for attr, value in title.items() -%}
+        {%- if attr not in ["text", "link_attrs", "class"] -%}
+          {{ attr }}="{{ value }}"
+        {%- endif -%}
+      {%- endfor -%}>
+      {{- title.get("text", "") | trim -}}
+    </h3>
```
muhammad-ali-pk: We need to make heading level of title customizable.
Context: For the resources pattern, the resources can be with or without categories. If the category is there, it is going to be h3, therefore the resource title needs to be h4 per UX.

I have added a [commit](https://github.com/canonical/vanilla-framework/pull/5659/commits/17d7355911f0091428cc020b957b12236d641c85) in your PR for this.

kim-isaac:
Thanks @jmuzina , looks great! I'll add +1

jmuzina:
Passing UX and Design, now requesting code and QA review - @muhammad-ali-pk could you help with this?

templates/_macros/vf_blog.jinja
```diff
@@ -0,0 +1,138 @@
...
+
+  {%- if articles | length == 3 -%}
+    {%- set layout = "3-blocks" -%}
+  {%- elif articles | length == 4 -%}
+    {%- set layout = "4-blocks" -%}
+  {%- endif -%}
+
+  {%- set template_mode = template_config.get("enabled", False) -%}
+
+  {%- if template_mode and (not articles | length) -%}
```
muhammad-ali-pk: Out of curiosity, what happens if a user passes in articles, but also enables template mode?

muhammad-ali-pk: Also, do we really need to check "not articles | length" in this conditional? From what it looks like down below, the template mode gets precedence, so maybe no need to check for articles length here?

muhammad-ali-pk:
Minor:

There are empty spaces in some classes, and some indentation issues
image

muhammad-ali-pk:
The code examples render macro usage code, with no indentations. Is it possible to have the code examples indent properly? Please ignore if the effort is too much and out of scope of this PR.

For example, 
```
{{ vf_blog(
  title={"text": "Latest from our blog"},
  articles=[
    {
      "title": {"text": "How to enable Real-time Ubuntu on your machine", "link_attrs": {"href": "#"}},
"description": {"text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit."},
"metadata": {
"authors": [
{
"text": "John Doe",
"link_attrs": {"href": "#"}
}
],
"date": {
"text": "15 March 2025"
}
}
},
{
"title": {"text": "How to enable Real-time Ubuntu on your machine", "link_attrs": {"href": "#"}},
"description": {"text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit."},
"metadata": {
"authors": [
{
"text": "John Doe",
"link_attrs": {"href": "#"}
}
],
"date": {
"text": "15 March 2025"
}
}
},
{
"title": {"text": "How to enable Real-time Ubuntu on your machine", "link_attrs": {"href": "#"}},
"description": {"text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit."},
"metadata": {
"authors": [
{
"text": "John Doe",
"link_attrs": {"href": "#"}
}
],
"date": {
"text": "15 March 2025"
}
}
},
{
"title": {"text": "How to enable Real-time Ubuntu on your machine", "link_attrs": {"href": "#"}},
"description": {"text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit."},
"metadata": {
"authors": [
{
"text": "John Doe",
"link_attrs": {"href": "#"}
}
],
"date": {
"text": "15 March 2025"
}
}
}
]
) }}
```

templates/_macros/vf_blog.jinja
jmuzina: > Out of curiosity, what happens if a user passes in articles, but also enables template mode?


The pattern ignores the articles list if template mode is enabled, and behaves as if template mode is enabled.

>  do we really need to check "not articles | length" in this conditional? From what it looks like down below, the template mode gets precedence, so maybe no need to check for articles length here?

Good point, we don't need to check article length here. I've removed that part of the check.

jmuzina:
> The code examples render macro usage code, with no indentations. Is it possible to have the code examples indent properly? 

@muhammad-ali-pk  Yes, looks like the indendations in the macro usage examples were not being rendered in the example iframe, even though they were in the example file, because I was embedding properties inside of objects declared on the same line. Ex:

```jinja2
"title": {"text": "lorem ipsum....", "link_attrs": {"href": "#"}},
```

I expanded this to always nest objects and properties on new lines and things now appear to be displaying properly. 

```jinja2
"title": {
  "text": "lorem ipsum....", 
  "link_attrs": {
    "href": "#"
  }
},
```

```jinja
{{ vf_blog(
  title={
    "text": "Latest from our blog"
  },
  articles=[
    {
      "title": {
        "text": "How to enable Real-time Ubuntu on your machine",
        "link_attrs": {
          "href": "#"
        }
      },
      "description": {
        "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit."
      },
      "metadata": {
        "authors": [
          {
            "text": "John Doe",
            "link_attrs": {
              "href": "#"
            }
          }
        ],
        "date": {
          "text": "15 March 2025"
        }
      }
    },
    {
      "title": {
        "text": "How to enable Real-time Ubuntu on your machine",
        "link_attrs": {
          "href": "#"
        }
      },
      "description": {
        "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit."
      },
      "metadata": {
        "authors": [
          {
            "text": "John Doe",
            "link_attrs": {
              "href": "#"
            }
          }
        ],
        "date": {
          "text": "15 March 2025"
        }
      }
    },
    {
      "title": {
        "text": "How to enable Real-time Ubuntu on your machine",
        "link_attrs": {
          "href": "#"
        }
      },
      "description": {
        "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit."
      },
      "metadata": {
        "authors": [
          {
            "text": "John Doe",
            "link_attrs": {
              "href": "#"
            }
          }
        ],
        "date": {
          "text": "15 March 2025"
        }
      }
    },
    {
      "title": {
        "text": "How to enable Real-time Ubuntu on your machine",
        "link_attrs": {
          "href": "#"
        }
      },
      "description": {
        "text": "Lorem ipsum dolor sit amet, consectetur adipiscing elit."
      },
      "metadata": {
        "authors": [
          {
            "text": "John Doe",
            "link_attrs": {
              "href": "#"
            }
          }
        ],
        "date": {
          "text": "15 March 2025"
        }
      }
    }
  ]
) }}
```

jmuzina:
> Minor: There are empty spaces in some classes, and some indentation issues 

@muhammad-ali-pk I've fixed these, good find - thanks!

PR #5672, opened 2025-10-17
fix: Support legacy imports of scss

edlerd:
## Done

* Existing usages use `@import "vanilla-framework/scss/build";`, so we have to export those nested paths as well.


edlerd:
duplicate of https://github.com/canonical/vanilla-framework/pull/5673

PR #5661, opened 2025-10-08
[WIP] feat: added ESM support

muhammad-ali-pk:
## Done

- Added ESM support

Fixes [#WD-29317](https://warthogs.atlassian.net/browse/WD-29317)

## QA

- This [demo PR](https://github.com/canonical/ubuntu.com/pull/15713) depicts an example of a project using bundler, like esbuild
  - Check out the [demo](https://ubuntu-com-15713.demos.haus/vanilla-demo) and verify the tabs are working as expected
- This [demo PR](https://github.com/canonical/canonical.com/pull/1996) depicts an example of a project without a bundler
  - Check out the [demo](https://canonical-com-1996.demos.haus/vanilla-demo) and verify the tabs are working as expected

### Check if PR is ready for release

If this PR contains Vanilla SCSS or macro code changes, it should contain the following changes to make sure it's ready for the release:

- [ ] PR should have one of the following labels to automatically categorise it in release notes:
  - `Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.
- [ ] Vanilla version in `package.json` should be updated relative to the [most recent release](https://github.com/canonical/vanilla-framework/releases/latest), following semver convention
  - if existing APIs (CSS classes & macro APIs) are not changed it can be a bugfix release (x.x.**X**)
  - if existing APIs (CSS classes & macro APIs) are changed/added/removed it should be a minor version (x.**X**.0)
  - see the [wiki for more details](https://github.com/canonical/vanilla-framework/wiki/Release-process#pre-release-tasks)
- [ ] Any changes to component class names (new patterns, variants, removed or added features) or macros should be listed on the [what's new page](https://github.com/canonical/vanilla-framework/blob/main/releases.yml).

## Screenshots

[if relevant, include a screenshot or screen capture]

PR #5673, opened 2025-10-17
Fix: exports the scss folder explicitly in package.json

advl:
## Done

- Explicitly exports the scss folder to maintain compatibility with deep import paths (e.g. `@import "vanilla-framework/scss/some-file";`), following the explicit exports introduced in #5661 

Fixes the breaking change introduced by #5661 for some applications (identified by @edlerd )

## QA

- Import a nested file of the vanilla framework from a consuming application, `@import "vanilla-framework/scss/some-file";` should work

### DIFF
--- a/templates/docs/patterns/segmented-control/index.md
+++ b/templates/docs/patterns/segmented-control/index.md
@@ -94,4 +94,4 @@ or reference directly from `node_modules`
 <script src="/node_modules/vanilla-framework/js/tabs.js"></script>
 ```
 
-For more information, see [Building Vanilla with JavaScript](http://localhost:8101/docs/building-vanilla#javascript)
+For more information, see [Building Vanilla with JavaScript](/docs/building-vanilla#javascript)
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -27,6 +27,9 @@
     ".": {
       "sass": "./_index.scss"
     },
+    "scss/*": {
```
edlerd:
```suggestion
    "./scss/*": {
```

edlerd: Also: shouldn't this be ordered alphabetically after the two js lines below?

advl: Good catch

I would keep the scss exports before the js for clarity

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vanilla-framework",
-  "version": "4.34.2",
+  "version": "4.34.3",
```
bartaz: I believe the whole ESM change should have been released as at least a minor release (not patch as it was), so I'd suggest doing it as `4.35.0`.

advl: Updated, thanks for the suggestion

advl:
Merged with a temporarily disabled percy check

PR #5683, opened 2025-10-31
chore(Tiered List): Refactor tiered list macro to extract duplicated code blocks

### LINKED ISSUE #5530: Tiered list: Extract tiered list item blocks
This is a minor code quality update proposal for the tiered list pattern.

This block is duplicated (appears twice) in the macro:
```jinja
<div class="row">
  <div class="col-9 col-start-large-4">
    {%- if number == 1 %}
      {# Skip the first HR #}
    {%- else %}
      <hr class="is-muted">
    {%- endif %}
  </div>
</div>
```

This should be extracted into an "inner" (private) macro, to avoid duplication. See [this example](https://github.com/canonical/vanilla-framework/blob/573e6d604a006259883c1995eaf5f8f98019901d/templates/_macros/vf_pricing-block.jinja#L33) of how we declare inner macros for this kind of use case. 

[This block](https://github.com/canonical/vanilla-framework/blob/3b382533f474804c699c008b2db3c248a6714287/templates/_macros/vf_tiered-list.jinja#L85-L89) is also mostly duplicated by [this block](https://github.com/canonical/vanilla-framework/blob/3b382533f474804c699c008b2db3c248a6714287/templates/_macros/vf_tiered-list.jinja#L101-L109) - the only difference is column classes. This is another good candidate for an inner macro.

First noticed [here](https://github.com/canonical/vanilla-framework/pull/5529#discussion_r2113934623)

Nayab-Gauhar:
Extract duplicated code blocks into private helper macros to improve maintainability.

Created two private helper macros:
- `_vf_tiered_list_item_separator()`: Renders the HR separator between list items
- `_vf_tiered_list_item()`: Renders list item content with parameterized column classes

This eliminates code duplication while maintaining the exact same functionality.

Fixes #5530

## Done
- Extracted HR separator block into `_vf_tiered_list_item_separator()` macro
- Extracted list item content block into `_vf_tiered_list_item()` macro with parameterized column classes
- Replaced duplicated code blocks with macro calls
- Verified that functionality remains unchanged

## QA
- The changes are purely structural/refactoring - no visual or functional changes
- All existing tiered list examples should continue to work exactly as before
- The macro API remains unchanged

- [ ] PR should have one of the following labels: `Maintenance 🔨` (suggested for this refactoring)
- [ ] Vanilla version in `package.json` should be updated (this is a maintenance/refactoring change, so a bugfix release x.x.X would be appropriate)
- [ ] No changes to component class names or macro APIs, so no updates needed to what's new page


templates/_macros/vf_tiered-list.jinja
```diff
@@ -9,6 +9,24 @@
...
+      <hr class="is-muted"/>
+    </div>
+  </div>
+{%- endmacro %}
+
+{#- Private helper macro for rendering list item content -#}
+{% macro _vf_tiered_list_item(title_col_classes, description_col_classes, title_content, description_content) -%}
+  <div class="grid-row">
+    <div class="{{ title_col_classes }}">{{ title_content }}</div>
+    <div class="{{ description_col_classes }}">{{ description_content }}</div>
```
jmuzina, line 26: nitpick: can we only add the `class` attributes here if the class parameters are defined and have length > 0?

PR #5689, opened 2025-10-31
Fix #5594: Align pre element with baseline grid using nudge system

### LINKED ISSUE #5594: Component amendment: Pre
## Component/pattern to amend

THe pre element is not on the baseline grid. It's padding-top and margin bottom are not based on the nudge system

## Visual

```
pre {
    background-color: var(--vf-color-background-code);
    color: var(--vf-color-text-default);
    display: block;
    margin-bottom: 1.5rem;
    margin-top: 0;
    overflow: auto;
    padding: .5rem 1rem;
    text-align: left;
    text-shadow: none;
    white-space: pre;
}
```

This leads to displacement from the baseline grid on following elements.

## Context

Image

Nayab-Gauhar:
## Summary
This PR fixes issue #5594 by aligning the `<pre>` element with the baseline grid using the nudge system.

## Changes
- Updated `padding` to use `$spv-nudge` for top padding and `($spv--small - $spv-nudge)` for bottom padding
- Adjusted `margin-bottom` to `$spv--x-large - $spv-nudge` to compensate for the nudge value

## What this fixes
The `<pre>` element was not properly aligned with the baseline grid because it wasn't using the nudge system that other text elements use. This fix applies the standard nudge pattern:
1. Adds nudge value to padding-top to shift content down
2. Compensates at padding-bottom to maintain total height
3. Adjusts margin-bottom to account for the nudge

## QA
- Open any page with `<pre>` elements in the documentation
- Verify that the pre elements now align properly with the baseline grid
- Check that the visual appearance and spacing remains correct

## Checklist
- [ ] PR should have the `Bug 🐛` label
- [ ] Package version update (if needed for release)


dgtlntv:
@jmuzina It looks like thats due to the margin of the `<pre>` element being 2 pixels too large. Since the height of the element is divisible by the baseline the margin in this case must be divisible by the baseline as well. (If the element height would not be divisible by the baseline then the margin would need to account for it so element + margin is on the baseline again).

PR #5694, opened 2025-11-17
feat(Quote): Quote wrapper pattern uses new grid; build quote block

### LINKED ISSUE #5589: Update quote wrapper pattern to use 4/4/8 grid
Update [quote wrapper](https://vanillaframework.io/docs/patterns/tiered-list) to use the new `grid-row` (4/4/8 grid)   instead of `row-` (legacy 4/6/12 grid)

jmuzina:
## Done

- Enables embedding the pattern in future patterns
  - Updates the quote wrapper pattern to use the new 4/4/8 grid.
  - Refactors the quote wrapper pattern to use blocks. These blocks could be consumed by other patterns in the future.
  - Simplifies the use of separators (p-rule) inside the quote wrapper (see [thread](https://www.figma.com/design/1brHCsyeTE6FU7hfJd6ao2?node-id=303-12548&m=dev#1517835614))
    - Top rule is a default rule, instead of a highlighted rule.
    - Removed muted rules within the quote block. Muted rules will only be used to separate quote blocks from each other when the pattern supports multiple quotes. 
  - Removed the shallow section spacing around quote text (it is no longer present in the [design](https://www.figma.com/design/n7klX1aSaAZjqTwf3ZSsKS/25.04-Things-to-fix-%F0%9F%9B%A0%EF%B8%8F---Sites?node-id=91-2499&m=dev))

Fixes #5589 
Fixes [WD-23969](https://warthogs.atlassian.net/browse/WD-23969)

## QA

- Use [quote examples](https://vanilla-framework-5694.demos.haus/docs/examples/patterns/quote-wrapper/combined) and [Percy](https://percy.io/bb49709b/web/vanilla-framework/builds/44819398?utm_campaign=bb49709b&utm_content=vanilla-framework&utm_source=github_status_private) to compare the examples against the design. 
  - **Please be sure to check on all 3 breakpoints** and carefully look at all examples in the example set. Pay special attention to the use of horizontal rules, spacing beneath quote text, and grid column distribution of signpost column & quote body. 
  - To see the grid distribution change, compare [before](https://vanillaframework.io/docs/examples/patterns/quote-wrapper/large?theme=light) and [after](https://vanilla-framework-5694.demos.haus/docs/examples/patterns/quote-wrapper/large?theme=light) on the medium breakpoint.
- Open quote block examples and match it against the [design](https://www.figma.com/design/1brHCsyeTE6FU7hfJd6ao2/%F0%9F%8C%90-Sites---Core-component-library?node-id=303-12548&m=dev). 
  - [default](https://vanilla-framework-5694.demos.haus/docs/examples/patterns/quote-wrapper/block?theme=light)
  - [no signpost](https://vanilla-framework-5694.demos.haus/docs/examples/patterns/quote-wrapper/block-no-signpost?theme=light)
- Review [release notes](https://vanilla-framework-5694.demos.haus/docs/whats-new)

[WD-23969]: https://warthogs.atlassian.net/browse/WD-23969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/docs/examples/patterns/quote-wrapper/block.html
+++ b/templates/docs/examples/patterns/quote-wrapper/block.html
@@ -1,7 +1,7 @@
 {% extends "_layouts/examples.html" %}
 {% from "_macros/vf_quote-wrapper.jinja" import vf_quote_block %}
 
-{% block title %}Quote block / No signpost{% endblock %}
+{% block title %}Quote block{% endblock %}
 {% block standalone_css %}patterns_all{% endblock %}
 
 {% block content %}
--- a/templates/docs/examples/patterns/quote-wrapper/block-no-signpost.html
+++ b/templates/docs/examples/patterns/quote-wrapper/block-no-signpost.html
@@ -1,7 +1,7 @@
 {% extends "_layouts/examples.html" %}
 {% from "_macros/vf_quote-wrapper.jinja" import vf_quote_block %}
 
-{% block title %}Quote block{% endblock %}
+{% block title %}Quote block / No signpost{% endblock %}
 {% block standalone_css %}patterns_all{% endblock %}
 
 {% block content %}


templates/_macros/vf_quote-wrapper.jinja
```diff
@@ -28,146 +266,50 @@
...
-          {% if has_image -%}
-            {#-  Optional image block  -#}
-            {{ image_content }}
-          {% endif -%}
-        {% endif -%}
-      </div>
-    </div>
+          <hr class="p-rule is-fixed-width u-hide--medium u-hide--large">
+      {% endif %}
+        {#- TODO resolve https://www.figma.com/design/1brHCsyeTE6FU7hfJd6ao2?node-id=303-12548&m=dev#1517835614 and adjust show_divider accordingly -#}
```
jmuzina: @kim-isaac Can you please have a look at the [referenced Figma comment](https://www.figma.com/design/1brHCsyeTE6FU7hfJd6ao2?node-id=303-12548&m=dev#1517835614)?

jmuzina: @kim-isaac thanks for the comment reply, can you have a look at this PR's demo and let me know if things look as you expect?

kim-isaac:
Thanks @jmuzina , Looks nice!
I have two comments below:

- [ ] At the medium dimension, the title and link should each expand to 100% width and stack vertically. (please check the [figma file](https://www.figma.com/design/n7klX1aSaAZjqTwf3ZSsKS/25.04-Things-to-fix-%F0%9F%9B%A0%EF%B8%8F---Sites?node-id=54-19208&t=jel3QriMAs8y6vPo-11))
- [x] Could I we have an example screens showing 2 or more blocks together? I'd like to verify whether the muted rule appears between them

Thanks again!

jmuzina:
> At the medium dimension, the title and link should each expand to 100% width and stack vertically. (please check the [figma file](https://www.figma.com/design/n7klX1aSaAZjqTwf3ZSsKS/25.04-Things-to-fix-%F0%9F%9B%A0%EF%B8%8F---Sites?node-id=54-19208&t=jel3QriMAs8y6vPo-11))

Thanks for pointing this out, I've adjusted accordingly:

[Screencast from 2025-11-20 09-55-44.webm](https://github.com/user-attachments/assets/d803c86f-0965-4778-9920-59e6d4ab6c92)



> Could I we have an example screens showing 2 or more blocks together? I'd like to verify whether the muted rule appears between them

@kim-isaac This PR doesn't add support for quote wrappers with multiple quote blocks, as that would change the API of the pattern. I'd suggest this can be done in a follow-up. In the current PR, the quote wrapper will never have muted rules, because there is only ever one block. However, I have prepared the pattern to support multiple blocks by exposing a `has_divider` setting at the block level, which the quote wrapper pattern could set to `True` for quote blocks after the first one.

https://github.com/jmuzina/vanilla-framework/blob/bebae8899eec8a2823095e396a6584874f59e8ea/templates/_macros/vf_quote-wrapper.jinja#L284-L285

kim-isaac:
@jmuzina Thank you for the update!

My concern about the second part is that, if two or more quotes appear within a single section, they may all be treated as the first block and come without the rule for all of them ([live site example](https://ubuntu.com/robotics/community#:~:text=Hear%20from%20our%20community))

If that’s the case, then until we support multiple quote blocks within the pattern, I think we should keep the rule visible even on the first block.

jmuzina:
@kim-isaac 

> My concern about the second part is that, if two or more quotes appear within a single section, they may all be treated as the first block and come without the rule for all of them ([live site example](https://ubuntu.com/robotics/community#:~:text=Hear%20from%20our%20community))
> 
> If that’s the case, then until we support multiple quote blocks within the pattern, I think we should keep the rule visible even on the first block.

I'd like to clarify which rules we're talking about here. This pattern currently has two HRs:

- default rule(formerly highlighted) at the top of the pattern above the title, spanning the grid
- muted rule above quote text, spanning the second 75%. Should only be shown on non-first blocks.

Since this pattern's initial implementation, that top (default) rule has only been shown if a title is present. This remains the case now. 

I've taken advantage of this for the time being by only showing the muted rule if no title row is present, which seems to be the case for users who are rendering quote patterns one after another, as you linked. The result is that we can remove the muted rule from quotes with title rows (these are always the first block as far as I understand) and keep it for quotes without title rows. Here's an [example](https://vanilla-framework-5694.demos.haus/docs/examples/patterns/quote-wrapper/multiple.html). Does this work? If so, quote wrapper could benefit from the new approach to rules early (refactoring it to use two blocks will be much more effort).

Alternatively, I can always render the muted rule.

kim-isaac:
@jmuzina Thanks for the explanation!
Given the structure you described, it sounds like updating the pattern won’t cause any issues on the live pages. That’s reassuring :) I'll add +1 with the current structure you've shared!

templates/docs/examples/patterns/quote-wrapper/block.html
```diff
@@ -0,0 +1,38 @@
+{% extends "_layouts/examples.html" %}
+{% from "_macros/vf_quote-wrapper.jinja" import vf_quote_block %}
+
+{% block title %}Quote block / No signpost{% endblock %}
```
advl: the titles of the examples should be exchanged in {% block title %}

PR #5736, opened 2026-01-16
fix(Table): Add keyboard support to table sorting

### LINKED ISSUE #5735: [Bug]: Table sorting is not keyboard accessible
**Describe the bug**

The [sortable table](https://vanillaframework.io/docs/base/tables#sortable)'s sorting functionality is not keyboard-accessible.

The table header elements have no role or tabindex, so they are not interactable for screen readers. Our example implementation uses only JS to fire the sort events. 

**To Reproduce**

1. Go to [table example](https://vanillaframework.io/docs/examples/patterns/tables/table-sortable?theme=light)
2. Try to use a keyboard to sort the data - see that you cannot focus or interact with the headers.
3. Use devtools to set `tabindex="0"` on the `th` elements, to make the headers keyboard interactable.
4. Focus a sortable tab header and press ENTER.
5. Verify the table still has not been sorted (it is only listening for click events) 

**Expected behavior**

I should be able to tab through a table's sortable headers, and press ENTER to change the sort direction. 

Sort direction should be announced with `aria-sort` (it already is).

Reference for example: [Material UI sortable table](https://mui.com/material-ui/react-table/#sorting-amp-selecting)

**Additional context**

Reported by @paul-geoghegan

jmuzina:
## Done

Makes sortable table headers keyboard-accessible by wrapping the text in button elements. Follows [w3c example](https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/) which recommends wrapping the header text in a button, inside the th.

Fixes https://warthogs.atlassian.net/browse/WD-32943
Fixes #5735 

React Components will require an update (updating the table headers markup) to benefit from this fix.

## QA

- Open [table example](https://vanilla-framework-5736.demos.haus/docs/examples/patterns/tables/table-sortable)
- Use the keyboard to sort the table. You can use SPACE or ENTER on a header to sort it.
- Verify sorting still works, identical to using the mouse.
- Verify focus ring appears around the header when focused, and not when hovered.
- Open [legacy sortable table example](https://vanilla-framework-5736.demos.haus/docs/examples/patterns/tables/table-sortable-legacy) and verify it can still be sorted (with the mouse, not with the keyboard).
- Use a screen reader like ORCA and verify the table headers are semantically described as table headers with interactive buttons, and they announce when the table is sorted. 
- Use dev tools to add lots of text to a table header so that the text is ellipsized. Verify that text is still ellipsized and sorting still works (though the sorting icon is hidden by the ellipses.... this is a pre-existing issue that should be handled separately). Verify the text doesn't overflow (it is truncated)
- Review [sortable table docs](https://vanilla-framework-5736.demos.haus/docs/base/tables#sortable-updated)
- Review [4.39.1 release notes](https://vanilla-framework-5736.demos.haus/docs/whats-new)

## Design

- I used the `button` element for proper interactivity semantics, but this carries with it the default button styling. I made the table sort button inherit the table header label's styling to negate this, but have a look and make sure I didn't miss anything.
- I handled the focus ring by applying a 2x4x2x4 padding around the button to make the focus ring appear like a link focus ring (since this visually appears like a link, not a button). Then I cancelled out the padding with negative margin to avoid affecting the size of the header itself. This is worth extra attention.

### DIFF
--- a/scss/_patterns_table-sortable.scss
+++ b/scss/_patterns_table-sortable.scss
@@ -22,7 +22,6 @@
   // stylelint-disable selector-max-type -- table elements can use selector types
   table th {
     &[aria-sort] {
-      align-items: center;
       cursor: pointer;
       outline-color: $colors--theme--focus;
       white-space: nowrap;
@@ -39,7 +38,6 @@
         color: inherit;
         display: inline-block;
         font: inherit;
-        height: 100%;
         // Cancel the focus outline space to avoid affecting layout
         margin: -2px -#{$sph--x-small};
         max-width: 100%;


juanruitina:
The keyboard navigation looks good to me: there's focus states, chevrons indicate sorting, hitting space works.

There is a bug: I've played a bit with it and sorting seems to be alphabetical; it should be smart enough to detect numbers so in those cases it sorts from higher to lower (instead of "1, 10, 12, 140, 2, 20, 3"). Happy to help test this.

jmuzina:
> There is a bug: I've played a bit with it and sorting seems to be alphabetical; it should be smart enough to detect numbers so in those cases it sorts from higher to lower (instead of "1, 10, 12, 140, 2, 20, 3"). Happy to help test this.

@juanruitina  I agree - though I believe this warrants a separate issue focusing on the sorting logic itself, rather than the keyboard accessibility. Can you file an issue?

dgtlntv:
LGTM. I was going to change what the focus state for a table header in terms of sorting is going to look like in the new table that I was designing which you can find here: https://www.figma.com/design/2MYJvamCVgbyWXZRT7IPan/25.10-Data-table---Design-system-contributions?node-id=4351-96714&t=1oohgcxOtOvarVUn-4.
But i think the more link inspired focus state here is probably appropriate for Vanilla as the clickable thing in Vanilla looks more like a link. So I think this is good, thanks for the a11y improvement! :)

scss/_patterns_table-sortable.scss
```diff
@@ -24,7 +24,38 @@
...
+      outline-color: $colors--theme--focus;
       white-space: nowrap;
+
+      &:hover,
+      &:focus-visible {
+        color: $colors--theme--link-default;
+      }
+
+      .p-table__sort-button {
+        // "inherit" styles are used to ensure this uses th styles instead of button defaults
```
bartaz, line 35: Can be worth looking at "link button" to see if there is anything that can be shared from there.

It doesn't look to extendable at the first glance, but maybe at least to verify if you override the same set of props:

https://github.com/canonical/vanilla-framework/blob/main/scss/_patterns_buttons.scss#L121-L147

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vanilla-framework",
-  "version": "4.39.0",
+  "version": "4.39.1",
```
bartaz: Maybe not completely necessary, but it's kind of more of a new feature than just a patch, so maybe 4.40 would be more fitting/

releases.yml
```diff
@@ -1,3 +1,9 @@
+- version: 4.39.1
+  features:
+    - component: Table
+      url: /docs/base/tables#sortable-updated
+      status: Updated
+      notes: Improved keyboard accessibility by enabling sorting via the Enter and Space keys.
```
bartaz: I'd probably focus more on addition of the native button. To also make it clearer that potential update to HTML is needed to enable this.

scss/_patterns_table-sortable.scss
```diff
@@ -39,7 +70,6 @@
     }
 
     &[aria-sort]:hover {
```
bartaz, line 70: I wonder if it would be worth removing hover styles from cell and move them to button, but that would technically be a breaking change, so maybe not worth it in the scope of this.

jmuzina: Yes I agree, I'll avoid a breaking change

scss/_patterns_table-sortable.scss
jmuzina: I had a look and it seems I'm overriding the same set of properties: background, border, margin, and padding, and then applying some additional styles to allow the contents of the button to be ellipsized as the `th` is.

scss/_patterns_table-sortable.scss
```diff
@@ -24,7 +24,38 @@
     &[aria-sort] {
       align-items: center;
       cursor: pointer;
+      outline-color: $colors--theme--focus;
       white-space: nowrap;
+
+      &:hover,
+      &:focus-visible {
+        color: $colors--theme--link-default;
+      }
```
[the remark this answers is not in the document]
jmuzina: kept for backwards compatibility

Issue #5475, opened 2025-02-19
[Bug]: Wrong active state in side navigation / table of contents


dgtlntv:
**Describe the bug**
On some of our pages / apps the side navigation / table of contents shows the wrong active state.

**To Reproduce**
On a page like CVEs there is a table of contents on the left. If you click on resources and then on description you can see that resources is still in active state, even though you clicked on description.
https://ubuntu.com/security/CVE-2025-21703

I have seen this issue in the 360 tool as well.

i am filing this here since it seems to be an issue we have in multiple site and apps. So it should probably be fixed at a DS level.

**Expected behavior**
The thing that is actually currently in view / what was clicked on is the active state in the table of contents / side navigation.

**Screenshots**
![Image](https://github.com/user-attachments/assets/fee05987-e01d-4740-838d-dc253c594eb5)

jmuzina:
Thanks for the bug report and repro steps @dgtlntv . I can indeed reproduce this on the [cves page](https://ubuntu.com/security/CVE-2025-21703#references). It seems the correct section is highlighted active when you scroll the page just after clicking on the "description" section.

Digging into this, I think this may be an issue with some `IntersectionObserver` code that [many of our sites have been copying and re-using](https://github.com/search?q=org%3Acanonical%20intersectionobserver&type=code), not with Vanilla itself. 

Here's what that code looks like in the u.com CVE page: https://github.com/canonical/ubuntu.com/blob/a1a4b104e00e60f20ad7704cbac77cbc5eca44a4/templates/security/cves/cve.html#L613-L631

I did some browser debugging and found that the intersection observer is not triggered after clicking the "description" or "status" sections on the CVE you linked, until after scrolling the page. This may have something to do with the intersection observer threshold settings. I'll send this to the sites team and see if they can investigate the JS on their end.

immortalcodes:
Hey, the issue is a bit different for me
https://www.loom.com/share/16f2aa50f9ac427b8a55c0c9407bc8d0?sid=ddede9c8-6812-42b5-b68e-84ae17e06e58

PR #5755, opened 2026-02-06
feat: in-page navigation component

petesfrench:
## Done

- Creates in page navigation component based on [figma specification](https://www.figma.com/design/gvs2FfJuQiSG9PhDTxsadn/Sites-in-page-navigation?node-id=272-43244&p=f&m=dev)

## QA

[Documentation page](https://vanilla-framework-5755.demos.haus/docs/patterns/in-page-navigation)
[Full-page example](https://vanilla-framework-5755.demos.haus/docs/examples/patterns/in-page-navigation/full-page)
[Manual example](https://vanilla-framework-5755.demos.haus/docs/examples/patterns/in-page-navigation/manual)
[Component example](https://vanilla-framework-5755.demos.haus/docs/examples/patterns/in-page-navigation/component)
[Component with optional heading example](https://vanilla-framework-5755.demos.haus/docs/examples/patterns/in-page-navigation/component-with-optional-heading)
[Changelog](https://vanilla-framework-5755.demos.haus/docs/whats-new)

- Check the implementation aligns with the design
- Check the automated navigation captures all headings
- Check navigation to headings through clicking the link or scrolling updates the navigation appropriately
- Check it works with keyboard navigation
- Review the updated documentation

- [ x ] PR should have one of the following labels to automatically categorise it in release notes:
- [ x ] Vanilla version in `package.json` should be updated relative to the [most recent release](https://github.com/canonical/vanilla-framework/releases/latest), following semver convention
- [ x ] Any changes to component class names (new patterns, variants, removed or added features) or macros should be listed on the [what's new page](https://github.com/canonical/vanilla-framework/blob/main/releases.yml).

### Fixes

https://github.com/canonical/vanilla-framework/issues/5475

### DIFF
--- a/tests/parker.js
+++ b/tests/parker.js
@@ -12,7 +12,7 @@ function generateMetrics(file, metricsArray) {
     {
       name: 'Stylesheet size',
       benchmark: 150000,
-      threshold: 522850,
+      threshold: 532480,
       result: results['total-stylesheet-size'],
     },
     {


kim-isaac:
Hi @petesfrench Thanks a lot for this work! It looks amazing.

I have some comments below:

**Documentation**

- [x] Since the full-page mode without a heading is the default, could we make this example ane place it at the top?

- [x] Please rename the version with a heading to “Manual mode, with optional heading.” Also, could you add a short description below? It’s meant to clarify what the heading is and when to use it.

> There is an option to add a heading on top of the in-page navigation component, which can be helpful in cases where the in-page navigation targets a specific section, not the entire page.

- [x] To better communicate that the heading is optional, could this be changed to “Optional heading” instead of "In this page"?
image


**Example**

Would it be possible to switch the examples so that:
* the full-page mode shows the version without a heading, and
* the manual mode shows the version with a heading? Please set the manual mode title to “What is the Linux kernel?” 
Also, could you remove the newsletter sign up from the second and third groups?

**Component**

Interaction:

- [x] Would it be possible to enable smooth scrolling when an item is clicked?

- [x] Earlier, I mentioned adding an 80px offset to account for the fixed global navigation. However, in cases where the global navigation is not sticky, could we exclude that offset?

Offset behavior proposal:
* nav sticky = true
    * Desktop: 80px
    * Mobile / Tablet: 120px
* nav sticky = false
    * Desktop: 24px
    * Mobile / Tablet: 72px

- [x] Separately from the offset, could we change the detection line for the navigation item highlight from the top to the center of the viewport? We’ve noticed cases where lower items pass by without being highlighted, and since users typically focus on the center of the screen while scrolling, this would likely be a more appropriate trigger point.

Mobile / Tablet:

- [x] In the Collapsed variant, there is currently 6px padding between the items and the bottom of the container. Could this padding be removed? (The bar height is 48px.)
image

- [x] The toggle button currently sits inside the grid, creating 1em spacing on the right. Could this button be moved outside the grid and fixed to the far right edge?
as-is:
image
to-be:
image

- [x] Please use the alternate background in the expanded version.

Thanks again!!

eliman11:
Thanks @petesfrench! I messaged @kim-isaac directly on questions/comments so will just +1 to everything she's listed already and have another look when you've made the changes

eliman11:
Hi Pete sorry looked into accessibility best practices a little more - I'm not sure why but sometimes on loading the demo page, tabbing through the navigation comp keeps the focus area there, and sometimes using one of the anchor links to go into a content section shifts the focus to the content area. Could we ensure that we allow for the latter - i.e. tabbing through and jumping into content sections via the page navigation comp takes you through the links in the content area, rather than back to the navigation comp? Example functionality [here](https://www.w3.org/WAI/WCAG22/Understanding/focus-order#intent)

Also @kim-isaac pointed out that when you tab through links in the content and the page scrolls down, the active tab in the navigation component doesn't seem to update - could we have the active tabs update as it would if you were scrolling through the page? Thank you!

kim-isaac:
Thanks @petesfrench !! A few comments are left, but I think we're almost there

**Documentation:**

- [x] Could we change the first "Manual" to Full-page?

image

It would also be great if the example title could be “In-page navigation / full-page”.

- [x] For the second one, the example title could be “In-page navigation / manual”.
image



**Examples:**

- [x] For the truncation example heading, would it be possible to change to a more formal version, such as “General Availability (GA) and variant Ubuntu kernels for different hardware and workloads”?

- [x] Also, on the manual example page, it would be good to move the newsletter sign-up to the very top of the content. (ATM it sits in the middle alongside the in-page navigation.)

**Component:**

- [x] On the smaller screen, the rule seems to extend beyond the viewport. Do you think this could impact the actual page width? From a visual perspective, spanning 100% of the viewport width should be sufficient.
image

- [x] Also, both collapsed and expanded variants at tablet size doesn’t appear to be 100% width. It looks like the left and right margins might still be set to -16px (the mobile value).
image

- [x] On mobile, the chevron button height is currently 56px. With a 16px icon and 16px padding on all sides, it should instead be 48*48px.
image


- [x] On the manual example page, the indicator bar is duplicated. Also, the heading should sit within the grid with 1em left padding. Could you help fix this?
image

Thank you!

eliman11:
Thanks Pete! 
Sorry just noticed the typo, I believe it should be "in-page" rather than "in page" everywhere we mention the component - is this correct @kim-isaac?

I think on tablet and mobile the expected behaviour is for the heading to be highlighted on the nav component only when the heading is at the top of the viewport? Atm the heading is highlighted in the midpoint of the previous section - see here:
Screenshot 2026-02-11 at 17 00 19

But will wait for input from @kim-isaac on the second part too in case this is what was intended.

kim-isaac:
@petesfrench Related to @eliman11 's comments, it was part of what I asked for in my initial comment.

> Separately from the offset, could we change the detection line for the navigation item highlight from the top to the center of the viewport? We’ve noticed cases where lower items pass by without being highlighted, and since users typically focus on the center of the screen while scrolling, this would likely be a more appropriate trigger point.

I’ve aligned with Elaine on this, and for touch-based breakpoints (mobile and tablet), having the scroll detection line aligned with the offset feels more appropriate.

So, would it be possible to keep the detection line centered in the viewport on desktop as originally requested, and move it to the top for the mobile and tablet breakpoints(same with the offset)?

I should have considered and explained this more clearly at the beginning - so so sorry for causing the back and forth!!

petesfrench:
> Could we change the first "Manual" to Full-page?

This example uses 'Manual' mode. Using 'Full-page' as an embedded iframe would be difficult due to the extra content that would need to be added.

> For the second one, the example title could be “In-page navigation / manual”.

This is dependent on my above comment.

For the scroll detection. I have limited it to just the top of the screen. Let me know if it needs adjustment. (if needed, I can include this [debugger script](https://github.com/pomber/intersection-observer-debugger?tab=readme-ov-file) which helps vizualise the detection points)

kim-isaac:
@petesfrench Thanks!! The debugger script was helpful to check detection points 😄 
Still, could we exclude this on the final publishing version?

**Documentation:**

- [x] It seems the tooltip doesn’t appear for the overflow in the example screen under the docs. Could we show the full hover state there as well?
- [x] Also, since ‘no heading’ would be the default state, could we rename the example screen title as below?
As-is: In-page navigation / Component, no heading
To-be: In-page navigation / manual

image

 For the second section, 
- [x] heading: In-page navigation / manual(as-is) → Manual with optional heading(to-be)
- [x] Example screen title : In-page navigation / Component, no heading(as-is) → In-page navigation / manual with optional heading
image

**Examples:**

- [ ] The tooltip is under the list layer -- could it be fixed?
image

- [x] Also could you change the heading "General Availability (GA) and variant Ubuntu kernels bla bla bla" to “General Availability (GA) and variant Ubuntu kernels for different hardware and workloads”?
image


- [ ] It seems the overflow’s independent scroll isn’t applied when the screen gets smaller. Could you add that, please?
image

- [x] On smaller screens, it looks like the link width is set to 2-col. Could we make it span the full width of the item?
image
image

- [x] On mobile, the chevron button has 24px padding on the left and right. Could we change it to 16px, and keep 24px only for tablet?
image


Thank you!!

kim-isaac:
@petesfrench Hey! sorry for adding comments -- I just noticed a few typos:

Clickling → Clicking ([link](https://vanilla-framework-5755.demos.haus/docs/patterns/in-page-navigation#:~:text=Clickling,-the%20links%20will))

navigaton → navigation ([1](https://vanilla-framework-5755.demos.haus/docs/patterns/in-page-navigation#:~:text=the%20page.%20The-,navigaton,-also%20highlights%20the),[2](https://vanilla-framework-5755.demos.haus/docs/patterns/in-page-navigation#:~:text=auto%2Dgenerating%20the-,navigaton.,-secondary_heading),[3](https://vanilla-framework-5755.demos.haus/docs/patterns/in-page-navigation#:~:text=auto%2Dgenerating%20the-,navigaton.,-Should%20follow%20%60primary_heading))

colums → columns ([link](https://vanilla-framework-5755.demos.haus/docs/patterns/in-page-navigation#:~:text=screens%2C%20and%204-,colums,-on%20medium%20and))

heirachy → hierarchy ([link](https://vanilla-framework-5755.demos.haus/docs/patterns/in-page-navigation#:~:text=an%20object.%20The-,heirachy,-and%20order%20of))

Srting → String ([link](https://vanilla-framework-5755.demos.haus/docs/patterns/in-page-navigation#:~:text=No-,Srting,-%22full%2Dpage%22))

string → String ([link](https://vanilla-framework-5755.demos.haus/docs/patterns/in-page-navigation#:~:text=navigation_items%7B%7D.id,the%20target%20heading))

TYSM!!

eliman11:
LGTM thank you so much for all your work and patience on this Pete!

kim-isaac:
@petesfrench There were a lot of comments and thank you so much for taking care of all of them! I'll also add +1

immortalcodes:
image
Observation: even though I am at the last heading, the side navigation does not highlight it correctly.
Also the case with smaller screens
image

immortalcodes:
https://vanilla-framework-5755.demos.haus/docs/examples/patterns/in-page-navigation/component-no-heading
The URL gives 404

immortalcodes:
image
What happens when we don't pass the navigation items, can we still use the pattern in intended way, if not; wouldn't it make it required?

immortalcodes:
image

Since the primary heading could only be h2/h3 , shall we mention that in description?

immortalcodes:
An example for this parameter would be nice
image

immortalcodes:
To maintain consisteny with other patterns, for params that are limited to specific values, we can exclusively mention the possible values in `type`
e.g.
image

templates/docs/examples/patterns/in-page-navigation/_in-page-navigation.js
```diff
@@ -0,0 +1,432 @@
...
+  return excludeList;
+}
+
+/**
+ * Generate CSS selectors for query based heading retrieval.
+ * @param {HTMLElement} navRoot
+ * @returns {Object} An object containing primarySelector, secondarySelector, and a query string
+ */
+function generateSelectors(navRoot) {
+  const primarySelector = navRoot.dataset.inPageNavigationPrimary;
```
immortalcodes, line 292: `data-in-page-navigation-primary` could be null as well

templates/_macros/vf_in-page-navigation.jinja
```diff
@@ -0,0 +1,87 @@
...
+    <template class="js-in-page-nav-template-item">
+      <li class="p-in-page-navigation__item p-tooltip--right" aria-describedby="#">
+        <a class="p-in-page-navigation__link" href="#"></a>
+        <span class="p-tooltip__message u-hide" role="tooltip" id="#"></span>
+      </li>
+    </template>
+    <template class="js-in-page-nav-template-sublist">
+      <ul class="p-in-page-navigation__list">
+      </ul>
+    </template>
```
immortalcodes: Curious why do we use JS auto-gen for full-page?

petesfrench:
> https://vanilla-framework-5755.demos.haus/docs/examples/patterns/in-page-navigation/component-no-heading The URL gives 404

The link changed, this should work: https://vanilla-framework-5755.demos.haus/docs/examples/patterns/in-page-navigation/component-with-optional-heading?theme=light

petesfrench:
> image Observation: even though I am at the last heading, the side navigation does not highlight it correctly. Also the case with smaller screens image

Due to how the [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) (IO) works there are going to be some compromises in the navigation tracking. There are some ways around it but I was trying to avoid over-engineering the the JS.

One solution here is the track if we are at the bottom of the page and then override the IO. Items near the bottom but not the last would still not be tracked though.

cc: @kim-isaac

petesfrench:
> image What happens when we don't pass the navigation items, can we still use the pattern in intended way, if not; wouldn't it make it required?

The [full-page example](https://vanilla-framework-5755.demos.haus/docs/examples/patterns/in-page-navigation/full-page?theme=light) auto-generates the navigation and does not require passing nav_items to the jinja pattern

templates/docs/examples/patterns/in-page-navigation/_in-page-navigation.js
petesfrench: We set a default and check any passed values are valid in the [macro](https://github.com/canonical/vanilla-framework/blob/wd-30490/templates/_macros/vf_in-page-navigation.jinja#L23C10-L23C15), so it should never be `null`

templates/_macros/vf_in-page-navigation.jinja
petesfrench: Do you mean why don't we auto build the navigation in jinja? This is because jinja can't parse html and the rendered server side

kim-isaac:
> Due to how the [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) (IO) works there are going to be some compromises in the navigation tracking. There are some ways around it but I was trying to avoid over-engineering the the JS.

> One solution here is the track if we are at the bottom of the page and then override the IO. Items near the bottom but not the last would still not be tracked though.

Sounds good to me. Also, on the actual page the footer will be present at the bottom, so all sections will cross the scroll detection line and all items will be highlighted in practice.

petesfrench:
> > One solution here is the track if we are at the bottom of the page and then override the IO. Items near the bottom but not the last would still not be tracked though.
> 
> Sounds good to me. Also, on the actual page the footer will be present at the bottom, so all sections will cross the scroll detection line and all items will be highlighted in practice.

Thanks for your input @kim-isaac! In that case, I would suggest we keep the implementation as is and iterate as needed. As not to block this work further.

templates/docs/examples/patterns/in-page-navigation/full-page.html
```diff
@@ -0,0 +1,254 @@
...
+      <div class="grid-col-6 grid-col-start-large-3">
+        <h1>What is the Linux kernel?</h1>
+      </div>
+    </div>
+  </section>
+
+  <section class="p-section">
+    <div class="grid-row">
+      <div class="grid-col-2 grid-col-medium-4 grid-col-small-4">
+        {{ vf_in_page_navigation(navigation_items=page_navigation_data,
```
immortalcodes: `page_navigation_data` seems undefined here

templates/docs/patterns/in-page-navigation/index.md
```diff
@@ -0,0 +1,239 @@
...
+        </td>
+        <td>
+          <code>String</code>
+        </td>
+        <td>
+          <code>"full-page"</code>
+        </td>
+        <td>
+          <strong>"full-page"</strong> - Automatically generates the in-page navigation and attaches heading IDs. Max 1 navigation per page. Requires associated JS.<br>
+          <strong>"manual"</strong> - Requires passing `navigation_items` dictionary. No limit on number of navigations per page.
```
immortalcodes, line 159: After understanding the entire implementation, the name `full-page` seems a bit confusing to me because you can still have a navigation that spreads full-page using `manual` mode, just that it would not be auto generated.
I feel `auto-gen`/`manual` could be good options for params.
cc: @kim-isaac

templates/docs/patterns/in-page-navigation/index.md
immortalcodes, line 1: Also worth mentioning in the docs page that interactions require the JS logic and may be we can even link the JS file from github so that devs could easily copy and use that logic

immortalcodes:
Just an idea, what if we only had auto-generation as a mode.
We could provide the `id` of the parent element where we wanted the side navigation, it will auto-detect the headings and form the navigation.
If the id was root, it would mean full page and the JS would ignore any other instances of` in-page navigation `and if it was say `#section-2` it would pull all the headings of just section-2 and create` in-page navigation` only in section-2.

Again this has nothing to do with current implementation just me brainstorming.

templates/docs/patterns/in-page-navigation/index.md
petesfrench: There are already a few explicit mentions of it and it is include in the snippet. But I have added an extra sentence and emphasized another.

petesfrench:
> Just an idea, what if we only had auto-generation as a mode. We could provide the `id` of the parent element where we wanted the side navigation, it will auto-detect the headings and form the navigation. If the id was root, it would mean full page and the JS would ignore any other instances of`in-page navigation`and if it was say `#section-2` it would pull all the headings of just section-2 and create` in-page navigation` only in section-2.
> 
> Again this has nothing to do with current implementation just me brainstorming.

I like the idea! For now, I think we should exclude it as this work is required by some other projects and I don't want to block its release with additional features.

It could be worth taking a note of it somewhere

PR #5326, opened 2024-08-26
Add theme support to tooltip component

jmuzina:
## Done

- Updates tooltips to invert their ancestor's color theme. I.E.: tooltips on dark use light theme, tooltips on light/paper use dark theme.
- Tooltips can be set to use a different theme than the inverse of the body, to handle cases where the tooltip is contained within an element that has a theme different from the body.

Fixes [WD-11871](https://warthogs.atlassian.net/browse/WD-11871)

## QA

- Verify that all tooltip examples look correct on all color themes. They should always use the opposite color theme.
  - [Default](https://vanilla-framework-5326.demos.haus/docs/examples/patterns/tooltips/default?theme=light)
  - [Detached](https://vanilla-framework-5326.demos.haus/docs/examples/patterns/tooltips/detached?theme=light)
  - Theme overrides - ensure that the "correct" tooltip inverts the theme of its surrounding strip, not the page body.
    - [Light](https://vanilla-framework-5326.demos.haus/docs/examples/patterns/tooltips/nesting-theme-override-light)
    - [Dark](https://vanilla-framework-5326.demos.haus/docs/examples/patterns/tooltips/nesting-theme-override-dark)
  - Review [combined tooltips example](https://vanilla-framework-5326.demos.haus/docs/examples/patterns/tooltips/combined) and verify the new examples have been added.
  - Review [theming tooltips docs](https://vanilla-framework-5326.demos.haus/docs/patterns/tooltips#theming)

### Check if PR is ready for release

If this PR contains Vanilla SCSS code changes, it should contain the following changes to make sure it's ready for the release:

- [x] PR should have one of the following labels to automatically categorise it in release notes:
  - `Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.
- [ ] Vanilla version in `package.json` should be updated relative to the [most recent release](https://github.com/canonical/vanilla-framework/releases/latest), following semver convention:
  - if CSS class names are not changed it can be bugfix release (x.x.**X**)
  - if CSS class names are changed/added/removed it should be minor version (x.**X**.0)
  - see the [wiki for more details](https://github.com/canonical/vanilla-framework/wiki/Release-process#pre-release-tasks)
- [ ] Any changes to component class names (new patterns, variants, removed or added features) should be listed on the [what's new page](https://github.com/canonical/vanilla-framework/blob/main/releases.yml).

## Screenshots

image
Screenshot 2024-08-30 at 10 19 36 AM

[...]

PR #5774, opened 2026-03-17
feat: add circle-of-friends icon and branded chip component

### LINKED ISSUE #5699: [New component]: <Branded chip>
A branded chip is needed for sites (and possibly also apps, so I would suggest this is a global component) to label items in relationship to Ubuntu Pro.
Discussed with Lidia, Mattea, Elaine. The chip will first live on sites, on in the Hero section of pages that need to be related to the Pro offering. it will also be used in dataviz or other instances where an item needs to signal this relationship. 

The proposal is a high contrast black chip with the circle of friends icon. 

This contribution includes adding the Circle of friends as an icon to the design system. 

Differences from the chip: 
- text uses "medium" because the regular text feels unbalanced next to the circle of friends; semibold starts feeling too heavy > see type tokens implications
- the left padding is 4px instead of 8 so tat the circle of friends is centered in the curve oif the chip
- the dense chip is used as a baselin
- the chip is 24px high instead of 23 (propose to also change the chip to 24).

## Visual

Image
Image

## Context

In hero sections, under the title and subitile, in data viz, in tables, in cards, etc.

Image
Image
Image
Image

## State

No states. 

Unoptimized svg icon for circle of friends: 
Image

Image

mariadias143:
## Done

- Creates branded chip component based on [figma specification](https://www.figma.com/design/2kjeRr23M9qh93UlBzQqGR/Branded-chip?node-id=1-8&p=f&m=dev)
- Adds Circle of Friends icon to standard icons
- Removed design guideline that prohibits usage of chip component for branding. If this were to be updated to a "when to use" scenario, then I would need a new image and description to go with.
- Drive by: fix links in CTA section which were breaking linkchecker CI step (caused by changes in the last merged PR)

Fixes [#5699](https://github.com/canonical/vanilla-framework/issues/5699)

## QA

- Check [the implementation](https://vanilla-framework-5774.demos.haus/docs/patterns/chip#branded-chip-new) aligns with the design
- Open [icons docs](https://vanilla-framework-5774.demos.haus/docs/patterns/icons) and verify p-icon--circle-of-friends has been added to the standard icons sections ([standard](https://vanilla-framework-5774.demos.haus/docs/patterns/icons#standard) and [dark](https://vanilla-framework-5774.demos.haus/docs/patterns/icons#dark-theme)).
- Open the [standard icons example](https://vanilla-framework-5774.demos.haus/docs/examples/patterns/icons/standard) and verify the circle of friends example appears as expected in all themes.
- Review updated documentation

## Screenshots
image
image
image
image
image
image
image


diana-stanciulescu:
@mariadias143 thanks! The icon-only version should be a perfect circle. Othewr than this, this looks great. 
I assume users can change the text if needed? Product managers may have different needs too.

scss/standalone/patterns_icons-additional.scss
jmuzina, line 1: This adds the CoF icon to the "additional" icons set. This is fine, but comes with the tradeoff that projects that want to consume it need to import the icon manually - and projects that aren't using SASS [will not be able to easily use the icon](https://chat.canonical.com/canonical/pl/fquf8ethsj85dpuke94aoqzrna). 

I guess it comes to the question of frequency - @diana-stanciulescu how heavily do we expect this icon to be used - and in which types of projects (sites / apps)? If we foresee heavy usage, it might make sense to add it to the standard icons set.

mariadias143:
> @mariadias143 thanks! The icon-only version should be a perfect circle. Othewr than this, this looks great. I assume users can change the text if needed? Product managers may have different needs too.

@diana-stanciulescu Yes, users can use whatever text they would like. Regarding the chip not being a perfect circle, I think one of the screenshots I added in the PR description (second one specifically) looked weird because of my monitor but it looks good on my laptop (see below). I updated the screenshot in the description as well. What do you think?


ACCESSIBILITY
Pasted Graphic 2

scss/standalone/patterns_icons-additional.scss
diana-stanciulescu: @jmuzina I think it makes sense to add it to our icons set - this chip will be used on both sites and apps so it's global. we won't use it all over the place but still nice to have available.

diana-stanciulescu:
@mariadias143 thanks for checking, the circle part seems ok, but what I noticed now is the icon looks small , can you please double check: 

Screenshot 2026-03-17 175348

scss/_patterns_chip.scss
```diff
@@ -273,3 +274,59 @@
...
+    align-items: center;
+    background-color: $colors--dark-theme--background-alt;
+    gap: 4px;
+    height: 24px;
+    justify-content: center;
+    margin-bottom: $spv-nudge-compensation;
+    min-width: max-content;
+    padding: 2.5px 8px 2.5px 4px;
+
+    [class*='p-icon--'],
```
jmuzina: We don't generally (unless absolutely necessary) target component classnames from sass files for other components (in this case you are applying Icon styles from within the chip styles). Could you share a bit on why this is necessary? Maybe there is a way we can minimize this together so we can keep this closer to compliance with Vanilla's general practices

scss/_patterns_chip.scss
```diff
@@ -273,3 +274,59 @@
...
+    .p-chip__value {
+      font-weight: $font-weight-medium;
+    }
+
+    & > [class*='p-icon--'] {
+      margin: 0;
+    }
+
+    & > .p-icon--circle-of-friends {
+      @include vf-icon-circle-of-friends($colors--dark-theme--text-default);
```
jmuzina: Why do we need to re-apply the icon styles here - aren't they already applied in `_patterns_icons`?

mariadias143: The issue I'm having is that the branded chip has an inverted palette (it has a dark background in light mode and a light background in dark mode), so the icon inside it must also invert its color. The standard applied in `_patterns_icons` follows the page theme but inside the branded chip we must invert it. Do you have a different approach in mind?

scss/_patterns_chip.scss
mariadias143: You're right to flag this. Looking at it again I realize it actually isn't needed at all since this rule only targets the color of the text inside the chip. The latest commit addresses this.

mariadias143:
> @mariadias143 thanks for checking, the circle part seems ok, but what I noticed now is the icon looks small , can you please double check:
> 
> Screenshot 2026-03-17 175348

@diana-stanciulescu can you check again? It should be fixed now
previously:
i p-icon--circle-of-friends 16×16

with latest changes:
ACCESSIBILITY

scss/_patterns_chip.scss
jmuzina: This sounds like the same requirement we used a while ago for the [tooltip](https://vanillaframework.io/docs/patterns/tooltips#theming). There is a slightly simpler way to do this, see [PR](https://github.com/canonical/vanilla-framework/pull/5326) for details. 

A short summary:

You can use `vf-theme-light` and `vf-theme-dark` to set the active color variables used by `$colors--theme` variables. This can be used to invert the active theme for some node and all its descendants. 

One drawback from using just pure CSS is that we have no contextual awareness of exactly what color theme is applied at the level of the tooltip/chip. We can make a (mostly safe) assumption that the active theme is whatever is set on `body` (most projects do not have nested areas of their app with different themes than the rest of the app). 

To handle the case where that assumption breaks, such as when `body.is-light`, `section.is-dark`, and there's a chip inside that section, it needs to invert to use the light theme - so we can ask and expect for the user to set the `is-dark` themself on the chip so it knows to use the light theme.

mariadias143: Thank you for the suggestion and clear explanation! I was a bit reluctant given the stylesheet bloat this causes but in the end I see this is consistent with how theme overrides are handled all over vanilla and probably the most idiomatic approach, so I decided to follow your rec. I updated the css and the component documentation to clarify expected usage :)

jmuzina: Thanks I'll have a look today!

templates/docs/examples/patterns/chip/branded-theme-override.html
```diff
@@ -0,0 +1,26 @@
+{% extends "_layouts/examples.html" %}
+{% block title %}Chip / Branded{% endblock %}
```
jmuzina: This example has the same title as your other example, so they're indistinguishable on the examples list:

Image

releases.yml
```diff
@@ -1,3 +1,13 @@
+- version: 4.47.0
+  features:
+    - component: Branded chip
+      url: /docs/patterns/chip
```
jmuzina: We generally keep the release notes entry titles to match the name of a docs page (a component, base element, pattern, etc). Then use the url and notes to lead the user to what specifically was changed there.


```suggestion
    - component: Chip
      url: /docs/patterns/chip#branded-chip-new
```

("new") is included in the URL due to [this issue](https://github.com/canonical/vanilla-framework/issues/5393) not being solved yet

scss/_patterns_chip.scss
jmuzina: nice work :)

mariadias143:
@jmuzina hi, sorry for all the back and forth. Latest comments have been addressed, can you review again pls?

jmuzina:
> Latest comments have been addressed, can you review again pls?

Looking good to me. Just needs a +1 from design cc @diana-stanciulescu

scss/_patterns_chip.scss
```diff
@@ -273,3 +274,54 @@
...
+    @extend %vf-button-dense-vertical-padding;
+
+    align-items: center;
+    background-color: $colors--theme--background-alt;
+    gap: 4px;
+    height: 24px;
+    justify-content: center;
+    margin-bottom: $spv-nudge-compensation;
+    min-width: max-content;
+    padding: 2.5px 8px 2.5px 4px;
```
jmuzina: One thing I just noticed, i'm curious where these specific values for `gap`, `height`, `padding` came from?

mariadias143: I directly used the values in the figma design, which I guess is not the best option. I see now that padding is unnecessary, I can use the values provided by default, but I do need to override the value for padding-left because the branded chip needs the icon a bit shifted to the left. Gap and height are necessary to ensure the component is compliant with the design provided (the chip with icon has a margin-right in the icon element but I override that so that the chip looks ok when it is icon only by setting margin: 0, so gap would be a cleaner way to ensure the spacing between icon and label).
What do you think if I went with the following approach, using relative values for height and gap? This would ensure the design is respected while at the same time not using fixed px values.

Image

diana-stanciulescu (approved):
design looks good to me. thanks!

scss/_patterns_chip.scss
jmuzina: Yes I would go for this

mariadias143: done

jmuzina (approved):
nice work!

PR #5777, opened 2026-03-19
fix: a11y low contrast in dark mode for code [AC-4348]

Stefan3002:
## Done

- Increased the contrast for the <code> code </code> element in dark mode.
- Screenshots show it was too low.
- The contrast is 5.13 now.

## QA

- [Add QA steps]
Look at a confirmation modal that says: Next time press SHIFT, the contrast of the SHIFT text should be increased and pass a11y guidelines

Fixes: [AC-4348](https://warthogs.atlassian.net/browse/AC-4348)

## Screenshots

image

[AC-4348]: https://warthogs.atlassian.net/browse/AC-4348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scss/_settings_colors.scss
```diff
@@ -58,7 +58,7 @@ $hover-background-opacity-percentage: $hover-background-opacity-amount * 100%;
 // NON-SEMANTIC COLOURS
 $color-label-validated: #006b75;
 $color-code-background: rgba($color-x-dark, 0.03);
-$color-code-background-dark: rgba($color-x-light, 0.3);
+$color-code-background-dark: rgba($color-x-light, 0.1);
```
jmuzina: While indeed the code text now has greater contrast against its own background, this also **decreases** the contrast between code and non-code text - it seems quite a bit harder to distinguish them.

#### Before
Image

#### After
Image

@juanruitina does this new background color look OK to you?

juanruitina: It is more important to be able to read the text than to tell the semantics apart (and in this case, semantics are even reinforced by the monospace font). The contrast lower than 4.5 is not acceptable.

That said, @Stefan3002 it seems the case you are sharing in your screenshot is a corner case caused by the use of [muted text](https://vanillaframework.io/docs/base/typography#muted-text). `rgba($color-x-light, 0.15)` would be sufficient to make the contrast greater than 4.5, see below.  We could apply this to `code` tags only inside muted text, or for all code for the sake of simplicity and increased text contrast in all cases. Both options would be compliant.

image

Interestingly, this is not an issue in the light theme:

image

Stefan3002: Indeed, the problem is only present in dark mode :smile:

Stefan3002:
Hey all, sorry for the huge delay. I was in medical leave unfortunately.

Could I get another review on this please? :smile:

jmuzina:
No worries, apologies for the delay also, I was on sick leave!

I'll give it a look tomorrow :)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vanilla-framework",
-  "version": "4.47.0",
+  "version": "4.47.1",
```
jmuzina: 4.47.0 hasn't been released yet - so doing this would actually skip 4.47.0 in the release order.

I'd recommend to keep this as 4.47.0

```suggestion
  "version": "4.47.0",
```

scss/_utilities_muted-text.scss
```diff
@@ -1,5 +1,9 @@
 @mixin vf-u-text-muted {
   .u-text--muted {
     @extend %muted-text;
+
+    code {
+      background-color: rgba($color-x-light, 0.15);
+    }
```
jmuzina: This should be applied inside `%muted-text` placeholder, not the muted text class selector, so it applies to elements that extend `%muted-text` but don't have `u-text--muted` class (currently only [notification timestamps](https://github.com/Stefan3002/vanilla-framework/blob/df439579798f5202138ecfed9c4e9afe9f6f1df2/scss/_patterns_notifications.scss#L177)).

scss/_utilities_muted-text.scss
```diff
@@ -1,5 +1,9 @@
 @mixin vf-u-text-muted {
   .u-text--muted {
     @extend %muted-text;
+
+    code {
+      background-color: rgba($color-x-light, 0.15);
```
jmuzina: `$color-x-light` is not color-theme sensitive. So, If you test this on light theme, the code background becomes indistinguishable from the background.

Image

To solve this we'd need an opinion from design on what the right background color is for code inside muted text on light mode. But, @juanruitina mentioned this is fine, so I'm OK to leave this mostly as-is, but we should add a comment here explaining why we are using a non-themable variable here.

scss/_utilities_muted-text.scss
jmuzina, line 1: This PR produces no visual diffs because Vanilla has no examples that use `code` inside of muted text (I don't think Vanilla was aware of this as a use case). Could you add an example for Code that shows it inside an element with muted text? That way we can capture visual regressions of this use case in future. 

It doesn't need to be shown on the docs page - just adding it to `examples/` and the `combined.html` example for Code will cause it to be visually tested.

scss/_utilities_muted-text.scss
```diff
@@ -1,5 +1,9 @@
 @mixin vf-u-text-muted {
   .u-text--muted {
     @extend %muted-text;
+
+    code {
```
jmuzina: Explicitly targeting element tags causes a stylelint error (we try not to apply styles to tags unless we absolutely have to).

I think this is a valid use case for a tag selector though. To ignore the error and pass CI, see one of [these examples](https://github.com/search?q=repo%3Acanonical%2Fvanilla-framework%20selector-max-type&type=code) to temporarily disable (then re-enable) this rule.

scss/_utilities_muted-text.scss
juanruitina: Little misunderstanding. The quick fix I suggest is to apply the subtlier background of `rgba(255, 255, 255, 0.15)` to code blocks in the dark theme, regardless of muted or not. I think the proposed code instead targets muted code regardless of theme, resulting in a regression in the light theme.

For the record, I think I was looking at the live version, where there is such a distinction because of the use of `var(--vf-color-background-code)`, `rgba(0, 0, 0, 0.03` and should remain unchanged:

Screenshot From 2026-04-06 10-14-46

This is what it looks like in the demo (which is a regression):

Screenshot From 2026-04-06 10-14-52

Stefan3002:
Hello @jmuzina and @juanruitina. I have addressed the comments and went with an approach that targets code blocks in dark theme, regardless of them being muted or not. How does this look now?

Stefan3002:
Hey there @jmuzina and @juanruitina, sorry to bother. I know this PR is a bit old, but could you have a look when you get some time? I would like to merge it if you agree with the changes. Thanks!!

scss/_settings_colors.scss
jmuzina, line 1: Can we set the version to `4.50.1` now that `4.50.0` has been released?

jmuzina (approved):
lgtm, please just update the package version before merging (see below)

scss/_settings_colors.scss
Stefan3002: Sure, done

PR #5782, opened 2026-03-27
Wd 31301 card dev

Skazitron:
## Done

  - Implemented the **Card** pattern (`.p-content-card`).
  - Created a flexible Jinja macro (`vf_card`) supporting configurable elements: headings, images, author/date metadata, and a footer with Vanilla icons and status chips.
  - Built 3 responsive layout variants using a mobile-first SCSS approach:
      - **2-column**: Default vertical layout.
      - **4-column**: Horizontal layout (collapses to 2-col on small screens `< 768px`).
      - **6-column**: Expansive horizontal layout (collapses to 2-col on medium screens `< 1024px`).
  - Implemented a CSS pseudo-element overlay trick to make the entire card a clickable link while maintaining valid HTML (no nested `<a>` tags) and preserving the clickability of nested elements (like the author link).
  - Applied a CSS `mask-image` gradient to the footer to gracefully fade out overflowing chips/text.
  - Wrote comprehensive documentation and added embedded Jinja examples for all layout variations.

## QA

  - Open [demo](https://vanilla-framework-5782.demos.haus/docs/patterns/content-card)
  - **Check Layouts:** Verify the 2-column, 4-column, and 6-column examples render correctly with and without images.
  - **Check Responsiveness:** Resize your browser window. Verify the 4-column card snaps to a vertical 2-column layout below `768px`, and the 6-column card snaps to vertical below `1024px`.
  - **Check Links:** - Click anywhere on the card's background or image to ensure it triggers the main heading link.
      - Hover and click the **Author** name to ensure the nested link works independently of the main card link.
  - **Check Overflow:** Shrink the card width or add long text to the footer to verify the horizontal scroll and transparent fade-out mask work as expected.
      - `docs/patterns/content-card/index.md` (Check that the macro API table and examples are accurate).
 - Cross reference with [figma doc](https://www.figma.com/design/VEWPciJ3sMhKOmUelKE7Q8/Card-component---Sites?node-id=701-12554&m=dev)

## Fixes
https://warthogs.atlassian.net/browse/WD-31301


scss/_patterns_content-card.scss
```diff
@@ -0,0 +1,268 @@
...
+        Modifier for 4-column and 6-column horizontal layouts.
+*/
+
+@mixin vf-p-content-card {
+  .p-content-card {
+    align-items: flex-start;
+    border: 1px solid var(--color-border-low-contrast, rgba(0, 0, 0, 0.1));
+    display: flex;
+    flex-direction: column;
+    height: 376px;
```
samhotep: The values for `height:` and `width:`  here, and elsewhere, should be set to variables, and should also be in `rem`

scss/_patterns_content-card.scss
```diff
@@ -0,0 +1,268 @@
...
+    &.p-content-card__has-image {
+      padding-top: 0;
+    }
+
+    @media screen and (min-width: 768px) {
+      &.p-content-card--4-cols {
+        align-items: flex-start;
+        align-self: stretch;
+        flex-direction: row;
+        gap: 16px;
```
samhotep: `gap` should also use `rem`, and probably a variable if it will be reused

mattea-turic:
Thank youu @Skazitron !

---
**General:**
Heading
- [x] Could the heading pls not be linked, so it's just plain text H2
- [ ] I realise the 2-col cards wouldn't work great on mobile – is it possible for these to occupy the full-width of the page, similar to our other patterns like the basic section
- [ ] Similar to the above, if we apply a 4-col or a 2-col card to the tablet, which has 4-columns anyway – these will also resize to occupy 100% or 50% (respectively) of the screen, right?

Footer
- [x] Icon + Category: Could the icon and category be closer together? I understand that there is baked-in padding, so can we remove any other spacing/gutter aside from this?
- [x] Would it also be possible to see different icons, as center-alignment might look more balanced rather than baseline 
- [x] I wonder if it's worth adding some space to the right of the container holding the tag and chip, so the elements can be pushed further to the left, avoiding the fade-out upon swiping to the end of the footer
- [x] More of a design question – @eliman11 dyt it would look nicer to have more spacing between the tag and chip? The 8 px I intended is a little claustrophobic


---
**4-Column Card Without Image:**
Heading
- [x] There seems to be excess bottom padding here

---
Could we see this as part of a mockup, so e.g. like in the Figma here: <br>
Screenshot 2026-03-27 at 2 46 54 PM

Just to see how it would look in practice, especially important for the animation (here I mean the border, if not the desc.)or with varying content length – and for better understanding as to how it'll look on mobile and tablet

scss/_patterns_content-card.scss
```diff
@@ -0,0 +1,268 @@
...
+        height: 187.25px;
+        padding: 16px 16px 0;
```
samhotep: should use default padding e.g ` $spv--large`

scss/_patterns_content-card.scss
```diff
@@ -0,0 +1,268 @@
...
+        width: var(--size-width-grid-col-4, 600px);
+
+        &:not(.p-content-card__has-image) .p-content-card__hr {
+          margin-left: -16px;
+          margin-right: -16px;
```
samhotep: should use default margin

templates/_macros/vf_card.jinja
```diff
@@ -0,0 +1,78 @@
...
+                    {%- else -%}
+                        <span class="p-content-card__author-link">{{ author_and_date.author }}</span>
+                    {%- endif -%} &nbsp; &middot;
+                    <span class="u-text--muted">{{ author_and_date.date }}</span> 
+                </small>
+            </div>
+            {%- endif -%}
+        </div>
+        <div>
+            {%- if footer -%}
```
samhotep: this should include the div on line `57`

scss/_patterns_content-card.scss
```diff
@@ -0,0 +1,268 @@
...
+
+    &__image-wrapper {
+      align-items: flex-start;
+      display: flex;
+      flex-direction: column;
+      gap: var(--spacing-vertical-image-container, 10px);
+      padding-bottom: var(--spacing-vertical-large, 16px);
+      width: 284px;
+
+      @media screen and (min-width: 768px) {
```
samhotep: should use default breakpoint variables

eliman11:
Thanks @Skazitron! Noting that tooltips on hover and the description on hover are pending. Other than that - 

- [x] Headings in the card are cropped unexpectedly. For example in the first screenshot it crops the top of the title, and in the second it shows the top of the next line
Screenshot 2026-03-27 at 12 23 30
Screenshot 2026-03-27 at 12 23 54

- [x] Headings shouldn't be links per design
- [x] At the moment scrolling to the end of the "footer" in the card still shows the white gradient on the right, giving the impression there is more content on the side when there isn't. Spoke to @mattea-turic about this, I think she had a solution in mind
Screenshot 2026-03-27 at 16 58 17

Researched keyboard navigation a bit more and found we'll need to make the following changes:
- [x] The whole card should be clickable - pressing "Enter" should take you to the link
- [x] Instead of separating the various elements in the card, the tab order should be: 1) Focus lands on the card, 2) Focus lands on the author name, which should take you to the link and 3) focus moves completely out of the card and onto the next element on the page (could be the next card or the next title, etc)

scss/_patterns_content-card.scss
Skazitron: > should use default margin

@samhotep I've spoken to design about the rule element that it's being applied to and they seem to be okay with it.

Skazitron:
@eliman11 Thanks for your suggestions. One thing to note: there is no way to make the footer clickable without losing scrollability due to limitations with HTML and CSS. I've implemented the other changes though.

eliman11:
Thanks @Skazitron! Not sure I understand - I've recorded this Loom to explain what I meant https://www.loom.com/share/c85d39673dcb4b48bc3a59b937f7941e

scss/_patterns_content-card.scss
samhotep: @Skazitron I see, okay then maybe worth adding variables for these values

mattea-turic:
Thank you @Skazitron ! Animation looks super cool :D

This is prob a dumb q, but say for tablet, would we be able to use the 4 col cards, or switch variants from e.g. 2 col card on desktop -> 4 col card on tablet, as tablet [uses the 4 col grid](https://vanillaframework.io/docs/patterns/grid)

Just a quick question for the 4 col card here:
Screenshot 2026-03-31 at 1 09 33 PM

Is there a reason that "MLOps..." seems to wrap early? As there seems to be space following the previous word

Skazitron:
- [x] Fixed the text wrap issue

As for turning the 2-col cards into 4 col cards we may able to do it but we'll have to hide the author and the title.

mattea-turic:
Thanks @Skazitron for the quick changes!

Another question – when we specify e..g 2 col, 4 col, etc. – those are technically 25% and 50%, respectively (for tablet). But, when adjusting the window size, they remain a fixed width. Could you make sure they comply with our grid, so are responding to the containers there.

E.g. [equal heights](https://vanilla-framework-5782.demos.haus/docs/patterns/equal-heights)

mattea-turic:
@Skazitron  

I think we can move onto code and QA now, but just wanted to go back to the question/note I added here, as I think it was accidentally missed :')

> Another question – when we specify e..g 2 col, 4 col, etc. – those are technically 25% and 50%, respectively (for tablet). But, when adjusting the window size, they remain a fixed width. Could you make sure they comply with our grid, so are responding to the containers there.
> 
> E.g. [equal heights](https://vanilla-framework-5782.demos.haus/docs/patterns/equal-heights)

Thanks!

scss/_patterns_content-card.scss
```diff
@@ -0,0 +1,434 @@
...
+        align-items: flex-start;
+        align-self: stretch;
+        flex-direction: row;
+        gap: $sp-medium; /* 16px */
+        height: 11.703125rem; /* 187.25px */
+        padding: $sp-medium $sp-medium 0;
+        width: var(--size-width-grid-col-4, 37.5rem); /* 600px */
+      }
+
+      &.p-content-card--4-cols:not(.p-content-card__has-image),
```
Onibenjo: For these classes

These should be `p-content-card--has-image` (BEM modifier) or simply `has-image` / `has-desc`.

Across the Vanilla codebase, state classes use either:                          
  - Standalone: has-image, has-search-open (see [_patterns_pull-quotes.scss](https://github.com/canonical/vanilla-framework/blob/dab84ddbd7cb924ace010e5bc364eb0663f6102c/scss/_patterns_pull-quotes.scss#L16))                                        
  - is- prefix: is-active, is-selected (see https://github.com/canonical/vanilla-framework/blob/dab84ddbd7cb924ace010e5bc364eb0663f6102c/scss/_patterns_navigation.scss#L273)

templates/_macros/vf_card.jinja
```diff
@@ -0,0 +1,92 @@
...
+  col-4: 4-column width card layout (Horizontal)
+    Params: columns (4), heading (req), link (req), image (opt), footer (opt), description (opt)
+  
+  col-6: 6-column width card layout (Horizontal)
+    Params: columns (6), heading (req), link (req), image (req), author_and_date (opt), footer (opt), description (opt)
+#}
+
+{%- macro vf_card(columns="2", link=None, heading=None, image=None, author_and_date=None, footer=None, description=None) -%}
+
+{%- set layout = 'p-content-card--' ~ columns ~ '-cols' -%}
```
Onibenjo, line 25: The layout class (e.g. p-content-card--2-cols) is applied to nearly every child element (__image-wrapper,  __image, __body, __heading, etc.) in the Jinja template. Modifiers belong on the block only; children should be styled via descendant selectors:                                                                                 
                                                       
  // Anti-pattern (current):
`  &__image { &.p-content-card--4-cols { ... } }   `
                                                                                                                   
  // Correct:
`  .p-content-card--4-cols &__image { ... }       `                                                                  
                                                       
I see you already use the correct pattern in some places (e.g. [here](https://github.com/canonical/vanilla-framework/pull/5782/changes#diff-91e35986f1d5c98a32db4ea67acfd75a8e5a02c94978cfcfa61d514b24f155f7R185)).

scss/_patterns_content-card.scss
```diff
@@ -0,0 +1,434 @@
...
+    Content:
+      .p-content-card__content:
+        Flex container wrapping the main body and footer of the card.
+      .p-content-card__body:
+        Grid container holding both the primary and hover content areas.
+      .p-content-card__primary-content:
+        The default visible content area (heading, author, date).
+      .p-content-card__hover-content:
+        The content area that slides up on hover (description).
+      .p-content-card__heading-wrapper:
```
Onibenjo: we can remove this __heading-wrapper since it is defined in SCSS and @classreference but never used in the template (dead code).

templates/_macros/vf_card.jinja
```diff
@@ -0,0 +1,92 @@
...
+{%- set img_class = 'p-content-card__has-image' if image else '' -%}
+{%- set has_desc = 'p-content-card__has-desc' if description else '' -%}
+
+<div class="p-content-card {{ layout }} {{ img_class }} {{ has_desc }}" aria-describedby="default-tooltip">
+
+    {%- if link -%}
+        <a href="{{ link }}" target="_blank" class="p-content-card__overlay-link" tabindex="-1" aria-hidden="true"></a>
```
Onibenjo:
```suggestion
        <a href="{{ link }}" class="p-content-card__overlay-link" tabindex="-1" aria-hidden="true"></a>
```

It's not our common practice for links to open in new tabs because of accessibility concerns - see https://www.w3.org/TR/WCAG20-TECHS/G200.html

Links shouldn't open in new tabs else users might lose progress on the current page, so links that we typically have to other web pages shouldn't open in new tabs

templates/_macros/vf_card.jinja
```diff
@@ -0,0 +1,92 @@
...
+    Params: columns (6), heading (req), link (req), image (req), author_and_date (opt), footer (opt), description (opt)
+#}
+
+{%- macro vf_card(columns="2", link=None, heading=None, image=None, author_and_date=None, footer=None, description=None) -%}
+
+{%- set layout = 'p-content-card--' ~ columns ~ '-cols' -%}
+{%- set img_class = 'p-content-card__has-image' if image else '' -%}
+{%- set has_desc = 'p-content-card__has-desc' if description else '' -%}
+
+<div class="p-content-card {{ layout }} {{ img_class }} {{ has_desc }}" aria-describedby="default-tooltip">
```
Onibenjo: `aria-describedby="default-tooltip"` references an ID that isn’t defined in this macro. Is this expected?

templates/_macros/vf_card.jinja
```diff
@@ -0,0 +1,92 @@
+{#
+  All Params
+  columns: Grid column span for the card. Options are 2, 4, 6. Default is 2.
+  heading: H4 title text. Truncated at 3 lines. Required.
+  link: Link to referenced item. Makes the entire card clickable. Required.
+  image: Dictionary containing 'url' and 'alt' for the 16:9 image. Required for col-6.
+  author_and_date: Dictionary containing 'author', 'date', and 'link'. Hidden if no image is provided.
```
Onibenjo: I see `link` but markup doesn’t implement it.

Onibenjo (commented):
some comments

templates/_macros/vf_card.jinja
Skazitron: You're right. The tooltip was a part of the first rendition of the card. I've gone and removed it.

templates/_macros/vf_card.jinja
Skazitron: Removed (it was part of the first rendition).

Skazitron:
@Onibenjo thanks for your review 🙏🏽 I've implemented your suggestions. Please take another look.

templates/_macros/vf_card.jinja
```diff
@@ -0,0 +1,92 @@
+{#
+  All Params
+  columns: Grid column span for the card. Options are 2, 4, 6. Default is 2.
+  heading: H4 title text. Truncated at 3 lines. Required.
+  link: Link to referenced item. Makes the entire card clickable. Required.
+  image: Dictionary containing 'url' and 'alt' for the 16:9 image. Required for col-6.
```
Onibenjo: Can we rename this to `src` instead of url?

This is because we usually use the image template when rendering images as it parses the url and returns a cloudinary link which serves the image via CDN.

Now the image template can output an object. Like [this](https://github.com/canonical/canonicalwebteam.image-template?tab=readme-ov-file#attribute-output)
and src and alt is part of that object
this would make it easier for anyone calling this jinja pattern to pass the image template output to it

templates/_macros/vf_card.jinja
```diff
@@ -0,0 +1,92 @@
...
+  author_and_date: Dictionary containing 'author' and 'date'. Hidden if no image is provided.
```
Onibenjo: I think we can name this something else like `article`

The "and" signals that one parameter is doing double duty for two distinct concepts. 

We can either:
- Split into separate parameters (preferred — matches how other macros handle distinct data)             
```  
macro vf_card(                                                                                                   
    ...                                                                                                            
    author="",                                                                                                     
    date="",                                                                                                       
    ...     
  )                                                                                                                
   ```                                                    
- Use a grouping noun (if they must stay bundled as a dict)
```  
macro vf_card(                                                                                                   
    ...         
    metadata={},   {# or "article" #} or something similar                                                                               
    ...                                                
  )                                                                                                                
  ``` 
Looking at the naming patterns across the codebase, we usually use simple nouns for objects: blocks, articles, tiers, footer, image   or they stay separated

Onibenjo (commented):
final comments

Onibenjo (approved):

PR #5784, opened 2026-03-27
fix: restrict textarea resize to vertical and enable auto-grow

### LINKED ISSUE #5713: [Component amendment]: <Multi-line text field>
## Component/pattern to amend

Multi-line text field

## Visual

Image

(https://vanillaframework.io/docs/base/forms)

## Context

Currently the multi line text field allows dragging the corner to adjust the height and width of the text field. It relies on individual implementations to decide if the vertical and horizontal resizing should be allowed. Horizontal dragging many times breaks the UI and the vertical is easily missed. A lot of times we end up with a small height of the box, which makes it hard for users to scroll within a small available space to see what they wrote. 

My proposal is that we enlarge the vertical height as the user types their text, so that the text field automatically grows with height together with text. Fort certain use cases there might be a need to still define the max height of the text field.

Aswath-Senthilkumar:
## Done
* Restricted textarea resize to vertical only (`resize: vertical`)
* Enabled auto-grow in height as the user types (`field-sizing: content`)
* Added configurable `$textarea-min-height` (defaults to 3 lines) and `$textarea-max-height` (defaults to `none`) variables in `_settings_spacing.scss` with `!default` flag for user override
* Updated docs heading with `{{ status('updated') }}` label
* Added entry to `releases.yml`
* `field-sizing: content` has ~80% global browser support as of March 2026 ([[caniuse](https://caniuse.com/mdn-css_properties_field-sizing)]). Unsupported browsers fall back gracefully, the textarea remains functional with fixed sizing and vertical-only resize still applies.

## QA
* Verify textarea only resizes vertically, the horizontal drag handle should be gone
* Verify textarea grows in height automatically as text is typed
* Verify textarea starts at a 3-row minimum height
* Tested locally using `dotrun` on WSL with the docs site at `localhost:8101`

#### Before:
Resizing_before

#### After:
Resizing_after

#### Before:
Auto-grow_before

#### After:
Auto-grow_after

---

Fixes #5713

### DIFF
--- a/releases.yml
+++ b/releases.yml
@@ -11,7 +11,7 @@
     - component: Multi-line text field
       url: /docs/base/forms#multi-line-text-field
       status: Updated
-      notes: Textarea now auto-grows in height as the user types and is restricted to vertical resizing only.
+      notes: Textarea now auto-grows in height as the user types and is restricted to vertical resizing only. Default min and max height can be customised via $textarea-min-height and $textarea-max-height variables.
 - version: 4.46.0
   features:
     - component: CTA section
--- a/releases.yml
+++ b/releases.yml
@@ -8,6 +8,10 @@
       url: /docs/patterns/icons
       status: New
       notes: Added icon for circle of friends.
+    - component: Multi-line text field
+      url: /docs/base/forms#multi-line-text-field
+      status: Updated
+      notes: Textarea now auto-grows in height as the user types and is restricted to vertical resizing only.
 - version: 4.46.0
   features:
     - component: CTA section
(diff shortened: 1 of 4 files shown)


Aswath-Senthilkumar:
Should `$textarea-min-height` and `$textarea-max-height` be documented on the spacing settings page? I noticed the existing form input variables (`$input-margin-bottom`, `$input-border-thickness`) aren't listed there either, so I wasn't sure if that's the right place. Happy to add it in this PR or as a follow-up if needed.

Issue #5067, opened 2024-04-30
Implement Expand and Copy functionality for Code snippets


minkyngkm:
In the support portal's comment sections, code snippets are squished and users have to scroll horizontally to read them. 

![image](https://github.com/canonical/vanilla-framework/assets/57550290/5bb1d413-fe7f-4951-9c02-df36aa028a7f)

Would it be possible to implement an "Expand Snippet" feature to view the code on a larger screen?
Something like this. This is "Expand snippet" from [StackOverflow](https://stackoverflow.com/questions/78370393/making-list-group-in-one-column-scroll-with-page-in-line-with-another-column-tha)

<img width="727" alt="Screen Shot 2024-04-30 at 12 47 20 PM" src="https://github.com/canonical/vanilla-framework/assets/57550290/f6b8e889-3c56-4eb5-9184-bb2897cd6169">

Additionally, would it be possible to include a "Copy Code" function to easily copy code snippets?

danielmutis:
* Nina to explore visual solutions for this. Ideally within the toolbar that this component already has. 
* Copy buttons are something we don't do as a policy from the gods.

dgtlntv:
Design systems contribution meeting:
- Generally think this could be valuable for users but we need to come up with a strong case to challenge existing policy.
- We currently don't have immediate resources or need to work on this right now.

PR #5795, opened 2026-04-22
feat: HMR for scss changes + copy code button

Skazitron:
## Done
- Added a hot module replacement (HMR) for scss.
- Added a `Copy code` button/link to code examples.

## QA
- Run the application locally
- For HMR:
  - Navigate to a page that has content with its own SCSS. Example: the Card Vanilla pattern
  - Open the corresponding `scss` file for the pattern in your favorite editor.  For the Card it would be `scss/_patterns_content-card.scss`
  - Change the contents of the scss. Adding a bright red background is my go-to in these scenarios.
  - Wait a few seconds and do a hard refresh for on the page.
- For 'Copy code':
   - This should be right next to the `Edit on Codepen` link under example code blocks.
   - Try pasting the copied code somewhere: formatting should remain the same.

## Fixes
https://warthogs.atlassian.net/browse/WD-36246
https://warthogs.atlassian.net/browse/WD-36255


package.json
```diff
@@ -95,5 +100,8 @@
   },
   "peerDependencies": {
     "sass": "^1.79.0"
+  },
+  "dependencies": {
+    "concurrently": "^9.2.1"
```
alvaromateo, line 105: Why there's concurrently as a dev dependency?
I don't see it used anywhere.

releases.yml
```diff
@@ -1,3 +1,9 @@
+- version: 4.51.0
+  features:
+    - component: Card pattern
```
alvaromateo, line 3: This PR doesn't modify the card pattern component, so I don't think it should be mentioned here.

I don't know what's the right approach to add a feature in a release for which changes are just development improvements. I would not add any new release for this change, as it is used only in development so there's no need to release it for people to take advantage of it.

alvaromateo (changes requested):
I'm concerned also about all the changes in yarn.lock file. It seems most of the packages have changed from `registry.yarnpkg.com` to `registry.npmjs.org`. Theoretically the packages should be the same... But is this a change that happened accidentally? Do you have any `npmRegistryServer` defined somewhere that might have caused this?

package.json
```diff
@@ -47,7 +47,10 @@
     "build-class-references": "node scripts/create-class-references.js",
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
-    "serve": "./entrypoint 0.0.0.0:${PORT}",
+    "serve": "npm-run-all --parallel serve:flask watch:scss serve:sync:conditional",
+    "serve:flask": "./entrypoint 0.0.0.0:${PORT:-8101}",
```
jmuzina, line 51: `serve` command has been a bit of an artifact of older days of Vanilla, we [document](https://github.com/canonical/vanilla-framework/blob/16db4952c5911b4b505b157bf8d78d8eee0d4e39/guides/hacking.md#running-the-project) for devs to use `dotrun` to run the project. 

So, now we're adding a dependency and updating `serve` scripts, breathing some more life into them. My concern: we should be careful here about introducing alternate methods to do the same task in the dev process, it may add confusion to the developer - I.e. now they could start the server with `dotrun` or `serve`. Do you see this as a path to fully replacing dotrun in the dev workflow?

package.json
```diff
@@ -47,7 +47,10 @@
...
+    "serve:sync": "browser-sync start --proxy \"127.0.0.1:${PORT:-8101}\" --port ${BROWSER_SYNC_PORT:-54567} --ui-port 8012 --files \"build/css/**/*.css\"",
+    "serve:sync:conditional": "if [ \"$FLASK_DEBUG\" = \"true\" ]; then yarn serve:sync; else echo \"FLASK_DEBUG is not true. Skipping BrowserSync.\"; fi",
```
jmuzina, line 53: These commands conflict with our [guidance](https://github.com/canonical/vanilla-framework/blob/16db4952c5911b4b505b157bf8d78d8eee0d4e39/guides/hacking.md#running-the-project) to use `dotrun watch`. I think this functionality should either be added to dotrun itself, or we should remove dotrun entirely (not easy tasks I know).

Do you think we really need a new dependency just to support scss reload? If so, can we also support a watch mode for [essential scss files only](https://github.com/canonical/vanilla-framework/blob/16db4952c5911b4b505b157bf8d78d8eee0d4e39/guides/hacking.md#running-the-project)?

templates/static/js/example.js
```diff
@@ -461,12 +469,45 @@
...
       handleCodePenClick(link, form);
+
+      // --- 2. Setup Copy Link (Right next to CodePen) ---
+      const copyLink = document.createElement('a');
+      copyLink.innerHTML = 'Copy code';
+
+      // Inline styling to match the CodePen link exactly
+      copyLink.style.cssText = 'display: block; padding: 0.5rem 0; cursor: pointer; margin: 0;';
+      copyLink.setAttribute('aria-label', 'Copy code to clipboard');
+      copyLink.setAttribute('title', 'Copy code');
```
jmuzina, line 486: Please check with Design before implementing a snippet copy button. Last I heard (see [issue](https://github.com/canonical/vanilla-framework/issues/5067#issuecomment-3004059665)), there is company policy to avoid code copy buttons, but I don't know the details. To quote @dgtlntv :

> this could be valuable for users but we need to come up with a strong case to challenge existing policy

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vanilla-framework",
-  "version": "4.50.0",
+  "version": "4.51.0",
```
jmuzina, line 3: No external API changes here, no version bump needed

```suggestion
  "version": "4.50.0",
```

templates/static/js/example.js
jmuzina: Also, I would suggest that even if we were to add a code copy button, it should be done in a separate PR from a SCSS HMR PR, for separation of concerns / ease of review :)

package.json
Skazitron: I see. I was thinking about not upgrading the release version (since there are no API changes), but a few devs recommended that I do it. Thanks for the clarification.

templates/static/js/example.js
Skazitron: I will be reaching out to @dgtlntv on MM.

package.json
Skazitron: @jmuzina adding another function to dotrun imo would defeat the purpose of having server + HMR started using a single command. Maybe it's alright if we stick to `dotrun watch` for now.

I have worked on implementing `Taskfile` for a smaller website, so maybe I could work on implementing that for Vanilla.

package.json
Skazitron: I don't see this as fully replacing `dotrun`, no.

PR #5833, opened 2026-08-19
feat(typography): adopt Baseline Foundry editorial type scale

lyubomir-popov:
Adopts the Baseline Foundry editorial type scale for headings, body and small text, and replaces the hand-tuned baseline nudges with values generated from the font's real vertical metrics by `@lyubomir-popov/baseline-nudge-generator`.

## Values

| level | font-size | line-height | space after | nudge |
|---|---|---|---|---|
| display | 5rem → **5.25rem** | 11 → **12 units** | 3 units | 0.35 → **0.1644rem** |
| display-mobile | 4rem | 9 units | 3 units | 0.25 → **0.34313rem** |
| h1, h2 | 2.5rem → **2.625rem** | 6 units (mobile 5 → **6**) | 3 → **2 units** | 0.55 → **0.56881rem** |
| h3, h4 | 1.5rem | 4 units | 3 → **2 units** | 0.45 → **0.46917rem** |
| h5, h6 | 1rem | 3 units | 3 → **2 units** | 0.375 → **0.41rem** |
| p | 1rem | 3 units | 3 → **2 units** | 0.375 → **0.41rem** |
| small | 0.875rem | 2.5 units | 2 → **1.5 units** | 0.05 → **0.0775rem** |
| x-small | 0.75rem | 2 units | 2 units | 0.25 → **0.245rem** |

Sizes are steps on the Mortensen modular scale the Canonical type-scale spec is built on. `display` was 80px, which is not on the scale; it moves to 84px, exactly one octave above h1.

## Behaviour changes to look at

**Headings sit closer to the text above them.** `sp-before` is now `0`, so the extra padding on `p + h1` is gone and spacing is owned by each element rather than by the pair. Combined with space-after dropping from 3 to 2 units, headings are noticeably tighter than before. This is the main thing to eyeball.

**`small` text blocks changed height.** Space-after moves from 2 to 1.5 units so a one-line block is a whole number of baseline units — it was 4.5 before, which pushed everything below it half a unit sideways.

**Form controls, buttons, table cells shifted very slightly.** `$spv-nudge` follows the paragraph nudge and moved 0.375rem → 0.41rem, so anything deriving padding from it moves by about half a pixel.

## Baseline accuracy

No regression. Measured mean absolute first-baseline error against the grid is 0.526px, versus 0.535px for the current hand-tuned values. Per level it is mixed and all sub-pixel. Block heights are exact whole baseline units at any line count for every level except `small`, whose 20px leading on an 8px grid re-syncs every second line — that is inherent and documented in the settings file.

A four-column rhythm audit was added to the top of the `typographic-spacing` example, with fixed 4/4/8 line counts and a rule above them, to make this reviewable by eye with the existing **Toggle baseline grid** switch.

## Notes

- Nudges are tied to the font binary. Vanilla loads `Ubuntu variable` v0.869; these are generated against the current Ubuntu Sans v1.006 release, whose vertical metrics were re-cut. The difference is under a pixel at every level and is documented in `_settings_spacing.scss`.
- No `releases.yml` entry or version bump yet — this is a breaking type-scale change and should take the version of whichever release it lands in.

## Testing

- `tests/typography.test.js` (new) — snapshots font-size, line-height, nudge and space-after for all eleven levels; asserts each occupies a whole number of baseline units; pins `small` as the only level allowed a half-unit line-height
- `tests/parker.js` — stylesheet size threshold raised to 558560
- `tests/snapshots.test.js` fails with a doubled-path `ENOENT` in `getExampleFiles`, pre-existing on `main` and unrelated


lyubomir-popov:
Superseded by #5834, which drops  + 'aseline-foundry' +  from the branch name and reverts the spacing changes so only font-size, line-height and padding-top move.

PR #5834, opened 2026-08-19
feat(typography): adopt the new Canonical type scale

lyubomir-popov:
Updates heading font sizes to the new Canonical type scale, and replaces the hand-tuned baseline nudges with values generated from the font's real vertical metrics by `@lyubomir-popov/baseline-nudge-generator`.

**Spacing is deliberately untouched.** `sp-after` and `sp-before` keep their existing values, so the `p + h1` pair rules and all margins behave exactly as they do on `main`. The only changes are font-size, line-height and padding-top.

## Values

| level | font-size | line-height | space after | nudge (padding-top) |
|---|---|---|---|---|
| display | 5rem → **5.25rem** | 11 → **12 units** | unchanged | 0.35 → **0.1644rem** |
| display-mobile | 4rem | 9 units | unchanged | 0.25 → **0.34313rem** |
| h1, h2 | 2.5rem → **2.625rem** | 6 units | unchanged | 0.55 → **0.56881rem** |
| h1, h2 mobile | 2rem | 5 units | unchanged | 0.55 → **0.53875rem** |
| h3, h4 | 1.5rem | 4 units | unchanged | 0.45 → **0.46917rem** |
| h5, h6 | 1rem | 3 units | unchanged | 0.375 → **0.41rem** |
| p | 1rem | 3 units | unchanged | 0.375 → **0.41rem** |
| small | 0.875rem | 2.5 units | unchanged | 0.05 → **0.0775rem** |
| x-small | 0.75rem | 2 units | unchanged | 0.25 → **0.245rem** |

Sizes are steps on the Mortensen modular scale the type-scale spec is built on. `display` was 80px, which is not a step; it moves to 84px, exactly one octave above h1. At 84px an 11-unit line-height no longer contains the glyphs, so it goes to 12.

## What to look at

**h1 and h2 are 2px larger** (40px → 42px) and **display is 4px larger** (80px → 84px). Everything else keeps its size.

**Headings sit ~8px lower inside their own block.** h1/h2 take an extra baseline unit of padding-top, because at 42px in a 48px line box the generated nudge is only ~1px, which left a 42px cap with less clearance above it than 16px body text has. The block height is unchanged — `margin-bottom` is derived as `sp-after - nudge`, so it absorbs the difference.

**Form controls, buttons and table cells shift by about half a pixel.** `$spv-nudge` follows the paragraph nudge and moves 0.375rem → 0.41rem, so anything deriving padding from it moves slightly.

## Baseline accuracy

No regression. Measured mean absolute first-baseline error against the grid is 0.526px, versus 0.535px for the current hand-tuned values — mixed per level, all sub-pixel. Every level still occupies a whole number of baseline units, so multi-line blocks stay in rhythm.

A four-column rhythm audit was added to the top of the `typographic-spacing` example, with fixed 4/4/8 line counts and a rule above them, to make this reviewable by eye with the existing **Toggle baseline grid** switch.

## Notes

- Nudges are tied to the font binary. Vanilla loads `Ubuntu variable` v0.869; these are generated against the current Ubuntu Sans v1.006 release, whose vertical metrics were re-cut. The difference is under a pixel at every level and is documented in `_settings_spacing.scss`.
- `small` (14px on 20px leading) is 2.5 baseline units, so it cannot sit on the 8px grid. Its nudge is generated on the 0.25rem sub-grid, which is the best available. Its 4.5-unit block is pre-existing and left alone.
- No `releases.yml` entry or version bump yet — this is a breaking type-scale change and should take the version of whichever release it lands in.

## Testing

- `tests/typography.test.js` (new) — snapshots font-size, line-height, nudge and space-after per level; asserts each occupies a whole number of baseline units; pins `small` as the only level allowed a half-unit line-height
- `tests/parker.js` — stylesheet size threshold raised to 558560
- `tests/snapshots.test.js` fails with a doubled-path `ENOENT` in `getExampleFiles`, pre-existing on `main` and unrelated


muhammad-ali-pk (approved):

muhammad-ali-pk:
@lyubomir-popov Maybe we should we also reflect these changes in the release notes as well as typography pages documentation.
Will take care of it.

muhammad-ali-pk (approved):

PR #5840, opened 2026-09-02
Fix: Tooltip text overlapping issue (#5757)

### LINKED ISSUE #5757: [Bug]: If a tooltip goes over text, the text overlaps the tooltip
**Describe the bug**

If a tooltip overlaps text in a document, the text itself (not the containers/element) is visible through the tooltip. See image below.

**To Reproduce**

1. Go to https://vanilla-framework-5755.demos.haus/docs/patterns/in-page-navigation
2. Make sure you are on large screen size
3. Hover over 'General Availability (GA) and variant Ubuntu...' to trigger the tooltip
4. Scroll the page so the tooltip overlaps text

**Screenshots**

Image

**Additional context**

This only exists in a [PR](https://github.com/canonical/vanilla-framework/pull/5755) currently, if it is merged it will no longer be reproducible in the demo link provided

SajanBisht:
Fixes #5757

Summary of changes:
Enforced opacity: 1 and isolation: isolate on .p-tooltip__message to ensure tooltips form an isolated stacking context and maintain full background opacity over overlapping content.

### DIFF
--- a/scss/_patterns_tooltips.scss
+++ b/scss/_patterns_tooltips.scss
@@ -35,8 +35,8 @@ $tooltip-overlay-top-border-radius: 0% 0% 100% 100%;
     @extend %small-text;
 
     background-color: $colors--theme--background-alt;
-    opacity: 1; // FIX #5757: Ensure full background opacity over overlapping text
-    isolation: isolate; // FIX #5757: Create isolated stacking context for pseudo-elements
+    opacity: 1; 
+    isolation: isolate; 
     border: 0;
     color: $colors--theme--text-default;
     display: none;
--- a/scss/_patterns_tooltips.scss
+++ b/scss/_patterns_tooltips.scss
@@ -35,6 +35,8 @@ $tooltip-overlay-top-border-radius: 0% 0% 100% 100%;
     @extend %small-text;
 
     background-color: $colors--theme--background-alt;
+    opacity: 1; // FIX #5757: Ensure full background opacity over overlapping text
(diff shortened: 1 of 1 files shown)


scss/_patterns_tooltips.scss
```diff
@@ -35,6 +35,8 @@ $tooltip-overlay-top-border-radius: 0% 0% 100% 100%;
     @extend %small-text;
 
     background-color: $colors--theme--background-alt;
+    opacity: 1; // FIX #5757: Ensure full background opacity over overlapping text
```
petesfrench: There is no need for the `// FIX` comments here

petesfrench:
Hi @SajanBisht, thank you for the contribution!

I tried to replicate this issue by going to https://vanillaframework.io/docs/patterns/in-page-navigation and adding some custom `<code>` blocks, but it seems the issue doesn't persist. Were you able to reproduce it?

Screenshot:
image

scss/_patterns_tooltips.scss
SajanBisht: Hi @petesfrench, thanks for the review!

I've updated the PR and removed the inline comment from scss/_patterns_tooltips.scss.

Reproduction context: The overlap occurs when tooltips hover over elements that establish their own stacking context or positioning (such as inline code blocks <code>, positioned wrappers, or elements with custom z-index/isolation). Adding isolation: isolate on .p-tooltip__message ensures the tooltip forms an independent stacking context regardless of sibling layering.

